PR #2, opened 2015-06-18
Comments around environment and user values

smarterclayton:
Clarify environment more strictly, and propose changes to user to separate "id" and "name" as distinct entities.  I think it is important to separate uid and name explicitly, given that the uid can be known outside the container, but the name may only have meaning inside the container (in the case of Linux containers that have custom /etc/nsswitch.conf)

Signed-off-by: Clayton Coleman ccoleman@redhat.com

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -320,7 +320,10 @@ See links for details about mountvol in Windows.
     "processes": [
         {
             "tty": true,
-            "user": "daemon",
+            "user": {
+              "name": "daemon",
+              "id": 0,
+            },
             "args": [
                 "sh"
             ],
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -320,7 +320,10 @@ See links for details about mountvol in Windows.
     "processes": [
         {
             "tty": true,
-            "user": "daemon",
+            "user": {
+              "name": "daemon",
+              "id": 0,
```
philips: this should be a string so it can be "optional"

runtime.md
```diff
@@ -338,11 +341,9 @@ The processes to be created inside the container are specified in a processes ar
 ```
 The command to start a process is specified in an array of args. It will be run in the working directory specified in the string cwd.
 
-Environment variables are specified is an array called env.
+Environment variables are specified is an array called `env`. Elements in the array are specified as `strings` in the form "KEY=value", and are passed unmodified to the underlying process.
```
jonboulle: is an -> in an

runtime.md
```diff
@@ -338,11 +341,9 @@ The processes to be created inside the container are specified in a processes ar
...
 
-Elements in the array are specified as Strings in the form "KEY=value"
-
-The user inside the container under which the process is running is specified under the user key.
+The user inside the container under which the process is running is specified under the `user` object. The `user` object may specify a `name` which is the string value of the user name, or the `id` which is an integer representing the Unix user identifier of the user. This is the expected value of the user as seen from the inside of the container - the runtime may define a mapping between the value seen inside the container and the value the host operating system sees.
```
jonboulle: what if both are supplied?

crosbymichael:
@smarterclayton I'm going to close this one because we have the same discussion going on in #10 about the user uid/gid vs username.  

There is nothing wrong with the PR, I just want to keep the discussion in one place and your PR lost the coin flip, sorry ;)

smarterclayton:
I'll survive :). Will split the env changes out

On Jun 29, 2015, at 4:48 PM, Michael Crosby <notifications@github.com>
wrote:

@smarterclayton <https://github.com/smarterclayton> I'm going to close this
one because we have the same discussion going on in #10
<https://github.com/opencontainers/specs/issues/10> about the user uid/gid
vs username.

There is nothing wrong with the PR, I just want to keep the discussion in
one place and your PR lost the coin flip, sorry ;)

—
Reply to this email directly or view it on GitHub
<https://github.com/opencontainers/specs/pull/2#issuecomment-116841773>.

PR #9, opened 2015-06-25
Update runtime.md

hannibalhuang:
Add a section on Modification

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -15,3 +15,7 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+### Modify
+
+Modify the container: 


philips:
@hannibalhuang Was this an accidental PR? What are you trying to express with modify?

hannibalhuang:
@philips For life cycle management usually we got create,modify and terminate, my thinking was simply adding a modification section currently is missing. But if there is more to come then I'm willing to close it :)

philips:
@hannibalhuang what sorts of things do you want to modify? In the create side it is mostly setting up namespace before exec'ing the target app. On the terminate side it is cleaning up said namespaces.

hannibalhuang:
@philips so what I have in mind are stuff like change container size due to Mesos policy, storage back-end change, network port change and so forth.

philips:
@hannibalhuang I think it is a bit early for any of these concepts. At some point we might turn the discussion towards things like updating cgroups isolation, etc.

hannibalhuang:
@philips sounds reasonable to me :) BTW updating isolation is still one kind of modification operation right?

jonboulle:
@hannibalhuang see #17

philips:
I am going to close this in favor of #17.

PR #16, opened 2015-06-25
config: Two small copyedits

wking:
Commit messages have terse descriptions ;).


philips:
We did a major rewrite to this stuff so none of this merges right now. Sorry about that. :(

PR #18, opened 2015-06-25
config-linux: Consistent namespace description wording

wking:
Format is:

    * {os-agnostic-name} -> {linux-name}: {description}

where `{linux-name}` is what you have at `/proc/{pid}/ns/{linux-name}`
(see namespaces(7)).


philips:
We decided to make namespaces OS specific. I will remove this section.

PR #21, opened 2015-06-25
README: Focus on local runtime (create/start/stop)

wking:
These actions need consistent tooling, and they're the sweet spot for
tools like runc.  Once you map the container spec to the filesystem,
the other actions (copy, snapshot, upload, download) can all be
handled by existing tools (e.g. POSIX's 'cp', 'btrfs subvolume
snapshot', netcat, etc.).  If existing tooling for those actions is
not sufficiently portable, new portable tools can be written, but that
should be separate from the container-specific stuff here.

I'm not sure about the intended meaning of the 'tagged' action.  Was
that intended for discoverability ("I want a container named 'Nginx'")
or auth ("Is my local /srv/nginx the same container 'Nginx' that Bob
audited?").  I think both are independent of the container runtime
though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ A great analogy for this is the shipping container. Just like how Standard Conta
 
 ## 1. Standard operations
 
-Just like shipping containers, Standard Containers define a set of STANDARD OPERATIONS. Shipping containers can be lifted, stacked, locked, loaded, unloaded and labelled. Similarly, Standard Containers can be started, stopped, copied, snapshotted, downloaded, uploaded and tagged.
+Just like shipping containers, Standard Containers define a set of STANDARD OPERATIONS. Shipping containers can be lifted, stacked, locked, loaded, unloaded and labelled. Similarly, Standard Containers can be created, started, and stopped using standard container tools (what this spec is about); copied and snapshotted using standard filesystem tools; and downloaded and uploaded using standard network tools.
 
 ## 2. Content-agnostic
 


README.md
```diff
@@ -19,7 +19,7 @@ A great analogy for this is the shipping container. Just like how Standard Conta
 
 ## 1. Standard operations
 
-Just like shipping containers, Standard Containers define a set of STANDARD OPERATIONS. Shipping containers can be lifted, stacked, locked, loaded, unloaded and labelled. Similarly, Standard Containers can be started, stopped, copied, snapshotted, downloaded, uploaded and tagged.
+Just like shipping containers, Standard Containers define a set of STANDARD OPERATIONS. Shipping containers can be lifted, stacked, locked, loaded, unloaded and labelled. Similarly, Standard Containers can be creates, started, and stopped using standard container tools (what this spec is about); copied and snapshotted using standard filesystem tools; and downloaded and uploaded using standard network tools.
```
whyrusleeping: `s/creates/created`

wking: On Thu, Jun 25, 2015 at 12:06:21PM -0700, Jeromy Johnson wrote:
> `s/creates/created`

Thanks.  Fixed and force-pushed.

philips:
Sure, this change seems fine to me.

taylorb-microsoft:
lgtm

mrunalp:
LGTM

PR #22, opened 2015-06-25
Placeholder for Windows configuration and a few clean up items.

taylorb-microsoft:
Adding a placeholder for the Windows Configuration.


philips:
I would rather wait until we have a better idea of the separation between "generic" and OS specific before then the first commit of config-windows can have some content.

taylorb-microsoft:
Fair enough - we should have some draft content in the next week or two.

PR #25, opened 2015-06-25
Clarifying and cleaning up root filesystem section

taylorb-microsoft:
Making the root filesystem examples match the implementation (which is cleaner) and disambiguating the notion of read-only.

https://github.com/opencontainers/specs/issues/23#issuecomment-115429350

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,7 @@ Just like shipping containers, Standard Containers are CONTENT-AGNOSTIC: all sta
 
 ## 3. Infrastructure-agnostic
 
-Both types of containers are INFRASTRUCTURE-AGNOSTIC: they can be transported to thousands of facilities around the world, and manipulated by a wide variety of equipment. A shipping container can be packed in a factory in Ukraine, transported by truck to the nearest
-
-routing center, stacked onto a train, loaded into a German boat by an Australian-built crane, stored in a warehouse at a US facility, etc. Similarly, a standard container can be bundled on my laptop, uploaded to S3, downloaded, run and snapshotted by a build server at Equinix in Virginia, uploaded to 10 staging servers in a home-made Openstack cluster, then sent to 30 production instances across 3 EC2 regions.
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Jun 25, 2015 at 04:41:48PM -0700, Taylor Brown wrote:
>   * Adding placeholder for Windows configuration
>   * Manifest version should be platform agnostic.
> …
>   * Revert "Adding placeholder for Windows configuration"

You probably want to rebase these out of this PR to avoid pulling
these in (they're from #22?).

>   * cleaning up example for filesystem to better match runc example and improve clarity.
>   * Attempting to clarify read-only FS.  (Issue 23)

These look pretty good to me.  Only suggestion is to change:

  “Some platforms may not support read-only file systems or read-only
  container file systems.”

to:

  “Some platforms may not support read-only root filesystems, and some
  may not support read-only file systems at all.”

I think that's what you were saying, and it seems like a useful
distinction to draw.

wking:
Or maybe you're saying that the filesystem containing the OC container
definition (vs. the filesystems mounted into running containers) must
be read-write?  So we have three filesystem cases:

1. The filesystem specifying the container (the “image” in Docker's
   lingo).
2. The filesystems mounted into a created container (the “container”
   in Docker's lingo).
   i. The root of the created container.
   ii. Additional filesystems mounted into the created container.

My understanding is that Windows requires 2.i to be writable [1].  In
my previous comment here, I felt like you were distinguishing between
2.i and 2.ii, but it's possible you were talking about 1?

I guess I'm not clear on how container creation is going to handle
these filesystems.  Are they just bind-mounted from their current
locations (in which case a writable 2.i requires a writable 1)?  Are
they copied off (hopefully via CoW) to some special location for the
running containers (like Docker does when launching images now)?  I
think I like the bind approach better (leaving CoW copies up to the
user and avoiding the “how do I get information out of a volume” issue
that Docker sometimes struggles with), but I'm not sure which way this
project will go.  I guess I'll spin up runc to find out ;).

[1]: https://github.com/opencontainers/specs/issues/23#issuecomment-115419866

wking:
On Thu, Jun 25, 2015 at 05:42:09PM -0700, W. Trevor King wrote:
> I guess I'm not clear on how container creation is going to handle
> these filesystems.  Are they just bind-mounted from their current
> locations (in which case a writable 2.i requires a writable 1)?

This is how runc seems to work (yay :).

taylorb-microsoft:
I'll update the PR and include the clarification you suggested, I'm new to git so apologies for the learning curve...  

To your second question - I am trying to clarify between 2.i and 2.ii.  Windows (today) can not boot off of a read only volume, it can (though rarely used directly) support secondary/additional read-only volumes.  As currently being implemented in the Windows OS the root file system (as were referring to it in this discussion) is always COW, with the express ability to mount additional file system locations (directories or volumes) from the host into the container read/write.

Issue #20, opened 2015-06-25
Support Lifecycle Hooks


vishh:
Capturing requirements for Lifecycle hooks discussed during the summit.

Lifecycle hooks was agreed to be a useful feature and support for a few basic hooks is expected to be in place. Specifically, PreStart, PreStop and PostStop hooks will be introduced to start with. Supporting hooks might require separating the lifetime of namespaces from that of the processes in the container. 

Hooks will be binaries that will be exec'ed. The Spec and current State of the container will be provided as arguments to the hook (via fds?). Are the exec hooks executed synchronously? 

This issue is meant to define the Spec for Hooks and discuss the hooks that are necessary to begin with.

LK4D4:
I think makes sense to add `PreCheckpoint`, `PostCheckpoint`, `PreRestore`, `PostRestore` at least for linux extension. ping @boucher @avagin
I think hooks must be synchronous and receive state as json to stdout.

wking:
On Thu, Jun 25, 2015 at 11:06:52AM -0700, Alexander Morozov wrote:
> I think makes sense to add `PreCheckpoint`, `PostCheckpoint`,
> `PreRestore`, `PostRestore` at least for linux extension…

Hmm, do we even have checkpoint/restore lifecycle actions [1]?  Maybe
that should be a separate issue before we talk about hooks for them
;).

[1]: https://github.com/opencontainers/specs/blob/e7e34ba6f4787abae20f408ffc8f529f3549b959/runtime.md

boucher:
It would definitely be useful to have pre/post checkpoint/restore.

philips:
@wking @LK4D4 @boucher Lets concentrate on what we know we must do: prestart and poststop before defining hooks for other stuff first.

@vishh What is a prestop hook used for in your mind? We had a debate on pre-stop and just held off on defining one in appc: https://github.com/appc/spec/issues/188#issuecomment-73999312

mrunalp:
I agree that prestart and poststop are what we call agree on as useful and can start with adding those.

mrunalp:
Should also we define what arguments are passed to these hooks as well? On Linux, we will have to pass the pid of the container process.

LK4D4:
@mrunalp I thought we will pass whole state to all hooks.

mrunalp:
@LK4D4 We could do that but wouldn't the state be implementation specific?

LK4D4:
@mrunalp But it still will be a string, right?

[...]

PR #34, opened 2015-06-29
hooks: start with pre-start and post-stop hooks.

mrunalp:
See #20 
Starting with pre-start and post-stop and then will add more specific hooks tied to namespaces in Linux.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -15,3 +15,45 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from runc cli. Process is killed from the outside.
 
 This event needs to be captured by runc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container.
+The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+Hook paths are absolute and are executed from the host's filesystem.
+
+### Pre-start
+The pre-start hooks are called after the container process is spawned, but before the user supplied command is executed.
+They are called after the container namespaces are created on Linux, so they provide an opportunity to customize the container.
+In Linux, for e.g., the network namespace could be configured in this hook.
+
+If a hook returns a non-zero exit code, then an error including the exit code and the stderr is returned to the caller and the container is torn down.
+
+### Post-stop
+The post-stop hooks are called after the container process is stopped. Cleanup or debugging could be performed in such a hook.
+If a hook returns a non-zero exit code, then an error is logged and the remaining hooks are executed.
+
+*Example*
+
+```json
+    "hooks" : {
+        "prestart": [
+            {
+                "path": "/usr/bin/fix-mounts",
+                "args": ["arg1", "arg2"],
+                "env":  [ "key1=value1"]
+            },
+            {
+                "path": "/usr/bin/setup-network"
+            }
+        ],
+        "poststop": [
+            {
+                "path": "/usr/sbin/cleanup.sh",
+                "args": ["-f"]
+            }
+        ]
+    }
+```
+
+`path` is required for a hook. `args` and `env` are optional.
--- a/spec.go
+++ b/spec.go
@@ -15,6 +15,16 @@ type Spec struct {
 	Hostname string `json:"hostname"`
 	// Mounts profile configuration for adding mounts to the container's filesystem.
 	Mounts []Mount `json:"mounts"`
+	// Hooks are the commands run at various lifecycle events of the container.
+	Hooks Hooks `json:"hooks"`
+}
+
+type Hooks struct {
+	// Prestart is a list of hooks to be run before the container process is executed.
+	// On Linux, they are run after the container namespaces are created.
+	Prestart []Hook `json:"prestart"`
+	// Poststop is a list of hooks to be run after the container process exits.
+	Poststop []Hook `json:"poststop"`
 }
 
 // Mount specifies a mount for a container.
@@ -60,4 +70,11 @@ type Platform struct {
 	OS string `json:"os"`
 	// Arch is the architecture
 	Arch string `json:"arch"`
+}
+
+// Hook specifies a command that is run at a particular event in the lifecycle of a container.
+type Hook struct {
+	Path string   `json:"path"`
+	Args []string `json:"args"`
+	Env  []string `json:"env"`
 }


runtime.md
```diff
@@ -15,3 +15,20 @@ Runs a process in a container. Can be invoked several times.
...
+
+### Pre-start
+The pre-start hook is called before the container process is started. It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
+
+### Post-stop
+The post-stop hook is called after the container process is stopped. Cleanup or debugging could be performed in such a hook.
+
+```
+    "hooks" : {
+        "pre-start": "/path/to/pre-start-hook",
```
crosbymichael: Should this be an array for the args?

runtime.md
```diff
@@ -15,3 +15,20 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allows one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over a well-known fd, so the hooks could get the information they need to do their work.
+
```
crosbymichael: We should also specify and provide an example of the payload that it sent to stdin

runtime.md
mrunalp: As per the discussion in #20, have stuck with not using arguments but passing over a fd like stdin.

runtime.md
mrunalp: We can provide an example. However, does it make sense to standardize what all is passed?
I think that will be platform specific for sure but also implementation specific up-to some extent. 
Right now I can think of this at minimum for Linux, if we were to define it:
```
{
    "pid": 1234,
    "root": "/path/to/container/root",
    "namespaces" : [
       {
        "type": "process",
        "path": "/proc/1234/ns/pid"
       },
     ...
    ]
     
}

WDYT?

File: runtime.md (hunk 1 above)
Comment (LK4D4) [inline runtime.md]: Now I think that args might be useful. You can have one binary for multiple hooks.

Comment (mrunalp) [inline runtime.md]: How about allowing arguments to be anything that the hooks want but still pass state over a fd?

Comment (LK4D4) [inline runtime.md]: Perfect

Comment (mrunalp) [inline runtime.md]: Updated.

**Comment by eyakubovich** [comment]
If containers are launched under some daemon that also handles coordination, I would rather let it take care of all the lifecycle related hooks. For example, if running under systemd, why not let it invoke the pre-start and post-exit events with its `ExecStartPre` and `ExecStopPost`. It just needs an ability to "nsenter" the container by its ID. This way we don't have to try to think of all the hooks that we might need. For example, `ExecStartPost` is also useful from systemd experience, do we add it? What about when we need to coordinate across multiple containers -- execute this action once these 3 containers have exited (e.g. shutdown load balancer once all the slaves have exited).

Granted, my examples might be "orchestration events" rather than container life-cycle events but is there a compelling reason to explicitly add hooks for just the container life-cycle. Finally, if these hooks are being declared in the container manifest, why can't the container just have a script such as this as the entry point:
```
#!/bin/sh
/path/to/pre-start-hook arg1 arg2
/path/to/executable
/path/to/post-stop
```

Just want to push back to make sure these are truly needed.

**Comment by mrunalp** [comment]
@eyakubovich perhaps the `pre-start` hook isn't named as it should. It is more of a `pre-exec` hook.
So, the intention is to call it after the `init` process is called, but before the user supplied command is exec'ed. 

In case of user namespaces, this allows a hook to customize the network namespace (as an examplle) before the container process is exec'ed. The counter argument to this is that why can't one just create a network namespace outside and then setns to it. However, that doesn't quite work, as each namespace has an owning user namespace and only the root user in that user namespace has all the privileges to make modifications to that namespace. Also, setns itself isn't allowed to a namespace belonging to parent user namespace or a sibling user namespace. 

So, such a `pre-exec` hook allows customization after the namespace fd's are created, but before the process is exec'ed. 

The flow is that container `create` process will launch an `init` process which clones into all the namespaces. The `init` process then waits on a pipe. The container `create` process then calls the `pre-exec` hook, that joins all the namespaces of `init` and make modifications to the namespaces. The `create` process then writes to the pipe to unblock the `init` process to continue on and exec the user supplied command.

One more thing of use that could be done during `pre-exec` is mounting the namespace fds on the host so these namespaces exist even after the container process exits (of course some caveats apply here). A good use-case (as mentioned by @vishh @rjnagal ) is keeping the mount namespace fd around and then examine the rootfs afterwards in a `post-stop` hook to debug if the container process exited abnormally.

@LK4D4 please correct or amend as necessary. I will add this to the document.
cc: @philips

File: runtime.md [hunk 3]
```diff
@@ -15,3 +15,20 @@ Runs a process in a container. Can be invoked several times.
...
+### Pre-start
+The pre-start hook is called before the container process is started. It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
+
+### Post-stop
+The post-stop hook is called after the container process is stopped. Cleanup or debugging could be performed in such a hook.
+
+```
+    "hooks" : {
+        "pre-start": ["/path/to/pre-start-hook", "arg1", "arg2"],
+        "post-stop": ["/path/to/post-stop-hook", "arg1"]
```
philips: We should consider the lessons learned from the docker plugins and make this more abstract names instead of absolute paths: https://github.com/docker/docker/pull/13514#issuecomment-111317974  and https://github.com/docker/docker/issues/13859

mrunalp: @philips I don't mind doing that. We would be prescribing standard locations to load the hooks from then, right? Also, that doesn't gel quite well with allowing arbitrary arguments, IMO (which isn't necessarily bad but we need to define useful arguments).

LK4D4:
@mrunalp Yeah, so `pre-exec` is after container creation, but before process start - for initial container setup, for linux it will be network creation for example.

mrunalp:
Pushed update to specify directories for hooks. Still need to resolve whether we need any arguments at all? The case that @LK4D4 mentioned could be solved by symlinks or copying the hooks with different names.

runtime.md
```diff
@@ -15,3 +15,27 @@ Runs a process in a container. Can be invoked several times.
...
+
+### Pre-start
+The pre-start hook is called after the container process is started but before the user supplied command is execed. It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
+
+### Post-stop
+The post-stop hook is called after the container process is stopped. Cleanup or debugging could be performed in such a hook.
+
+```
+    "hooks" : {
+        "pre-start": ["pre-start-hook", "arg1", "arg2"],
```
crosbymichael: This should be an array so that you can specify multiple binaries for a single hook.

```json
"pre-start": [
    ["ps", "aux"],
    ["ip", "a"]
]
```

runtime.md
```diff
@@ -15,3 +15,27 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allows one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over a well-known fd, so the hooks could get the information they need to do their work.
+
+Hooks are placed under a well known directories under which they are search for and executed. The list of directories that will be looked up are:
```
crosbymichael: We also need to specify what we will do if a hook returns a non-zero exit status.  Should we stop container execution or continue on?  I think we should stop and propagate the error back up to the caller if a hook fails.

runtime.md
mrunalp: :+1: Updating

runtime.md
mrunalp: Agreed. I have added a couple of lines to that effect.

runtime.md
```diff
@@ -15,3 +15,36 @@ Runs a process in a container. Can be invoked several times.
...
```
crosbymichael: Do we need a specific directory to store the hooks in?  Why can a hook not be any file on the filesystem?

mrunalp: @crosbymichael That is up for debate. I am leaning towards any file. @philips do you have any arguments against?

runtime.md
```diff
@@ -15,3 +15,36 @@ Runs a process in a container. Can be invoked several times.
...
+### Post-stop
+The post-stop hook are called after the container process is stopped. Cleanup or debugging could be performed in such a hook.
+
+```
+    "hooks" : {
+        "pre-start": [
+            ["fix-mounts", "arg1", "arg2"],
+            ["setup-network", "eth0"]
+        ],
+        "post-stop": [
```
rghv: So, does `post-stop` mean after the container stops? Or when the handler receives the instruction to stop the running container? Am trying to take the analogy of the libc `atexit`. So, if it's like `atexit`, would naming it `on-stop` make it easier to understand?

Could it be even better to expose different hooks for multiple types of termination signals/events? Just a thought.

mrunalp: `post-stop` means after the container is stopped. What you are describing sounds more like a `pre-stop` hook. 

> Could it be even better to expose different hooks for multiple types of termination signals/events?

Could you elaborate?

LK4D4: I think we will pass state of stopped container to hook, with signal. So hook could decide itself what to do for different signals.

crosbymichael: the hook should get the exit status of the container, not necessarily the signal that make it stop

mrunalp: We will probably have to preserve the state before the `post-stop` hooks are done.

mrunalp:
Any more thoughts for these hooks?

wking:
On Wed, Jul 01, 2015 at 01:55:54PM -0700, Mrunal Patel wrote:
> Pushed update to specify directories for hooks…

I'd rather not specify well-known directories for hooks, since that
seems like it may be tricky to maintain.  Who installs the hooks?
Most users will be running some high-level tool and not runc directly.
Do they put their hooks somewhere else, and have the high-level tool
copy them into the right spot (this sounds wasteful)?  Do they push
the hooks directly to the runc locations (now you have to explain to
the user that you're using runc under the hood)?

To make it easy to distribute hooks, I'd rather have relative hook
paths be based on the config's directory (like root.path).  But I'd be
ok with “check through a set of well-known directories for hooks
matching the relative path, and make the config's directort the last
well-known directory in that list”.

In any case, I feel like allowing absolute-paths for hooks makes sense
and is independent of how we decide to handle relative paths.

jonboulle:
On Wed, Jul 8, 2015 at 4:14 AM, W. Trevor King <notifications@github.com>
wrote:

> I'd rather not specify well-known directories for hooks, since that
>

+1, location of hooks should be at implementation's discretion

mrunalp:
Updated to take out the hooks directory paragraph.

wking:
On Wed, Jul 08, 2015 at 09:47:38AM -0700, Mrunal Patel wrote:
> Updated to take out the hooks directory paragraph. 

Hmm.  Will runc take an argument or a RUNC_HOOK_PATH environment
variable?  Will it just search the PATH?  Are we looking for the hooks
in the host filesystem?  In the container's filesystem?

Since this all seems complicated, I went back and reread
@eyakubovich's request for motivation [1] and your response [2].  If
the pre-start and post-stop hooks are executed in the same namespace
as the main process, I still don't see why users who need them can't
bundle them into a wrapper script.  With something like:

  $ cat rootfs/bin/wrap.sh
  #!/bin/sh
  fix-mounts arg1 arg2 &&
  setup-network eth0 &&
  "$@"
  cleanup -f --exit-code "$?"

You could:

  # runc wrap docker --daemon

or some such to get full control over your setup/teardown without any
spec-side complication.  I'm not sure how a pre-exec hook that has
joined all the init namespaces would be mounting namespace fds on the
host anyway.  So:

* We should decide if the hook-like functionality is executed inside
  or outside the container.
* If it's inside, I suggest we just drop hooks from the spec in favor
  of wrapper scripts.
* If it's outside, I suggest we interpret hooks like any other
  host-side executable (absolute paths are absolute, relative paths
  are relative to $PATH), with an additional convenience fallback that
  we expand them from the config directory if we can't find a match in
  PATH.  That's like appending ‘.’ to the PATH when launching the hook
  processes, but keeping the original PATH for any children of the
  hook processes.

[1]: https://github.com/opencontainers/specs/pull/34#issuecomment-117334152
[2]: https://github.com/opencontainers/specs/pull/34#issuecomment-117345662

mrunalp:
@wking The hooks won't be run in the container's namespaces. They will be run from the host and they would get the state data including the container pid, so they could join the namespaces that they want to.
The hooks will also be run from the host's filesystem. If we require the path to the hooks to be absolute then there is no need for setting a special PATH variable. We could also just use the default PATH to search for the hooks when it is not absolute.

I will clarify this in the doc if there is no objection.
Thanks!

mrunalp:
Pushed update with clarification about where the hooks will be executed from.

mrunalp:
Fixed.

mrunalp:
For post-stop I think it makes sense to log errors and continue executing the remaining hooks. Updated to reflect that in the document and moved the previous generic section to pre-start section.

wking:
On Wed, Jul 08, 2015 at 11:45:09AM -0700, Mrunal Patel wrote:
> I will clarify this in the doc…

It looks better to me now :).

My main concern with host-PATH lookups that don't include the root
container-bundle directory (where the config file lives) is that
you're injecting additional unversioned dependencies.  On the one
hand, using host-installed software makes it easier for the host to
trust the hooks to not abuse their outside-the-container privileges.
On the other hand, it means you need to be able to count on the weak
API-compatibility link between the argument syntax that container
author wrote for and the version of a tool that the host sysadmin has
actually installed (assuming they've installed it at all :p).  Since
you're exposing your whole filesystem to arbitrary args anyway, folks
will already be able to do things like:

  "hooks" : {
    "pre-start": [
      ["rm", "-rf", "/"]
    ]
  }

I think it makes more sense to include the root container-bundle
directory in the lookup chain, which will give the container author a
say in which tool is run.  But for folks like me that are only running
trusted containers (and not offering container-running-as-a-service),
I guess:

  $ PATH=".:$PATH" runc

is easy enough ;).

Another thing to consider here is whether the hook-block should be
platform-specific or not.  I expect it should be a single Hooks type
that's used as a field in the per-platform types (e.g. Linux.Hooks,
Windows.Hooks, …).  And then handle inconsistencies within the
platform in the hook-executables themselves?

mrunalp:
> I think it makes more sense to include the root container-bundle directory in the lookup chain, which will give the container author a say in which tool is run

+1. I like that idea. I will add it if no one has any objections.

mrunalp:
> Another thing to consider here is whether the hook-block should be platform-specific or not. I expect it should be a single Hooks type that's used as a field in the per-platform types (e.g. Linux.Hooks, Windows.Hooks, …). And then handle inconsistencies within the platform in the hook-executables themselves?

Right now we can keep them in the generic section and then move them around if required.

mrunalp:
> Do we tear down the init-ed container on hook error? E.g. unmounting mounts, closing namespaces,
Yes, we do cleanup on any pre-start failure.

wking:
On Wed, Jul 08, 2015 at 02:03:02PM -0700, Mrunal Patel wrote:
> > Do we tear down the init-ed container on hook error?
> > E.g. unmounting mounts, closing namespaces,
>
> Yes, we do cleanup on any pre-start failure.

Sounds good to me.  Can we mention that in the docs?

mrunalp:
Updated.

wking:
On Thu, Jul 09, 2015 at 04:24:08PM -0700, Mrunal Patel wrote:
> Updated.

Looks good to me :).  I left a tiny typo comment in the diff, but I
can't think of anything else…

mrunalp:
Any new comments here? @crosbymichael @philips @LK4D4

runtime.md
```diff
@@ -15,3 +15,33 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over a well-known fd, so the hooks could get the information they need to do their work.
```
philips: just call it stdin?

mrunalp: done.

mrunalp:
I have updated the PR as per the discussion today. @philips @LK4D4 @vmarmol PTAL

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+The path to the hooks could be absolute or relative and are executed from the host's filesystem. If they are relative, then they are searched for within the PATH environment variable on the host.
+
+The hooks are executed from the container's root filesystem by default, unless overriden in the configuration.
```
LK4D4: Didn't undestand this. Like `PATH=rootfs:$PATH` or like we're calling it after `chroot rootfs`? Also how it can be overridden in configuration?

mrunalp: This isn't nicely worded. I have pushed an update.

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+The path to the hooks could be absolute or relative and are executed from the host's filesystem. If they are relative, then they are searched for within the PATH environment variable on the host.
```
LK4D4: I don't think that it called `relative` :) relative can be `../x` or something like this. But I'm not sure how call-by-name is really called.

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
...
+
+The working directory for the hooks is set to the container's root filesystem unless overriden in the configuration.
+
+### Pre-start
+The pre-start hooks are called after the container process is started but before the user supplied command is executed. It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
+
+If a hook returns a non-zero exit code, then an error is returned to the caller and the container is torn down.
```
LK4D4: Is stderr source of info about error?

runtime.md
mrunalp: Updated :)

runtime.md
mrunalp: I think it is up to the implementation to do what it wants to with the stderr; whether to log it or encapsulate it in the returned error. The critical aspect here is that the container start up fails if any hook returns non-zero.

LK4D4: Probably we should standardize this. Because it makes sense for implementation to know error text in addition to status code.

mrunalp: Updated.

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
...
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+The path to the hooks could be absolute or just the name of the file. Hooks are executed from the host's filesystem. If the path to a hook isn't absolute then it is searched for within the PATH environment variable on the host.
+
+The working directory for the hooks is set to the container's root filesystem unless overriden in the configuration.
+
+### Pre-start
+The pre-start hooks are called after the container process is started but before the user supplied command is executed. It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
+
+If a hook returns a non-zero exit code, then an error including the exit code and the stderr is returned to the caller and the container is torn down.
```
LK4D4: Why don't mention `stderr`?

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+The path to the hooks could be absolute or just the name of the file. Hooks are executed from the host's filesystem. If the path to a hook isn't absolute then it is searched for within the PATH environment variable on the host.
```
wking: I think this is still missing [“lookup relative to the config-file's directory”][1] (motivation [here][2]).  How about:

> Hooks are executed from the hosts's filesystem and environment.  Hook paths can be:
>
> * an absolute path,
> * a bare command name (without path separators), in which case the command is searched for within the `PATH` environment variable, or
> * a relative path, in which case the command is found relative to the directory containing the config file.  Relative paths cannot reference parent directories (i.e., no `../…`).

That way I can have (with paths like `hook/pre-start.sh`):

    .
    ├── config.json
    ├── hook
    │   ├── post-stop.sh
    │   └── pre-start.sh
    └── rootfs

That way I can have (with paths like `./pre-start.sh`):

    .
    ├── config.json
    ├── post-stop.sh
    ├── pre-start.sh
    └── rootfs

I'm also not sure how any of the non-bare-command cases will be handled in a cross-OS way.  Do we require slash-separators (vs. Windows with it's backslashes)?  How do folks handle Windows drives?  Do we need OS-specific hook sections so folks can use their native path syntax (and possibly different commands / arguments)?

[1]: https://github.com/opencontainers/specs/pull/34#issuecomment-119729756
[2]: https://github.com/opencontainers/specs/pull/34#issuecomment-119714814

runtime.md
mrunalp: @LK4D4 I think I did mention it :)
"...including the exit code and the stderr.."

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
...
+
+The working directory for the hooks is set to the container's root filesystem unless overriden in the configuration.
```
wking: Is this [“the bundle directory”][1] (e.g. `/path/to/bundle`) or [“the container's root filesystem”][2] (e.g. `/path/to/bundle/rootfs`)?  I'd guess it's the latter, but to make things very clear I'd suggest something like:

    … the container's [root filesystem][rootfs] unless overriden in the configuration.

    [rootfs]: config.md#root-configuration

[1]: https://github.com/opencontainers/specs/blob/5c98addb2b07a600c57e3f7229041103abb5c4af/bundle.md
[2]: https://github.com/opencontainers/specs/blob/5c98addb2b07a600c57e3f7229041103abb5c4af/config.md#root-configuration

runtime.md
mrunalp: I like the idea of having relative paths. Incorporated your input. Thanks!

runtime.md
```diff
@@ -15,3 +15,46 @@ Runs a process in a container. Can be invoked several times.
...
+The pre-start hooks are called after the container process is started but before the user supplied command is executed. It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
+
+If a hook returns a non-zero exit code, then an error including the exit code and the stderr is returned to the caller and the container is torn down.
+
+### Post-stop
+The post-stop hooks are called after the container process is stopped. Cleanup or debugging could be performed in such a hook.
+
+If a hook returns a non-zero exit code, then an error is logged and the remaining hooks are executed.
+
+```
```
wking: You probably want a new header here.  `### Example`?  `config.md` uses `*Example*`, but I think it's worth pulling this out of the post-stop section.

runtime.md
mrunalp: I think having paths as string handles Windows.

wking: On Wed, Jul 22, 2015 at 04:26:14PM -0700, Mrunal Patel wrote:
> I think having paths as string handles Windows.

Ah, I was thinking that we might have fat containers that are
simultaneously compatible with several OSes.  For example:

  .
  ├── config.json
  └── rootfs
      ├── linux
      ├── os-x
      └── windows

But looking over the spec again, it seems like the separate Linux /
Windows / … sections are just for easy validation, and not a way to
setup things like:

  {
    …
    "linux": {
      "root": {
        "path": "rootfs/linux",
        …
      },
      …
    },
    "windows": {
      "root": {
        "path": "rootfs\\windows",
        …
      },
      …
    }
  }

Since that's not what we're shooting for, there's no need for per-OS
hook sections or cross-OS path-separator conventions.

runtime.md
```diff
@@ -15,3 +15,51 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+Hooks are executed from the host's filesystem. Hook paths can be:
+* An absolute path,
+* a file name (without path separators), in which case the file is searched for within the PATH environment variable, or
```
wking: In my [earlier suggestion][1], I'd also changed “Hooks are executed from the host's filesystem” to “Hooks are executed from the host's filesystem and environment” because I'd dropped the host-environment reminder from your earlier:

> … within the PATH environment variable on the host.

[1]: https://github.com/opencontainers/specs/pull/34#discussion_r35275069

runtime.md
```diff
@@ -15,3 +15,51 @@ Runs a process in a container. Can be invoked several times.
 Not sure we need that from oc cli. Process is killed from the outside.
 
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+Hooks are executed from the host's filesystem. Hook paths can be:
+* An absolute path,
```
wking: You need a blank line before the start of the list for proper Markdown rendering.

runtime.md
```diff
@@ -15,3 +15,51 @@ Runs a process in a container. Can be invoked several times.
...
+* a file name (without path separators), in which case the file is searched for within the PATH environment variable, or
+* a relative path, in which case the file is found relative to the directory containing the config file. Relative paths cannot reference parent directories (i.e., no ../ ...).
```
wking: You probably want backticks, and may want an ellipsis character after the slash: `../…`.  If you don't want non-ASCII characters, I'd probably just say “no `../`”.

runtime.md
wking: On Wed, Jul 22, 2015 at 04:47:37PM -0700, W. Trevor King wrote:
> Ah, I was thinking that we might have fat containers that are
> simultaneously compatible with several OSes.

And thinking this over some more, I guess we could still be shooting
for this goal, just via several per-platform configs, instead of a
single, multi-platform config:

  .
  ├── config-linux.json
  ├── config-os-x.json
  ├── config-windows.json
  └── rootfs
      ├── linux
      ├── os-x
      └── windows

The current path handling is fine, but I thought I'd post this
follow-up comment to avoid spreading my earlier confusion about if/how
fat containers would be supported.

wking:
Recent commits (at least c64292ccd5) have summaries starting with
“ooks: …”, not “hooks: …”.  Too many rerolls?  ;)

LK4D4:
LGTM

lizf-os:
How about make it clear that there can be more than one pre-start/post-stop hooks, and those hooks are executed in what order, and happens to the remaining hooks if one hook returns error.

runtime.md
```diff
@@ -15,3 +15,52 @@ Runs a process in a container. Can be invoked several times.
...
 This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container. The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+Hooks are executed from the host's filesystem. Hook paths can be:
+
+* An absolute path,
+* a file name (without path separators), in which case the file is searched for within the PATH environment variable, or
+* a relative path, in which case the file is found relative to the directory containing the config file. Relative paths cannot reference parent directories (i.e., no ../).
```
philips: now that I think about this more lets just start with an absolute path and see if we need the other two. If we need them we can add them later.

mrunalp:
Added the code changes here.

runtime.md
LK4D4: +1
should be simpler

mrunalp: I don't mind that. done.

crosbymichael:
The struct still seems pretty complex.  Why have path and args, why not just Args like how the Process is in the spec?  argv[0] is the path.  Also why change the env format to a map when the Process spec is a []string?

I think it should just be 
```json
{
    "args": ["ip", "link", "add", "eth0", "blablabla"],
    "env": ["IP=192.168.1.5"]
}
```

I don't think we need dir or anything like that because it's something that we cannot guarantee because we want to invoke some hooks on the outside of the namespace and inside.  Most of the data the hook will car about comes via stdin with the container state.

wking:
On Tue, Jul 28, 2015 at 11:45:22AM -0700, Brandon Philips wrote:
> > @@ -15,3 +15,52 @@ Runs a process in a container. Can be invoked several times.
> > +Hooks are executed from the host's filesystem. Hook paths can be:
> > +
> > +* An absolute path,
> > +* a file name (without path separators), in which case the file is searched for within the PATH environment variable, or
> > +* a relative path, in which case the file is found relative to the directory containing the config file. Relative paths cannot reference parent directories (i.e., no ../).
>
> now that I think about this more lets just start with an absolute
> path and see if we need the other two. If we need them we can add
> them later.

That makes me concerned again about what the bundle-author can assume
about the host system (see my earlier motivation for in-bundle paths
[1]).  As it stands (with commit 8e319e10), the example calls for
/usr/bin/fix-mounts and /usr/sbin/cleanup.sh, neither of which exist
on my system.  With config-relative paths (or bundle-relative paths),
we could have:

  .
  |-- bin
  |   |-- cleanup.sh
  |   `-- fix-mounts
  |-- config.json
  `-- rootfs

which gives bundle authors a lot more flexibility (they can write
shell scripts or compile binaries appropriate to a given platform)
while also making life easier for bundle consumers (they don't have to
install fix-mounts locally).

I guess I can always come back and argue for this again when I
personally need to use a hook though…

[1]: https://github.com/opencontainers/specs/pull/34#issuecomment-119714814

mrunalp:
@crosbymichael updated.

crosbymichael:
LGTM

we can specifiy what the state looks like that is passed on STDIN after we, you know, figure out the state format in the other issues/prs

spec.go
```diff
@@ -61,3 +65,8 @@ type Platform struct {
 	// Arch is the architecture
 	Arch string `json:"arch"`
 }
+
+type Command struct {
```
crosbymichael: nit: maybe call this `Hook`

spec.go
```diff
@@ -15,6 +15,10 @@ type Spec struct {
 	Hostname string `json:"hostname"`
 	// Mounts profile configuration for adding mounts to the container's filesystem.
 	Mounts []Mount `json:"mounts"`
+	// Prestart is a list of commands to be run before the container process is executed.
+	Prestart []Command `json:"prestart"`
```
crosbymichael: These two fields should be embeded in a `Hooks` struct with the `json:"hooks"` tag

mrunalp:
@crosbymichael updated.

crosbymichael:
LGTM

runtime.md
```diff
@@ -14,4 +14,40 @@ Runs a process in a container. Can be invoked several times.
 
 Not sure we need that from runc cli. Process is killed from the outside.
 
-This event needs to be captured by runc to run onstop event handlers.
+This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container.
+The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
```
dqminh: should we document what's the expected format of the input ? (i.e. all of state.json)

runtime.md
```diff
@@ -14,4 +14,40 @@ Runs a process in a container. Can be invoked several times.
...
+
+Hook paths are absolute and are executed from the host's filesystem.
+
+### Pre-start
+The pre-start hooks are called after the container process is started but before the user supplied command is executed.
+It provides an opportunity to customize the container. In Linux, for e.g., the network namespace could be configured in this hook.
```
dqminh: i think we should also document the expected current context for script here (i.e. root directory). This should be before we change root right, so scripts can be placed on the host.

runtime.md
mrunalp: We will document that once we finalize the format which is being discussed in other issues.

runtime.md
mrunalp: See @crosbymichael comment above. We aren't guaranteeing the cwd for the hook. However, it gets the container rootfs path as part of the state, so it can switch to that directory if desired.

LK4D4:
@crosbymichael @mrunalp I think `Path` was okay. For example we want to invoke docker binary with argv[0] `setup-namespace`(this how reexec in docker works now). Also I think makes sense to be as close to `exec.Cmd` as possible.

crosbymichael:
@LK4D4 that works the same way, we can setup the command with argv0 as path

LK4D4:
@crosbymichael Maybe I missing something. So, I need to run `/usr/bin/docker` with `os.Args[0] == "setup-namespace"`, how can I do this with array of args?

crosbymichael:
@LK4D4 if argv0 is an abs path then just set it as the path.

LK4D4:
@crosbymichael Okay, but I still think that Args+Path more understandable, then "if first element is abs path, then argv[0] is actually argv[1]"

mrunalp:
@crosbymichael WDYT? Need to decide one way or the other :)

crosbymichael:
@mrunalp I think @LK4D4 is right, sorry for the back and forth

mrunalp:
@crosbymichael No problem ;) I think it makes sense to add Path, too. Updated. 
@LK4D4

LK4D4:
LGTM

runtime.md
```diff
@@ -14,4 +14,45 @@ Runs a process in a container. Can be invoked several times.
...
+This event needs to be captured by oc to run onstop event handlers.
+
+## Hooks
+Hooks allow one to run code before/after various lifecycle events of the container.
+The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+
+Hook paths are absolute and are executed from the host's filesystem.
+
+### Pre-start
+The pre-start hooks are called after the container process is started but before the user supplied command is executed.
```
crosbymichael: I would change this to `after the container's namespaces are created` and not say started because it's confusing with the hooks name as prestart

runtime.md
```diff
@@ -14,4 +14,45 @@ Runs a process in a container. Can be invoked several times.
 
 Not sure we need that from runc cli. Process is killed from the outside.
 
-This event needs to be captured by runc to run onstop event handlers.
+This event needs to be captured by oc to run onstop event handlers.
```
crosbymichael: I dont think this sentence makes sense

mrunalp: Unrelated to this PR. Might be a rebase issue.

runtime.md
mrunalp: How about
```
The pre-start hooks are called after the container process is spawned, but before the user supplied command is executed.
They are called after the container namespaces are created on Linux, so they provide an opportunity to customize the container.
In Linux, for e.g., the network namespace could be configured in this hook.

```

mrunalp:
@crosbymichael updated

crosbymichael:
LGTM

PR #50, opened 2015-07-01
Adds a section for user namespace mappings

mrunalp:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -118,6 +118,27 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
 
 rlimits allow setting resource limits. The type is from the values defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html). The kernel enforces the soft limit for a resource while the hard limit acts as a ceiling for that value that could be set by an unprivileged process.
 
+## Linux user namespace mappings
+
+```
+    "uidMappings": [
+        {
+            "from"  : 1000,
+            "to"    : 0,
+            "count" : 10
+        }
+    ],
+    "gidMappings": [
+        {
+            "from"  : 1000,
+            "to"    : 0,
+            "count" : 10
+        }
+    ]
+```
+
+uid/gid mappings describe the user namespace mappings from the host to the container. *from* is the starting uid/gid on the host to be mapped to *to* which is the starting uid/gid in the container and *count* refers to the number of ids to be mapped. The Linux kernel has a limit of 5 such mappings that can be specified.
+
 ## Security
 
 **TODO:** security profiles


config-linux.md
```diff
@@ -118,6 +118,21 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
 
 rlimits allow setting resource limits. The type is from the values defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html). The kernel enforces the soft limit for a resource while the hard limit acts as a ceiling for that value that could be set by an unprivileged process.
 
+## Linux user namespace mappings
+
+```
+    "userMappings": [
```
LK4D4: btw I wonder why we don't use capital-first camelcase.

LK4D4:
How about groupMappings?

config-linux.md
```diff
@@ -118,6 +118,21 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
...
+        {
+            "from"  : 1000,
+            "to"    : 0,
+            "count" : 10
```
LK4D4: Not sure about `count`, it called
```
The length of the range of user IDs
```
in man page.

config-linux.md
mrunalp: Just matching what we have in spec_linux.go. We could even go with user_mappings.

crosbymichael: in proper json you don't capitalize the first letter

config-linux.md
mrunalp: That's how we have it today in the spec. We could either go with how it is in the kernel or what we have in golang.

mrunalp:
Added gidMappings and changed from userMappings to uidMappings.

LK4D4:
LGTM

cbinion46511:
???

crosbymichael:
@mrunalp do you also want to change the types now that they are in the repo?

mrunalp:
@crosbymichael Yep, will update this in a few.

mrunalp:
@crosbymichael Actually spec_linux.go matches what we have here unless I am misreading it..

config-linux.md
```diff
@@ -118,6 +118,27 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
...
+    "gidMappings": [
+        {
+            "from"  : 1000,
+            "to"    : 0,
+            "count" : 10
+        }
+    ]
+```
+
+uid/gid mappings describe the user namespace mappings from the host to the container. *from* is the starting uid/gid on the host to be mapped to *to* which is the starting uid/gid in the container and *count* refers to the number of ids to be mapped. The linux kernel has a limit of 5 such mappings that could be specified.
```
jonboulle: The Linux kernel has a limit of 5 such mappings that can be specified.

mrunalp: Updated.

jonboulle: nit, you missed linux -> Linux

mrunalp: done.

wking:
I'm proposing changes to the field names in #63 and
opencontainers/runc#98.  If those branches look promising, we'll want
to reroll this branch to match the new names.

crosbymichael:
LGTM

PR #55, opened 2015-07-03
bundle.md: various updates to latest spec

jonboulle:
- Move to single root filesystem, as the configuration specification now
  only supports a single process
- Clarify that signatures are just another kind of content directory
- Cross-reference configuration specification for config.json

I think my understanding w.r.t. the rootfs stuff is correct, but I could be totally off-base on this; if so it's not at all clear to me what the purpose of multiple rootfs directories is.

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -12,20 +12,19 @@ A standard container bundle is made of the following 3 parts:
 
 # Directory layout
 
-A Standard Container bundle is a directory containing all the content needed to load and run a container. This includes its configuration file, content directories, and cryptographic signatures. The main property of this directory layout is that it can be moved as a unit to another machine and run the same container.
+A Standard Container bundle is a directory containing all the content needed to load and run a container. This includes its configuration file (`config.json`) and content directories. The main property of this directory layout is that it can be moved as a unit to another machine and run the same container.
 
-One or more *content directories* may be adjacent to the configuration file. This at least includes the root filesystem (referenced in the configuration by the *rootfs* field) and other related content (signatures, other configs, etc.). The interpretation of these resources is specified in the configuration.
+The syntax and semantics for `config.json` are described in [this specification](config.md).
+
+One or more *content directories* may be adjacent to the configuration file. This must include at least the root filesystem (referenced in the configuration file by the *root* field) and may include other related content (signatures, other configs, etc.). The interpretation of these resources is specified in the configuration.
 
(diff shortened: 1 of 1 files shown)


bundle.md
```diff
@@ -12,20 +12,19 @@ A standard container bundle is made of the following 3 parts:
...
 ```
 /
 !
 -- config.json
 !
---- rootfs1
+--- rootfs
 !
---- rootfs2
+--- sigs
```
stevvooe: `signatures`. No need for brevity ;).

bundle.md
```diff
@@ -12,20 +12,19 @@ A standard container bundle is made of the following 3 parts:
...
+
+One or more *content directories* may be adjacent to the configuration file. This must include at least the root filesystem (referenced in the configuration file by the *root* field) and may include other related content (signatures, other configs, etc.). The interpretation of these resources is specified in the configuration.
 
 ```
 /
 !
 -- config.json
 !
---- rootfs1
+--- rootfs
```
stevvooe: Do we name this `rootfs` by convention or is the the specified name for this directory? Do we need to clarify that above? Perhaps a note that explains these names are a convention is required.

jonboulle: @stevvooe If my understanding is correct and bundles only support a single rootfs directory then it seems to me like the natural next step would just be standardising/mandating its location at a well-known path.

jonboulle: (in the meantime I'll add a note on convention, that would clarify "sigs" as well ;-)

stevvooe:
@jonboulle Some background here: I think the original idea behind multiple root filesystems was to be able to support POD-like functionality. I am not sure where the conversation on PODs has headed, so it's possible we're going in a different direction.

crosbymichael:
LGTM

mrunalp:
LGTM

jonboulle:
Thanks for the merge! I'm still curious on thoughts on https://github.com/opencontainers/specs/pull/55#discussion_r33834913 - would it be simpler to just mandate a well-known location for the rootfs given there can be only one?

mrunalp:
I think that keeping it configurable would help later if we support multiple of them down the line.

stevvooe:
@jonboulle I agree with @mrunalp. However, we should have a list of strongly encouraged conventions:

```
rootfs
signatures
```

PR #58, opened 2015-07-06
Add memory swappiness to linux spec

lizf-os:
I'm not able to start runc on ubuntu 14.04 with kernel version 3.13, and it's because runc fails to set swappiness.

This patch adds memory.swappiness to the spec, and then I'll send out patches to support setting swappiness in runc so we'll be able to leave swappiness as the system default value.


crosbymichael:
LGTM

PR #61, opened 2015-07-07
Fix Windows path escaping in example mount JSON

titanous:


philips:
@titanous ha, what a mess of escaping. Thank you!

philips:
@titanous actually, since this is in a code-block shouldn't all of the escaping be removed?

titanous:
@philips This is JSON escaping, not markdown escaping.

philips:
@titanous gah, you are right! Thanks!

PR #63, opened 2015-07-08
spec_linux.go: Rename IDMapping fields to follow syscall.SysProcIDMap

wking:
'From' and 'To' are potentially ambiguous for a one-to-one map like
this, and there's already an established name convention in
SysProcIDMap [1].  This commit removes the mental overhead of two
separate naming schemes for the same information.  I'd like to drop
IDMapping entirely in favor of SysProcIDMap, but SysProcIDMap doesn't
give the JSON hints we need for (de)serializing.

[1]: https://golang.org/pkg/syscall/#SysProcIDMap

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -123,16 +123,16 @@ rlimits allow setting resource limits. The type is from the values defined in [t
 ```
     "uidMappings": [
         {
-            "from"  : 1000,
-            "to"    : 0,
-            "count" : 10
+            "hostID": 1000,
+            "containerID": 0,
+            "size": 10
         }
     ],
     "gidMappings": [
         {
-            "from"  : 1000,
-            "to"    : 0,
-            "count" : 10
+            "hostID": 1000,
+            "containerID": 0,
+            "size": 10
         }
     ]
 ```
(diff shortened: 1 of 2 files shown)


mrunalp:
I don't mind this change as it keeps it the same as SysProcIDMap.

mrunalp:
@crosbymichael @LK4D4 WDYT?

wking:
Rerolled to also update the docs that just landed with #50.

philips:
seems fine to me, crosby has final say

lizf-os:
+1

crosbymichael:
@philips i do not! ;)

I vote for whatever is most explicit and easy for people to understand.  If everyone thinks this is more clear then SGTM

mrunalp:
Looks like we all agree :) LGTM

wking:
On Thu, Jul 09, 2015 at 11:45:41AM -0700, Mrunal Patel wrote:
> Merged #63.

You probably want to merge opencontainers/runc#98 too (or at least,
review it, so I can fix any issues, and then we can merge it ;).  Note
that besides updating runc to match the more-explicit names, that PR
also fixes a semantic bug (notes in wking/runc@41979ca).

PR #64, opened 2015-07-08
spec: linux: add support for the PIDs cgroup

cyphar:
Add support for the PIDs cgroup as a cgroup resource constraint in the
Linux container specification. Since PIDs are a real resource, we need
to support the ability to limit them.

The PIDs cgroup subsystem is available in Linux 4.3+.

I realised while implementing opencontainers/runc#58 that if we want to
have PIDs support in the `config.json` we need to update the spec too.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -128,7 +128,7 @@ Next parameters can be specified:
 ## Control groups
 
 Also known as cgroups, they are used to restrict resource usage for a container and handle
-device access.  cgroups provide controls to restrict cpu, memory, IO, and network for
+device access.  cgroups provide controls to restrict cpu, memory, IO, pids and network for
 the container. For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
 ## Sysctl
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -126,6 +126,11 @@ type CPU struct {
 	Mems string `json:"mems"`
 }
 
+type Pids struct {
+	// Maximum number of PIDs. A value < 0 implies "no limit".
+	Limit int64 `json:"limit"`
+}
+
 // Network identification and priority configuration
 type Network struct {
 	// Set class identifier for container's network packets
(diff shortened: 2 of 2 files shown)


cyphar:
Please note that Linux 4.3 still hasn't hit, so I'd recommend abstaining from merging this until we actually hit 4.3-rc1 (which will be in 6 weeks or so), unless we're okay with merging a no-op configuration feature.

lizf-os:
LGTM, but let's wait until upstream kernel merges the pids controller.

mrunalp:
Thanks @cyphar. We will wait for the upstream.

philips:
I am marking this for the post-draft release in hopes it will be merged with no problems by then :)

spec_linux.go
```diff
@@ -127,6 +127,13 @@ type CPU struct {
 	Mems string `json:"mems"`
 }
 
+type Pids struct {
+	// Maximum number of PIDs. A value <= 0 implies "no limit".
+	// While technically 0 means "no tasks allowed in the cgroup", this does
+	// not make sense in the context of containers.
```
mikedanese: If the pid limit is not immutable, it might make sense to set to 0 for some use cases. Should we be restricting the input domain in the spec?

cyphar: I guess that's a fair point. It does make it slightly annoying because the default value will have to be `-1` (which means we can't do `new(spec.Linux)` without modifying it each time). I'll fix this later.

cyphar: Actually, I'm not sure that you're right. In order for the PID limit to make sense, you have to have a process in the cgroup. Nothing stops you from attaching more processes to a cgroup, so it's illogical to claim that a PID limit of 0 has any *real* use. Combine that with the annoyance of not being able to do `new(spec.Linux)` or `spec.Linux{}`, and it's clear that a limit of 0 is not useful.

cyphar:
The PIDs changes have been merged into Linus' tree as of torvalds/linux@	8bdc69b764013a9b5ebeef7df8f314f1066c5d79.

vbatts:
nicely

philips:
merged upstream, lgtm.

vbatts:
Will some of these features set platform version boundaries?

philips:
@vbatts I feel it is sort of up to the runtime to decide whether it will/won't run something based on its configuration.

vbatts:
Fair

LK4D4:
So, cool.
LGTM

Issue #73, opened 2015-07-23
Multi platform container


sladage:
From what I understood from the specs is that a container targets a single platform? If so, wouldn't it be better to allow for different configurations depending on the platform, all packaged into one container?

wking:
On Thu, Jul 23, 2015 at 10:12:43AM -0700, Stefan Ladage wrote:
> From what I understood from the specs is that a container targets a
> single platform? If so, wouldn't it be better to allow for different
> configurations depending on the platform, all packaged into one
> container?

I was recently confused about this too, but I think the current spec
*almost* allows this already.  We just need to remove the requirement
for the config.json name [1].  The ‘runc’ command already supports
launching a container from an alternative spec [2], although I haven't
been able to find clear docs describing this functionality.

However, this still requires the launcher to figure out which of
several configs to use for their platform.  I think it might make more
sense to replace the default config.json with a config/ directory.
The launcher (e.g. runC) would walk that directory to find potential
config files (anything ending in ‘.json’?) and then figure out which
of the found spec files was the best platform match (erroring out if
there were no compatible specs).

To keep the config DRY, it would be nice if there was also a way to
share code between configs.  For example, inheritance:

  {
    …
    "parents": [
      "../base.json",
      "./amd64.json"
    ],
    …
  }

or embedding (e.g. with JSON Schema's $ref [3]:

  {
    …
    "linux": {
      "$ref": "./amd64.json#linux"
    }
    …
  }

Since inheritance and embedding are potentially complicated, I'm fine
punting on those for now, but I do think that a walkable config
directory is a clear win for multi-platform fat containers.

Of course, keeping the option to manually specify a config file when
invoking runC is good too.  For example, you may have one config to
launch your indended application, and another with something like:

  {
    "version": "pre-draft",
    "parents": ["./main.json"],
    "process": {
      "args": ["bash"]
    }
  }

that's identical to the app config but just launches a shell.  For the
default platform-matching logic, you'd want a way to break ties like
this.  I'd suggest either C-locale sorts of the relative path from
config/ (earlier paths win) or an explicit “"priority": <number>”
setting (higher priorities win).

[1]: https://github.com/opencontainers/specs/blob/5c98addb2b07a600c57e3f7229041103abb5c4af/config.md#configuration-file
[2]: https://github.com/opencontainers/runc/blob/dfc8d4aa26f69dfea40f4d5861c01d6291583e76/run.go#L69
[...]

PR #74, opened 2015-07-24
Multi-platform support

sladage:
Allowing for a single container to contain multiple platform-specific configurations.

### DIFF
--- a/config.md
+++ b/config.md
@@ -137,7 +137,7 @@ For Linux-based systems the user structure has the following fields:
 
 ## Platform-specific configuration
 
-The *platforms* array should contain a list of platforms which the container supports.
+The *platforms* array should contain a list of platforms which the container supports. Only the main configuration file `config.json` may contain a *platforms* section.
 
 * **os** (string, required) specifies the operating system family this image must run on. Values for os must be in the list specified by the Go Language document for [`$GOOS`](https://golang.org/doc/install/source#environment).
 * **arch** (string, required) specifies the instruction set for which the binaries in the image have been compiled. Values for arch must be in the list specified by the Go Language document for [`$GOARCH`](https://golang.org/doc/install/source#environment).
--- a/config.md
+++ b/config.md
@@ -137,14 +137,20 @@ For Linux-based systems the user structure has the following fields:
 
 ## Platform-specific configuration
 
+The *platforms* array should contain a list of platforms which the container supports.
+
 * **os** (string, required) specifies the operating system family this image must run on. Values for os must be in the list specified by the Go Language document for [`$GOOS`](https://golang.org/doc/install/source#environment).
 * **arch** (string, required) specifies the instruction set for which the binaries in the image have been compiled. Values for arch must be in the list specified by the Go Language document for [`$GOARCH`](https://golang.org/doc/install/source#environment).
+* **config** (string, optional) specifies the config file containing platform-specific overrides and/or platform-specific sections. File path must be relative to the main config (`config.json`) file.
 
 ```json
-"platform": {
-    "os": "linux",
-    "arch": "amd64"
-}
+"platforms": [
+    {
+        "os": "linux",
+        "arch": "amd64",
+        "config": "linux64.json"
+    }
+]
 ```
 
-Interpretation of the platform section of the JSON file is used to find which platform-specific sections may be available in the document. For example, if `os` is set to `linux`, then a JSON object conforming to the [Linux-specific schema](config-linux.md) SHOULD be found at the key `linux` in the `config.json`.
+In this case a JSON object conforming to the [Linux-specific schema](config-linux.md) SHOULD be found at the key `linux` in the `linux64.json` config file.
--- a/spec.go
+++ b/spec.go
@@ -5,8 +5,8 @@ package specs
 type Spec struct {
 	// Version is the version of the specification that is supported.
 	Version string `json:"version"`
-	// Platform is the host information for OS and Arch.
-	Platform Platform `json:"platform"`
+	// Platforms supported by the container.
+	Platform []Platform `json:"platforms"`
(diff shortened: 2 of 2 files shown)


wking:
Ah, I see you didn't indent it to be recursive ;).  I think this
explicit-child approach makes it too hard to handle per-combination
tweaks (see [1]), but I don't have a better idea that doesn't make
all-combination tweaks difficult.  I'm fine waiting until someone does
have a better idea (to avoid putting work into a likely dead end), and
until then think we should just walk a config directory looking for
the best platform match (so change the bundle spec) without altering
the config spec itself (see [1]).

But if we do land this approach, I'd suggest at least making the
"platforms" behavior recursive.  Special-casing the top-level config
will just make life harder on implementors and container authors (and
I don't see an upside to the special case).

[1]: https://github.com/opencontainers/specs/issues/73#issuecomment-124722204

sladage:
I had not thought about making it recursive, that could actually work. Or if we don't make it recursive, it could load all matching configs. (each platform section being more specific)

As for the walking trough a config dir solution, although simple, it could get messy having to update all configs which mostly contain the same info anyway. (not very efficient either)

But yeah, let's see if someone has a better idea.

wking:
On Fri, Jul 24, 2015 at 02:28:27PM -0700, Stefan Ladage wrote:
> As for the walking trough a config dir solution, although simple, it
> could get messy having to update all configs which mostly contain
> the same info anyway. (not very efficient either)

Yeah, but Dockerfile has had reasonable success with the KISS
approach, and folks can always add external tooling like [1] if they
want to build those repetitive configs from a DRYer template.

[1]: https://github.com/wking/dockerfile/blob/269a64f1f54b8f5077bbf27d6b93eabffdb4fa9b/build.sh#L219-L228

sladage:
True, the simple approach is generally the better approach,just less flexible in this case. Even if we don't need the flexibility at this moment, we might in the future. Having said that, I actually like the config dir idea. We just have to be careful that we don't, in an attempt to keep things simple, end up with a specification that is too inflexible. Not that I think it is an issue here, just something to keep in mind.

wking:
On Fri, Jul 24, 2015 at 03:18:41PM -0700, Stefan Ladage wrote:
> … I actually like the config dir idea…

I'll write up a PR.

sladage:
Looks good, although, I'm not entirely sure where you were going with the version check. Also, the naming convention should be made more clear I think. Like:
```
[platform]_[arch]_[???].json

linux_x64.json
linux_arm64.json
```

Looking at both solutions I can't really say one is more simple than the other though. :) It just moves the problem from the config file spec to the directory layout spec.

wking:
On Sun, Jul 26, 2015 at 04:23:22AM -0700, Stefan Ladage wrote:
> Looks good, although, I'm not entirely sure where you were going
> with the version check.

In #76?  I'm not sure if I can add much to clarify my commit-message
comments (repeated in the initial PR post), but the version ranking
lets you write a bundle compatible with both 1.x and 2.x of this spec
(by supplying both a v1.0 and a v2.0 config for a given platform in
your config/ directory).  Does that help?

> Also, the naming convention should be made more clear I think. Like:
> ```
> [platform]_[arch]_[???].json
> 
> linux_x64.json
> linux_arm64.json
> ```

Again, I think you're talking about #76.  With that approach, the
filenames (under config/) don't matter at all.  Use whatever you like
best.  If folks want to pick a convention, that's fine with me, but I
don't see a need to suggest something in the spec.  If we do want to
suggest something, I'll happily merge a PR against my config-directory
branch ;).

> Looking at both solutions I can't really say one is more simple than
> the other though. :) It just moves the problem from the config file
> spec to the directory layout spec.

The approach in #76 makes it possible to support multiple,
incompatible versions (e.g. 1.x and 2.x) at the same time.  It also
doesn't require any config-file merging, so the existing runC can
*already* launch containers from such a bundle (you just have to do
the config-choosing work by hand until we get the proposed mpm tool
[1]).  And the only problem we need to solve for mpm is matching
platforms (we don't need to worry about combining the parent spec with
the child specs).  Of course, “combining the parent spec with child
specs” (or handling $ref replacement) can be *useful* for reducing
duplication, but as I tried to show earlier [2], I'm not convinced
that either is sufficiently powerful to be worth the complication.  I
do think *automatic platform matching* is sufficiently powerful to be
worth the complication though, and that's the only thing #76 adds.

And as I tried to say in [3], I doubt we'll be able to leap
immediately to consensus solutions for all the issues facing container
authoring and management.  I think we need an ecosystem of tools that
each address a single facet of the problem (runC for launching a
container from a given config, mpm for finding the config that best
matches the current platform and runC implementation, …), and we need
a clearer framework to flexibly combine those tools.  Then I can try
my way, and if you're not convinced, you can try your way, bundle
authors can choose whichever works for them, bundle consumers can push
back if installing the required tools is too troublesome, etc., etc.
And we'll eventually consolidate around a smaller set of great tools.
If we end up with something like the Mercurial / Git divide while a
particular pair of tools takes a few years to sort out the popularity
contest, that's fine.

[1]: https://github.com/opencontainers/specs/pull/76#issuecomment-124913227
[2]: https://github.com/opencontainers/specs/issues/73#issuecomment-124722204

wking:
On Mon, Jul 27, 2015 at 03:19:43PM -0700, W. Trevor King wrote:
> And as I tried to say in [3], …
> …
> [1]: https://github.com/opencontainers/specs/pull/76#issuecomment-124913227

Oops, should have been [1].

philips:
Can someone write-out a use case for this? What is an example use case for a bundle that supports multiple archs? Why is having two bundles too complicated?

philips:
I am going to close this out and repeat my questions on #73 until we decide whether this is desirable or not.

PR #75, opened 2015-07-24
Add prestart/poststop hooks to the spec

mrunalp:
Related to #34 


wking:
On Fri, Jul 24, 2015 at 12:04:06PM -0700, Mrunal Patel wrote:
> Related to #34 

Why not just add this to #34?  I don't see a need to land the docs and
spec changes in independent PRs ;).

philips:
In fact I think we should land docs and spec code at the same time. Makes it easier to read the git history and understand what/why together.

Issue #11, opened 2015-06-25
bundle: filesystem metadata format


philips:
@stevvooe and I caught-up in person about our digest discussion and the need for serialize file-system metadata. If you want to read my attempt it is found here: https://github.com/opencontainers/specs/issues/5#issuecomment-114208979

Problem: a rootfs for a container bundle sitting on-disk may not reflect the exact intended state of the bundle when it was copied to its current location. Possible causes might include: running on filesystems with varying levels of metadata support (nfs w/o xattrs), accidental property changes (chown -R), or purposeful changes (xattrs added to enforce local policies).

Obviously the files contents will be identical so that isn't a concern.

Solution: If we hope to create a stable digest of the bundle in the face of these likely scenarios we should store the intended filesystem metadata into a file itself. This can be done in a variety of ways and this issue is a place to discuss pros/cons. As a piece of prior-art @vbatts has implemented https://github.com/vbatts/tar-split and we have the linux package managers with tools to verify and restore filesystem metadata from a database with `rpm -a --setperms` and `rpm -V`.

stevvooe:
After some thought, we need a format that does the following:

1. A _manifest_ enumerates files in a bundle.
    1. Provides a content hash.
    2. Provides a path.
    3. Provides a file type.
    4. Provides standard file mode.
    5. Provides xattr.
    6. Provides an extension mechanism. 
         1. Geared towards multiple OS support.
         2. Not infinitely extendable but should be easy to add new field.
2. Bundle contents attributes can be reset to contents of file manifest.
    1. Bundle is scanned and any differences from manifest are rectified.
    2. Unames/Gnames/Uid/Gid can be mapped during "reset".
3. Bundle contents can be verified against manifest.
    1. Content hash can be checked.
    2. Attributes can be checked.
        1. Certain attributes can be checked against machine-local mapping (uid/gid, etc.).
    3. Manifest can optionally be signed.

Meeting the above three use cases with this format puts us above the test for Cost problems with the ideas from #5. Requirement 2 above has a common use case, by avoiding placing unreasonable requirements on transports. Requirement 3 above gives us the functionality of #5 with the extra benefits of Requirement 2.

@shykes @philips @crosbymichael

philips:
@stevvooe I am having a hard time parsing this sentence: "Meeting the above three use cases with this format puts us above the test for Cost problems with the ideas from #5."

I agree with all three needs overall despite my confusion above.

stevvooe:
[...]

PR #76, opened 2015-07-25
bundle: Add platform-matching config directories

wking:
To allow a single bundle to be cross-platform.  I've tried to add
enough context to motivate the additional complexity without adding so
much that the context distracts from the spec changes.

The tie-breaking version ranking (step (2) for picking the best config
file) also make it possible to write backwards-compatible bundles that
still take advantage of new features when possible.  For example,
placing v1.0, v1.6, and v2.0 configs in the same directory would let
you run the same container on all v1.* and v2.* runtimes while still
letting you take advantage of v1.6 and v2.0 features for compatible
runtimes.  After explaining the multi-platform advantages, the
multi-version example seemed obvious enough to not be worth cluttering
the bundle.md description, but commit-message space is cheap so I'm
talking about it explicitly here ;).

There was discussion about schemes for sharing content between config
files ([JSON Schema's `$ref`][1] and [explicit child
declarations][2]).  However, [neither approach makes it convenient to
both make mass tweaks across a family of related configs and make
targetted tweaks to a single leaf][3], so for now we'll follow the
Dockerfile example and have [simple, stand-alone configs][4].  Folks
who find this tedious or redundant are free to automate it with
external tooling, and if a given external tool gains enough mass we
can roll it into the spec later.

See also: #73 and #74.

[1]: https://github.com/opencontainers/specs/issues/73#issuecomment-124192589
[2]: https://github.com/opencontainers/specs/pull/74
[3]: https://github.com/opencontainers/specs/issues/73#issuecomment-124722204
[4]: https://github.com/opencontainers/specs/pull/74#issuecomment-124732349

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -12,19 +12,72 @@ A standard container bundle is made of the following 3 parts:
 
 # Directory layout
 
-A Standard Container bundle is a directory containing all the content needed to load and run a container. This includes its configuration file (`config.json`) and content directories. The main property of this directory layout is that it can be moved as a unit to another machine and run the same container.
+A Standard Container bundle is a directory containing all the content needed to load and run a container. This includes its configuration file(s) and content directories. The main property of this directory layout is that it can be moved as a unit to another machine and run the same container.
 
-The syntax and semantics for `config.json` are described in [this specification](config.md).
-
-One or more *content directories* may be adjacent to the configuration file. This must include at least the root filesystem (referenced in the configuration file by the *root* field) and may include other related content (signatures, other configs, etc.). The interpretation of these resources is specified in the configuration. The names of the directories may be arbitrary, but users should consider using conventional names as in the example below.
+*Example*
 
 ```
 /
-!
--- config.json
-!
---- rootfs
-!
---- signatures
+|-- config.json
+`-- rootfs
 ```
 
+## Configuration
+
+The config file's syntax and semantics are described in [this specification](config.md).  By default, containers will use `config.json` in the bundle root:
+
+```
+/
+|-- config.json
+`-- rootfs
+```
+
+However, sometimes you need a more flexible configuration than you can get from a single static file.  Runtime's should use the following logic to select which config file to use:
+
+1. If a `config.json` file exists (or you were passed a file path), use that.
+2. If a `config` directory exists (or you were passed a directory path), walk it looking for the best platform match and use that.
+
+### Alternative configurations
+
+Runtimes like [runC][] allow you to specify a different config file explicitly, so you may find it convenient to place additional config files somewhere in your bundle.  For example, with a bundle like:
+
+```
+/
+|-- config.json
+|-- config-shell.json
+`-- rootfs
+```
+
+Then you could use `runc` to launch your application, and `runc config-shell.json` to launch a shell in a similar container environment for poking around.
+
+### Multiple platforms
+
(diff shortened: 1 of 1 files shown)


wking:
Thinking this over some more, the config-choosing functionality doesn't *have* to get rolled into the runtime.  It could happen in a separate step.  So it might be better to split this off into a separate spec layer (opencontainers/multi-platform-spec?).  Then bundle-authors who want to use this functionality can skip `config.json` and write their `config` directory.  A meta-runtime that understands that extension would:

1. Notice the presence of the `config` directory
2. Launch the config-matching utility, and write it's stdout to `config.json`.
3. Launch the container-launching utility (e.g. runC)

That approach also makes it easy for folks to plug in alternative or additional layers (e.g. `$ref` replacement, explicit-child joining, environment-variable injection, signature checking #11, …).  I like the Unix-philosophy approach here, but we'd want a way to:

1. Declare requirements so the meta-runtime knows what to do.
2. Make the per-tool specs discoverable.
3. Provide examples tying the tools together to perform real-world things.

For (2), what do we think about a “Related tools” section in the opencontainers/specs README that has a list of per-tool specs, a brief description of the idea, and then known implementations of the spec?  For example:

    * Multi-platform spec matching ([spec][mpms]).  Implementations: [mpm][]
    * Namespace joining ([spec][nsj]).  Implementations: [util-linux][], [jpetazzo][]
    …

    [mpms]: https://github.com/wking/multi-platform-match-spec
    [mpm]: https://github.com/wking/multi-platform-match-spec
    [nsj]: http://man7.org/linux/man-pages/man1/nsenter.1.html
    [util-linux]: https://git.kernel.org/cgit/utils/util-linux/util-linux.git/
    [jpetazzo]: https://github.com/jpetazzo/nsenter

Then (1) can be a shell script in the root of the container (`main.sh`)?.  The meta-runtime can launch from a restricted environment if it wants better security.  This sort of thing could be handled  by hooks in `config.json`, but if the purpose of a given script is to *build* `config.json` you have a chicken/egg problem.  The POSIX shell syntax (likely with a reduced set of available [utilities][1]) is well established and independent of containerization, so I think that makes the most sense for the lowest orchestration layer.  Going down this path will probalby lead to container systems like [ipfs-exec][2], which is just “fetch a bundle, `cd` into it, and run something (maybe in a restricted environment)”.  If we have a consistent name for what that “run something” looks like (and I'd recommend `main.sh`), life gets easier for everyone.

How does this sound?  Is it crazy?  If so, what are our alternatives for orchestrating multi-spec/component runtimes without building a single monolithic tool?

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/idx/utilities.html
[2]: https://github.com/wking/ipfs-exec

philips:
I think we are a ways off from this being a concern. My preference is that we build something simple enough that we only add new fields and remain backwards compatible at the config level forever. This may take a few releases but I think we know enough about the problem space to do it.

Put another way we need to treat the spec JSON like a protobuf and only add new fields, and deprecate old fields. We should never repurpose a field based on the version number.

philips:
Closing until we get further along.

wking:
On Tue, Jul 28, 2015 at 12:31:12PM -0700, Brandon Philips wrote:
> Put another way we need to treat the spec JSON like a protobuf and
> only add new fields, and deprecate old fields. We should never
> repurpose a field based on the version number.

Even without breaking strict backwards compatibility, I think we'll
still want a way to be compatible with multiple config versions.  For
example, say we add a more convenient way to mount sysfs with the
right permissions (#65).  What happens to my current sysfs mount [1]?
How do I stay compatible with the current no-special-sysfs-handling
runC while also taking advantage (for newer runtimes) of the new,
convenient sysfs handling?  Without providing a way to select between
a few provided config versions, it seems like bundles will be stuck
jumping through all the hoops required by the oldest minor version of
the spec that they need to support.

[1]: https://github.com/opencontainers/specs/pull/65#issuecomment-119757802

PR #81, opened 2015-07-26
specs: align the ascii graph

laijs:
The three files/directory are at the same level in the bundle directory,
they should be aligned.


LK4D4:
LGTM

PR #85, opened 2015-07-29
bundle: add initial run use case

philips:
To get everyone on the same page on what the bundle spec can do I want to enumerate reasonable use cases. I have two follow-up PRs that are discussion points for use cases that I don't think are reasonably covered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,14 @@ Table of Contents
   - [Linux Specific Configuration](config-linux.md)
 - [Runtime and Lifecycle](runtime.md)
 
+## Use Cases
+
+To provide context for users the following section gives example use cases for each part of the spec.
+
+### Filesystem Bundle & Configuration
+
+- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+
 # The 5 principles of Standard Containers
 
 Define a unit of software delivery called a Standard Container. The goal of a Standard Container is to encapsulate a software component and all its dependencies in a format that is self-describing and portable, so that any compliant runtime can run it without extra dependencies, regardless of the underlying machine and the contents of the container.


bundle.md
```diff
@@ -10,7 +10,13 @@ A standard container bundle is made of the following 3 parts:
 - One or more content directories
 - A configuration file
 
-# Directory layout
+## Use Cases
+
+This bundle specification enables users in the following cases:
+
+- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runner.
```
mrunalp: Maybe "implementation" instead or "runner" ?

wking: On Wed, Jul 29, 2015 at 09:38:06AM -0700, Mrunal Patel wrote:
> > +- A user can create a root filesystem and configuration, with
> > low-level OS and host specific details, and launch it as a
> > container under an Open Container runner.
> 
> Maybe "implementation" instead or "runner" ?

The current spec seems to use both ‘runtime’, ‘runc’, and
‘implementation’.  I don't have a strong preference for any of these
choices, but it would be nice to pick a single name to standardize on.
Since ‘runtime’ is currently the most popular (by grep hits), can we
use that?

bundle.md
```diff
@@ -10,7 +10,13 @@ A standard container bundle is made of the following 3 parts:
 - One or more content directories
 - A configuration file
 
-# Directory layout
+## Use Cases
+
+This bundle specification enables users in the following cases:
```
wking: How about “Possible use cases:”?

wking:
On Wed, Jul 29, 2015 at 08:30:56AM -0700, Brandon Philips wrote:
> To get everyone on the same page on what the bundle spec can do I
> want to enumerate reasonable use cases. I have two follow-up PRs
> that are discussion points for use cases that I don't think are
> reasonably covered.

I think listing out a few examples is good, but I'd also like to see
the unsupported cases listed too (vs. just not listing them under the
list of examples).

I also think it may be better to spin off this list into it's own
file, so bundle.md can stay focused on the bundle-specific details
that a bundle or runtime author needs to know.  Sprinking a few
informative sections into the normative parts of the spec is fine, but
this block seems decoupled enough that it's worth keeping it separate.

bundle.md
philips: runtime makes since to me too. I will fix that up here and then do a PR to make it consistent.

philips:
@wking I agree we need to list unsupported use cases also; the current bundle only covers a very small number of use cases for a standard container. We need to start discussing how this all fits together with  layers, images, etc

philips:
@mrunalp @wking thoughts?

mrunalp:
@philips SGTM

wking:
On Thu, Jul 30, 2015 at 11:45:39AM -0700, Brandon Philips wrote:
> I agree we need to list unsupported use cases also; the current
> bundle only covers a very small number of use cases for a standard
> container.

Are you looking for suggested bullet points?  In your initial post for
this PR, you mention “two follow-up PRs that are discussion points for
use cases that I don't think are reasonably covered”.  I don't see any
PRs that sound like those.  Maybe you forgot to push them?

Beyond that, I think the scope limitations for this spec is anything
beyond the standard operations (creating, starting, and stopping
containers [1]).  In #21 we pushed snapshots, transfer, and tagging
out of those goals, so I think we should list them in non-use-cases
(ideally pointing to external resources for folks who are interested
in that functionality).  The other functionality that I understand to
be out-of-scope for this spec is cross-platform operations [2] and
bundle-verification [3].

> We need to start discussing how this all fits together with layers,
> images, etc

Is that Docker layers/images?  I think that falls under
tagging/transport (with maybe some dabbling into bundle-verification).
I'm fine mentioning that in the out-of-scope list, but I'd probably
organize the statement to focus on the functionality we don't specify
(e.g. how to locate and download a bundle) and then mention that folks
who want to do that sort of thing might be interested in using the
Docker registry (although I think they're still working on search for
v2, docker/distribution#206).  And I'm not entirely clear on how you'd
decouple a Docker-fetched container from the Docker daemon so you
could launch it directly with runC.  But I haven't gone over that
section of the Docker implementation, so maybe some of the Docker
folks can chip in with a quick Docker-over-runC sketch.

[1]: https://github.com/opencontainers/specs/tree/d619f2fccb13385e5c6bfcbf3ac1a3dcd36f314b#1-standard-operations
[2]: https://github.com/opencontainers/specs/issues/73#issuecomment-126421828
[3]: https://github.com/opencontainers/specs/issues/5#issuecomment-115423504

crosbymichael:
LGTM

vbatts:
LGTM

vbatts:
@wking lots of good points. The initial piece being ironed out is runtime, and really it needs a rootfs and configuration of what the application needs/is-asking-for. 
The function of layering to produce the rootfs, is a related topic, but as far as runtime is concerned, provides a rootfs. Distribution is another topic, just like verification is. All good topics. :-)

vbatts:
@wking also, we have a list that we could discuss this on https://groups.google.com/a/opencontainers.org/forum/#!forum/dev

wking:
On Fri, Jul 31, 2015 at 07:15:44PM -0700, Vincent Batts wrote:
> @wking also, we have a list that we could discuss this on
> https://groups.google.com/a/opencontainers.org/forum/#!forum/dev

I'm happy to post my out-of-scope use cases to that list or to open
them as per-point pull requests.  Just let me know the preferred
communication channel for things like this (maybe it's the mailing
list?).  I'd just started discussing them here because I read
Brandon's “we need to list” and “we need to start discussing” as “in
this PR” and not “in this project”.

Also, should I open a follow-up PR for moving this informative
(vs. normative) discussion to a separate file [1]?  I didn't see any
direct push-back after I suggested that earlier, but I'm not sure if
that means “suggestion considered and rejected” or “suggestion not
noticed” ;).

[1]: https://github.com/opencontainers/specs/pull/85#issuecomment-126034306

wking:
On Sat, Aug 01, 2015 at 11:07:58AM -0700, W. Trevor King wrote:
> Also, should I open a follow-up PR for moving this informative
> (vs. normative) discussion to a separate file [1]?

Ah, I see it was shifted from bundle.md to the README.  So scratch
this paragraph from my previous comment.

PR #87, opened 2015-07-29
Add runtime state configuration and structs

### LINKED ISSUE #41: runtime: create state directory and structure
For supporting things like hooks and updates we need a way to define runtime state of the container.  This can be used to pass information to hooks so that they know the runtime state of the container and also used to dynamically make changes to resource allocations by running an update command for the runtime and have it reconcile the changes against the container's current state and the update requested.  

For the hooks usecase the state should contain information such as:

* pid of the container's process
* paths to the namespaces (linux)
* paths to the cgroups setup for the container

For supporting updates the state should container information such as:

* pid of the container's process
* cgroup paths
* container root path
* cgroup/resource information 
* device nodes
* rlimits 
* sysctls
* mounts

Some of this information would be duplicated from the initial container's config so it maybe worth it to look into embedding the original config into the state structure.

### LINKED ISSUE #72: Storing/retrieving container state
We need to standardize on how to store/retrieve the state of a running container. runc today allows one to specify an id during invocation and then stores the state of that container under ```/var/run/ocf/<id>```.

I think we need to agree on a directory and the mechanism if there are any alternate suggestions/ideas.

@crosbymichael @vmarmol @LK4D4 @philips

crosbymichael:
This adds runtime state information for oci container's so that it can
be persisted and used by external tools.

Closes #72
Closes #41 

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -1,5 +1,34 @@
 # Runtime and Lifecycle
 
+## State
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information should be stored in `/run/oci/containers`.
+The directory structure for a container is `/run/oci/containers/<containerID>/state.json`.
+By providing a default location that container state is stored external applications can find all containers running on a system.
+
+* **version** (string) Version of the OCI specification used when creating the container.
+* **id** (string) ID is the container's ID.
+* **pid** (int) Pid is the ID of the main process within the container.
+* **root** (string) Root is the path to the container's bundle directory.
+
+The ID is provided in the state because hooks will be executed with the state as the payload.
+This allows the hook to perform clean and teardown logic after the runtime destroys its own state.
+
+The root directory to the bundle is provided in the state so that consumers can find the container's configuration and rootfs where it is located on the host's filesystem.
+
+*Example*
+
+```json
+{
+    "id": "oci-container",
+    "pid": 4422,
+    "root": "/containers/redis"
+}
+```
+
 ## Lifecycle
 
 ### Create
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -2,6 +2,9 @@ package specs
 
 import "os"
 
+// LinuxStateDirectory holds the container's state information
+const LinuxStateDirectory = "/run/oci/containers"
+
 // LinuxRuntimeSpec is the full specification for linux containers.
 type LinuxRuntimeSpec struct {
 	RuntimeSpec
--- a/config.go
+++ b/config.go
@@ -56,3 +56,15 @@ type MountPoint struct {
 	// Path specifies the path of the mount. The path and child directories MUST exist, a runtime MUST NOT create directories automatically to a mount point.
 	Path string `json:"path"`
 }
+
+// State holds information about the runtime state of the container.
+type State struct {
+	// Version is the version of the specification that is supported.
+	Version string `json:"version"`
+	// ID is the container ID
+	ID string `json:"id"`
+	// Pid is the process id for the container's main process.
+	Pid int `json:"pid"`
+	// Root is the path to the container's bundle directory.
+	Root string `json:"root"`
+}


wking:
On Wed, Jul 29, 2015 at 02:10:46PM -0700, Michael Crosby wrote:
> This adds runtime state information for oci container's so that it can
> be persisted and used by external tools.
> 
> Closes #72

I think closing #41 (the state spec) is important, but I'm still not
sold on the directory-binding proposed in #72.  Besides comments
supporting storage-agnostic state by me, @LK4D4 [1,2], and
@kunalkushwaha [3] in #72, see the discussion in
opencontainers/runc#159.

Can we restrict this PR to just address the spec content (#41), which
I think we all agree should be addressed in this spec)?  Then the more
controversial storage location (#72) won't bog down these changes.

[1]: https://github.com/opencontainers/specs/issues/72#issuecomment-123965223
[2]: https://github.com/opencontainers/specs/issues/72#issuecomment-124134519
[3]: https://github.com/opencontainers/specs/issues/72#issuecomment-124452794

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+## State
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information is stored in `/run/oci`.
+The directory structure for a container is `<root>/<containerID>/state.json`.
+
+* **id** (string) ID is the container's ID.
+* **pid** (int) Pid is the main processes id within the container.
```
wking: “processes” → “process's”.  And also probably “Pid” → “PID” (at least, that's what kill(1) uses) and “id” → “ID” (that's what chown(1) uses).

spec_linux.go
```diff
@@ -150,3 +153,16 @@ type Resources struct {
 	// Network restriction configuration.
 	Network Network `json:"network"`
 }
+
+type LinuxState struct {
```
wking: I'd rather add this to a new `state_linux.go`, since it's not part of the spec.  It's still in the `specs` package, so you could still use `spec_linux.go`-defined types in the `LinuxState` structure.

state.go
```diff
@@ -0,0 +1,11 @@
+package specs
+
+// State holds information about the runtime information of the container.
```
wking: Information about information?  How about “State holds information about a container instance.”?

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+*Example Linux*
+
+```json
+{
+    "namespaces": {
+        "process": "/proc/33/ns/pid",
+        "net": "/proc/33/ns/net"
+    },
+    "cgroups": {
+        "device": "/sys/fs/cgroup/devices/oci-container",
```
wking: Is the difference between `device` (in the key) and `devices` (in the value) intentional?

It would be nice if the spec went into the same detail for cgroups [that it does for namespaces][1], but maybe that's a separate PR.

[1]: https://github.com/opencontainers/specs/blob/d619f2fccb13385e5c6bfcbf3ac1a3dcd36f314b/config-linux.md#namespace-types

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+    "pid": 4422,
+    "root": "/containers/redis"
+}
+```
+
+Linux systems add some platform specific information to the state.
+
+* **namespaces** Paths to the Linux namespaces setup for the container.
+* **cgrousps** Paths to the container's cgroups.
+* **externalFds** Paths to the container's main processes STDIO and other external fds.
```
wking: “processes” → “process's”.  And I'd probably just write out “file descriptors”, at least in the description sentence.

spec_linux.go
```diff
@@ -150,3 +153,16 @@ type Resources struct {
...
+	State
+	// Linux holds platform specific state information for linux.
+	Linux struct {
+		// Namespaces holds paths to the container's namespaces on disk.
+		Namespaces map[string]string `json:"namespaces"`
+		// Cgroups holds paths to the container's cgroup paths.
+		Cgroups map[string]string `json:"cgroups"`
+		// ExternalFds holds paths to the container's STDIO.
```
wking: I'd replace “STDIO” here with “open file descriptors”.  You mention “STDIO and other…” in `runtime.md`, which is a nice hint, but I'd rather avoid being overly-specific in the Go-struct comments.

wking:
The state content here looks good to me (modulo a few minor copy-edits
which I commented on inline).  The only things missing from #41 are:

* device nodes
* rlimits
* sysctls
* mounts

Folks might also be interested in things like the TTY associated with
the main process.  But I'd expect all of that can be extracted
canonically from the kernel itself (once you know the main PID), so we
don't have to pass it along in the state JSON.  On the other hand, you
can also get the cgroups and namespaces from the main PID by looking
through /sys, so I'm not sure where we draw the line between “we'll
bundle that in the state JSON” and “look it up yourself”.

mrunalp:
@wking We are being conservative in what we add to the state right now.

wking:
On Wed, Jul 29, 2015 at 03:59:11PM -0700, Mrunal Patel wrote:
> We are being conservative in what we add to the state right now.

That makes sense, but then why not leave cgroups and namespaces out of
the spec?  Interested hooks can always use ‘ps -o cgroup $PID’, ‘ps -o
ipcns $PID’, etc. to figure those things out given the main process's
PID.

wking:
On Wed, Jul 29, 2015 at 04:11:49PM -0700, W. Trevor King wrote:
> … but then why not leave cgroups and namespaces out of the spec?

And the external file descriptors, since ‘lsof -p $PID’ should be able
to give you those.

runtime.md
```diff
@@ -1,5 +1,51 @@
...
 
+## State
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information is stored in `/run/oci`.
+The directory structure for a container is `<root>/<containerID>/state.json`.
+
+* **id** (string) ID is the container's ID.
```
jonboulle: id is a bit under specified; are there any restrictions on the format of the string? what is the uniqueness domain?

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+    "id": "oci-container",
+    "pid": 4422,
+    "root": "/containers/redis"
+}
+```
+
+Linux systems add some platform specific information to the state.
+
+* **namespaces** Paths to the Linux namespaces setup for the container.
+* **cgrousps** Paths to the container's cgroups.
```
jonboulle: cgrousps

crosbymichael:
any other comments, questions, concerns?

wking:
On Tue, Aug 04, 2015 at 11:23:02AM -0700, Michael Crosby wrote:
> any other comments, questions, concerns?

If that's “can we merge this now?”, I'd suggest at least addressing
the typos first [1,2,3].  The other comments that still apply to [4]
all look reasonable and easy to address as well.

I still think the issue of “how much more than the main PID do we
need?” is open [5], since we're currently professing conservatism [6]
but still seem to be supplying some PID-calculable data [7,8].  We can
continue that discussion in a follow-up issue if we want, but it's
hard to pull information back out of a spec that's intended to be
infinitely forward compatible [9].

[1]: https://github.com/opencontainers/specs/pull/87/files#r35819305
[2]: https://github.com/opencontainers/specs/pull/87/files#r35828059
[3]: https://github.com/opencontainers/specs/pull/87/files#r35820092
[4]: https://github.com/opencontainers/specs/pull/87/files
[5]: https://github.com/opencontainers/specs/pull/87#issuecomment-126119707
[6]: https://github.com/opencontainers/specs/pull/87#issuecomment-126121373
[7]: https://github.com/opencontainers/specs/pull/87#issuecomment-126123937
[8]: https://github.com/opencontainers/specs/pull/87#issuecomment-126124977
[9]: https://github.com/opencontainers/specs/pull/76#issuecomment-125729729

crosbymichael:
`If that's “can we merge this now?”, I'd suggest at least addressing
the typos first [1,2,3]. `


@wking Yes, I know a little bit about contributing to open source, I like to get additional feedback before I start addressing the comments so I don't have to update, push, feedback, update, push, feedback multiple times.

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information is stored in `/run/oci`.
+The directory structure for a container is `<root>/<containerID>/state.json`.
+
+* **id** (string) ID is the container's ID.
+* **pid** (int) Pid is the main processes id within the container.
+* **root** (string) Root is the path to the container's root filesystem specified in the configuration.
```
wking: Maybe replace “**$ENTRY** ($TYPE) $ENTRY is the …” with “**$ENTRY** ($TYPE) The …”.

wking:
On Tue, Aug 04, 2015 at 12:40:41PM -0700, Michael Crosby wrote:
> Yes, I know a little bit about contributing to open source, I like
> to get additional feedback before I start addressing the comments so
> I don't have to update, push, feedback, update, push, feedback
> multiple times.

That makes sense, although rebase/push is reasonably cheap when
there's just a single commit.

While I'm summarizing issues that could be addressed in a reroll [1],
I think the only one left out was my suggestion to drop the storage
phrasing (or at least pull it out into a separate PR) [2].

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-127707178
[2]: https://github.com/opencontainers/specs/pull/87#issuecomment-126114680

runtime.md
mrunalp: How about all lowercase alphabets and digits?

wking: On Wed, Aug 05, 2015 at 09:19:02AM -0700, Mrunal Patel wrote:
> > +* **id** (string) ID is the container's ID.
> 
> How about all lowercase alphabets and digits? 

That sounds good to me, although we might want to allow hyphens for
folks who want to use UUIDs.  And currently runC defaults to the
bundle's directory name, so we'd need to figure out how we wanted to
map other characters into the valid character set.

Do we want to add requirements for the uniqueness domain?  Unique to
the host seems like a minimum, but we can require universal uniqueness
(e.g. “container IDs must be UUIDs [1]”) or leave that up to the
implementation.  I'd rather leave it up to the implementation, since
as long as UUIDs are *valid* container identifiers, runtime callers
that want universal uniqueness are free to use them.

[1]: https://tools.ietf.org/html/rfc4122

crosbymichael:
I updated this PR addressing comments and rebasing.

wking:
On Wed, Aug 05, 2015 at 01:09:49PM -0700, Michael Crosby wrote:
> I updated this PR addressing comments and rebasing.

Do we have a reason for not moving LinuxState from spec_linux.go to
state_linux.go [1]?

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r35819480

crosbymichael:
@wking because I don't want alot of files

wking:
On Wed, Aug 05, 2015 at 01:23:35PM -0700, W. Trevor King wrote:
> Do we have a reason for not moving LinuxState from spec_linux.go to
> state_linux.go [1]?

And I didn't mention this explicitly on the first pass, but I'd move
LinuxStateDirectory from spec_linux.go to state_linux.go as well (if
we don't drop the storage stuff or spin it off to a separate PR [1]).

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-126114680

wking:
On Wed, Aug 05, 2015 at 01:09:49PM -0700, Michael Crosby wrote:
> I updated this PR addressing comments and rebasing.

Also no longer visible after the rebase [1]:

  Maybe replace “$ENTRY ($TYPE) $ENTRY is the …” with “$ENTRY ($TYPE) The …”.

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r36233315

wking:
On Wed, Aug 05, 2015 at 01:27:58PM -0700, Michael Crosby wrote:
> @wking because I don't want alot of files

Then why not squash state.go into spec.go?  I think “we don't separate
the spec from the config” makes sense, and “we do separate the spec
from the config” makes sense, but I'd rather avoid “we sometimes
separate the spec from the config”.

spec.go
```diff
@@ -78,3 +78,13 @@ type Hook struct {
 	Args []string `json:"args"`
 	Env  []string `json:"env"`
 }
+
+// State holds information about the runtime state of the container
```
wking: We introduce the `Spec` entry with:

> Spec is the base configuration for the container.  It specifies platform independent configuration.

Since the `State` type is the root of a new JSON structure, independent of `Spec`, we may want to echo that here with something like:

> State is the base type for the container's runtime state.  It specifies platform independent state.

Or something like that to help folks avoid confusing the spec types and the state types.

crosbymichael:
Yes, i moved state into the spec.go file now

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+    "pid": 4422,
+    "root": "/containers/redis"
+}
+```
+
+Linux systems add some platform specific information to the state.
+
+* **namespaces** Paths to the Linux namespaces setup for the container.
+* **cgroups** Paths to the container's cgroups.
+* **externalFds** Paths to the container's open file descriptors.
```
philips: what is the utility of this? Why can't it be derived from the pid?

wking: On Wed, Aug 05, 2015 at 03:38:33PM -0700, Brandon Philips wrote:
> > +Linux systems add some platform specific information to the state.
> > +
> > +* **namespaces** Paths to the Linux namespaces setup for the container.
> > +* **cgroups** Paths to the container's cgroups.
> > +* **externalFds** Paths to the container's open file descriptors.
> 
> what is the utility of this? Why can't it be derived from the pid?

+1 to dropping these (previous discussion in [1,2,3,4]).

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-126119707
[2]: https://github.com/opencontainers/specs/pull/87#issuecomment-126121373
[3]: https://github.com/opencontainers/specs/pull/87#issuecomment-126123937
[4]: https://github.com/opencontainers/specs/pull/87#issuecomment-126124977

spec_linux.go
```diff
@@ -2,6 +2,9 @@
 
 package specs
 
+// LinuxStateDirectory holds the container's state information
+const LinuxStateDirectory = "/run/oci"
```
vbatts: i'm cool with this, as long as it's expected that it is a tmpfs state directory.

runtime.md
crosbymichael: @avagin @boucher Can you comment on this?

spec_linux.go
crosbymichael: it is

vbatts:
what's left here?

wking:
On Mon, Aug 10, 2015 at 11:38:25AM -0700, Vincent Batts wrote:
> what's left here?

* A decision on whether or not to drop the Linux-specific fields
  (discussion starting with [1], currently waiting on comments from
  @avagin @boucher [2])

* Clarification on the ID format and uniqueness domain (discussion
  starting with [3]).

* Clearer docs (or rejecting the need to clarify the docs) showing
  that spec.go's `State` is the root of a structure independent of
  `Spec` [4].

* Bumping (or rejecting the need to bump) the directory-binding to a
  separate PR [5].

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r36363285
[2]: https://github.com/opencontainers/specs/pull/87#discussion_r36540805
[3]: https://github.com/opencontainers/specs/pull/87/files#r35828046
[4]: https://github.com/opencontainers/specs/pull/87/files#r36351375
[5]: https://github.com/opencontainers/specs/pull/87#issuecomment-126114680

wking:
On Mon, Aug 10, 2015 at 11:48:13AM -0700, W. Trevor King wrote:
> Mon, Aug 10, 2015 at 11:38:25AM -0700, Vincent Batts:
> > what's left here?
>
> …

I've opened a PR against this branch [1] to collect my suggested
resolutions for these issues, in case that makes it easier to keep
track and move forward.

[1]: https://github.com/crosbymichael/specs/pull/1

runtime.md
philips: @avagin @boucher ping

runtime.md
philips: We need to add a comment that the reason we have an ID is because the hooks only get an open fd to this json document so they won't know the unique ID which may be useful for removal, etc.

runtime.md
```diff
@@ -1,5 +1,51 @@
 # Runtime and Lifecycle
 
+## State
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information is stored in `/run/oci`.
+The directory structure for a container is `<root>/<containerID>/state.json`.
```
philips: we should document a potential use case here.

philips: the potential use case is to have things like cadvisor find the container

runtime.md
```diff
@@ -1,5 +1,51 @@
 # Runtime and Lifecycle
 
+## State
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information is stored in `/run/oci`.
```
philips: To make this fit with the suggestion in the previous sentence perhaps: "On Linux based systems the state information SHOULD BE stored in `/run/oci`."

LK4D4:
What about version? :)

runtime.md
philips: Also need to note that containerID MUST match the containerID subdirectory field.

runtime.md
philips: @vbatts suggested that we use `/run/oci/containers/` so we don't consume the whole namespace.

wking:
On Wed, Aug 26, 2015 at 10:30:42AM -0700, Alexander Morozov wrote:
> What about version? :)

I think we should just use whichever aproach we decide on for
config.json in #110.

runtime.md
```diff
@@ -1,5 +1,51 @@
...
+The directory structure for a container is `<root>/<containerID>/state.json`.
+
+* **id** (string) ID is the container's ID.
+* **pid** (int) Pid is the ID of the main process within the container.
+* **root** (string) Root is the path to the container's root filesystem specified in the configuration.
```
philips: can we document the handful of use cases: 

- hook that wants to setup things in the root filesystem

wking: On Wed, Aug 26, 2015 at 10:43:24AM -0700, Brandon Philips wrote:
> > +* **id** (string) ID is the container's ID.
> > +* **pid** (int) Pid is the ID of the main process within the container.
> > +* **root** (string) Root is the path to the container's root filesystem specified in the configuration.
> 
> can we document the handful of use cases: 

I put some notes in the meeting's Google docs [1], in case folks want
reminders when writing this up for an informative section of the spec.

[1]: https://docs.google.com/document/d/1niKxf4-JWZxHh45bnke4Dl8xoR0QS1ctJ8kVPqS-OKQ/edit?usp=sharing

runtime.md
wking: On Wed, Aug 26, 2015 at 10:37:33AM -0700, Brandon Philips wrote:
> > @@ -1,5 +1,51 @@
> >  # Runtime and Lifecycle
> >  
> > +## State
> > +
> > +The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
> > +The runtime state is stored in a JSON encoded file.
> > +It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
> > +On Linux based systems the state information is stored in `/run/oci`.
> 
> @vbatts suggested that we use `/run/oci/containers/` so we don't consume the whole namespace.

In further meeting discussion, I think we stepped back from that and
collected around:

* /run/oci/{container-id} is a flat namespace for all containers,
  although the ID can contain forward slashes so it might involve
  several directories.  But there's no OCI-specified namespacing.
* Any important information (e.g. “this state is for a container, not
  a hypervisor”) will be stored in the state file itself.

I think my namespaced-version proposal [1] will make it easy to
collect information like “hypervisor vs. container”, with version
entries like:

  "version": {
    "ocsContainerState": "1.0.0"
  }

or:

  "version": {
    "runVState": "0.3.0"
  }

etc.

[1]: https://github.com/opencontainers/specs/issues/110#issuecomment-133587646

runtime.md
wking: On Wed, Aug 26, 2015 at 10:12:52AM -0700, Brandon Philips wrote:
> > +* **id** (string) ID is the container's ID.
> 
> We need to add a comment that the reason we have an ID is because
> the hooks only get an open fd to this json document so they won't
> know the unique ID which may be useful for removal, etc.

I wasn't clear on how it would be useful for removal.  Everyone got on
board when someone mentioned network teardown, but the connection
between that and the container ID wasn't clear to me.  Not that we
have to be completely specific when motivating this entry.  I'm just
pointing out that if we want to explain why we need the container ID
in a post-stop hook, we might want to provide more detail than we had
in today's meeting.

crosbymichael:
I just updated this PR will all of the feedback from today.

wking:
On Wed, Aug 26, 2015 at 03:09:01PM -0700, Michael Crosby wrote:
> I just updated this PR will all of the feedback from today. 

Now that #88 has landed (with a runtime tree in runtime_config*.go),
it seems even more strange to squash the state tree into the
config*.go files.  Previous discussion around [1].

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-128139632

wking:
On Wed, Aug 26, 2015 at 03:09:01PM -0700, Michael Crosby wrote:
> I just updated this PR will all of the feedback from today. 

I'd still recommend the small rephrasing from
wking/opencontainer-specs@098ba7ec, which I'd floated in
crosbymichael/specs#1.

runtime.md
```diff
@@ -1,5 +1,47 @@
...
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information should be stored in `/run/oci/containers`.
+The directory structure for a container is `/run/oci/containers/<containerID>/state.json`.
+By providing a default location that container state is stored external applications can find all containers running on a system.
+
+* **version** (string) Version of the OCI specification used when creating the container.
+* **id** (string) ID is the container's ID.
+* **pid** (int) Pid is the ID of the main process within the container.
+* **root** (string) Root is the path to the container's bundle directory.
```
wking: My understanding of today's meeting was that we did need host-side path to the container's root filesystem here. Folks were motivating this entry for mounting additional host-side content into the container's filesystem.  I don't see how the bundle root would help much there.  Maybe bundle-root → `config.json` → `root.path`?  But that assumes that the container wasn't launched with an alternate config, that the config hasn't been tweaked since launch, ….  It seems better to just point folks directly at the container's root filesystem.

runtime.md
```diff
@@ -1,5 +1,47 @@
...
+
+The ID is provided in the state because hooks will be executed with the state as the payload.
+This allows the hook to perform clean and teardown logic after the runtime destroys it's own state.
+The root directory to the bundle is provided in the state so that consumers can find the container's configuration and rootfs.
```
wking: I don't see any value to finding the configuration, and I explained why I think we should link directly to the rootfs [above][1].

[1]: https://github.com/opencontainers/specs/pull/87/files#r38043014

runtime.md
SaiedKazemi: @philips @crosbymichael @avagin
For checkpoint restore, we need to pass the external pathnames to CRIU via --ext-mount-map.

Docker containers, for example, currently have two sets of external paths: (1) statically named external bind mounts /etc/hostname, /etc/hosts, and /etc/resolv.conf which are in config.json as HostnamePath, HostsPath, and ResolvConfPath (2) dynamically named external volumes (which may or may not exist for a container) and are listed under Volumes.

In short, for checkpoint restore via CRIU, the runtime state configuration has to have all external pathnames.

runtime.md
```diff
@@ -1,5 +1,47 @@
...
+Linux systems add some platform specific information to the state.
+
+* **externalFds** Paths to the container's open file descriptors.
+
+*Example Linux*
+
+```json
+{
+    "externalFds": [
+        "/proc/33/fd/1"
```
wking: Is this path in the context of the host's process namespace or the container's process namespace?  If it's in the host's context, is PID 33 one of the application processes, or an out-of-container process who's stdout is being read by an application process?

wking: For example, with one process piping its stdout into another's stdin, I see:

    $ ls -l /proc/19119/fd/1 /proc/19120/fd/0
    l-wx------ 1 wking wking 64 Aug 26 15:38 /proc/19119/fd/1 -> pipe:[83842]
    lr-x------ 1 wking wking 64 Aug 26 15:38 /proc/19120/fd/0 -> pipe:[83842]

You can also find the pipe name *and* the associated process at the other end using `lsof`:

    $ lsof -p 19120 +E | grep 0r
    cat     19119 wking    1w  FIFO   0,10      0t0    83842 pipe 19120,less,0r
    less    19120 wking    0r  FIFO   0,10      0t0    83842 pipe 19119,cat,1w

Since the CRIU bindings can pull this information from the kernel, and I'd suggest not supplying it in the state JSON.

runtime.md
```diff
@@ -1,5 +1,47 @@
 # Runtime and Lifecycle
 
+## State
+
+The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
+The runtime state is stored in a JSON encoded file.
+It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the state information should be stored in `/run/oci/containers`.
+The directory structure for a container is `/run/oci/containers/<containerID>/state.json`.
```
wking: After the meeting, my understanding was that we wanted a  flat `/run/oci/{containerID}/state.json` namespace, with internal-to-the-state content for figuring out the semantics of the state content.  If the `state.json` requires a version field that clarifies the associated spec (e.g. [my suggestion here][1]), there would be no need for the `containers/` path to namespace different types of container specs.

[1]: https://github.com/opencontainers/specs/issues/110#issuecomment-133587646

runtime.md
```diff
@@ -1,5 +1,47 @@
...
+By providing a default location that container state is stored external applications can find all containers running on a system.
+
+* **version** (string) Version of the OCI specification used when creating the container.
+* **id** (string) ID is the container's ID.
+* **pid** (int) Pid is the ID of the main process within the container.
```
wking: Maybe clarify the host's process namespace?  For example:

> **pid** (int) The ID of the application process in the host's process namespace.

assuming we land something like #107 to get a definition for “application”.  Although this somewhat conflates the initial process and the whole tree descending from that process under application.  And I'm not sure what happens if there are fork/exec calls to split off new trees and the container *isn't* using process namespacing.

runtime.md
SaiedKazemi: My earlier comment about external pathnames and how they should be passed to CRIU for checkpoint restore is not showing up anymore.  Was it deleted?

Re /proc/<pid>/fd/<fd>, this information is available only for opened files and the paths will point inside the container namespace.  Also, not all of the external paths remain open during the entire lifetime of the container.

If this information is not provided in the runtime state configuration file, it will have to be looked up via /proc/<pid>/mountinfo.

wking: On Wed, Aug 26, 2015 at 04:05:46PM -0700, Saied Kazemi wrote:
> Re /proc/<pid>/fd/<fd>, this information is available only for
> opened files and the paths will point inside the container
> namespace.

I can confirm this when accessed from the host:

  # lsof -p 20050 +E
  COMMAND   PID  USER   FD   TYPE DEVICE SIZE/OFF     NODE NAME
  …
  exe     20050 wking    0r   REG   0,19   140520 10149337 /lib64/ld-2.20.so
  …

where I'm looking at that file in my container's mount namespace, not
the host's.  So if we go with the /proc approach, users that want
host-side paths will need to unpack the mount namespacing to find the
host-side path.

> Also, not all of the external paths remain open during the entire
> lifetime of the container.

Why would this matter?  You're using CRIU to checkpoint/restore the
container as it stands, not for some pre-checkpoint state.

> If this information is not provided in the runtime state
> configuration file, it will have to be looked up via
> /proc/<pid>/mountinfo.

That seems like the only reasonable approach to me, but I can see sane
people differing on who does that direct-kernel lookup (i.e. the
runtime or the state consumer).  If you're just caching
container-creation-time config settings, you'll miss any subsequent
mounts that may have been made since creation time (e.g. via hooks
[1]).

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r38043014

SaiedKazemi:
During checkpoint, CRIU has to parse the process's mountinfo table.  When it hits an external bind mount, it will fail unless it is "mapped" via the --ext-mount-map command line option.  So, the issue is not the state of container open files at the time of checkpoint.  It's how the container filesystem (including mount points) are checkpointed and restored.

wking:
On Wed, Aug 26, 2015 at 04:40:49PM -0700, Saied Kazemi wrote:
> During checkpoint, CRIU has to parse the process's mountinfo table.
> When it hits an external bind mount, it will fail unless it is
> "mapped" via the --ext-mount-map command line option.

So the CRIU caller has to parse /proc/{pid}/mountinfo as well and
setup --ext-mount-map entries for any mounts?  From my host,
inspecting an oci-gentoo-minimal [1] container that mounts a bundled
directory into /root:

  $ cat /proc/20704/mountinfo 
  71 43 0:18 /home/wking/src/opencontainers/oci-gentoo-minimal/rootfs / ro,noatime - btrfs /dev/sda4 rw,ssd,space_cache
  …
  87 71 0:18 /home/wking/src/opencontainers/oci-gentoo-minimal/root /root rw,noatime - btrfs /dev/sda4 rw,ssd,space_cache
  …

and that's all you need (as far as I can tell) to figure out that you
should be dumping with:

  # criu dump ... --ext-mount-map /root:rootmount

and restoring with:

  # criu restore ... --ext-mount-map rootmount:/path/to/bundles/oci-gentoo-minimal/root

[1]: https://github.com/wking/oci-gentoo-minimal

runtime.md
avagin: It's hard to understand what you are talking about.
I want to show an example how it works in libcontainer:

```
        var stdout bytes.Buffer

        pconfig := libcontainer.Process{
                Args:   []string{"cat"},
                Env:    standardEnvironment,
                Stdout: &stdout,
        }

        err = container.Start(&pconfig)
```

```
"init_process_pid":7725,
"external_descriptors":["/dev/null","pipe:[66755]","/dev/null"]
```

```
# ls -l /proc/7725/fd
total 0
lr-x------ 1 root root 64 Aug 27 16:14 0 -> /dev/null
l-wx------ 1 root root 64 Aug 27 16:14 1 -> pipe:[66755]
lr-x------ 1 root root 64 Aug 27 16:14 2 -> /dev/null
l-wx------ 1 root root 64 Aug 27 16:14 255 -> pipe:[66755]
```
In this example, the user creates a pipe and executes a process. The user holds one end of a pipe and the other end gives to the process as stdout. Then the user does checkpoint/restore. Here is a question how to restore descriptors which are associated with the user pipe.
Currently the user creates this pipe and execute "criu restore". But CRIU needs to know which file descriptors should be replaced by this pipe. For that we use externalFds. When CRIU dumps processes, it marks all descriptors from externalFds as external, criu restore takes external descriptors from the user and restore them accoding with externalFds.

wking: On Thu, Aug 27, 2015 at 06:32:33AM -0700, Andrew Vagin wrote:
> It's hard to understand what you are talking about.  I want to show
> an example how it works in libcontainer:

Thanks :).  It helps to have something concrete to make sure we're all
talking about the same things ;).

> "init_process_pid":7725,
> "external_descriptors":["/dev/null","pipe:[66755]","/dev/null"]
>
> # ls -l /proc/7725/fd
> total 0
> lr-x------ 1 root root 64 Aug 27 16:14 0 -> /dev/null
> l-wx------ 1 root root 64 Aug 27 16:14 1 -> pipe:[66755]
> lr-x------ 1 root root 64 Aug 27 16:14 2 -> /dev/null
> l-wx------ 1 root root 64 Aug 27 16:14 255 -> pipe:[66755]

This makes it seem like all the information about open file
descriptors is included in /proc/{pid}/fd.  Are you saying you need
something that's not in /proc?  Or is the issue file descriptor 255,
which you don't have in your external_descriptors list?

> Currently the user creates this pipe and execute "criu restore". But
> CRIU needs to know which file descriptors should be replaced by this
> pipe. For that we use externalFds.

But it seems like you could almost as easily use the values available
via /proc/{pid}/fd.  Why not get them directly from /proc instead of
requiring them to be listed in the state JSON?

avagin: Yes, the issue file descriptor is 255. How can we use /proc/{pid}/fd, if the process has closed the 1 descriptor.
```
 lr-x------ 1 root root 64 Aug 27 16:14 0 -> /dev/null
 lr-x------ 1 root root 64 Aug 27 16:14 2 -> /dev/null
 l-wx------ 1 root root 64 Aug 27 16:14 255 -> pipe:[66755]
``` 
How can we understand that 255 is external file descriptor, which is associated with the initial stdout?

wking: On Thu, Aug 27, 2015 at 10:35:58PM -0700, Andrew Vagin wrote:
>  lr-x------ 1 root root 64 Aug 27 16:14 0 -> /dev/null
>  lr-x------ 1 root root 64 Aug 27 16:14 2 -> /dev/null
>  l-wx------ 1 root root 64 Aug 27 16:14 255 -> pipe:[66755]
>
> How can we understand that 255 is external file descriptor…

Because it's an unnamed pipe associated with a process outside of the
container (see my ‘lsof -p {pid} +E’ for how to find the other
processes associated with the pipe [1]).

> … which is associated with the initial stdout?

For the general case (what is that thing on the other end?), I think
the semantic meaning is outside our specification [2].

For this specific case, #113 says that we'll pass through file
descriptors 0, 1, 2, and possibly more when we launch the application
process.  So we can find the semantics relative to the runtime by
inspecting the runtime or other hints.  Here's BusyBox running with
process.terminal set to true:

  # pstree -sSp 3958
  init(1)───login(2172)───…───sudo(3950)───runc(3954)───busybox(3958,ipc,mnt,pid,user,uts)

  # lsof -p 3958 +E
  COMMAND  PID  USER   FD   TYPE DEVICE SIZE/OFF     NODE NAME
  busybox 3958 wking  cwd    DIR   0,19       56 10149051 /
  busybox 3958 wking  rtd    DIR   0,19       56 10149051 /
  busybox 3958 wking  txt    REG   0,19  1947624 10149332 /bin/busybox
  busybox 3958 wking  mem    REG   0,18          10149332 /bin/busybox (path dev=0,19, inode=3013809)
  busybox 3958 wking    0u   CHR  136,6      0t0        9 /6
  busybox 3958 wking    1u   CHR  136,6      0t0        9 /6
  busybox 3958 wking    2u   CHR  136,6      0t0        9 /6
  busybox 3958 wking   10u   CHR    5,0      0t0       33 /dev/tty

From the DEVICE column, we can see that its standard streams are wired
up to /dev/pts/6.

Here's BusyBox running with process.terminal set to false:

  $ sudo lsof -p 4196 +E
  COMMAND  PID  USER   FD   TYPE DEVICE SIZE/OFF     NODE NAME
  runc    4192  root    3r  FIFO   0,10      0t0    18132 pipe 4192,runc,4w 4196,busybox,0r
  runc    4192  root    4w  FIFO   0,10      0t0    18132 pipe 4192,runc,3r 4196,busybox,0r
  runc    4192  root    5r  FIFO   0,10      0t0    18133 pipe 4192,runc,6w 4196,busybox,1w
  runc    4192  root    6w  FIFO   0,10      0t0    18133 pipe 4192,runc,5r 4196,busybox,1w
  runc    4192  root    7r  FIFO   0,10      0t0    18134 pipe 4192,runc,8w 4196,busybox,2w
  runc    4192  root    8w  FIFO   0,10      0t0    18134 pipe 4192,runc,7r 4196,busybox,2w
  busybox 4196 wking  cwd    DIR   0,19       56 10149051 /
  busybox 4196 wking  rtd    DIR   0,19       56 10149051 /
  busybox 4196 wking  txt    REG   0,19  1947624 10149332 /bin/busybox
  busybox 4196 wking  mem    REG   0,18          10149332 /bin/busybox (path dev=0,19, inode=3013809)
  busybox 4196 wking    0r  FIFO   0,10      0t0    18132 pipe 4192,runc,3r 4192,runc,4w
  busybox 4196 wking    1w  FIFO   0,10      0t0    18133 pipe 4192,runc,5r 4192,runc,6w
  busybox 4196 wking    2w  FIFO   0,10      0t0    18134 pipe 4192,runc,7r 4192,runc,8w

So you can see that its standard streams are wired up to runc's 3/4,
5/6, and 7/8, and that runC (as of v0.0.3) needs to do a better job
opening one-directional streams for this purpose ;).

[1]: https://github.com/opencontainers/specs/pull/87/files#r38044364
[2]: https://github.com/opencontainers/specs/pull/113#issuecomment-135639909

avagin: "lsof -p 4196 +E" enumerates all processes, so this operation can require dozen seconds on hosts with 20K processes. Any solutions which require to enumerate all tasks in procfs doesn't work, because a container should be frozen for this time.

> Because it's an unnamed pipe associated with a process outside of the
container (see my ‘lsof -p {pid} +E’ for how to find the other
processes associated with the pipe [1]).

If we will know the other processes associated with the pipe, how does this help us to restore the pipe? Our idea is that "Start" and "Restore" operations should be symmetrical. If we create descriptors before starting a container, we need to create the same set of descriptors before restoring the container.

>So you can see that its standard streams are wired up to runc's 3/4,
5/6, and 7/8, and that runC (as of v0.0.3) needs to do a better job
opening one-directional streams for this purpose ;).

Do you suggest to remember this mappings for each version of runC? And this approach doesn't work for cases when we set a few extra descriptors to Process.

Let's I show another example and you will try to explain how to handle it without externalFds.
```
conn, err := net.Dial("unix", "/monitor.sock")
sk, err := conn.File()

// runc start
process := libcontainer.Process{
    Args: ...,
    ExtraFiles: []*os.File{sk},
}
err = container.Start(&process)
```

```
$ ls -l /proc/2573/fd/385
lrwx------ 1 root root 64 Aug 28 12:01 /proc/2573/fd/385 -> socket:[3197260]
lrwx------ 1 root root 64 Aug 28 12:01 /proc/2573/fd/389 -> socket:[3197260]
lrwx------ 1 root root 64 Aug 28 12:02 /proc/2573/fd/405 -> socket:[3197298]
```

"externalFds" : [" socket:[3197260]"]

We create a socket, set it as a third descriptor for a process. Then  a process creates one more socket for it's own needs. /monitor.sock is unreachable from a container.

Then we decided to dump this container. CRIU will check that it know how to restore all sockets. It's possible to restore socket:[3197298], because it has been created inside CT, but  socket:[3197260] should be marked as external. Currently we get this information from externalFds.

Then we prepare a new socket and restore a container. All descriptors which were linked with  "socket:[3197260]" will be linked with the new socket.

```
// runc checkpoint
container, err := getContainer(context)
container.Checkpoint(options);
```

```
conn, err := net.Dial("udp", "monitor:8080")
sk, err := conn.File()

// runc restore
process := libcontainer.Process{
    Args: ...,
    ExtraFiles: []*os.File{sk},
}
err = container.Restore(&process, checkpointOpts)
```

wking: On Fri, Aug 28, 2015 at 02:53:56AM -0700, Andrew Vagin wrote:
> "lsof -p 4196 +E" enumerates all processes, so this operation can
> require dozen seconds on hosts with 20K processes. Any solutions
> which require to enumerate all tasks in procfs doesn't work, because
> a container should be frozen for this time.

Ah, I wasn't sure how it was working.  It makes sense that we might
want to cache information that would let us avoid that loop.  Although
it seems like the useful information to record would be what we know
about the other end of the file descriptor (e.g. “18132 pipe
4192,runc,3r 4192,runc,4w” in [1]), not information that can be
extracted directly from the application's PID (e.g. “pipe:[18132]” in
[1]).  But maybe the important information in externalFds is just the
order and fact that a given file descriptor was passed from the
launching runtime?

> lrwx------ 1 root root 64 Aug 28 12:01 /proc/2573/fd/385 -> socket:[3197260]
> lrwx------ 1 root root 64 Aug 28 12:01 /proc/2573/fd/389 -> socket:[3197260]
> lrwx------ 1 root root 64 Aug 28 12:02 /proc/2573/fd/405 -> socket:[3197298]
>
> "externalFds" : [" socket:[3197260]"]
>
> We create a socket, set it as a third descriptor for a process. Then
> a process creates one more socket for it's own needs. /monitor.sock
> is unreachable from a container.
> 
> Then we decided to dump this container. CRIU will check that it know
> how to restore all sockets. It's possible to restore
> socket:[3197298], because it has been created inside CT, but
> socket:[3197260] should be marked as external. Currently we get this
> information from externalFds.
> 
> Then we prepare a new socket and restore a container. All
> descriptors which were linked with "socket:[3197260]" will be linked
> with the new socket.

This is the tricky part.  How does the restoring application know to
create that socket?  For things like #113's additional file
descriptors, you may have access to some user-supplied sockets, etc.,
and you may also have some runtime-created pipes that you intend to
connect to the application's standard streams.  How do you know which
of those should replace socket:[3197260]?  I'd rather see something
that tagged the supplied objects:

  "externalFds": {
    "stdin": "pipe:[18132]",
		"stdout": "pipe:[18133]",
		"stderr": "pipe:[18134]",
		"socket": "socket:[3197298]"
  }

where the standard stream keys would default to being created by the
runtime (when process.terminal was false), but all keys could be set
by the runtime-caller on invocation.  For example, if you were passing
a systemd-created socket through to runc on fd 8, you'd use:

  runc start --file-descriptor socket:8

and runc would inspect file descriptor 8, see that it was
socket:[3197298], and write that entry in externalFds.

On restore time, maybe you were passing the systemd-created socket
through to runc on fd 3, and you'd use:

  runc restore --file-descriptor socket:4

and whatever you passed in would be wired up where socket:[3197298]
had been on the checkpointed container.

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r38173424

avagin: > But maybe the important information in externalFds is just the
order and fact that a given file descriptor was passed from the
launching runtime

externelFds defines correspondence between descriptor numbers and file objects in a moment of start.

"externalFds": {"pipe:[18132]", "pipe:[18132]", "pipe:[18132]", "socket[123456]"}
This means that a process was executed with stdout:"pipe:[18132]", ..., 3:"socket[123456]"

> How does the restoring application know to create that socket?
The user knows how to start a process and how to prepare descriptors for it. If he wants to dump and restore the process, he must know how to create these descriptors in this case too.

> How do you know which of those should replace socket:[3197260]?

Lets imagine that we start a process with one extra descriptor:
```go
process := libcontainer.Process{
    Args: ...,
    Stdin: stdin,
    Stdout: stdout,
    Stderr: stderr,
    ExtraFiles: []*os.File{sk},
}
err = container.Start(&process)
```

Now we have:
  "externalFds": {
    "stdin": "pipe:[18132]",
    "stdout": "pipe:[18133]",
    "stderr": "pipe:[18134]",
    "socket": "socket:[3197260]"
  }

Then we make dump and restore this CT. On restore we need to set new descriptors for stdin, stdout, stderr ans sk.

```go
process := libcontainer.Process{
    Args: ...,
    Stdin: stdin_new,
    Stdout: stdout_new,
    Stderr: stderr_new,
    ExtraFiles: []*os.File{sk_new},
}
err = container.Restore(&process)
```
ExtraFiles doesn't include stdin, stdout, stderr, so the full list of descriptors looks like this:

Files: {stdin_new, stdout_new, stderr_new, sk_new}

"CRIU restore" will restore Files[i] instead of externalFds[i].

crosbymichael: Thanks for the clarification.

crosbymichael:
@vbatts @philips I updated this PR. PTAL

runtime.md
```diff
@@ -1,5 +1,47 @@
...
+* **root** (string) Root is the path to the container's bundle directory.
+
+The ID is provided in the state because hooks will be executed with the state as the payload.
+This allows the hook to perform clean and teardown logic after the runtime destroys it's own state.
```
mrunalp: typo: its

runtime_config_linux.go
```diff
@@ -195,3 +198,12 @@ type Syscall struct {
 	Action Action `json:"action"`
 	Args   []*Arg `json:"args"`
 }
+
+type LinuxState struct {
```
wking: You're defining `State` in `config.go` (part of the `Spec` tree) and `LinuxState` in `runtime_config_linux.go` (part of the `RuntimeSpec` tree).  I'd [still prefer a separate `state.go` and `state_linux.go`][1], but if you want to [avoid adding more files][2], I'd shift `State` to `runtime_config.go`.

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-128137008
[2]: https://github.com/opencontainers/specs/pull/87#issuecomment-128137602

runtime.md
wking: In fact, I don't see why we'd need `root` at all, because it's easy to find the path to *all* container-mounted filesystems by looking in `/proc/{pid}/mountinfo`, and I don't see why the root filesystem would deserve special treatment.

runtime.md
wking: On Fri, Aug 28, 2015 at 08:24:59AM -0700, Andrew Vagin wrote:
> ExtraFiles doesn't include stdin, stdout, stderr…

Ok, I can live with that.  So the idea is that externalFds[i] tells
you where to connect the restoring runtime's FD i+2.  That doesn't
cleanly handle cases where we have gaps in the file descriptors:

  # runc config.json 123<foo

I'd still prefer something like:

  "externalFds": {
    "123": "/home/wking/foo",
  }

to:

  "externalFds": [-1, -1, -1, …, -1, "/home/wking/foo"],

to setup that high-valued file descriptor.

I'm also not clear on what the value should be (“/home/wking/foo”
above).  For examples like 123<foo, where we're passing in a plain
file's descriptory, it's reasonably clear (modulo seeks and such).
But how do we handle unnamed pipes?  Is “pipe:[18132]” the read or
write end of that pipe?  What if the client is duping descriptors?
What if we pass a Unix socket in, and then (after start-time), pass
*another* file descriptor down that socket with sendmsg?  What if the
client has a listening socket and has accepted a new connection?

I'm having trouble imagining a system that works for all cases.
That's fine, but I'd like externalFds to land with docs that describes
its intended value clearly, and lays out the logic for deciding
whether or not a particular application's file descriptor
manipulations will break the checkpoint/restore functionality.
Linking to CRIU docs would be fine for the parts of this that are
covered in the CRIU docs.

avagin: > So the idea is that externalFds[i] tells you where to connect the restoring runtime's FD i+2.

... where to connect the restoring runtime's FD [i].

> That doesn't cleanly handle cases where we have gaps in the file descriptors

The libcontainer interface doesn't assume gaps in the file descriptors
https://github.com/golang/go/blob/92390e47d81713d13b0d9890bf8b153f142a5b8a/src/os/exec/exec.go#L82

> 

> Is “pipe:[18132]” the read or write end of that pipe?

It doesn't matter. We always can reopen it via /proc.

> What if the client is duping descriptors?

It's the main problem which we solved with help of externalFds. We saves uniq id for each file object, so on restore we can find all descriptors which were linked with this file object.

> What if we pass a Unix socket in, and then (after start-time), pass *another* file descriptor down that socket with sendmsg?  What if the client has a listening socket and has accepted a new connection?

If criu found sockets from external net namespaces and they are not marked as external, it will fail. You will know that this container can't be dumped or migrated and this operation doesn't affect the container state.
You are right. It doesn't work for all cases. 

> Linking to CRIU docs would be fine for the parts of this that are covered in the CRIU docs.

http://www.criu.org/Inheriting_FDs_on_restore
http://criu.org/External_UNIX_socket

wking: On Sat, Aug 29, 2015 at 02:58:58AM -0700, Andrew Vagin wrote:
> > So the idea is that externalFds[i] tells you where to connect the
> > restoring runtime's FD i+2.
> 
> ... where to connect the restoring runtime's FD [i].

So externalFds is all the file descriptors, but ExtraFiles is
everything 2 and up.

> > That doesn't cleanly handle cases where we have gaps in the file descriptors
> 
> The libcontainer interface doesn't assume gaps in the file
> descriptors
> https://github.com/golang/go/blob/92390e47d81713d13b0d9890bf8b153f142a5b8a/src/os/exec/exec.go#L82

This isn't a libcontainer spec ;).  I can imagine an OCS runtime that
supports sparse FD arrays, and it's trivial to support in the spec, so
why not allow it here?  If it's too much work for libcontainer to
migrate their array to a map (which seems unlikely), just have
libcontainer explain that it adds an additional continuity requirement
to the externalFds keys.

> > Is “pipe:[18132]” the read or write end of that pipe?
> 
> It doesn't matter. We always can reopen it via /proc.

Who's we?  The runtime?  CRIU?  It's probably not going to be the
application ;).

> > Linking to CRIU docs would be fine for the parts of this that are
> > covered in the CRIU docs.
> 
> http://www.criu.org/Inheriting_FDs_on_restore
> http://criu.org/External_UNIX_socket

While I'm personally interested in those docs, I was asking for docs
in the PR itself (“but I'd like externalFds to land with docs…” [1]).
Nobody will ever come back and read this thread after this PR lands,
but I imagine folks will still want to understand the intended
semantics for externalFds and any implementation limitations.

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r38260688

runtime.md
vbatts: >
> In fact, I don't see why we'd need root at all, because it's easy to find
the path to all container-mounted filesystems by looking in
/proc/{pid}/mountinfo, and I don't see why the root filesystem would
deserve special treatment.

That is assuming that root has been mounted?

runtime.md
avagin: > So externalFds is all the file descriptors, but ExtraFiles is everything 2 and up.

yes

> This isn't a libcontainer spec ;).

I am not an author of this pull request;))

> I can imagine an OCS runtime that
supports sparse FD arrays, and it's trivial to support in the spec, so
why not allow it here?  If it's too much work for libcontainer to
migrate their array to a map (which seems unlikely), just have
libcontainer explain that it adds an additional continuity requirement
to the externalFds keys.

I think your understanding of externalFds is incomplete. The index in this array isn't interested for us. Only the order of elements is impotant. We generate two arrays of externalFds. The first one is generated when a container is started, and the second one is generated when a container is restored. We use these array to map elements from the first array to the second array (f[i] -> s[i]).
ExternalFds can be continuous even when FD array is sparce.

> Who's we?  The runtime?  CRIU?  It's probably not going to be the application ;).

CRIU will restore a proper end of pipe for each descriptor.

runtime.md
philips: @vbatts Can we get the host path to the bundle root from /proc/{application pid}/mountinfo? I think that is the point of this field.

philips: @crosbymichael can you clarify if this is the root of the bundle or the rootfs of the bundle? We should probably call it bundlepath or rootfs or something to clarify.

philips:
Remaining todos as far as I can see:

- [ ] Provide extreme clarity on how to handle the externalFDs stuff in the doc (or move to a follow-on PR)
- [ ] Clarify the root field as the bundle path, rootfs, etc

LGTM after that. I would suggest you drop the externalFDs in this PR because it is distracting from the main thrust and perhaps @avagin @SaiedKazemi could more clearly document the use case.

runtime.md
vbatts: @philips sorry. formatting got jacked up. That was my point in replying to @wking 

Deriving the `root` of a container from `/proc/pid/mountinfo` is not guaranteed. So this field _is_ needed.

philips: @vbatts ack, ok, all we need now is for @crosbymichael to clarify where this actually points in the bundle

wking: On Sat, Aug 29, 2015 at 09:36:51AM -0700, Vincent Batts wrote:
> Deriving the `root` of a container from `/proc/pid/mountinfo` is not
> guaranteed. So this field _is_ needed.

You mentioned root needing to be mounted for this [1], but I'm not
clear on what that means.  In what situation would the container's
root not be mounted?  For example, here's a container with a mount
namespace:

  $ cat /proc/6468/mountinfo 
  67 41 0:18 /home/…/oci-gentoo-minimal/rootfs / ro,noatime - btrfs /dev/sda4 rw,ssd,space_cache
  …

showing that the container's root is at the host's
/home/…/oci-gentoo-minimal/rootfs.

And here's a container without a mount namespace:

  $ cat /proc/6304/mountinfo 
  18 0 0:18 / / rw,noatime - btrfs /dev/sda4 rw,ssd,space_cache
  …

showing that the container's root is the same as my host's root.
Although without a mount namespace, I don't know what the
config.json's ‘root’ setting does (this is similar to, but not the
same as my not thinking that we should require a rootfs directory in
the bundle [2]).

[1]: https://github.com/opencontainers/specs/pull/87/files#r38261424

runtime.md
wking: On Sat, Aug 29, 2015 at 05:34:30AM -0700, Andrew Vagin wrote:
> > This isn't a libcontainer spec ;).
> 
> I am not an author of this pull request;))

True ;).  I agree with @philips [1] that it would be good to drop
externalFds from this pull request, get clarification on the
platform-independent fields, and get that merged.  Then we can land
the externalFds via a follow-up pull request that just focuses on that
issue, explaining how it's supposed to work and linking to the CRIU
docs for caveats and additional details.

> I think your understanding of externalFds is incomplete…

Before this conversation it was nonexistent, so at least I'm moving in
the right direction ;).  Hopefully we can boil my confusions down into
docs that will help future newcomers avoid the same traps.

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-136007462

crosbymichael:
Updated

wking:
On Wed, Sep 02, 2015 at 11:16:05AM -0700, Michael Crosby wrote:
> Updated

With a79d565→180df9d, LinuxState and ExternalFds have been punted.
But I think we still need resolution on:

* whether we need to record the path to the bundle root [1,2], and if
  so, what it might be used for.

* whether we need the /containers/ namespacing for
  `/run/oci/containers/<containerID>/state.json`, and if so, under
  what conditions would an application chose something other than
  /containers/ [3].

* whether it makes sense to put State in config.go (part of the Spec
  tree) and LinuxStateDirectory in runtime_config_linux.go (part of
  the RuntimeSpec tree). I'd still prefer a separate state.go and
  state_linux.go, but if you want to avoid adding more files, I'd
  shift State to runtime_config.go [4].

All the other changes I'm interested in are wording changes that don't
impact the semantics, so I'm happy to PR them separately after this
lands.

[1]: https://github.com/opencontainers/specs/pull/87/files#diff-b84a8d65d8ed53f4794cd2db7e8ea731R15
[2]: https://github.com/opencontainers/specs/pull/87/files#r38269006
[3]: https://github.com/opencontainers/specs/pull/87#discussion_r38045268
[4]: https://github.com/opencontainers/specs/pull/87#discussion_r38259067

mrunalp:
LGTM ping @vbatts @philips

vbatts:
LGTM

PR #82, opened 2015-07-28
README: add a rule for paragraph formatting in markdown

philips:
This is something we learned in the appc spec: collaborating on english prose in plain-text with git is hard, and you need rules to make it easier.

Issue #70, opened 2015-07-15
Add rootfs Mount Propagation 


rootfs:
Open up new discussion from #56 

Rootfs mount propagation deserves a new spec to set rootfs's mount propagation to slave, private, or shared.

philips:
Closing this. I think we all agree we need this, we just need to add it to the spec_linux.go.

PR #89, opened 2015-07-30
Add security section 

mrunalp:
This is WIP for adding security settings as discussed in the meeting.

The settings are based on security profiles in nsinit and Seccomp configuration in https://github.com/opencontainers/runc/pull/70

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -143,7 +143,7 @@ the container. For more information, see the [kernel cgroups documentation](http
 ## Linux capabilities
 
 Capabilities is an array that specifies Linux capabilities that can be provided to the process
-inside the container. Valid values are the string after `CAP_` for capabilities defined 
+inside the container. Valid values are the string after `CAP_` for capabilities defined
 in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html)
 
 ```json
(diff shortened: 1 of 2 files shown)


config-linux.md
```diff
@@ -148,5 +134,60 @@ rootfsPropagation sets the rootfs's mount propagation. Its value is either slave
 
 ## Security
 
-**TODO:** security profiles
+### Linux capabilities
+
+Capabilities is an array that specifies Linux capabilities that can be provided to the process
+inside the container. Valid values are the string after `CAP_` for capabilities defined
```
wking: Probably a good time to migrate these to the new Markdown paragraph format (#82).

config-linux.md
```diff
@@ -148,5 +134,60 @@ rootfsPropagation sets the rootfs's mount propagation. Its value is either slave
...
+            "AUDIT_WRITE",
+            "KILL",
+            "NET_BIND_SERVICE"
+        ]
+    }
+```
+
+### Selinux process label
+
+Selinux process label specifies the label with which the processes in a container are run.
```
wking: I think this would read easier as:

> The SELinux label under which a container's processes are run.

config-linux.md
```diff
@@ -148,5 +134,60 @@ rootfsPropagation sets the rootfs's mount propagation. Its value is either slave
...
+For more information about SELinux, see  [Selinux documentation](http://selinuxproject.org/page/Main_Page)
```
wking: A more specific link would be nice.  It might be obvious to SELinux users (I'm not one), but the main page didn't say anything about process labels.

wking:
On Thu, Jul 30, 2015 at 01:46:35PM -0700, Mrunal Patel wrote:
> The settings are based on security profiles in nsinit and Seccomp
> configuration in #70

I think you meant opencontainers/runc#70.

I realize this spec isn't going to be a tutorial on container
security, but linking to something more specific than each tool's main
page would be nice (although I also like having main-page links).  In
particular, notes about what sort of strings are valid would be
welcome (like the current the CAP_ note for Linux capabilities).  So
there's no need to explain the security model, but we do need to
explain the mapping from the security model to the configuration JSON.

For example, the ‘3’ value for the
security.seccomp.syscalls[name=getcwd].action, seems like it's the
decimal version of the SECCOMP_RET_ACTION portion of SECCOMP_RET_TRAP.
It would be nice to allow string forms there (`"action": "trap"`?),
but if we're going to have folks matching kernel #defines, we should
explain the mapping.

mrunalp:
@wking Yeah, I will flesh out the Seccomp section some more.

philips:
I like adding the apparmor, selinux, seccomp and capabilities stuff but I don't see any reason to group them in a security section. rlimits may be thought of as a security mechanism to avoid denial-of-service to the rest of the system, for example.

mrunalp:
Got rid of the separate Security struct.

philips:
Overall this change makes sense to me.

config-linux.md
```diff
@@ -146,7 +146,39 @@ rootfsPropagation sets the rootfs's mount propagation. Its value is either slave
...
+   "apparmorProfile": "acme_secure_profile"
+```
+
+## Seccomp
+
+Seccomp provides application sandboxing mechanism in the Linux kernel.
+Seccomp configuration allows one to configure actions to take for matched syscalls and furthermore also allows
+matching on values passed as arguments to syscalls.
+For more information about Seccomp, see [Seccomp kernel documentation](https://www.kernel.org/doc/Documentation/prctl/seccomp_filter.txt)
+The actions are strings that match the definitions in seccomp.h and are translated to corresponding values.
```
lizf-os: There are two seccomp.h that are visible to users. One is exported by linux kernel and the other is defined by libseccomp. I think we should mention libseccomp?

mrunalp:
@lizf-os Updated to mention that we use the header from libseccomp.

LK4D4:
@mrunalp IIRC we also have `MountLabel` for SELinux, should it be part of `specs.Mount`?

mrunalp:
@LK4D4 Yes, I think we should probably add that as well. I think I will do a follow-on PR for that.

vbatts:
overall looks fine. rebase needed though.

mrunalp:
Rebased.

vbatts:
LGTM

crosbymichael:
LGTM

PR #93, opened 2015-08-04
Implementations page

vbatts:
Adding an implementations page.

### DIFF
--- a/implementations.md
+++ b/implementations.md
@@ -1,0 +1,16 @@
+# Implementations
+
+The following sections link to associated projects, some of which are maintained by the OCI and some of which are maintained by external organizations.
+If you know of any associated projects that are not listed here, please file a pull request adding a link to that project.
+
+## Runtime (Container)
+
+* [opencontainers/runc](https://github.com/opencontainers/runc) - Reference implementation of OCI runtime
+
+## Runtime (Virtual Machine)
+
+* [hyperhq/runv](https://github.com/hyperhq/runv) - Hypervisor-based runtime for OCI
+
+## Testing
+
+* [huawei-openlab/oct](https://github.com/huawei-openlab/oct) - Open Container Testing framework for OCI configuration and runtime
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@ Table of Contents
 - [Container Configuration](config.md)
   - [Linux Specific Configuration](config-linux.md)
 - [Runtime and Lifecycle](runtime.md)
+- [Implementations](implementations.md)
 
 ## Use Cases
 


implementations.md
```diff
@@ -0,0 +1,13 @@
+# Implementations of Container Specification
+
+## Runtime
+
+* [opencontainers/runc](https://github.com/opencontainers/runc) - reference implementation of OCF runtime
```
wking: “OCF” → “OCI”.

implementations.md
```diff
@@ -0,0 +1,13 @@
...
+* [hyperhq/runv](https://github.com/hyperhq/runv) - Hypervisor-based runtime for OCF
+* [huawei-openlab/ocp-testing](https://github.com/huawei-openlab/ocp-testing) - Testing framework for OCF runtime
```
wking: I'd put this in a separate section (`## Testing`?), since it's not an implementation of the runtime.

implementations.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Configuration
```
wking: What is this this section about?  Maybe add a short into unpacking “Configuration”.  Maybe it's about tools to help authors write configurations, in which case I'd point out `runc spec` directly (the rest of `runc` does not seem particularly relevant to authoring bundles, except in its role as a runtime for testing the bundle.

vbatts: We con drop it for now. In my mind they were different aspects, but I realize their basically overlapped for now.

vbatts:
updated. PTAL

implementations.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Runtime
+
+* [opencontainers/runc](https://github.com/opencontainers/runc) - reference implementation of OCI runtime
+* [hyperhq/runv](https://github.com/hyperhq/runv) - Hypervisor-based runtime for OCI
+
+## Testing
+
+* [huawei-openlab/ocp-testing](https://github.com/huawei-openlab/ocp-testing) - Testing framework for OCI configuration and runtime
+
```
wking: Extra blank line.

implementations.md
```diff
@@ -0,0 +1,11 @@
+# Implementations of Container Specification
+
+## Runtime
+
+* [opencontainers/runc](https://github.com/opencontainers/runc) - reference implementation of OCI runtime
+* [hyperhq/runv](https://github.com/hyperhq/runv) - Hypervisor-based runtime for OCI
+
+## Testing
+
+* [huawei-openlab/ocp-testing](https://github.com/huawei-openlab/ocp-testing) - Testing framework for OCI configuration and runtime
```
wking: Does it test configs independent of the runtime?  That would be a nice feature to have, but I'd be surprised if ocp-testing had a static config validator.  If it doesn't, I'd suggest just “Testing framework for OCI runtimes”.

implementations.md
```diff
@@ -0,0 +1,11 @@
+# Implementations of Container Specification
+
+## Runtime
+
+* [opencontainers/runc](https://github.com/opencontainers/runc) - reference implementation of OCI runtime
+* [hyperhq/runv](https://github.com/hyperhq/runv) - Hypervisor-based runtime for OCI
```
wking: For consistency, probably capitalize the first word after the dash in each case (so “Reference implementation of OCI runtime” for runC).  And if we're setting runV apart based on the underlying technology, it would be nice to list runC's supported systems too, so folks know when to chose one over the other.

wking:
On Tue, Aug 04, 2015 at 11:01:31AM -0700, Vincent Batts wrote:
> updated. PTAL

I added a few more pedantic comments.  But I like this PR, and am fine
with it landing in any form with ≥ it's current level of polish ;).

implementations.md
vbatts: Having not run it, I can't confirm. But their page says "The testing includes: ocp specification, container function and container performance."

vbatts:
Updated (again ;-). PTAL

crosbymichael:
LGTM

implementations.md
wking: On Tue, Aug 04, 2015 at 11:14:27AM -0700, Vincent Batts wrote:
> > +* [huawei-openlab/ocp-testing](https://github.com/huawei-openlab/ocp-testing) - Testing framework for OCI configuration and runtime
> 
> Having not run it, I can't confirm. But their page says "The testing
> includes: ocp specification, container function and container
> performance."

I haven't run the tester either, but from skimming [1] it seemed like
the specs tests were “is this runtime compatible with
opencontainers/specs”, not “is this bundle-specification compatible
with opencontainers/specs”.

@liangchenye, @zenlinTechnofreak, @ChengTiesheng, @maquanyi and other
@huawei-openlab folks, can you clarify?

[1]: https://github.com/huawei-openlab/ocp-testing/tree/master/Cases/Specstest

mrunalp:
Maybe squash :)

implementations.md
```diff
@@ -0,0 +1,13 @@
+# Implementations of Container Specification
```
wking: Maybe just “Implementations” here?  The fact that we're talking about this specification should be clear from the context.

mrunalp: +1

vbatts: done

vbatts:
done

implementations.md
liangchenye: We are working on  “is this runtime compatible with opencontainers/specs”.

The 'bundle-specification compatibility' is not planed yet.

zenlinTechnofreak: @vbatts  @wking @crosbymichael @mrunalp  Glad to have advise from you nice guys, any suggestion and considering are welcomed, expect more advices...
Thanks.

philips: Looking through the code I don't think that the ocp-testing implements what we need for testing the open container spec and we shouldn't point people at it for now. For example see this [source file](https://github.com/huawei-openlab/ocp-testing/blob/master/Cases/Benchmark/monitor/cpumonitor/source/cpu_usage_breakdown.go)

Things I would like to see on an OCI testing tool:

- Validate the configuration of a bundle
- Generate and test a bundle/image cryptographic identity
- Generates an OCI bundle that tests that the runtime is compliant
- A tool for testing fetching/unpacking of OCI images

And these things are things we should probably add to this repo as unit tests for the specification.

Right now this seems to do a lot of testing of docker, which is great, but it isn't directly related to what we are doing with the spec at the moment and is confusing.

zenlinTechnofreak: @philips thanks. 
The project containers three parts, OCI specs test, container's network test and perfomace test.
the [source file](https://github.com/huawei-openlab/ocp-testing/blob/master/Cases/Benchmark/monitor/cpumonitor/source/cpu_usage_breakdown.go) you mationed  is testcase for perfomace but not for spec test.
Could u go through the [testcase](https://github.com/huawei-openlab/ocp-testing/tree/master/Cases/Specstest/root_readonly_true).

It do testing as below steps,
- Set the config.json item linuxspec.Spec.Root.Readonly = true,
- Run runc used the config.json
- Check weather the root dir in runc is readonly.

Once u go through it, please feedback to me if convinient.

zenlinTechnofreak:
I newed a issuse to talk about How to do bundle compatibility test? #78 
Expect anyone who are convenience to give some suggestion, thanks.

wking:
See also the discussion that lead to this PR (which also has a
discussion about whether or not this GitHub organization should
contain *any* implementation repositories):
https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Tjq5QJ_eez0

In case that link dies, or folks want to look up the conversation in
their local mail archives, here's some information from the message
spawning that thread:

  Date: Mon, 3 Aug 2015 16:20:04 +0800
  Message-ID: <CAHKD89TqBQvbcdvwt-g3J--sDyq=P50qi6CytokNJznVsei7YQ@mail.gmail.com>
  Subject: Contribute to OCI
  From: Thibault Bronchain
  To: dev@opencontainers.org

liangchenye:
Hi @philips , we are testing docker, rkt (and will test other runtimes which implement appc spec).
The rkt testing is on going, we plan to ship a first 0.1 version on 8.15th.

laijs:
`huawei-openlab/ocp-testing` should be changed to `huawei-openlab/oct`

LGTM

vbatts:
updated. PTAL

implementations.md
```diff
@@ -0,0 +1,18 @@
+# Implementations
+
+The following sections link to associated projects, some of which are
+maintained by the OCI and some of which are maintained by external
+organizations.  If you know of any associated projects that are not listed
+here, please file a pull request adding a link to that project.
```
wking: This paragraph should be re-wrapped to follow #82.

vbatts: done

mrunalp:
LGTM

implementations.md
wking: Looks good to me :).

philips:
LGTM

PR #94, opened 2015-08-05
Replace Linux.Device with more specific config

LK4D4:
It appeared that we need more control over devices, for example different permissions and choosing path on caller side.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -55,20 +55,82 @@ within the container.
 
 ### Access to devices
 
-Devices is an array specifying the list of devices from the host to make available in the container.
-By providing a device name within the list the runtime should look up the same device on the host's `/dev`
-and collect information about the device node so that it can be recreated for the container.  The runtime
-should not only create the device inside the container but ensure that the root user inside 
-the container has access rights for the device.
+Devices is an array specifying the list of devices to be created in the container.
+Next parameters can be specified:
+
+* type - type of device: 'c', 'b', 'u' or 'p'. More info in `man mknod`
+* path - full path to device inside container
+* major, minor - major, minor numbers for device. More info in `man mknod`.
+                 There is special value: `-1`, which means `*` for `device`
+                 cgroup setup.
+* permissions - cgroup permissions for device. A composition of 'r'
+                (read), 'w' (write), and 'm' (mknod).
+* fileMode - file mode for device file
+* uid - uid of device owner
+* gid - gid of device owner
 
 ```json
    "devices": [
-        "null",
-        "random",
-        "full",
-        "tty",
-        "zero",
-        "urandom"
+        {
+            "path": "/dev/random",
+            "type": "c",
+            "major": 1,
+            "minor": 8,
+            "permissions": "rwm",
+            "fileMode": 0666,
+            "uid": 0,
+            "gid": 0
+        },
+        {
+            "path": "/dev/urandom",
+            "type": "c",
+            "major": 1,
+            "minor": 9,
+            "permissions": "rwm",
+            "fileMode": 0666,
+            "uid": 0,
+            "gid": 0
+        },
+        {
+            "path": "/dev/null",
+            "type": "c",
+            "major": 1,
+            "minor": 3,
+            "permissions": "rwm",
+            "fileMode": 0666,
+            "uid": 0,
+            "gid": 0
+        },
+        {
+            "path": "/dev/zero",
+            "type": "c",
+            "major": 1,
+            "minor": 5,
+            "permissions": "rwm",
+            "fileMode": 0666,
+            "uid": 0,
+            "gid": 0
+        },
+        {
+            "path": "/dev/tty",
+            "type": "c",
+            "major": 5,
+            "minor": 0,
+            "permissions": "rwm",
+            "fileMode": 0666,
+            "uid": 0,
+            "gid": 0
+        },
+        {
+            "path": "/dev/full",
+            "type": "c",
+            "major": 1,
+            "minor": 7,
+            "permissions": "rwm",
+            "fileMode": 0666,
+            "uid": 0,
+            "gid": 0
+        }
     ]
 ```
 
--- a/spec_linux.go
+++ b/spec_linux.go
@@ -2,6 +2,8 @@
 
 package specs
 
+import "os"
+
 // LinuxSpec is the full specification for Linux containers
 type LinuxSpec struct {
 	Spec
@@ -27,7 +29,7 @@ type Linux struct {
 	// Capabilities are Linux capabilities that are kept for the container
 	Capabilities []string `json:"capabilities"`
 	// Devices are a list of device nodes that are created and enabled for the container
-	Devices []string `json:"devices"`
+	Devices []Device `json:"devices"`
 	// RootfsPropagation is the rootfs mount propagation mode for the container
 	RootfsPropagation string `json:"rootfsPropagation"`
 }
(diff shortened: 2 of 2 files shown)


config-linux.md
```diff
@@ -55,20 +55,20 @@ within the container.
...
 ```json
    "devices": [
-        "null",
-        "random",
-        "full",
-        "tty",
-        "zero",
-        "urandom"
+        {
+            "path": "/dev/random",
```
crosbymichael: We should probably specify or describe in this doc what each of the fields mean.  But overall this PR looks good

LK4D4: I'll combine description from docstrings.

philips: Is there ever a reason not to include null, random, full, tty, zero and urandom? It seems like we should just say every container can expect to have those device nodes available without specifying.

philips:
LGTM overall, I do think we should consider making the "obvious" device nodes available as they are a pretty accepted part of the POSIX API.

config-linux.md
LK4D4: Ah, okay, thought about this just like about json example, not real config :) I'll add all from runc defaults.

philips: @LK4D4 Yes, I will file a separate issue on this one.

wking:
On Wed, Aug 05, 2015 at 03:26:47PM -0700, Alexander Morozov wrote:
> It appeared that we need more control over devices, for example
> different permissions and choosing path on caller side.

Can you elaborate on this or provide a concrete example?  I'd guess
it's something like “I translate my UIDs so my container root isn't my
host root, but I still need access to a root-only host device”, but
that seems like it's handled by the current “The runtime should not
only create the device inside the container but ensure that the root
user inside the container has access rights for the device” phrasing.
Any changes you need to make after that can be handled by the
application itself (or a pre-start hook).  If it turns out that folks
do need this sort of detailed device control, can't they just populate
rootfs/dev in their bundled filesystem?

config-linux.md
philips: @LK4D4 filed https://github.com/opencontainers/specs/issues/95

spec_linux.go
```diff
@@ -157,3 +159,22 @@ type Resources struct {
...
+	// Major is the device's major number.
+	Major int64 `json:"major"`
+	// Minor is the device's minor number.
+	Minor int64 `json:"minor"`
+	// Cgroup permissions format, rwm.
+	Permissions string `json:"permissions"`
+	// FileMode permission bits for the device.
+	FileMode os.FileMode `json:"fileMode"`
+	// Uid of the device.
+	Uid uint32 `json:"uid"`
```
mrunalp: UID

spec_linux.go
```diff
@@ -157,3 +159,22 @@ type Resources struct {
...
+	// Gid of the device.
+	Gid uint32 `json:"gid"`
```
mrunalp: GID

LK4D4:
@wking For example we want different cgroups permissions for devices. Another case is to creating device in paths which doesn't exists on host. Also this interface maps pretty well to device creation process on the host, I don't see why we should try to cut it off.
PTAL @crosbymichael @mrunalp @philips

philips:
I don't think a full exhaustive list of many devices is necessary in the spec to illustrate but it LGTM.

config-linux.md
```diff
@@ -55,20 +55,80 @@ within the container.
...
 
-Devices is an array specifying the list of devices from the host to make available in the container.
-By providing a device name within the list the runtime should look up the same device on the host's `/dev`
-and collect information about the device node so that it can be recreated for the container.  The runtime
-should not only create the device inside the container but ensure that the root user inside 
-the container has access rights for the device.
+Devices is an array specifying the list of devices to be created in the container.
+Next parameters can be specified:
+
+* type - type of device: 'c', 'b', 'u' or 'p'. More info in `man mknod`
```
laijs: From my instinct,  the abbreviation of r/w/m is OK for permissions. But here, I think we should use "character", "block" and "fifo". I don't have strong reason, just feeling.

mrunalp: I think this is okay, as more information could be looked up in the man page.

wking:
On Wed, Aug 05, 2015 at 05:37:28PM -0700, Alexander Morozov wrote:
> For example we want different cgroups permissions for devices.
> Another case is to creating device in paths which doesn't exists on
> host.

For “devices in paths which don't exist on the host”, it seems like
the easiest approach would be to just bundle those device files
(e.g. in rootfs/dev/…).

For “I want special cgroups on $PATH”, it seems like that would be
better handled by either a pre-start hook or a more generic
linux.resources entry.  Something that directly maps to [1], like:

  {
    …
    "linux": {
      …
      "resources": {
        …
        "devices": [
          {
             "type": "all",
             "major": "*",
             "minor": "*",
             "access": "rwm",
          }
        ]
      }
    }
  }

As this PR stands, I'm not sure how you'd use wildcards for
major/minor cgroups rules.

[1]: https://www.kernel.org/doc/Documentation/cgroups/devices.txt

LK4D4:
@wking I can add to description, that -1 will mean wildcard.

wking:
On Thu, Aug 06, 2015 at 10:10:30AM -0700, Alexander Morozov wrote:
> @wking I can add to description, that -1 will mean wildcard.

If you have a -1 in the spec, and the device doesn't exist on the
host, what value will you pass to mknod?  That's why I want to
separate the mknod step (which can be the current “copy from the host”
or “handled by the bundle author when building the bundled
filesystem”) from the cgroups step (which I think needs a new spec
entry) [1].

[1]: https://github.com/opencontainers/specs/pull/94#issuecomment-128445736

LK4D4:
@wking It's up to you which value you will pass, I don't like any `copy from the host` in spec, because it's not flexible at all.

wking:
On Thu, Aug 06, 2015 at 10:29:10AM -0700, Alexander Morozov wrote:
> It's up to you which value you will pass, …

No, I meant “assuming the -1 wildcard [1], if I put a -1 major value
in my bundle config, what will the runtime use when it calls mknod?”

> I don't like any `copy from the host` in spec, because it's not
> flexible at all.

I'm ok dropping that section, since bundle authors that want to copy
device nodes from the host are expecting that those copied files are
portable enough to not need further specification.  If that's the
case, they are likely to have the device files on the host that
they're using to write the bundle, and they can copy them from there
into their bundled filesystem.

Maybe it would help if I just PRed my suggestion…

[1]: https://github.com/opencontainers/specs/pull/94#issuecomment-128446624

LK4D4:
@wking Yup, example of your thoughts in code would be perfect.

wking:
On Thu, Aug 06, 2015 at 10:57:22AM -0700, Alexander Morozov wrote:
> @wking Yup, example of your thoughts in code would be perfect.

Filed as #98.

crosbymichael:
LGTM

wking:
On Thu, Aug 06, 2015 at 03:41:04PM -0700, Michael Crosby wrote:
> Merged #94.

Hmm.  Can someone clarify how the -1 wildcards [1] are supposed to
work?  I still don't see how a single major (or minor) field is
sufficient for both mknod and cgroups [2] (which is why I floated #98
and #99).

[1]: https://github.com/LK4D4/specs/commit/5273b3d7855d1c79ff254bd9f35966bdfa043ab8#diff-428eec4013a655816cdefafd5d3505f1R64
[2]: https://github.com/opencontainers/specs/pull/94#issuecomment-128457618

LK4D4:
@wking Basically, if you want to use wildcard, then you need to skip Path, because otherwise runtime will try to mknod. If Permissions is not empty - then device is allowed. That's whole relationship between devices and cgroup as I see it. I don't see a big problem if it'll be documented.

wking:
On Thu, Aug 06, 2015 at 05:32:17PM -0700, Alexander Morozov wrote:
> Basically, if you want to use wildcard, then you need to skip Path,
> because otherwise runtime will try to mknod. If Permissions is not
> empty - then device is allowed. That's whole relationship between
> devices and cgroup as I see it. I don't see a big problem if it'll
> be documented.

I started working on a PR documenting this, but I'm not sure how to
distinguish between allow and deny cgroup rules [1].  Do you have
something equivalent to [2]?

[1]: https://www.kernel.org/doc/Documentation/cgroups/devices.txt
[2]: https://github.com/wking/opencontainer-specs/commit/dfb613a9e32f1fbfdfcbaff85ac67630eb12ad18#diff-36b33634fde0342d15add700c1eac653R108

LK4D4:
@wking basically idea was to `echo a > devices.deny` and then add to `allow` only devices with non-empty permissions.

wking:
On Fri, Aug 07, 2015 at 10:10:38AM -0700, Alexander Morozov wrote:
> @wking basically idea was to `echo a > devices.deny` and then add to
> `allow` only devices with non-empty permissions.

That may be overly strict for child containers.  What if you want to
allow your children to add further allows?

wking:
I've floated some more explicit docs in #101, if we want to talk about
the semantics and docs for allow vs. deny there.

LK4D4:
@wking Maybe it's better to discuss this in IRC. It's too much for closed PR.

PR #96, opened 2015-08-06
Move User struct to spec.go

gao-feng:
Otherwise User struct will not be found on other platform

Signed-off-by: Gao feng <feng@hyper.sh>

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -55,6 +55,16 @@ type Process struct {
 	Cwd string `json:"cwd"`
 }
 
+// User specifies user and group information for the container's main process
+type User struct {
+	// Uid is the user id
+	UID int32 `json:"uid"`
+	// Gid is the group id
+	GID int32 `json:"gid"`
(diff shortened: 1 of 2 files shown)


lizf-os:
Why don't you add spec_xxx.go to support other platforms? I don't think this User struct is applicable for Windows.

gao-feng:
@lizf-os Process in spec.go references to User

dqminh:
I think it's better to have different user definition for different platforms. I think it's enough to only have `user_unix` and `user_windows`, but conflating the unix default to spec.go is not a good idea

mrunalp:
@gao-feng Thanks for your PR. However, this was discussed earlier and it made sense to have a separate per platform User struct as it doesn't seem very portable right now. When we have more implementations, we will consider abstracting it and moving it into platform independent section if it makes more sense.

mrunalp:
Closing this for now. We will reopen it in the future. Thanks!

PR #99, opened 2015-08-06
Add linux.resources.devices

wking:
This cherry-picks the device-cgroup commit from #98, since I think the
approach is useful regardless of how we're creating the devices.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -74,9 +74,60 @@ the container has access rights for the device.
 
 ## Linux control groups
 
-Also known as cgroups, they are used to restrict resource usage for a container and handle
-device access.  cgroups provide controls to restrict cpu, memory, IO, and network for
-the container. For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt)
+Also known as cgroups, they are used to restrict resource usage for a container and handle device access.
+For more information, see the [kernel cgroups documentation][cgroups].
+You can configure a container's cgroups via the "resources" field of the Linux configuration.
+
+### Disable out-of-memory killer
+
+FIXME
+
+### Memory
+
+FIXME
+
+### CPU
+
+FIXME
+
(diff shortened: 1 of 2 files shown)


wking:
And all the pre-close pushback against #98 seemed to focus on the commit dropping linux.devices.  I didn't see any comments in #98 on the commit cherry-picked here (just the discussion in #94: [1][], [2][], [3][], [4][], [5][]).

[1]: https://github.com/opencontainers/specs/pull/94#issuecomment-128445736
[2]: https://github.com/opencontainers/specs/pull/94#issuecomment-128446624
[3]: https://github.com/opencontainers/specs/pull/94#issuecomment-128450677
[4]: https://github.com/opencontainers/specs/pull/94#issuecomment-128451434
[5]: https://github.com/opencontainers/specs/pull/94#issuecomment-128457618

wking:
I was hunting around for the device docs to link today, and was surprised to find them [in the `runtime.json` tree][1].  Besides avoiding the complication of overloading (see #101), I think splitting the mknod and cgroup functionality would let us move the mknod functionality back into the `config.json` tree (it certainly doesn't seem like a host-specific setting), while leaving the cgroup functionality in `runtime.json` (since that *does* seem like a host-specific setting).

[1]: https://github.com/opencontainers/specs/blob/cbda52164745ca5a179cc04e0ee36cbdc7ef46b1/runtime-config-linux.md#devices

crosbymichael:
I think this one can be closed now that we already have a Device struct and fields documented.

wking:
On Fri, Sep 25, 2015 at 01:45:09PM -0700, Michael Crosby wrote:
> I think this one can be closed now that we already have a Device
> struct and fields documented.

Not really documented, see #101 ;).  I still think the separation
between mknod and cgroups proposed here would make everyone's life
easier than landing the #101 docs for the current structure.  But it's
your project, I'm just floating suggestions :).

PR #101, opened 2015-08-07
Additional documentation for overloaded mknod / cgroups in linux.devices

wking:
Building on #94, although I still prefer handling these features
independently (possible specs for that are in #98 and #99).

I'm still not clear on allow vs. deny for cgroups rules, but there's
some discussion after [this comment][1].

[1]: https://github.com/opencontainers/specs/pull/94#issuecomment-128769183

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -58,9 +58,9 @@ within the container.
 Devices is an array specifying the list of devices to be created in the container.
 The following parameters can be specified:
 
-* type - type of device: 'c', 'b', 'u' or 'p'. More info in `man mknod`
+* type - type of device: 'c', 'b', 'u' or 'p'. More info in [mknod(1)][]
 * path - full path to device inside container
-* major, minor - major, minor numbers for device. More info in `man mknod`.
+* major, minor - major, minor numbers for device. More info in [mknod(1)][].
                  There is special value: `-1`, which means `*` for `device`
                  cgroup setup.
 * permissions - cgroup permissions for device. A composition of 'r'
@@ -212,3 +212,4 @@ rootfsPropagation sets the rootfs's mount propagation. Its value is either slave
 
 **TODO:** security profiles
 
+[mknod(1)]: http://man7.org/linux/man-pages/man1/mknod.1.html
--- a/config-linux.md
+++ b/config-linux.md
@@ -58,7 +58,7 @@ within the container.
 Devices is an array specifying the list of devices to be created in the container.
 The following parameters can be specified:
 
-* type - type of device: 'c', 'b', 'u' or 'p'.
+* type - type of device: 'b', 'c', 'u' or 'p'.
   More info in [mknod(1)][]
 * path - full path to device inside container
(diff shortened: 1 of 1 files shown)


wking:
Discussing this with @LK4D4 on IRC, I think if we want full cgroups handling here, we'll want to add a new `allow` field for non-path entries.  That would allow you to do things like:

```json
{"type": "a", "allow": false, "major": -1, "minor": -1, "permissions": "rwm"}
```

to block all devices.  @LK4D4 also prefers [blocking all devices by default][1], but the above rule is easy enough to write that I'd prefer the default to be “don't touch device cgroups”.

[1]: https://github.com/opencontainers/specs/pull/94#issuecomment-128769183

wking:
Oh, and this PR doesn't touch the Go type comments and whatnot.  It's just for discussing the Markdown documentation.  Once we reach a consensus there, I'll go back and update the Go comments.

config-linux.md
```diff
@@ -55,28 +55,45 @@ within the container.
...
+            "major": -1,
+            "minor": -1,
+            "permissions": "rwm",
+        },
+        {
             "path": "/dev/random",
             "type": "c",
             "major": 1,
             "minor": 8,
-            "permissions": "rwm",
```
crosbymichael: why did u delete these lines?

wking: On Mon, Oct 05, 2015 at 11:09:18AM -0700, Michael Crosby wrote:
> >              "path": "/dev/random",
> >              "type": "c",
> >              "major": 1,
> >              "minor": 8,
> > -            "permissions": "rwm",
> 
> why did u delete these lines?

My first device entry set rwm permissions for all major/minor types,
so there's no need to repeat that in the later entries.

config-linux.md
```diff
@@ -55,28 +55,45 @@ within the container.
...
+* type - type of device: 'a', 'b', 'c', 'u' or 'p'.
+  Required.
+  If path is given, only 'b', 'c', 'u', and 'p' are allowed (more info in [mknod(1)][]).
+  If permissions is given, only 'a', 'b', and 'c' are allowed (more info in the [cgroups devices documentation][cgroups-devices]).
+* path - full path to device inside container.
+  Optional.
+  If not given, the entry only creates a cgroups allow rule.
+* major, minor - major, minor numbers for device.
+  More info in [mknod(1)][].
+  There is special value: `-1`, which uses the wildcard `*` for the cgroups rule.
```
mrunalp: ...is a special value..

crosbymichael:
Lets not make any additional changes if the real issues that people are having are with the creation and permissions being in one field.  This was split before and added alot of verbosity but we can revisit a better way if it's an issue.  Lets discuss on the mailing list.

wking:
On Mon, Oct 05, 2015 at 11:46:54AM -0700, Michael Crosby wrote:
> Lets discuss on the mailing list.

Spun off into https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/y_Fsa2_jJaM

PR #102, opened 2015-08-08
README: Document pre-meeting agenda alteration

wking:
On Sat, Aug 08, 2015 at 08:23:21AM -0400, Vincent Batts [wrote][1]:
> We said the topics would be proposed on this list. Discussion notes have
> been shared as a Google doc so far. I'm not opposed to markdown notes, but
> initially that just seems like an repo to me.

While touching this paragraph, I also re-wrapped it to match
README.md#markdown-style.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/41jZ9Qe7R5c/ZInajC_0DgAJ

PR #103, opened 2015-08-10
Correct err in devices json in config-linux.md

zenlinTechnofreak:
Correct err in devices json in config-linux.md, type field should be top of the path field, according to the spec_linux.go.

Signed-off-by: LinZhinan(Zen Lin) <linzhinan@huawei.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -72,8 +72,8 @@ Next parameters can be specified:
 ```json
    "devices": [
         {
-            "path": "/dev/random",
             "type": "c",
+            "path": "/dev/random",
             "major": 1,
             "minor": 8,
             "permissions": "rwm",
@@ -82,8 +82,8 @@ Next parameters can be specified:
             "gid": 0
         },
         {
-            "path": "/dev/urandom",
             "type": "c",
+            "path": "/dev/urandom",
             "major": 1,
             "minor": 9,
             "permissions": "rwm",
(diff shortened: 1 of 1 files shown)


wking:
On Mon, Aug 10, 2015 at 04:51:27AM -0700, LinZhinan(Zen Lin) wrote:
> Correct err in devices json in config-linux.md, type field should be
> top of the path field, according to the spec_linux.go.

I don't think the key order actually matters for JSON (so it's
probably not an error as it stands).  But I think we might as well be
consistent in our ordering anyway, so I'm +1 on this change (I'd also
be fine with reordering the fields in spec_linux.go).

zenlinTechnofreak:
@wking Yes, so agree with u, maybe I have mistook to exp it as a error,thanks.

zenlinTechnofreak:
@wking To your opinion, which file to be changed is more reasonable, spec_linux.go or config-linux.md.
If change the order in spec_linux.go ,I am going to change the PR.

philips:
path seems like the "unique key" in this struct so I guess we should reorder the spec_linux.go instead.

zenlinTechnofreak:
@wking @philips I have newed a PR for changing it in the spec_linux.go file.
Change Device field order in spec_linux.go #106

wking:
On Mon, Aug 10, 2015 at 09:08:22PM -0700, Brandon Philips wrote:
> path seems like the "unique key" in this struct so I guess we should
> reorder the spec_linux.go instead.

I believe path is optional now that this structure also handles
cgroups rules (#94, proposed docs in #102).  So it's not a unique key.

I really have no preference for which order we chose ;).

PR #106, opened 2015-08-11
Change Device field order in spec_linux.go

zenlinTechnofreak:
Change Device field order in spec_linux.go, 'Path' should be top of the 'Type' field, according to the different of the config-linux.md, 'Path' field is the unique key.

Signed-off-by: LinZhinan(Zen Lin) <linzhinan@huawei.com>

### DIFF
--- a/spec_linux.go
+++ b/spec_linux.go
@@ -167,10 +167,10 @@ type Resources struct {
 }
 
 type Device struct {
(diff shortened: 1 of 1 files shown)


philips:
lgtm, the path is the "primary key" to me.

mrunalp:
LGTM

PR #107, opened 2015-08-11
glossary: Provide a quick overview of important terms

wking:
And link them to the more detailed specification.

This borrows from appc, which calls the launched process tree ["the
app"][1].  I don't see a point to abbreviating it, so I've gone with
"application".

Subsection titles for the entries will be obnoxiously spacious, but
the other alternatives seem worse:

* An HTML definition list (`<dl>`) would have nice default styling, but
  it's annoying to write raw HTML.  And we would have needed
  something like:

        <dt name="bundle">Bundle</dt>
        <dd>

        A [directory structure](bundle.md) that is...

        </dd>

  to get Markdown-style links in the defintion itself.

* A Markdown list (`* …`) would have reasonable default styling, but
  there's no Markdown syntax for adding anchors to the entries.  And a
  glossary is much less useful if you can't link to a specific entry.

Spun off from #88.

[1]: https://github.com/opencontainers/specs/pull/88#discussion_r36269387

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -13,7 +13,14 @@ The [`config.json`](config.md) and [`runtime.json`](runtime-config.md) files in 
 An environment for executing processes with configurable isolation and resource limitations.
 For example, namespaces, resource limits, and mounts are all part of the container environment.
 
+## JSON
+
+All configuration [JSON][] MUST be encoded in [UTF-8][].
+
 ## Runtime
 
 An implementation of this specification.
 It reads the [configuration files](#configuration) from a [bundle](#bundle), uses that information to create a [container](#container), launches a process inside the container, and performs other [lifecycle actions](runtime.md).
+
+[JSON]: http://json.org/
+[UTF-8]: http://www.unicode.org/versions/Unicode8.0.0/ch03.pdf
--- a/glossary.md
+++ b/glossary.md
@@ -1,0 +1,19 @@
+# Glossary
+
+## Bundle
+
+A [directory structure](bundle.md) that is written ahead of time, distributed, and used to seed the runtime for creating a [container](#container) and launching a process within it.
+
+## Configuration
+
+The [`config.json`](config.md) and [`runtime.json`](runtime-config.md) files in a [bundle](#bundle) which define the intended [container](#container) and container process.
+
+## Container
+
+An environment for executing processes with configurable isolation and resource limitations.
+For example, namespaces, resource limits, and mounts are all part of the container environment.
+
+## Runtime
+
+An implementation of this specification.
+It reads the [configuration files](#configuration) from a [bundle](#bundle), uses that information to create a [container](#container), launches a process inside the container, and performs other [lifecycle actions](runtime.md).
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ Table of Contents
 - [Runtime and Lifecycle](runtime.md)
   - [Linux Specific Runtime](runtime-linux.md)
 - [Implementations](implementations.md)
+- [Glossary](glossary.md)
 
 In the specifications in the above table of contents, the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 


glossary.md
```diff
@@ -0,0 +1,22 @@
+# Glossary
+
+## Application
+
+A process tree launched inside a [container](#container).
+
+## Bundle
+
+A [directory structure](bundle.md) that is written ahead of time, distributed, and used to seed the runtime for creating [containers](#container) and launching [applications](#application).
```
philips: launching an application? A bundle only specifies launches a single application process.

wking: On Tue, Aug 11, 2015 at 09:26:58PM -0700, Brandon Philips wrote:
> > +## Bundle
> > +
> > +A [directory structure](bundle.md) that is written ahead of time, distributed, and used to seed the runtime for creating [containers](#container) and launching [applications](#application).
> 
> launching an application? A bundle only specifies launches a single
> application process.

One application (which can result in a multi-process tree) per runtime
invocation yes.  But you can use the same bundle to launch several
copies of the application.

And you can also have several config choices in a single bundle.  My
proposed docs for this are in this subsection [1] of #76.  That PR was
rejected for other reasons, but that particular subsection still
applies to the current spec and at least the runC implementation.

[1]: https://github.com/wking/opencontainer-specs/blob/13349ba471ffb5a287527bf585fdb9cbcc0d7b95/bundle.md#alternative-configurations

philips: The intention is that a single bundle launches a single application. The bundle defines a single identity (bundle directory, cgroup hierarchy, etc) and a single lifecycle (hooks, etc).

wking: On Wed, Aug 12, 2015 at 09:42:27AM -0700, Brandon Philips wrote:
> > +## Bundle
> > +
> > +A [directory structure](bundle.md) that is written ahead of time, distributed, and used to seed the runtime for creating [containers](#container) and launching [applications](#application).
> 
> The intention is that a single bundle launches a single
> application. The bundle defines a single identity (bundle directory,
> cgroup hierarchy, etc) and a single lifecycle (hooks, etc).

Do we explicitly set that “single application” limit somewhere?  I
understand we have a default config location (config.json) for the
bundle [1], but I don't see why you couldn't store additional config
files in the bundle, and have optional runtime support for launching
from one of those additional configs [2].  Allowing the default config
filename to be overridden at runtime seems both useful and easy to
implement, so I don't see an upside to requiring a
single-application-per-bundle limit.

[1]: https://github.com/opencontainers/specs/blob/7e4a8779670cebc4125d352c0ca0c582474d5048/bundle.md#directory-layout
[2]: https://github.com/opencontainers/specs/pull/107/files#r36882200

wking: On Wed, Aug 12, 2015 at 02:37:00PM -0700, W. Trevor King wrote:
> Allowing the default config filename to be overridden at runtime
> seems both useful and easy to implement…

For a concrete example, see [1], which supports both the IPFS gateway:

  # runc

and a shell for poking around:

  # runc config-sh.json

[1]: https://github.com/wking/oci-gentoo-minimal/tree/ipfs-gateway

philips: a single bundle is intended to launch a single process at this point.

wking: On Wed, Aug 26, 2015 at 07:31:50AM -0700, Brandon Philips wrote:
> a single bundle is intended to launch a single process at this point.

Is there a reason to avoid allowing multiple application specs inside
a bundle?  Besides the fact that you can do that now (at least with
runC), and that I do this already (with the above-linked IPFS gateway
bundle), we also have @julz sounding positive about this sort of
use-case [1] in the discussion that I've spun off into [2].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/BIxya5eSNLo/gQQ6PHqPFAAJ
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/kPWu-ZavQjs

julz: the Garden API (which cloud foundry uses) has separate Create and Run calls (and can run multiple processes in a container, analogously to 'runC exec' or 'docker exec'). We'd ideally like, therefore, to maintain the flexibility to do this in the spec (as Trevor mentioned, this is already existing function in runC). We can work around not having this (e.g. By using runC directly and not supporting arbitrary OCS implementations, which would be a shame, or by waiting until the Run call to start the container which'd change our current semantics a bit) but I'd like to understand the reasoning behind not supporting these use cases in a common container spec.

wking:
While [commenting on #87][1], I realized that I don't understand what happens when a non-PID-namespaced process forks/execs a child that escapes its process tree.  Is it still part of the “application” (I'm not sure)?  Is that process something that gets terminated when we stop a container (I'm not sure, but this should probably be the same answer as for “application”)?  Is it checkpointed/restored when we move a container (you could tie checkpointing into just the freezer cgroup)?

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r38045786

glossary.md
wking: On Wed, Aug 26, 2015 at 02:09:43PM -0700, W. Trevor King wrote:
> Wed, Aug 26, 2015 at 07:31:50AM -0700, Brandon Philips:
> > a single bundle is intended to launch a single process at this
> > point.
> 
> Is there a reason to avoid allowing multiple application specs
> inside a bundle?

To tie this back into the current spec phrasing, I think:

  # Standard operations (lifecycle)
  
  ## Create
  
  Creates the container: file system, namespaces, cgroups, capabilities.
  
  ## Start (process)
  
  Runs a process in a container. Can be invoked several times.

is fairly clear in its desire for a separate container-creation
operation after which you can run several application-launching start
operations.  That wording landed in 77d44b10 (Update runtime.md,
2015-06-16) and is still current with today's 6a5a4454.

That doesn't mean that there portions of the spec that are phrased to
sound like there will only be a single application per bundle
(although I don't have one that I can quote here), but I think it
makes my plural in this PR plausible enough that we can merge it as it
stands ;).  We can always file a follow-up PR removing my plural and
rephrasing the lifecycle descriptions if/when we decide that we need a
single application per bundle.

laijs: >   ## Create
>  Creates the container: file system, namespaces, cgroups, capabilities.

When the namespaces have been created, it implies that at least one process has been started. (it is the `runc init` process for the runc implementation for example)

Although this process hasn't yet exec-ed to the container entrypoint, but the process did start.

wking: On Sat, Aug 29, 2015 at 01:40:07AM -0700, Lai Jiangshan wrote:
> When the namespaces have been created, it implies that at least one
> process has been started. (it is the `runc init` process for the
> runc implementation for example)

RunC currently conflates the container creation and application
execution, but that's not required.  For example, bocker:

1. creates a network namespace, cgroups, and networking here [1],
2. enters that container and executes the application [2], and
3. deletes the container [3,4]

It doesn't split those into separate commands (since it's based on the
Docker UI), but you *could* certainly split them into separate
commands, and that seems like the most sane approach to me.

[1]: https://github.com/p8952/bocker/blob/000633061c6cfcb99c8d9eef5aa483a44318f3e6/bocker#L62-L81
[2]: https://github.com/p8952/bocker/blob/000633061c6cfcb99c8d9eef5aa483a44318f3e6/bocker#L82-L87
[3]: https://github.com/p8952/bocker/blob/000633061c6cfcb99c8d9eef5aa483a44318f3e6/bocker#L88-L89
[4]: https://github.com/p8952/bocker/blob/000633061c6cfcb99c8d9eef5aa483a44318f3e6/bocker#L41

laijs: Thanks for pointing out.
But how to create other namespaces without starting any additional process?
Must be there some ways like network, but I can't find out any grace way currently.

------------
I found a bad way (untested)
the creating process()
  unshare()
  mount --bind to save the namespace on some where.
  setns() back or exit.

the starting init process()
  setns() the the save namespace.
-----
So I think it is better to create namespaces when starting the container.

wking: On Sat, Aug 29, 2015 at 02:26:41AM -0700, Lai Jiangshan wrote:
> But how to create other namespaces without starting any additional
> process?  Must be there some ways like network, but I can't find out
> any grace way currently.

They may not all have a way ;).  Any namespaces that turn out to
require a PID may require the runtime to store create-time
information, check that store when launching a new application, and
create any missing namespaces when launching that application.  That
gives us consistent client semantics (e.g. you can still join an
existing user namespace by listing it's path in application.json [1]).
And if/when the kernel gains the ability to create/destroy those
namespaces independently we can drop the associated runtime storage.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/kPWu-ZavQjs

laijs: so
>   ## Create
>  Creates the container: file system, namespaces, cgroups, capabilities.
is some kinds of misleading. Only the network namespace can be created
graciously before starting.

wking: On Sat, Aug 29, 2015 at 10:25:27AM -0700, Lai Jiangshan wrote:
> > ## Create
> > Creates the container: file system, namespaces, cgroups, capabilities.
>
> is some kinds of misleading. Only the network namespace can be
> created graciously before starting.

That may be.  But I think this is getting too far afield for this PR
(maybe shift it to [1]?).  I was only quoting that *existing spec*
(not suggesting we add that wording in this PR) to show that the
current spec seems to lean both ways (depending on which section
you're reading) about whether or not there will be multiple
applications in one container [2].  And if we accept that the
singular/plural application question is currently up in the air, the
‘applications’ plural suggested by this PR isn't a major concession.

But that cuts both ways, so I guess I'll just change to singular here
and isolate the singular/plural lifecycle discussion in [1].  Done in
the just pushed 5fc03cf→9db5fbe, which also rebases onto the current
master (not that there were any conflicts, but having a shorter
fork/merge delay makes for an easier-to-read history graph ;).

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/kPWu-ZavQjs
[2]: https://github.com/opencontainers/specs/pull/107/files#r38259905

wking:
Rebased onto the current master to get #88 and reference both `config.json` and `runtime.json`.  I also added a few links out to relevant sections of the spec.

In today's meeting, @vbatts was saying some of the wording here needed review, but after reading it over again I'm not sure what needs to change.  I'm happy to make any changes required to get this landed.

wking:
I added a new commit with the UTF-8 JSON requirement (since that was the maintainer-recommended approach in #146.

crosbymichael:
I'm -1 on making this change now.  This is adding new terms that don't exist, not being a glossary for existing things.  There is no "Application".  There are only containers and processes.  Also the glossary has a little container/application infinite loop going on which is not helpful.

wking:
On Mon, Oct 05, 2015 at 11:01:47AM -0700, Michael Crosby wrote:
> There is no "Application".

Dropped with b795e6f → a57ba05, which also breaks the container ↔
application loop.  Anything else that I need to change to get this
landed?

wking:
On Mon, Oct 05, 2015 at 11:09:12AM -0700, W. Trevor King wrote:
> Mon, Oct 05, 2015 at 11:01:47AM -0700, Michael Crosby:
> > There is no "Application".
> 
> Dropped with b795e6f → a57ba05…

I missed a few with that update.  Dropped the rest with a57ba05 →
6bddc28.

glossary.md
```diff
@@ -0,0 +1,25 @@
...
+
+A [directory structure](bundle.md) that is written ahead of time, distributed, and used to seed the runtime for creating a [container](#container) and launching a process within it.
+
+## Configuration
+
+The [`config.json`](config.md) and [`runtime.json`](runtime-config.md) files in a [bundle](#bundle) which define the intended [container](#container) and container process.
+
+## Container
+
+An environment for executing processes with configurable isolation and resource limitations (namespaces, resource limits, mounts, …).
```
philips: just say "for example" instead of the parenthetical and ellipses?

philips:
lgtm

glossary.md
wking: On Tue, Dec 22, 2015 at 10:28:22PM -0800, Brandon Philips wrote:
> > +An environment for executing processes with configurable isolation and resource limitations (namespaces, resource limits, mounts, …).
> 
> just say "for example" instead of the parenthetical and ellipses?

In 6bddc28 → 0f9ec22 I've spun the parenthetical out into a new
sentence:

  For example, namespaces, resource limits, and mounts are all part of
  the container environment.

Does that address your concerns?

hqhq:
LGTM. Seems all terms are existed in specs, and I think it's good enough for now, merging, thanks.

PR #113, opened 2015-08-21
Add fd section for linux container process

mrunalp:
Closes #100 

### DIFF
--- a/runtime-linux.md
+++ b/runtime-linux.md
@@ -1,0 +1,6 @@
+## File descriptors
+By default, only the `stdin`, `stdout` and `stderr` file descriptors are kept open for the application by the runtime.
+
+The runtime may pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
+
+Some of the file descriptors may be redirected to /dev/null even though they are open.


wking:
I think we need to specify the mechanism for configuring which file
descriptors should be leaked (although this is a runtime-caller issue
more than a bundle-author issue, so maybe it deserves a different spec
[1]).

The bundle author might want a way to map file descriptor names to a
particular semantic interpretation (e.g. so you can tell the
application that fd 3 is the listening socket), but you could handle
that with environment variables ([2] and opencontainers/runc#200), so
I don't think we need any bundle-author docs beyond this note that it
may be possible to get additional file descriptors (although it would
be nice to link to the runtime-caller docs explaining how to configure
which file descriptors are leaked).

[2]: https://github.com/opencontainers/specs/blob/9ad789fa16156841074275172530fe7db6bcb26b/config.md#process-configuration

runtime-linux.md
```diff
@@ -0,0 +1,3 @@
+## File descriptors
+By default, only the `stdin`, `stdout` and `stderr` file descriptors are kept open for the container process by the runtime.
+The runtime may leak additional file descriptors into the container process to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
```
philips: perhaps pass instead of leak?

philips: also, I think we should settle on something other than "container process" as the word container means too many things today. "application" seems reasonable to me.

dqminh: `container process` should be `container's init process` i think. Do we want to pass extra file descriptors for non-init process ?

philips: @dqminh what is an "init" process? In my mind "init" is /sbin/init and I don't think the runtime should care whether it is a non-init or init process.

dqminh: hmm i was thinking of init process as the first process that is running inside the container, as we can execute multiple processes over the lifetime of the container.

philips: @dqminh In the spec it is not the case that you can execute multiple processes

mrunalp: http://man7.org/linux/man-pages/man7/pid_namespaces.7.html calls it as the "init" process for the pid namespace. However, we could be in a situation where we don't use a new pid namespace.

dqminh: i'm fine with calling it an application. I think we should standardize one term for it through out the specs. `application` sounds fine as user may want to run it inside a VM instead of container for example. Then process/init process is not so approriate.

mrunalp: Well, application implies that there is only one application being run in the container whereas in reality there could be many.

conqer:
Since the use case here is running services inside containers with  a mapped  handle between the host and the container, why not use terminology like service, handles and host mapping?

runtime-linux.md
philips: @mrunalp As defined today an OCI bundle only has a single application.

mrunalp: @philips Updated.

wking:
The wording changes from 3174a03→22ede9f look good to me.

Do we want to suggest environment variables [1] and/or external specs
(like [2]) for associating semantic meaning with file descriptors?  Or
are we happy to leave that up to spec readers to figure out on their
own?

[1]: https://github.com/opencontainers/specs/pull/113#issuecomment-133534708
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/stdin.html

runtime-linux.md
```diff
@@ -0,0 +1,3 @@
+## File descriptors
+By default, only the `stdin`, `stdout` and `stderr` file descriptors are kept open for the application by the runtime.
```
wking: `stdin`, `stdout`, and `stderr` only seem to be attached between the runtime and the application when process.terminal is false.  When it's true, the applications standard streams are attached to a TTY.  See [(1)][1].

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r38173424

mrunalp: Yes, what those fds point to depends on the TTY setting, but the main point being that these are open for the application process.

mrunalp:
@philips @crosbymichael @vbatts ping

crosbymichael:
LGTM

LK4D4:
@mrunalp I'm sorta failing to understand how this closes #100. Shouldn't this include how you can pass additional file descriptors?

mrunalp:
@LK4D4 That is left as an implementation detail to the runtime. For e.g. using environment variables like in https://github.com/opencontainers/runc/pull/231. What we concluded was that having a serializable field in the spec json for ExtraFields doesn't make much sense as the fds and what values they have is dynamic and determined at runtime. What could make sense instead over there is passing fds to specific files.

LK4D4:
@mrunalp Hmm, I don't like implementation detail, why we need anything in spec then? Why this field can't be in runtime?

philips:
LGTM

philips:
I would like for the section to expand to explain that some of the fds may be opened but connected to /dev/null (like stdin).

mrunalp:
@LK4D4 Yeah, so we aren't adding anything to the spec. Just saying that the runtimes may pass additional fds. If we want to standardize that then we may have to prescribe environment variables like OCI_LISTEN_FDS and OCI_LISTEN_FDS_START but they will still have to set from the application specific ones to make sense.

LK4D4:
LGTM

runtime-linux.md
```diff
@@ -0,0 +1,6 @@
...
+
+The runtime may pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
+
+Some of the file descriptors may be redirected to /dev/null even though they are open.
```
wking: `/dev/null` should probably be in backticks.

runtime-linux.md
wking: On Fri, Aug 28, 2015 at 08:21:14AM -0700, Mrunal Patel wrote:
> Yes, what those fds point to depends on the TTY setting, but the
> main point being that these are open for the application process.

Ah, good point.

the standard streams should maybe not be in backticks.  Alternatively,
these existing occurrences should be in backticks [1,2] ;).

[1]: https://github.com/opencontainers/specs/blob/6a5a4454655b4ffacbba11d50471817bc11ed36c/runtime.md#hooks
[2]: https://github.com/opencontainers/specs/blob/6a5a4454655b4ffacbba11d50471817bc11ed36c/runtime.md#pre-start

philips:
Most everyone has LGTM'd this so I am going to fix the one nit about backticks around the `/dev/null` and merge.

philips:
Merged via https://github.com/opencontainers/specs/commit/e3ee431f39720a96412a036e6a9e72ce70214253

mrunalp:
@philips Thanks!

PR #122, opened 2015-08-29
runtime-config: Remove blank lines from the end of files

wking:
These snuck in with 7232e4b1 (specs: introduce the concept of a
runtime.json, 2015-07-30, #88) and 73bf1ba8 (JSON objects are easier
to parse/manipulate, 2015-08-27, #120).


laijs:
LGTM

PR #123, opened 2015-08-29
runtime-config: Drop optional 'options' when empty

wking:
The field is optional, so why bother giving an empty value?

I'd [suggested this][1] after looking over #120, but this change is
orthogonal to that PR's string→array change.

[1]: https://github.com/opencontainers/specs/pull/120#issuecomment-135641212

### DIFF
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -14,8 +14,7 @@ Additional filesystems can be declared as "mounts", specified in the *mounts* ar
     {
         "type": "proc",
         "source": "proc",
-        "destination": "/proc",
-        "options": []
+        "destination": "/proc"
     },
     {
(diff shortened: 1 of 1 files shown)


philips:
eh, I don't see how it maters either way. I am just going to close this.

wking:
On Sat, Aug 29, 2015 at 09:35:04AM -0700, Brandon Philips wrote:
> eh, I don't see how it maters either way. I am just going to close
> this.

I think giving explicit empty values for optional parameters is
useless noise, when the examples are trying to demonstrate the use of
the other fields.  But yeah, it's not a big deal either way.  If you
want to keep the explicitly empty values, I'll survive ;).

PR #124, opened 2015-08-29
config: Add example additionalGids

wking:
The field is optional, but it's nice to have at least one example of
it in use.  The GIDs I've chosen are currently "tty" and "disk" on
Gentoo (1 is "bin"), which may be remotely reasonable choices, but the
values we're using don't really matter without an example filesystem
to provide context.

Signed-off-by: W. Trevor King <wking@tremily.us>


philips:
LGTM.

mrunalp:
LGTM

PR #125, opened 2015-08-29
config: Add example cwd

wking:
The field is optional, but it's nice to have at least one example of
it in use.

Signed-off-by: W. Trevor King <wking@tremily.us>


philips:
LGTM, merging.

PR #126, opened 2015-08-29
Adjust filesnames to match associated JSON

wking:
And add more links to the runtime.json specs, since #88 didn't add
any.  More details and motivation for each step in the commit
messages.


wking:
The last commit (d83516ef) adjust the lifecycle vs. master, but as I explain in that commit message, I don't think the lifecycle explained in master makes sense.  If the explicit-destroy is unpalatable, we can use just “start” (for “create and start”) and “stop” (for “stop and destroy”) to get this PR landed while we work out the container/application separation (for more on this split, see [this thread][1]).

I'm also personally in favor of allowing multiple applications per container, in which case I think we'll want something like wking@4a87a44fa, but that's proved controversial in #107, so I'm leaving it out in the hopes that we can land this reorganization PR more quickly.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/kPWu-ZavQjs

README.md
```diff
@@ -5,9 +5,11 @@ This project is where the [Open Container Initiative](http://www.opencontainers.
 Table of Contents
 
 - [Filesystem Bundle](bundle.md)
-- [Container Configuration](config.md)
+- [Host-independent Container Configuration](config.md)
   - [Linux Specific Configuration](config-linux.md)
-- [Runtime and Lifecycle](runtime.md)
+- [Host-specific Container Configuration](runtime.md)
+  - [Linux Specific Configuration](runtime-linux.md)
```
laijs: you have two `Linux Specific Configuration` one for the `config-linux.md` and the other for `runtime-linux.md`

laijs:
```
- [Platform-independent Container Configuration](config.md)
 - [Linux Specific Configuration](config-linux.md)
- [Host-specific Container Configuration](runtime.md)
 - [Linux Host-specific Configuration](runtime-linux.md)

wking: On Sat, Aug 29, 2015 at 01:11:34AM -0700, Lai Jiangshan wrote:
> >    - [Linux Specific Configuration](config-linux.md)
> > -- [Runtime and Lifecycle](runtime.md)
> > +- [Host-specific Container Configuration](runtime.md)
> > +  - [Linux Specific Configuration](runtime-linux.md)
> 
> you have two `Linux Specific Configuration` one for the
> `config-linux.md` and the other for `runtime-linux.md`

They share the same title because they are namespaced by the
higher-level entry.  To be completely specific, it would be:

  - [Host-independent Container Configuration](config.md)
    - [Linux-specific, Host-independent Configuration](config-linux.md)
  - [Host-specific Container Configuration](runtime.md)
    - [Linux-specific, Host-specific Configuration](runtime-linux.md)

But that seemed too wordy.

Although I suppose I should at least convert to “Linux-specific” for
consistent hyphenation.

wking:
Rebased on the current master (mostly picking up #113), and adding some more consistent cross-linking.  For example, nothing was linking to the content that landed in #113.

wking:
I don't think I'm doing anything controversial here, but this large renaming is going to be difficult to rebase as other PRs land in front of it (and possibly for folks writing other PRs if this lands in front of theirs, depending on whether their work touches something I did inside a file).  It would be easiest if we could get an accept/reject on this one quickly to avoid needless rebasing work.  If there's anything I can do to make this easier to review, or split off smaller chunks, let me know.

laijs:
I think it will be better if the pr is split into smaller ones.
I'm glad if you do the rename as the first pr. I need the patch for RUNV.

wking:
On Mon, Aug 31, 2015 at 05:53:57PM -0700, Lai Jiangshan wrote:
> I think it will be better if the pr is split into smaller ones.

What do you think I can split off?  I'm fine with folks merging /
cherry-picking the earlier commits if there's anything later on that's
difficult to review or controversial.

laijs:
Your pr changes several things:
* update the lifecycle
* update headers
* misc
* rename

I think the PR can be split into these PRs.
If you don't mind, I can take the rename part and create a new PR.

wking:
On Mon, Aug 31, 2015 at 06:51:34PM -0700, Lai Jiangshan wrote:
> Your pr changes several things:
> * update the lifecycle
> * update headers
> * misc
> * rename

Which one of these are commits like 4cfa584?  Once the rename makes
the mapping between documentation and JSON files clear, you recognize
that things like the current hooks documentation is in the wrong place
(e.g. it's in the lifecycle docs, while it should be in the
runtime.json docs, 4cfa584).  Then you realize that the runtime docs
weren't linked from anywhere, so you add c5305f9, fcd496c.  Folks who
are used to having hook documentation in the lifecycle section will
probably want links to the new hook location, so you add enough of a
lifecycle overview that you can link to the hooks (8dae07a).  Etc.,
etc.  I'm fine splitting off smaller PRs, just tell me which commits
to put in which PRs.  I could open up 11 teensy PRs, but that just
seems silly ;).

I'd recommend folks just start in with the first PR, and then LGTM (or
push back against) each commit as they go, for as far as they get.
Once the early commits get enough LGTMs, someone can merge them, which
will make it more likely for others to get deeper in the stack
(assuming reviewing something like 4cfa584 takes more than 30 seconds
;).

vbatts:
going to close this for a couple of things. Some of the linking is now outdated. Also, with trying to overhaul the approach to be more "top down" a lot of this will likely get changed further.
Also, as a side note, golang file names standardly use "_", not "-". They used to not even compile with hyphens.

wking:
On Mon, Oct 05, 2015 at 11:54:44AM -0700, Vincent Batts wrote:
> going to close this for a couple of things.

Works for me, although I still think having runtime.md documenting
runtime.json makes sense.  Now that the linking is a bit better
anyway, would a PR that made just that change be acceptable?

PR #128, opened 2015-08-31
Header cleanup

laijs:


jonboulle:
NACK on d9ed9b04c6271561b3c8834bdf41efe0f22a1eca - that section is specifically cgroups, not resources generally (e.g rlimits provide another means of resource limiting)

laijs:
@jonboulle in the .go file, its name is "Resources", and the user should also define a "Resources" in his config.json (instead of "cgroup" or something else). The spec (*.md) and the user config (config.json) should be consistent.  Maybe we should also add an example for it in this .md. I will just drop this change and and update the other 2 commits as @wking suggested.

wking:
With the tip at 750763, this looks good to me.

laijs:
Anyone objects the first commit?

wking:
On Mon, Aug 31, 2015 at 06:56:30PM -0700, Lai Jiangshan wrote:
> Anyone objects the first commit?

I like both e93d463 and 750763a ;).

laijs:
ping

mrunalp:
@laijs could you squash the commits?

laijs:
@mrunalp done

wking:
On Wed, Sep 02, 2015 at 04:14:53PM -0700, Lai Jiangshan wrote:
> @mrunalp done

004ec55 still looks good to me :).

mrunalp:
LGTM

PR #129, opened 2015-08-31
Update cgroup

laijs:


laijs:
@jonboulle @wking the last commit of #128 is updated and moved here
conflicted with the ummerged #99 from @wking.

wking:
On Mon, Aug 31, 2015 at 07:24:33AM -0700, Lai Jiangshan wrote:
>   * rename the "Control groups" to "Resources" and update the description

“Control groups” are a Linux feature I'd heard of before runC.
“Resources” could be anything.  If we really do step outside of
cgroups here, I'm ok with a structure like:

  ## Resource

  Limit the resources that can be used by your container.  Bla bla
  bla…

  ### Control groups

  Bla bla bla…

  You can also setup device cgroups via
  [devices](config-linux.md#access-to-devices).

  ### Other resource limit that's not covered by control groups

  Bla bla bla…

Although with #99, I've proposed shifting the device-cgroup setup into
the cgroups structure, so it's not squashed in with mknod control.

wking:
On Mon, Aug 31, 2015 at 07:27:47AM -0700, Lai Jiangshan wrote:
> the last commit of #128 is updated and moved here conflicted with
> the ummerged #99 from @wking.

I like #99, but I wouldn't worry about conflicts with it in other pull
requests until it attracts more interest ;).  So far it's been almost
a month with no non-me comments, so blocking anything else on #99 will
probably just kill the blocked thing :p.

wking:
This change will conflict with #137 (although I like where they're
both going, so this is just a heads up that someone may have to rebase
;).

laijs:
a example is added by another pr, close it

PR #130, opened 2015-08-31
Fix LinuxRuntime field

LK4D4:
ping @mrunalp @crosbymichael

### DIFF
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -5,8 +5,8 @@ import "os"
 // LinuxRuntimeSpec is the full specification for linux containers.
 type LinuxRuntimeSpec struct {
 	RuntimeSpec
-	// Linux is platform specific configuration for linux based containers.
-	LinuxRuntime Linux `json:"linux"`
+	// LinuxRuntime is platform specific configuration for linux based containers.
+	Linux LinuxRuntime `json:"linux"`
 }
 
 type LinuxRuntime struct {


jonboulle:
more context? this is essentially breaking with the [config.json convention for platform-specific configuration](https://github.com/opencontainers/specs/blame/e9cb5644fa03cc66f46657266b724f74ee86b04e/config.md#L95) - so at a minimum, should have an equivalent stipulation (e.g. in the currently-underspecified runtime.md) - but I'm not clear from this PR why this change is necessary

LK4D4:
@jonboulle First it isn't working at all. Second it's runtime config and not runtime. I choose name for consistency because actually type name is `RuntimeLinux`. I don't care to name it `Linux`, but because other field named `RuntimeSpec` and not `Spec`, maybe it makes sense as it is.

LK4D4:
Actually I'm sorta lost about naming :) Any ideas are welcome

crosbymichael:
i like just `"linux"`

jonboulle:
I understand that runtime is distinct from config; I was highlighting that a) this introduces an inconsistency between the two for platform-specific configuration, and b) that the runtime configuration is not well specified outside the go implementation here. As for "it isn't working at all", it's still impossible to tell what this means without context...

LK4D4:
@jonboulle It's another golang type, not what was expected.

LK4D4:
@jonboulle @crosbymichael PTAL

mrunalp:
LGTM

jonboulle:
lgtm!

dqminh:
LGTM

LK4D4:
I'm merging then.

PR #134, opened 2015-09-02
Update config.md

liangchenye:
update config.md since the `spec.go` is now changed to `config.go`

### DIFF
--- a/bundle.go
+++ b/bundle.go
@@ -1,0 +1,9 @@
+package specs
+
+// Bundle is the metadata of the files in the container.  It specifies 
+// the attributes of files.
+type Bundle struct {
+	// Files metadata of the files in the container.
+	// Use the 'Device' in runtime_config_linux.go
+	Files []Device `json:"files"`
+}
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Configuration file
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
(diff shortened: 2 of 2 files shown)


wking:
On Tue, Sep 01, 2015 at 08:49:19PM -0700, 梁辰晔 (Liang Chenye) wrote:
> update config.md since the `spec.go` is now changed to `config.go`

+1.  I'd recommend we have two phases to PR review:

1. Is this change worthwhile?  What should it look like in broad
   strokes?
2. Do these commits cleanly implement the desired change?

(1) could be on the mailing list, and (2) a PR here that's just all
polishing.  I feel like getting through (1) is such a push that we
sometimes have stuff landed in order to finalize discussion, but
before we've actually looked over what we're landing with the eye to
quality control that would have caught things like this.

bundle.go
```diff
@@ -0,0 +1,9 @@
+package specs
+
+// Bundle is the metadata of the files in the container.  It specifies 
+// the attributes of files.
+type Bundle struct {
```
mrunalp: This change should be discussed on the mailing list first.

goonzoid:
Looks like the initial intention of this PR has now been resolved when #139 was merged, leaving only the new bundle struct, which, as @mrunalp notes, needs discussion on the mailing list.

Suggest closing this and a new one can be opened as appropriate.

PR #136, opened 2015-09-02
Change layout of mountpoints and mounts

LK4D4:
Added info about MountPoints to config.md.

Filled after discuss in opencontainers/runc#242
I noticed that `MountPoints` still has json tag `mounts`, should I rename it to `mountPoints`?

### DIFF
--- a/config.md
+++ b/config.md
@@ -34,6 +34,37 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 }
 ```
 
+## Mount Points
+
+You can add array of mount points inside container as `mounts`.
+Each record in this array must have configuration in [runtime config](runtime-config.md#mount-configuration).
+
+* **name** (string, required) Name of mount point. Used for config lookup.
+* **path** (string, required) Destination of mount point: path inside container.
+
+*Example*
+
+```json
+"mounts": [
+    {
+        "name": "proc",
+        "path": "/proc"
+    },
+    {
+        "name": "dev",
+        "path": "/dev"
+    },
+    {
+        "name": "devpts",
+        "path": "/dev/pts"
+    },
+    {
+        "name": "data",
+        "path": "/data"
+    }
+]
+```
+
 ## Process configuration
 
 * **terminal** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -1,6 +1,10 @@
 ## Mount Configuration
 
-Additional filesystems can be declared as "mounts", specified in the *mounts* array. The parameters are similar to the ones in Linux mount system call. [http://linux.die.net/man/2/mount](http://linux.die.net/man/2/mount)
+Additional filesystems can be declared as "mounts", specified in the *mounts* object.
+Keys in this object are names of mount points from portable config.
+Values are objects with configuration of mount points.
+The parameters are similar to the ones in [the Linux mount system call](http://man7.org/linux/man-pages/man2/mount.2.html).
+Only [mounts from the portable config](config.md#mount-points) will be mounted.
 
 * **type** (string, required) Linux, *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). Windows: ntfs
 * **source** (string, required) a device name, but can also be a directory name or a dummy. Windows, the volume name that is the target of the mount point. \\?\Volume\{GUID}\ (on Windows source is called target)
@@ -10,45 +14,40 @@ Additional filesystems can be declared as "mounts", specified in the *mounts* ar
 *Example (Linux)*
 
 ```json
-"mounts": [
-    {
+"mounts": {
+    "proc": {
         "type": "proc",
         "source": "proc",
-        "destination": "/proc",
         "options": []
     },
-    {
+    "dev": {
         "type": "tmpfs",
         "source": "tmpfs",
-        "destination": "/dev",
         "options": ["nosuid","strictatime","mode=755","size=65536k"]
     },
-    {
+    "devpts": {
         "type": "devpts",
         "source": "devpts",
-        "destination": "/dev/pts",
         "options": ["nosuid","noexec","newinstance","ptmxmode=0666","mode=0620","gid=5"]
     },
-    {
+    "data": {
         "type": "bind",
         "source": "/volumes/testing",
-        "destination": "/data",
         "options": ["rbind","rw"]
     }
-]
+}
 ```
 
 *Example (Windows)*
 
 ```json
-"mounts": [
-    {
+"mounts": {
+    "myfancymountpoint": {
         "type": "ntfs",
         "source": "\\\\?\\Volume\\{2eca078d-5cbc-43d3-aff8-7e8511f60d0e}\\",
-        "destination": "C:\\Users\\crosbymichael\\My Fancy Mount Point\\",
         "options": []
     }
-]
+}
 ```
 
 See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [SetVolumeMountPoint](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365561(v=vs.85).aspx) in Windows.
--- a/runtime_config.go
+++ b/runtime_config.go
@@ -1,8 +1,10 @@
 package specs
 
 type RuntimeSpec struct {
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-	Mounts []Mount `json:"mounts"`
+	// Mounts is a mapping of names to mount configurations.
+	// Which mounts will be mounted and where should be chosen with MountPoints
+	// in Spec.
+	Mounts map[string]Mount `json:"mounts"`
 	// Hooks are the commands run at various lifecycle events of the container.
 	Hooks Hooks `json:"hooks"`
 }
@@ -29,8 +31,6 @@ type Mount struct {
 	// Source specifies the source path of the mount.  In the case of bind mounts on
 	// linux based systems this would be the file on the host.
 	Source string `json:"source"`
-	// Destination is the path where the mount will be placed relative to the container's root.
-	Destination string `json:"destination"`
 	// Options are fstab style mount options.
 	Options []string `json:"options"`
 }
--- a/config.go
+++ b/config.go
@@ -14,7 +14,7 @@ type Spec struct {
 	// Hostname is the container's host name.
 	Hostname string `json:"hostname"`
 	// Mounts profile configuration for adding mounts to the container's filesystem.
-	MountPoints []MountPoint `json:"mounts"`
+	Mounts []MountPoint `json:"mounts"`
 }
 
 // Process contains information to start a specific application inside the container.


LK4D4:
ping @philips @mrunalp @crosbymichael

runtime-config.md
```diff
@@ -1,6 +1,12 @@
 ## Mount Configuration
 
-Additional filesystems can be declared as "mounts", specified in the *mounts* array. The parameters are similar to the ones in Linux mount system call. [http://linux.die.net/man/2/mount](http://linux.die.net/man/2/mount)
+Additional filesystems can be declared as "mounts", specified in the *mounts* object.
+Keys in this object are names of mount points from portable config.
```
wking: Link with “`… of [mount points from the portable config](config.md#mount-points)`”?

runtime-config.md
```diff
@@ -1,6 +1,12 @@
...
+Values are objects with configuration of mount points.
+The parameters are similar to the ones in Linux mount system call.
+[http://linux.die.net/man/2/mount](http://linux.die.net/man/2/mount)
+Note, that only mount points which listed in portable configuration will be
+really mounted.
```
wking: This sentence should all be on one line (README's Markdown formatting).  And I'd end with “will actually be mounted”, because things without `config.json` listings will not be mounted in any sense, so “really” doesn't seem to fit.

runtime-config.md
```diff
@@ -1,6 +1,12 @@
 ## Mount Configuration
 
-Additional filesystems can be declared as "mounts", specified in the *mounts* array. The parameters are similar to the ones in Linux mount system call. [http://linux.die.net/man/2/mount](http://linux.die.net/man/2/mount)
+Additional filesystems can be declared as "mounts", specified in the *mounts* object.
+Keys in this object are names of mount points from portable config.
+Values are objects with configuration of mount points.
+The parameters are similar to the ones in Linux mount system call.
+[http://linux.die.net/man/2/mount](http://linux.die.net/man/2/mount)
```
wking: Instead of splatting in the URL, I'd use: “`the ones in [the Linux mount system call](http://man7.org/linux/man-pages/man2/mount.2.html)`”.  That's also using Kerrisk's site for the page source, since it seems the most canonical.

config.md
```diff
@@ -34,6 +34,25 @@ Each container has exactly one *root filesystem*, specified in the *root* object
...
+in runtime config.
+* **path** (string, required) Destination of mount point: path inside container.
+
+*Example*
+
+```json
+"mounts": {
+    "name": "proc",
+    "path": "/proc",
+}
```
wking: If we want to list several mount points in the `runtime.json` docs, I think we might as well map all of those here.  If we like the abbreviated example here, I think we might as well abbreviate the `config.json` example to match ;).  So I don't mind which way we go, but I think the examples should fit together.

runtime-config.md
LK4D4: I don't get "README's Markdown formatting" part. Why they should be on one line?

runtime_config.go
```diff
@@ -1,8 +1,10 @@
 package specs
 
 type RuntimeSpec struct {
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-	Mounts []Mount `json:"mounts"`
+	// Mounts is a mapping of names to mount configuration.
+	// Which mounts will be mounted and where should be chosen with MountPoints
+	// in Spec.
```
wking: It's unfortunate that `*Spec` didn't change to `*Config` as part of 7232e4b (specs: introduce the concept of a runtime.json, 2015-07-30, #88), but that's probably orthogonal to this PR ;).

runtime-config.md
mrunalp: +1 to using man7.org

wking:
On Wed, Sep 02, 2015 at 08:47:03AM -0700, Alexander Morozov wrote:
> I noticed that `MountPoints` still has json tag `mounts`, should I
> rename it to `mountPoints`?

I don't think so.  We need different Go names because we're putting
all these types in one package, but the JSON is clearly namespaced by
the file it's in (config.json vs. runtime.json), so there's no need to
disambiguate there.

runtime-config.md
wking: On Wed, Sep 02, 2015 at 09:31:39AM -0700, Alexander Morozov wrote:
> I don't get "README's Markdown formatting" part. Why they should be on one line?

https://github.com/opencontainers/specs/tree/138deee1418d2376a4e9d4966e9f1543e717a641#markdown-style

LK4D4:
Actually we don't need different go name too.

wking:
On Wed, Sep 02, 2015 at 09:45:00AM -0700, Alexander Morozov wrote:
> Actually we don't need different go name too.

We don't?  They're both in the specs package in the same directory.
If they were both Mount, how would you access one or the other?

LK4D4:
@wking It's type name, have nothing in common with field name.

wking:
On Wed, Sep 02, 2015 at 09:51:08AM -0700, Alexander Morozov wrote:
> @wking It's type name, have nothing in common with field name.

Ah, right.  +1 to making both field names Mounts to match the JSON
keys.

LK4D4:
@wking @mrunalp I think I addressed comments.

runtime_config.go
```diff
@@ -1,8 +1,10 @@
 package specs
 
 type RuntimeSpec struct {
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-	Mounts []Mount `json:"mounts"`
+	// Mounts is a mapping of names to mount configuration.
```
mrunalp: s/configuration/configurations/

wking:
On Wed, Sep 02, 2015 at 11:21:13AM -0700, Alexander Morozov wrote:
> @wking @mrunalp I think I addressed comments.

Matching examples for mounts and mount-points is still open [1].  But
“it's fine the way it is” is a valid response ;).

[1]: https://github.com/opencontainers/specs/pull/136/files#r38553654

LK4D4:
@wking I actually added example of `runtime.conf` entry below that line.

wking:
On Wed, Sep 02, 2015 at 11:43:07AM -0700, Alexander Morozov wrote:
> @wking I actually added example of `runtime.conf` entry below that line.

But it only lists ‘proc’, while the config.json example lists proc,
dev, devpts, and data.  I'm suggesting we either list all those names
in both examples, or only proc in both examples.

config.md
```diff
@@ -34,6 +34,33 @@ Each container has exactly one *root filesystem*, specified in the *root* object
...
+```
+In `runtime.json`:
```
wking: Ah, I wasn't suggesting we copy the `runtime.json` side over here.  I want to [*link* to it][1].  But I think the `config.json` example here should be compatible with the `runtime.json` example we link to.  I'm happy to PR a fixup against your branch if that's easier ;).

[1]: https://github.com/opencontainers/specs/pull/136/files#diff-695bac0ce528e5c6765a6666fcccc425R14

LK4D4: okay

wking:
On Wed, Sep 02, 2015 at 11:50:11AM -0700, W. Trevor King wrote:
> On Wed, Sep 02, 2015 at 11:43:07AM -0700, Alexander Morozov wrote:
> > @wking I actually added example of `runtime.conf` entry below that line.
> 
> But it only lists ‘proc’…

I've been explaining this poorly, so here's a PR with my suggested
fixup: LK4D4/specs#1.  Feel free to squash into your commit if it
makes sense to you.

runtime-config.md
wking: The link is now on line 7 (“Only [mounts from …]”), so I'm fine leaving this line as it stands.

runtime_config.go
```diff
@@ -1,9 +1,11 @@
...
 
 type RuntimeSpec struct {
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-	Mounts []Mount `json:"mounts"`
-	// Hooks are the commands run at various lifecycle events of the container.
+	// Mounts is a mapping of names to mount configuration.
+	// Which mounts will be mounted and where should be chosen with MountPoints
+	// in Spec.
+	Mounts map[string]Mount `json:"mounts"`
+	// Hooks are the commands run at various lifecycle events of the containers.
```
wking: We don't want to touch this comment.

runtime_config.go
```diff
@@ -1,9 +1,11 @@
 package specs
 
 type RuntimeSpec struct {
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-	Mounts []Mount `json:"mounts"`
-	// Hooks are the commands run at various lifecycle events of the container.
+	// Mounts is a mapping of names to mount configuration.
```
wking: This still needs “configuration” → “configurations”.

runtime_config.go
LK4D4: I don't see what changed in this comment.

wking: On Wed, Sep 02, 2015 at 01:12:41PM -0700, Alexander Morozov wrote:
> > -	Mounts []Mount `json:"mounts"`
> > -	// Hooks are the commands run at various lifecycle events of the container.
> > …
> > +	Mounts map[string]Mount `json:"mounts"`
> > +	// Hooks are the commands run at various lifecycle events of the containers.
> 
> I don't see what changed in this comment.

The post-Mounts comment should be about *hooks*, and there's only one
*container* for lifecycle events.

> > +	// Mounts is a mapping of names to mount configuration.
> > +	// Which mounts will be mounted and where should be chosen with MountPoints
> > +	// in Spec.

The pre-Mounts comment^, on the other hand should be talking about
“mount configurations”, not “mount configuration”.

config.md
```diff
@@ -34,6 +34,37 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 }
 ```
 
+## Mount Points
+
+You can add array of mount points inside container as `mounts`.
+Each record in this array must have configuration in [runtime config](runtime-config.md#mount-configuration).
+
+* **name** (string, required) Name of mount point. Used for config lookup.
+* **path** (string, required) Destination of mount point: path inside container.
```
wking: `<nit>`Maybe “Mount destination in the container filesystem”?  The colon-separated form works, but it took me a reread to understand what it meant.`</nit>`.

runtime_config.go
```diff
@@ -1,8 +1,10 @@
 package specs
 
 type RuntimeSpec struct {
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-	Mounts []Mount `json:"mounts"`
+	// Mounts is a mapping of names to mount configurations.
+	// Which mounts will be mounted and where should be chosen with MountPoints
+	// in Spec.
```
wking: Maybe “MountPoints in Spec” → “Spec.Mounts”?

mrunalp:
LGTM

crosbymichael:
LGTM

runtime-config.md
```diff
@@ -10,45 +14,41 @@ Additional filesystems can be declared as "mounts", specified in the *mounts* ar
...
 *Example (Windows)*
 
 ```json
-"mounts": [
-    {
+"mounts": {
+    "myfancymountpoint": {
         "type": "ntfs",
         "source": "\\\\?\\Volume\\{2eca078d-5cbc-43d3-aff8-7e8511f60d0e}\\",
         "destination": "C:\\Users\\crosbymichael\\My Fancy Mount Point\\",
```
laijs: Here is a leftover "destination"

laijs:
I think the mount order should be **clearly** defined in the spec.

Alternative definitions in my mind:
* The mount order are exactly the same as this mount points array. The config file MUST list them in reasonable order.
* The mount order are mostly the same as this mount points array. But the longer path will be altered downward and be placed after the shorter path which is the ancestor of the longer path.

Any thoughts?

wking:
On Wed, Sep 02, 2015 at 04:53:54PM -0700, Lai Jiangshan wrote:
> * The mount order are exactly the same as this mount points
>   array. The config file MUST list them in reasonable order.

This is how I'd do it.

laijs:
> This is how I'd do it.

Wait until full discussion?

wking:
On Wed, Sep 02, 2015 at 05:30:18PM -0700, Lai Jiangshan wrote:
> > This is how I'd do it.
> 
> Wait until full discussion?

Probably wise ;).  One thing to consider for auto-ordering, is what
would happen if you did:

1. Bind mount /a/b/c to /d
2. Mount $x over /a/b

I'm not actually sure ;), but it's possible that /d still looks into
the original /a/b/c, and not into $x/c.  If so, that means
auto-ordering could be problematic, because it would see that (1)'s
/a/b/c seemingly depends on (2), when in fact the bundle-author
preferred (1) to happen first.

laijs:
I prefer the strict/exact order. But it needs to be documented in the Spec.

laijs:
LGTM (c18c283)

wking:
On Wed, Sep 02, 2015 at 09:25:54PM -0700, Lai Jiangshan wrote:
> LGTM (c18c283)

Looks good to me too.  I'm fine punting explicit ordering statements
(and my other comment cleanups [1,2]) to follow-up PRs, if that helps
this one land faster.

[1]: https://github.com/opencontainers/specs/pull/136/files#r38583332
[2]: https://github.com/opencontainers/specs/pull/136/files#r38583510

LK4D4:
I think we have enough LGTMs here and I want to fix spec for now. Feel free to post your suggestions as another PR.

Issue #114, opened 2015-08-25
Support for Quality of Service


vishh:
The API has to let users express the notion of priority among containers. Consider the case when a web server is run alongside a logging container. The web-server is more important than the logging side-car container. When there is a resource crunch, the user doesn't mind killing one of the lower priority containers, which is logging in this case. Ideally, if all containers were to be run with limits, this might not be necessary. In reality though, setting limits is hard and users tend to over provision resources and that leads to poor resource utilization.

Since this notion of priority can be expressed in different ways, I propose letting users handle cgroup management and only expose the following knobs:

1. **CgroupsPath** - This lets users manage cgroup hierarchies. 

2. **OomScoreAdj** - Apply a custom 'oom_score_adj' to the init process of the container.

wking:
On Tue, Aug 25, 2015 at 04:48:58PM -0700, Vish Kannan wrote:
> The API has to let users express the notion of priority among
> containers. Consider the case when a web server is run alongside a
> logging container…

We need to expose the underlying knobs to let the runtime-caller
manage this sort of thing when launching a container, but I think
explicitly handling relationships between containers is out-of-scope
for this spec [1].

> 1. **CgroupsPath** - This lets users manage cgroup hierarchies. 

We can do this with namespaces now [2].  Similar handling in
linux.resources for joining existing cgroups (in addition to creating
new ones, which is the focus of the current Resources [3]) sounds good
to me.

> 2. **OomScoreAdj** - Apply a custom 'oom_score_adj' to the init
> process of the container.

We have disableOOMKiller, which landed without docs in #51.  But yeah,
there appears to be no way to set oom_score_adj [4] directly.
Personally, I think this sort of thing is better handled via a
host-injected pre-start hook, since it's more of a deploy-time
decision than a bundle-author-time decision.  But I'm still unclear on
the intendend environemnt for those hooks [5].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/uo11avcWlQQ/7quJAJIZEgAJ
     From: W. Trevor King
     Subject: Re: appc + oci harmonization progress
     Date: Tue, 18 Aug 2015 11:20:12 -0700
     Cc: dev@opencontainers.org
     Message-ID: <20150818182012.GX21585@odin.tremily.us>
[2]: https://github.com/opencontainers/specs/blob/9ad789fa16156841074275172530fe7db6bcb26b/config-linux.md#linux-namespaces
[...]

PR #137, opened 2015-09-02
Adding cgroups path to the Spec.

vishh:
Related to #114
This change should be useful for exec'ing new processes in an existing container.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -131,6 +131,60 @@ Also known as cgroups, they are used to restrict resource usage for a container 
 cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
+The path to the cgroups can to be specified in the Spec via `cgroupsPath`.
+`cgroupsPath` is expected to be relative to the cgroups mount point.
+If not specified, cgroups will be created under '/'.
+Implementations of the Spec can choose to name cgroups in any manner.
+The Spec does not include naming schema for cgroups.
+The Spec does not support [split hierarchy](https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt).
+The cgroups will be created if they don't exist.
+
+```json
+   "cgroupsPath": "/myRuntime/myContainer"
+```
+
+`cgroupsPath` can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+
+Optionally, cgroups limits can be specified via `resources`.
+
+```json
+    "resources": {
+        "disableOOMKiller": false,
+        "memory": {
+            "limit": 0,
+            "reservation": 0,
+            "swap": 0,
+            "kernel": 0,
+            "swappiness": -1
+        },
+        "cpu": {
+            "shares": 0,
+            "quota": 0,
+            "period": 0,
+            "realtimeRuntime": 0,
+            "realtimePeriod": 0,
+            "cpus": "",
+            "mems": ""
+        },
+        "blockIO": {
+            "blkioWeight": 0,
+            "blkioWeightDevice": "",
+            "blkioThrottleReadBpsDevice": "",
+            "blkioThrottleWriteBpsDevice": "",
+            "blkioThrottleReadIopsDevice": "",
+            "blkioThrottleWriteIopsDevice": ""
+        },
+        "hugepageLimits": null,
+        "network": {
+            "classId": "",
+            "priorities": null
+        }
+    }
+```
+
+Do not specify `resources` unless limits have to be updated.
+For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
+
 ## Sysctl
 
 sysctl allows kernel parameters to be modified at runtime for the container.
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -24,7 +24,11 @@ type LinuxRuntime struct {
 	Sysctl map[string]string `json:"sysctl"`
 	// Resources contain cgroup information for handling resource constraints
 	// for the container
-	Resources Resources `json:"resources"`
+	Resources *Resources `json:"resources"`
+	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
+	// The path is expected to be relative to the cgroups mountpoint.
+	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
+	CgroupsPath string `json:"cgroupsPath"`
 	// Namespaces contains the namespaces that are created and/or joined by the container
 	Namespaces []Namespace `json:"namespaces"`
 	// Devices are a list of device nodes that are created and enabled for the container


runtime-config-linux.md
```diff
@@ -130,6 +130,13 @@ Next parameters can be specified:
 Also known as cgroups, they are used to restrict resource usage for a container and handle
 device access.  cgroups provide controls to restrict cpu, memory, IO, and network for
 the container. For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
+The path to the cgroups can be specified in the Spec. The path is expected to be relative
+to the cgroups mount point. The Spec does not support split hierarchy. The cgroups will be 
+created if they don't exist.
```
wking: This should be [one line per sentence][1].  I'll file a PR suggesting we move stuff like that from the README to a `CONTRIBUTING.md`

[1]: https://github.com/opencontainers/specs/tree/138deee1418d2376a4e9d4966e9f1543e717a641#markdown-style

runtime-config-linux.md
```diff
@@ -130,6 +130,13 @@ Next parameters can be specified:
 Also known as cgroups, they are used to restrict resource usage for a container and handle
 device access.  cgroups provide controls to restrict cpu, memory, IO, and network for
 the container. For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
+The path to the cgroups can be specified in the Spec. The path is expected to be relative
+to the cgroups mount point. The Spec does not support split hierarchy. The cgroups will be 
+created if they don't exist. 
+
+```json
+   "cgroupsPath": "/myRuntime/myContainer"
```
wking: I'd mirror [cgexec(1)][1] with something like:

    "cgroups": {
      "join": [
        {
          "path": "/myRuntime/test1",
          "controllers": ["cpu", "memory"]
        },
        {
          "path": "/myRuntime/test2",
          "controllers": ["swap"]
        }
      ]
    }

And it would be nice to move the current "resources" into:

    "cgroups": {
      "create": …,
    }

[1]: http://linux.die.net/man/1/cgexec

wking:
This might be worth floating on the list first [1] ;).  Reviewing
different proposals (e.g. your path in be6ec5ff vs. my array of
objects in [2]) is probably too much to wedge into GitHub's inline
comments :p.

[1]: https://github.com/opencontainers/specs/tree/138deee1418d2376a4e9d4966e9f1543e717a641#discuss-your-design
[2]: https://github.com/opencontainers/specs/pull/137/files#r38597501

runtime-config-linux.md
vishh: Won't this break with [unified hierarchy](https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt)?

vishh:
@wking: Apologies for not being updated on the new contribution workflows.

@mrunalp: Any thoughts on systemd compatibility?

runtime-config-linux.md
wking: On Wed, Sep 02, 2015 at 05:09:05PM -0700, Vish Kannan wrote:
> Won't this break with [unified
> hierarchy](https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt)?

Ah, that's interesting reading.  In that case, a single path is fine.

Can we link to those Kernel docs from our “The Spec does not support
split hierarchy.”?

runtime_config_linux.go
```diff
@@ -21,6 +21,9 @@ type LinuxRuntime struct {
 	// Resources contain cgroup information for handling resource constraints
 	// for the container
 	Resources Resources `json:"resources"`
+	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
+	// The path is expected to be relative to the cgroups mountpoint.
+	CgroupsPath string `json:"cgroupsPath"`
```
mrunalp: @vishh It will be either CgroupsPath or Resources, right? We should either clarify that in the spec or have a layer of indirection.

mrunalp:
@vishh I think that this should work with systemd as well. The only difference is the terminology AFAICS.

For e.g.
```
10:blkio:/system.slice/sshd.service
9:hugetlb:/
8:perf_event:/
7:devices:/system.slice/sshd.service
6:freezer:/
5:memory:/system.slice/sshd.service
4:net_cls,net_prio:/
3:cpuset:/
2:cpu,cpuacct:/system.slice/sshd.service
1:name=systemd:/system.slice/sshd.service

# mount
...
cgroup on /sys/fs/cgroup/systemd type cgroup (rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/usr/lib/systemd/systemd-cgroups-agent,name=systemd)
...
```

runtime_config_linux.go
dqminh: i agree that it should either be path or resources as it doesnt really make sense ( or even wrong ) to join with existing paths and then modify those resources.

vishh: @mrunalp: How about creating the cgroups for the user and applying the limits expressed via `Resources`? 
@dqminh: Why would it be wrong? AFAIK the Spec doesn't try to limit how the existing Linux functionalities can be consumed right?

mrunalp: @vishh There could be two interpretations. One is simple joining the cgroups (that were configured by some other tool or by another container) and other is using that path and then going in and modifying the values under that path. The first is necessary for us to support exec like functionality. Does that work for your use cases?

dqminh: Hmm actually i withdraw my comment about enforcing the limitation of cgroupPaths vs resources. I think there can be usecases where having that capabilities is helpful ( like launching an agent container that modifies the cgroup resources at boot time and/or during the agent runtime maybe ? ).

I think that clarifying the order of resources modification is good enough i.e, if cgroups path is specified then we join the paths, otherwise we create new paths, and then apply the resources limit.

vishh: @mrunalp: Your reasoning is sound! I see use cases for both `exec` and `resource updates`. Would making `Resources` optional resolve the issue?

wking: On Tue, Sep 08, 2015 at 03:23:20PM -0700, Daniel, Dao Quang Minh wrote:
> I think that clarifying the order of resources modification is good
> enough i.e, if cgroups path is specified then we join the paths,
> otherwise we create new paths, and then apply the resources limit.

See also comments in #155 and #158, where we're having the same
discussion about namespaces.

mrunalp: > @mrunalp: Your reasoning is sound! I see use cases for both exec and resource updates. Would making Resources optional resolve the issue? 

@vishh Sounds good to me!

philips:
nit: reformat to one sentence per line please

mrunalp:
Needs rebase as well.

vishh:
@mrunalp @crosbymichael: WDYT of making cgroups path required? Without that, the options are cgroups can created under under root `/`, or under a user specified hierarchy (which requires configuration).

mrunalp:
@vishh I think we could default to ` /` if cgroups path is not specified and a user specified path otherwise.

vishh:
What will be the identifier we use for the cgroups? Will it be the base
name of the bundle?

vbatts:
That is a common thought.

mrunalp:
@vishh runc uses the --id or the base name if not specified.

wking:
On Wed, Sep 09, 2015 at 04:55:45PM -0700, Vish Kannan wrote:
> What will be the identifier we use for the cgroups? Will it be the
> base name of the bundle?

Does this need to be spec'ed out?  There's not currently much language
around container IDs.  We just list them in state.json, and I've
pointed out that I'm not clear on why [1].  It makes sense that we'd
want some sort of handle for tracking “the same container” across a
checkpoint/restore migration (which is likely to change the PID), but
I don't see a need to bind the container ID in more tightly than that.
I think we should leave the cgroup/namespace name generation to the
runtime, and have consumers query (via the PID) to figure out the
cgroup/namespace name if they need it for something.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/z25xQsF3pHA/w0XqQhRyFwAJ
     Message-ID: <20150905041822.GA25638@odin.tremily.us>

vishh:
@wking: My understanding is that it is recommended for runtimes to manage cgroups. But for a good dev experience, we can default to creating cgroups under `/` with an easy to use identifier (base name of the bundle).

vishh:
@mrunalp @vbatts @dqminh Updated docs based on comments and rebased. PTAL.

vbatts:
Though base of the bundle would not be unique enough for more than 1
container runtime. Let's leave this id for the runtime. Won't the paths be
referencable through the /run/opencontainers/ path anyways.

vishh:
Yeah. The cgroup paths can be retrieved from the state.

wking:
On Wed, Sep 09, 2015 at 05:13:32PM -0700, Vish Kannan wrote:
> My understanding is that it is recommended for runtimes to manage
> cgroups. But for a good dev experience, we can default to creating
> cgroups under `/` with an easy to use identifier (base name of the
> bundle).

If folks can spawn new containers without setting ‘cgroupsPath’, then
folks using a runtime shouldn't be guessing “/{container.id}” (or
whatever) as a cgroup path, they'll have to look it up for each
container (which I think is a good thing).

So since users will have to lookup a process's cgroups path anyway,
why should we bother specifying a rule to auto-generate paths?  If
different runtimes pick different rules for the default cgroup path,
what use-cases will suffer?

Wed, Sep 09, 2015 at 06:43:27PM -0700, Vish Kannan:
> Yeah. The cgroup paths can be retrieved from the state. 

Not directly.  Folks need to use the PID to get them (and that seems
fine to me), at least with the spec that landed via #87, see [1].

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-126123937

runtime-config-linux.md
```diff
@@ -131,6 +131,58 @@ Also known as cgroups, they are used to restrict resource usage for a container
 cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
+The path to the cgroups can to be specified in the Spec via `cgroupsPath`.
+`cgroupsPath` is expected to be relative to the cgroups mount point.
+If not specified, cgroups will be created under '/' with the base name of the bundle.
```
dqminh: I think it's more reasonable / simpler to error out here rather than continuing and generate a path based on base name of the bundle. `cgroupsPath` should be a required data right ?

wking: On Thu, Sep 10, 2015 at 01:54:07AM -0700, Daniel, Dao Quang Minh wrote:
> > +If not specified, cgroups will be created under '/' with the base name of the bundle.
> 
> I think it's more reasonable / simpler to error out here rather than
> continuing and generate a path based on base name of the
> bundle. `cgroupsPath` should be a required data right ?

I often want cgroups for my containers, but don't care what the path
is.  I think we do want the ability to tell the runtime “I don't care,
just make something up”, which is what leaving cgroupsPath out does.
But I'd rather leave the “make something up” algorithm up to the
runtime [1].  See also the “several containers from one bundle” use
case [2].

[1]: https://github.com/opencontainers/specs/pull/137#issuecomment-139103731
[2]: https://github.com/opencontainers/specs/pull/137#issuecomment-139078372

vishh: I'm in favor of defaulting to sub-containers of root. Ping @crosbymichael @mrunalp @vbatts

wking: On Thu, Sep 10, 2015 at 10:46:59AM -0700, Vish Kannan wrote:
> > +If not specified, cgroups will be created under '/' with the base name of the bundle.
> 
> I'm in favor of defaulting to sub-containers of root.

If anyone disagrees with *that*, I think it's a whole nother can of
worms.  I think the disagreement so far has been about what the
sub-container is going to be called, not where it will be rooted.

mrunalp:
LGTM

crosbymichael:
LGTM

runtime-config-linux.md
```diff
@@ -131,6 +131,60 @@ Also known as cgroups, they are used to restrict resource usage for a container
 cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
+The path to the cgroups can to be specified in the Spec via `cgroupsPath`.
+`cgroupsPath` is expected to be relative to the cgroups mount point.
+If not specified, cgroups will be created under '/'.
+Implementations of the Spec can choose to name cgroups in any manner.
```
wking: This is only true if `cgroupsPath` is unset.  So maybe rephrase as “In that case, implementations can choose to name cgroups in any manner.” to tie it into the previous ”If not specified…” sentence.

vbatts: I feel it's clear enough for now.

vbatts:
LGTM

PR #139, opened 2015-09-03
Fix links to go files in config.md

goonzoid:
Looks like these didn't get updated when `spec.go` and `spec_linux.go` were renamed.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Configuration file
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Sep 03, 2015 at 02:21:04AM -0700, Will Pragnell wrote:
> Looks like these didn't get updated when `spec.go` and
> `spec_linux.go` were renamed.

This duplicates #134 (873effe), although that PR has since grown a
follow-up commit (3bae26c) that I don't understand, so I'm ok with
this one getting merged instead.  You also have a more discoverable PR
summary ;).

mrunalp:
LGTM

PR #142, opened 2015-09-03
config: Require the runtime to mount Spec.Mounts in order

wking:
If we don't specify this, some bundle-authors or runtime-implementers
might expect the runtime to intelligently order mounts to get the
"right" order.  But that's not possible because:

    $ mkdir -p a/b/c d/e/f h
    # mount --bind a/b h
    # mount --bind d a/b
    $ tree --charset=ascii h
    h
    `-- c

But in the other order:

    # umount a/b
    # umount h
    # mount --bind d a/b
    # mount --bind a/b h
    $ tree --charset=ascii h
    h
    `-- e
        `-- f

So there's no "right" order.  Allowing the bundle-author to specify
their intended order is both easy to implement and unambiguous.

Spun off from #136.

### DIFF
--- a/config.md
+++ b/config.md
@@ -38,6 +38,7 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 
 You can add array of mount points inside container as `mounts`.
 Each record in this array must have configuration in [runtime config](runtime-config.md#mount-configuration).
+The runtime MUST mount entries in the listed order.
 
 * **name** (string, required) Name of mount point. Used for config lookup.
 * **path** (string, required) Destination of mount point: path inside container.


wking:
Ping @laijs.

config.md
```diff
@@ -38,6 +38,7 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 
 You can add array of mount points inside container as `mounts`.
 Each record in this array must have configuration in [runtime config](runtime-config.md#mount-configuration).
+The runtime MUST mount entries in the listed order (`mounts[0]` first, `mounts[1]` second, …).
```
vbatts: Perhaps simpler to say "order matters"

wking: On Fri, Sep 04, 2015 at 12:02:02PM -0700, Vincent Batts wrote:
> > +The runtime MUST mount entries in the listed order (`mounts[0]` first, `mounts[1]` second, …).
> 
> Perhaps simpler to say "order matters"

That wouldn't say *how* the order matters ;).  I think it's hard to
have too much clarity, but I'm fine dropping the parenthetical if it
feels too wordy.  The sentence through “… listed order” should be
sufficient to stand alone.

wking:
#157 landed with similar phrasing but [no parenthetical][1].  I've dropped the parenthetical (and rebased onto master) with 19c3bce → b051802 in case that help get this merged.

[1]: https://github.com/opencontainers/specs/pull/142#discussion_r38806989

laijs:
LGTM

I was wondering if you could add "Suggested-by: Lai Jiangshan <jiangshanlai@gmail.com>"

wking:
On Thu, Sep 10, 2015 at 05:31:12PM -0700, Lai Jiangshan wrote:
> I was wondering if you could add "Suggested-by: Lai Jiangshan
> <jiangshanlai@gmail.com>"

Done with b051802 → d282a32, and I cited your initial comment as a
reference for the possible alternative interpretation that this PR
closes off.

mrunalp:
LGTM

vbatts:
LGTM

PR #143, opened 2015-09-03
state: Shift the state tree to it's own set of files

wking:
@crosbymichael was adding these to existing Go files [to avoid having
a lot of files][1].  But 7232e4b (specs: introduce the concept of a
`runtime.json`, 2015-07-30, #88) split the runtime-configuration into
two file sets (one rooted in `config.go` for `config.json` and another
rooted in `runtime_config.go` for `runtime.json`).  In the face of a
two-set split for a single task (feeding the runtime while it
manipulates the container lifecycle), it seems odd to push the type
definition for a completely different task (sharing
container/application state with other tools) into an existing file
set.

Spun off from [here][2].  See also #126, which standardizes the JSON /
doc / Go naming pattern.

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-128137602
[2]: https://github.com/opencontainers/specs/pull/87#issuecomment-137200444

### DIFF
--- a/config.go
+++ b/config.go
@@ -56,15 +56,3 @@ type MountPoint struct {
 	// Path specifies the path of the mount. The path and child directories MUST exist, a runtime MUST NOT create directories automatically to a mount point.
 	Path string `json:"path"`
 }
-
-// State holds information about the runtime state of the container.
-type State struct {
-	// Version is the version of the specification that is supported.
-	Version string `json:"version"`
-	// ID is the container ID
-	ID string `json:"id"`
-	// Pid is the process id for the container's main process.
-	Pid int `json:"pid"`
-	// Root is the path to the container's bundle directory.
-	Root string `json:"root"`
-}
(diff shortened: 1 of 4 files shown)


vbatts:
I really don't see the need in having this struct split out at all :-\

wking:
On Fri, Sep 04, 2015 at 07:19:12AM -0700, Vincent Batts wrote:
> I really don't see the need in having this struct split out at all :-\

There's no *need*, it's just consistent with the current config.go /
runtime_config.go split.  I'd also be ok just moving everything into a
single file-set (e.g. spec.go and spec_linux.go) that contain all of
Spec, RuntimeSpec, and State (although then it's a bit harder to link
in from the Markdown docs to the appropriate section).

crosbymichael:
The files are small so I agree with @vbatts that we don't really need this type of split with the structs.

PR #146, opened 2015-09-03
Specify UTF-8 for all our JSON

wking:
I wish there was a cleaner reference for what UTF-8 was.  But linking
[Wikipedia][1] seems too glib, and I can't find a more targetted link
than just dropping folks into a Unicode chapter (which is what
[Wikipedia][1] does):

> The Unicode Standard, Version 6.0, §3.9 D92, §3.10 D95 (2011)

With the current v8.0 (2015-06-17), it's still §3.9 D92 and §3.10 D95.

To get a place to say this for the runtime.json docs, I used config.md
as a template for the top-level header and blurb.  The “Host-specific
container configuration” phrasing comes from bundle.md, which has:

> The `runtime.json` file contains settings that are host specific...

While I was touching the config.md lead-in, I fixed “container” →
“bundle”.

[1]: https://en.wikipedia.org/wiki/UTF-8

### DIFF
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -1,3 +1,8 @@
+# Host-specific container configuration
+
+The bundle's top-level directory MUST contain a configuration file called `runtime.json` with [UTF-8][] [JSON][].
+For now the canonical schema is defined in [`runtime_config.go`](runtime_config.go) and [runtime_config_linux.go](runtime_config.go), but this will be moved to a formal JSON schema over time.
+
 ## Mount Configuration
 
 Additional filesystems can be declared as "mounts", specified in the *mounts* object.
@@ -51,3 +56,6 @@ Only [mounts from the portable config](config.md#mount-points) will be mounted.
 ```
 
 See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [SetVolumeMountPoint](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365561(v=vs.85).aspx) in Windows.
+
+[UTF-8]: http://www.unicode.org/versions/Unicode8.0.0/ch03.pdf
+[JSON]: http://json.org/
--- a/config.md
+++ b/config.md
@@ -1,6 +1,6 @@
 # Configuration file
 
-The container's top-level directory MUST contain a configuration file called `config.json`.
+The bundle's top-level directory MUST contain a configuration file called `config.json` with [UTF-8][] [JSON][].
 For now the canonical schema is defined in [spec.go](spec.go) and [spec_linux.go](spec_linux.go), but this will be moved to a formal JSON schema over time.
 
 The configuration file contains metadata necessary to implement standard operations against the container.
@@ -124,3 +124,6 @@ For Linux-based systems the user structure has the following fields:
 ```
 
 Interpretation of the platform section of the JSON file is used to find which platform-specific sections may be available in the document. For example, if `os` is set to `linux`, then a JSON object conforming to the [Linux-specific schema](config-linux.md) SHOULD be found at the key `linux` in the `config.json`.
+
+[UTF-8]: http://www.unicode.org/versions/Unicode8.0.0/ch03.pdf
+[JSON]: http://json.org/
--- a/runtime.md
+++ b/runtime.md
@@ -3,7 +3,7 @@
 ## State
 
 The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
-The runtime state is stored in a JSON encoded file.
+The runtime state is stored in a [UTF-8][] [JSON][] file.
 It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
 On Linux based systems the state information should be stored in `/run/oci/containers`.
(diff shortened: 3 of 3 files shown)


runtime-config.md
```diff
@@ -1,3 +1,8 @@
+# Host-specific container configuration
+
+The bundle's top-level directory MUST contain a configuration file called `runtime.json` with [UTF-8][] [JSON][].
+For now the canonical schema is defined in [`runtime_config.go`](runtime_config.go) and [runtime_config_linux.go](runtime_config.go), but this will be moved to a formal JSON schema over time.
```
vbatts: I would have the OS specific reference be more generic than pointing out linux individually.

vbatts: and "moved to a formal JSON schema over time" where and why this comment?

wking: On Fri, Sep 04, 2015 at 11:44:48AM -0700, Vincent Batts wrote:
> and "moved to a formal JSON schema over time" where and why this comment?

Both are just echoing the current host-independent phrasing [1].  I
tried to make that clear in the commit message [2] and original PR
post with the:

  “To get a place to say this for the runtime.json docs, I used
  config.md as a template for the top-level header and blurb.”

text, but I'm happy to use different commit-message phrasing if that
would help make the source more obvious.

[1]: https://github.com/opencontainers/specs/blob/9d59c53bbf117075424f8bbf82b605a5c59b53f7/config.md#configuration-file

vbatts: I see that phasing. This feels like something to have roadmapped and discussed rather than alluding to.

wking: On Tue, Sep 08, 2015 at 03:59:13PM -0700, Vincent Batts wrote:
> This feels like something to have roadmapped and discussed rather
> than alluding to.

Sounds reasonable to me.  Shall I open the mailing-list thread, or
would you like to?

And is there an alternative phrasing you'd like to see here in the
meantime, so we can land the UTF-8 notes in this PR without waiting on
the desired-schema-language discussion?  Copying the existing phrasing
from config.md seemed like the least opinionated approach, but maybe
just dropping the “For now the canonical schema…” line with it's
Go-file links would be better?

jonboulle:
peanut gallery, why not use the JSON RFC here? https://tools.ietf.org/html/rfc7159

wking:
On Tue, Sep 08, 2015 at 04:39:10PM -0700, Jonathan Boulle wrote:
> peanut gallery, why not use the JSON RFC here?
> https://tools.ietf.org/html/rfc7159

How does that help?  RFC 7159 has [1]:

  JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
  encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
  interoperable in the sense that they will be read successfully by
  the maximum number of implementations; there are many
  implementations that cannot successfully read texts in other
  encodings (such as UTF-16 and UTF-32).

  Implementations MUST NOT add a byte order mark to the beginning of a
  JSON text.  In the interests of interoperability, implementations
  that parse JSON texts MAY ignore the presence of a byte order mark
  rather than treating it as an error.

To test the Go handling in particular, ‘json.NewDecoder(os.Stdin)’
seems to be able to handle UTF-8 on stdin:

  $ cat sample.json | go run test.go

but crashes with UTF-16LE:

  $ cat sample.json | iconv -f UTF-8 -t UTF-16LE - | go run test.go
  2015/09/08 20:35:00 invalid character '\x00' looking for beginning of object key string
  exit status 1

I'd rather stay away from BOMs.  I'd be ok with allowing runtimes to
optionally be configured to read the configuration JSON from other
encodings (e.g. UTF-16LE, …), but then we'd need some bundle-side
logic or bundle-introspection (e.g. BOM checks) to decide which
encoding to tell the runtime to use.  It seems easier for everyone if
we just require UTF-8 (but I may be missing something obvious ;).

And I see nothing about UTF-8 on json.org or the ECMA-404 it
references [2], so we'd want to make sure to link the RFC for our JSON
definition if we go this route.

[1]: https://tools.ietf.org/html/rfc7159#section-8.1
[2]: http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf

philips:
we should just put something in a glossary saying that when we say json in the spec we mean uft8 encoded json.

wking:
On Wed, Sep 09, 2015 at 09:46:41AM -0700, Brandon Philips wrote:
> we should just put something in a glossary saying that when we say
> json in the spec we mean uft8 encoded json.

I'd rather keep the glossary informative, and put normative stuff in
the spec itself.

crosbymichael:
+1 for what @philips said and we are currently working on a protobuf idea and discussing this during the meetings.

wking:
On Fri, Sep 25, 2015 at 01:43:33PM -0700, Michael Crosby wrote:
> +1 for what @philips said…

Glossary entry it is.  I'll bump #107.

> … and we are currently working on a protobuf idea and discussing
> this during the meetings.

This is about what authors provide in their bundles.  I don't see
where protobuf comes in…

PR #148, opened 2015-09-04
runtime: use opencontainer vs oci

philips:
In an effort to be more discoverable and self-describing change out the
term "oci" for "opencontainer". We should consider using this as the
prefix for global namespaces going forward as oc has conflicts with
existing tools and oci is the name of the initiative not the specs.


vbatts:
LGTM

mrunalp:
LGTM

PR #149, opened 2015-09-04
Make examples and text of namespace types match

glestaris:
The example and the list of namespace types differ on what the network namespace should be called. "net" seems most consistent with the other types.

@julz & @glestaris


mrunalp:
LGTM

glestaris:
Superseded by #152. Closing.

PR #152, opened 2015-09-04
Make namespace names match runc

julz:
"mount" rather than "mnt", "network" rather than "net"

If we prefer the longer form names. This would supersede #149 and https://github.com/opencontainers/runc/pull/247


vbatts:
LGTM

crosbymichael:
LGTM

crosbymichael:
LGTM

Issue #155, opened 2015-09-06
what happen if set mount path in linux namespaces and mounts in spec at the same time


gao-feng:
Allow joining-to-mnt-ns to  change the existing mount info? or mounts in spec should be ignored?

```
"mounts": {
    "data": {
        "type": "bind",
        "source": "/data",
        "options": ["rw"]
    }
    ...
}
"linux": {
    "namespaces" : [
        {
            "type": "mount",
            "path": "/proc/1234/ns/pid"
        }
        ...
    ]
    ...
}

wking:
On Sun, Sep 06, 2015 at 02:10:59AM -0700, Gao feng wrote:
> Allow joining-to-mnt-ns to change the existing mount info? or mounts
> in spec should be ignored?

I see two options that sound reasonable:

a. Mount config.json's mounts and then join runtime.json's mount
   namespace(s).
b. Join runtime.json's mount namespace(s) and then mount config.json's
   mounts.

Folks who don't want to change mounts can always remove the mounts
entry from their config.json.

I don't have a strong preference between (a) and (b), but (a)'s
outside-of-namespace mounts could be handled by hooks (which I think
are executed from the host's namespaces [1]), so maybe it's better to
do (b) for config-side mounts.

[1]: https://github.com/opencontainers/runc/pull/160#issuecomment-133091157

gao-feng:
Allow to change the mounts info of existed mnt ns? what's happen if config.json specified rootfs? allow to change the rootfs of mnt ns?

wking:
On Sun, Sep 06, 2015 at 11:21:26PM -0700, Gao feng wrote:
> Allow to change the mounts info of existed mnt ns? what's happen if
> config.json specified rootfs? allow to change the rootfs of mnt ns?

What does the ‘root’ entry actually do anyway?  I'd guess it just sets
up the mount namespace and tweaks R/W, propagation, etc. when doing
that.  So I'd suggest we error out with an “invalid config” error if
we have both a ‘root’ entry and a mount-path under linux.namespaces.

mrunalp:
@gao-feng The idea behind joining the existing namespaces is that no setup is required to be done in those namespaces by the runtime. I will create a PR to clarify that in the spec.

mrunalp:
Created #158

PR #158, opened 2015-09-08
Clarify behavior around namespaces paths

mrunalp:
See #155 

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -7,6 +7,7 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man7/na
 Namespaces are specified in the spec as an array of entries.
 Each entry has a type field with possible values described below and an optional path element.
 If a path is specified, that particular file is used to join that type of namespace.
+Also, when a path is specified, a runtime MUST assume that the setup for that particular namespace has already been done and error out if the config specifies anything else related to that namespace.
 
 ```json
     "namespaces": [


wking:
Instead of “ignore any other config when a namespace path is set”, I'd
rather just error-out if some incompatible values are set [1] and
support the join-and-tweak (or tweak-and-join) where possible [2].
Silently ignoring portions of the config seems like it would be a
source of confusion.

[1]: https://github.com/opencontainers/specs/issues/155#issuecomment-138350425
[2]: https://github.com/opencontainers/specs/issues/155#issuecomment-138099758

gao-feng:
@wking Can not join-and-tweat be implemented by hook, Ignore configuation may confuse someone, how about warning or error-out.

laijs:
The clarify is definitely necessary and important! Thanks!

I think we should also point out which config are the "other config related to that namespace" clearly.
And the implementation should output warning or error if it is hit.

```
Spec.Root
Spec.Mount
LinuxSpec.Linux.RootfsPropagation
RuntimeSpec.Mounts
LinuxRuntime.Devices
...
```

I think this list hasn't contains all

Any thought?

runtime-config-linux.md
```diff
@@ -7,6 +7,7 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man7/na
 Namespaces are specified in the spec as an array of entries.
 Each entry has a type field with possible values described below and an optional path element.
 If a path is specified, that particular file is used to join that type of namespace.
+Also, when a path is specified, a runtime MUST assume that the setup for that particular namespace has already been done and ignore any other config related to that namespace.
```
philips: Hrm, if there is a join and a create in the runtime I would say it should error, not ignore.

vishh: +1 for validation and raising an error. We essentially need something like [proto oneof](https://developers.google.com/protocol-buffers/docs/proto3?hl=en#oneof) to get around this issue.

dqminh: Right. I was wondering if we should describe config and state in a richer language than go struct ( like protobuf ), which we can then serialize to/from json.

wking:
On Wed, Sep 09, 2015 at 12:52:40AM -0700, Gao feng wrote:
> @wking Can not join-and-tweat be implemented by hook…

Lots of existing config stuff *can* be implemented via hooks
(e.g. everything under ‘mounts’).  But we have spec entries for it,
and there's nothing non-sensical about “Join the mnt namespace at
foo/bar and mount a tmpfs at /a/b/c”.  If we thought the “mount a
tmpfs at /a/b/c” was common enough to be in the spec for container
creation, I see no reason to have it *not* be in the spec for
namespace joinin.

> Ignore configuation may confuse someone, how about warning or
> error-out.

Yeah, I prefer explicitly failing to silently ignoring non-sensical
content [1].  I just don't think all “joining an existing namespace
and manipulating that namespace” specs are non-sensical.  But in
today's call, the consensus seemed to be punting on cases like that
until someone tells us a convincing story about why they would be
useful.  Until then, I'm fine erroring out whenever a config specifies
a namespace path and also modifications for that namespace.

[1]: https://github.com/opencontainers/specs/pull/158#issuecomment-138687129

runtime-config-linux.md
mrunalp: @vishh Yes, that is the requirement.
@philips Updated.

runtime-config-linux.md
```diff
@@ -7,6 +7,7 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man7/na
 Namespaces are specified in the spec as an array of entries.
 Each entry has a type field with possible values described below and an optional path element.
 If a path is specified, that particular file is used to join that type of namespace.
+Also, when a path is specified, a runtime MUST assume that the setup for that particular namespace has already been done and error out if there is any other config related to that namespace specified.
```
wking: “namespace specified” reads a little awkwardly to me.  Maybe “if the config specifies anything else related to that namespace”?

gao-feng:
@wking 
`Yeah, I prefer explicitly failing to silently ignoring non-sensical
content [1].  I just don't think all “joining an existing namespace
and manipulating that namespace” specs are non-sensical.  But in
today's call, the consensus seemed to be punting on cases like that
until someone tells us a convincing story about why they would be
useful.  Until then, I'm fine erroring out whenever a config specifies
a namespace path and also modifications for that namespace.
`
Thanks for your information, 
Let's disallow this behavior to get rid of the possible mistake.

wking:
02b456e9 looks good to me.

gao-feng:
LGTM

vbatts:
LGTM

crosbymichael:
LGTM

Issue #109, opened 2015-08-12
Add config validation and environment  tests


mrunalp:
The spec should have testing tools for validating the config and testing that an implementation provides the expected environment for the container process. This will involve launching a container process from a bundle that will introspect to make sure that the runtime set it up correctly.

- [ ] Config validation (mrunalp)
- [ ] Environment testing

philips:
@mrunalp by environment testing you mean generating an actual bundle and testing, from inside of the container, that the runtime ran correctly, right?

mrunalp:
@philips Yes, I have updated the description to clarify that.

wking:
On Wed, Aug 12, 2015 at 04:43:28PM -0700, Mrunal Patel wrote:
> - [ ] Config validation (mrunalp)
> - [ ] Environment testing

I'm pretty sure we all agree that these are important, but I don't
think they're part of the *spec*.  Can we get new repositories
implementing each of these?  It's possible that [OCT][1] is *already*
the environment-testing portion of this issue.

For config validation, I'm open to including more machine-parsable
annotations for things like whether an attribute is required or
optional, but I'm also fine just leaving that in the human-readable
spec and leaving it to the config-testing implementation to write the
code enforcing those requirements by hand.

It would be nice if we could offer that config validation as a web
service on opencontainers.org.

Should this discussion be happening on the mailing list (see #104)?

[1]: https://github.com/huawei-openlab/oct

vbatts:
@wking having a validator in specs is most logical

wking:
On Fri, Aug 14, 2015 at 04:09:38PM -0700, Vincent Batts wrote:
> @wking having a validator in specs is most logical

I imagine the relationship between config-validators and specs is
going to be many-to-many.  For example, opencontainers.org will
probably have a web-service validator written in Go that validates all
released versions of the spec.  But other folks may want validators
written in Python or C.  Since validating a config (like writting a
runtime implementation) seems easy to separate from writing the spec,
it makes more sense to me to use a separate repository, which makes
the many-to-many relationship easy to deal with.

Can you give more background on why it seems more logical to you to
build the validator into the spec repository?  How do you intend to
validate past versions of the spec?  Do you intend to version-bump the
spec repository when you find and fix bugs in the validator, or change
the validator's user-facing API, but don't touch the spec itself?

vbatts:
On Sat, Aug 15, 2015 at 2:30 PM, W. Trevor King <notifications@github.com>
wrote:

[...]

PR #159, opened 2015-09-08
Change the rlimit type to string instead of int

mrunalp:

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -156,8 +156,8 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
 ```
 
 rlimits allow setting resource limits.
-The type is from the values defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html).
-The kernel enforces the soft limit for a resource while the hard limit acts as a ceiling for that value that could be set by an unprivileged process.
+`type` is a string with a value from those defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html).
+The kernel enforces the `soft` limit for a resource while the `hard` limit acts as a ceiling for that value that could be set by an unprivileged process.
 
 ## SELinux process label
 
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -79,7 +79,7 @@ type IDMapping struct {
 // Rlimit type and restrictions
 type Rlimit struct {
 	// Type of the rlimit to set
-	Type int `json:"type"`
+	Type string `json:"type"`
 	// Hard is the hard limit for the specified type
 	Hard uint64 `json:"hard"`
 	// Soft is the soft limit for the specified type


vbatts:
should we set some const for these? (`RLIMIT_CPU`, `RLIMIT_AS`, ...)

mrunalp:
@vbatts I think for this case, we could let the runtime map from string to the constants just like for capabilities. That way we don't have to track changes as newer constants are added.

runtime-config-linux.md
```diff
@@ -155,7 +155,7 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
    ]
 ```
 
-rlimits allow setting resource limits. The type is from the values defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html). The kernel enforces the soft limit for a resource while the hard limit acts as a ceiling for that value that could be set by an unprivileged process.
+rlimits allow setting resource limits. `type` is a string with a value from those defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html). The kernel enforces the `soft` limit for a resource while the `hard` limit acts as a ceiling for that value that could be set by an unprivileged process.
```
wking: Touching this line is probably a good enough excuse to reformat with one line per-sentence.

wking:
I'd rather drop the leading RLIMIT.  See [1] in #109 and [2] in #160.

[1]: https://github.com/opencontainers/specs/issues/109#issuecomment-138687234
[2]: https://github.com/opencontainers/specs/pull/160#issuecomment-138691658

vbatts:
@wking but that is less clear?

wking:
On Tue, Sep 08, 2015 at 01:42:00PM -0700, Vincent Batts wrote:
> @wking but that is less clear?

Why is it less clear?  The context (linux.rlimits) should be enough to
make it clear that these are rlimit values.  For example, may
discussions of kernel config parameters drop the leading CONFIG_.  If
there's a strong feeling that the RLIMIT_-less version is unclear,
then I'll survive adding it in, but it feels redundant to me ;).

vbatts:
Those dropped-prefixes are idioms elected for those projects internal
references. Externally their reference is the full name. Doing this is to
propose that we impose special prefix logic those that implement.

wking:
On Tue, Sep 08, 2015 at 02:20:01PM -0700, Vincent Batts wrote:
> Those dropped-prefixes are idioms elected for those projects
> internal references. Externally their reference is the full
> name.

That's not just because C doesn't namespace definitions?  Since we get
namespacing via the config's JSON keys, it seems redundant to also
have namespace prefixes on the values themselves.  Are you worried
about folks seeing something like ‘CPU’ out of context and not
realizing that it's a resource-limit at all?  Or are you worried that
it will be hard to grep for resource-limit-specific occurences of CPU
without also hitting other occurences of CPU (e.g. if there was a
CAP_CPU)?

vbatts:
Context is something to be aware of. Honestly, I'm not sure whether you're
arguing for actual technical reasons and simplistic design, or just to
derive another idiom for implementation. if the capability has CAP_ in
front, then have CAP in front. We do not control whether one day that
prefix changes, and now the semantic pattern is broken.

wking:
On Tue, Sep 08, 2015 at 03:39:11PM -0700, Vincent Batts wrote:
> Context is something to be aware of. Honestly, I'm not sure whether
> you're arguing for actual technical reasons and simplistic design,
> or just to derive another idiom for implementation. if the
> capability has CAP_ in front, then have CAP in front. We do not
> control whether one day that prefix changes, and now the semantic
> pattern is broken.

I think it's unlikely that the prefix will change ;).  But there's no
harm other than redundancy for keeping the CAP_ around, and I guess a
little redundancy and extra typing isn't a big deal :p.

mrunalp:
One could argue both ways but I think it is better to be explicit here. Also, some tooling could benefit by directly using these definitions from their header files.

wking:
On Tue, Sep 08, 2015 at 04:03:58PM -0700, Mrunal Patel wrote:
> One could argue both ways but I think it is better to be explicit
> here.

This seems fair.

> Also, some tooling could benefit by directly using these definitions
> from their header files.

Adding the prefix back in before searching headers seems easy enough,
so I'm not buying the tool-support argument ;).

But “the redunancy reduction from removing the namespacing prefix is
not useful enough to be worth trimming the upstream identifier” seems
sufficiently reasonable to carry this PR's approach without appealing
to tool-support.

wking:
On Tue, Sep 08, 2015 at 03:39:11PM -0700, Vincent Batts wrote:
> Honestly, I'm not sure whether you're arguing for actual technical
> reasons and simplistic design…

And I think everyone who's chimed in is on board with prefixed strings
now, but the seed of my initial resistance was preserving the current
capabilities deprefixing (which is going away with #160).  And we've
had that since the ancient days of 77d44b10 (Update runtime.md,
2015-06-16).  Unfortunately that commit landed in the pre-PR days, so
there's no motivation for its use of deprefixed capabilities strings.
Since that stuff was mostly from libcontainer, I looked there, and the
deprefixed form landed with docker/libcontainer@6415e8be (Initial
commit of libcontainer, 2014-02-18).  I haven't dug into the
provenance of *that* code, so I'm not sure what the original
motivation was for the deprefixed approach (maybe it was like my
preference, since sidelined, for avoiding redundancy).

philips:
Seems fine, needs a rebase.

mrunalp:
Rebased.

philips:
lgtm

vbatts:
LGTM

PR #160, opened 2015-09-08
Modify the capabilities constants to match header files like other constants

mrunalp:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -6,13 +6,13 @@ Additional information is needed for Linux over the [default spec configuration]
 ## Capabilities
 
 Capabilities is an array that specifies Linux capabilities that can be provided to the process inside the container.
-Valid values are the string after `CAP_` for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html).
+Valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html)
 
 ```json
    "capabilities": [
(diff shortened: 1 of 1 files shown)


config-linux.md
```diff
@@ -7,9 +7,8 @@ in order to configure these various kernel features.
 
 ## Capabilities
 
-Capabilities is an array that specifies Linux capabilities that can be provided to the process
-inside the container. Valid values are the string after `CAP_` for capabilities defined
-in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html)
+Capabilities is an array that specifies Linux capabilities that can be provided to the process inside the container.
+Valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html)
```
LK4D4: Maybe add (e.g. CAP_SYS_ADMIN), because those aren't strings in man.

mrunalp: I modified the example json.

vbatts:
LGTM

wking:
Why require the leading CAP_?  Isn't that already clear from the
context (linux.capabilities)?

philips:
to-may-to to-mah-to, is there a strong argument one way or the other?

mrunalp:
@philips Thinking it might be better to explicit in these cases and just match the values defined in the header files at the expense of typing few more characters.

Rebased.

vishh:
LGTM

philips:
lgtm

PR #164, opened 2015-09-09
config-linux: specify the default devices/filesystems available

### LINKED ISSUE #95: linux: specify the default devices/filesystems available

philips:
Fixes #95

Signed-off-by: Brandon Philips <brandon.philips@coreos.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -49,3 +49,25 @@ uid/gid mappings describe the user namespace mappings from the host to the conta
 The mappings represent how the bundle `rootfs` expects the user namespace to be setup and the runtime SHOULD NOT modify the permissions on the rootfs to realize the mapping.
 *hostID* is the starting uid/gid on the host to be mapped to *containerID* which is the starting uid/gid in the container and *size* refers to the number of ids to be mapped.
 There is a limit of 5 mappings which is the Linux kernel hard limit.
+
+## Default Devices and File Systems
+
+The Linux ABI includes both syscalls and several special file paths.
+Applications expecting a Linux environment will very likely expect these files paths to be setup correctly.
+
+The following devices and filesystems MUST be made available in each application's filesystem
+
+|     Path     |  Type  |  Notes  |
+| ------------ | ------ | ------- |
+| /proc        | [procfs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)    | |
+| /sys         | [sysfs](https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt)    | |
+| /dev/null    | [device](http://man7.org/linux/man-pages/man4/null.4.html)                 | |
+| /dev/zero    | [device](http://man7.org/linux/man-pages/man4/zero.4.html)                 | |
+| /dev/full    | [device](http://man7.org/linux/man-pages/man4/full.4.html)                 | |
+| /dev/random  | [device](http://man7.org/linux/man-pages/man4/random.4.html)               | |
+| /dev/urandom | [device](http://man7.org/linux/man-pages/man4/random.4.html)               | |
+| /dev/tty     | [device](http://man7.org/linux/man-pages/man4/tty.4.html)                  | |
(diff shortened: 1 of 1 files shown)


wking:
I'm still in favor of just mounting devtmpfs for the devices [2].
That's currently broken in runC (opencontainers/runc#243), but the
spec shouldn't block based on runtime implementations.  The only
devtmpfs pushback I've heard is that you don't want the mounted
devices to depend on optional kernel features [2,3], but it seems like
the two sticky cases would be things like:

a. The kernel doesn't support tmpfs, /dev/random, or some other
   feature this PR would require.
b. The kernel suppoers devices (like /dev/comedi0) that this PR
   doesn't require.

For (a), I'd rather allow those systems to use OCI bundles, even
though they have a very limited kernel.  Does that sound ok to you?

For (b), I'd rather get those devices auto-mounted via devtmpfs.  You
can always use cgroups to blacklist/whitelist devices you don't want
the container accessing, so I don't see a problem with exposing them
via devtmpfs.  Does that sound ok to you?

[1]: https://github.com/opencontainers/specs/issues/95#issuecomment-131265209
[2]: https://github.com/opencontainers/specs/issues/95#issuecomment-131266716
[3]: https://github.com/opencontainers/specs/issues/95#issuecomment-131266857
[4]: https://github.com/opencontainers/specs/issues/95#issuecomment-131265209

crosbymichael:
LGTM

vbatts:
LGTM

mrunalp:
LGTM

wking:
On Wed, Sep 09, 2015 at 10:16:49AM -0700, Mrunal Patel wrote:
> Merged #164.

Just to document the devtmpfs resolution, in today's call we decided
that:

* This is a required lower bar.  So trying to run an OCI runtime on
  kernels like (a) in my previous post without /dev/random [1] will
  result in errors.

* Runtimes will only create devices if the listed entry is missing.
  That way we don't clobber entries in the host's filesystem if the
  user launches a container without a mount namespace.

* Runtimes will create the listed devices before iterating through
  config-specified mounts.  So you can bind-mount the required entries
  to other locations via your config-specified mounts (although I'm
  not sure why you would want to do that ;).  This last point didn't
  get a lot of airtime in the call, so it may not be settled.

[1]: https://github.com/opencontainers/specs/pull/164#issuecomment-138978263

PR #165, opened 2015-09-09
runtime-config: Remove 'destination' docs from mounts

wking:
c18c283 (Change layout of mountpoints and mounts, 2015-09-02, #136)
removed the destination field from the Go type and examples, but
[forgot to remove it from the documentation][1].  Fix that with this
commit.

[1]: https://github.com/opencontainers/specs/issues/109#issuecomment-138531616


mrunalp:
LGTM

vbatts:
LGTM

PR #167, opened 2015-09-09
.tools: repo validation tool

vbatts:
Initially only a DCO validation for travis

https://github.com/opencontainers/specs/issues/90


.tools/validate.go
```diff
@@ -0,0 +1,196 @@
...
+	flag.Parse()
+
+	/*
+		if len(os.Getenv("TRAVIS")) > 0 {
+			fmt.Println(os.Getenv("TRAVIS_COMMIT_RANGE"))
+			if len(os.Getenv("TRAVIS_PULL_REQUEST")) > 0 {
+				fmt.Println(os.Getenv("TRAVIS_BRANCH"))
+			}
+		}
+	*/
```
wking: We should probably either drop this debugging comment or uncomment it and put it behind a flag before landing this PR.

.tools/validate.go
```diff
@@ -0,0 +1,196 @@
...
+		for _, commit := range c {
+			fmt.Printf(" * %s %s ... ", commit.AbbreviatedCommit, commit.Subject)
+			if err := ValidateCommit(commit, DefaultRules); err != nil {
+				invalids = append(invalids, InvalidCommit{commit, err})
+				fmt.Printf("FAIL: %s\n", err)
+			} else {
+				fmt.Println("PASS")
+			}
+		}
+	}
```
wking: Maybe error out if neither `TRAVIS_COMMIT_RANGE` nor the `--range` flag are set?

.tools/validate.go
```diff
@@ -0,0 +1,196 @@
...
+		os.Exit(1)
+	}
+}
+
+type ValidateRule func(CommitEntry) error
+
+func ValidateCommit(c CommitEntry, rules []ValidateRule) error {
+	for _, r := range rules {
+		if err := r(c); err != nil {
+			return err
```
wking: Instead of failing after the first invalid rule, maybe always iterate over all rules and return an array of failures?  Otherwise there will be noise in PRs as folks fix the initial error, repush, get notified of the previously-masked error, fix it, repush, ….

.tools/validate.go
```diff
@@ -0,0 +1,196 @@
...
+	output := buf.Bytes()
+	if err := json.Unmarshal(output, &c); err != nil {
+		fmt.Println(string(output))
+		return nil, err
+	}
+	output, err := exec.Command("git", "log", "-1", prettyLogSubject, commit).Output()
+	if err != nil {
+		return nil, err
+	}
+	c.Subject = strings.TrimSpace(string(output))
```
wking: Why not get the subject from the initial `prettyLogFormat` output?  Is the concern with using `%s` that there will be double-quotes in the subject, which would need to be escaped in the JSON?  It seems like there could be double-quotes in the author name or committer name too, so maybe it's best to pull those out as well?  I think a loop over all user-supplied fields with a single log call per-field/commit is probably best.  That's what you're doing here for the subject and body (but you've unrolled the loop).

.tools/validate.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func GitFetchHeadCommit() (string, error) {
+	output, err := exec.Command("git", "rev-parse", "--verify", "HEAD").Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+func GitHeadCommit() (string, error) {
```
wking: This doesn't seem to be used anywhere.  Maybe leave it out until we need it?

.tools/validate.go
```diff
@@ -0,0 +1,196 @@
...
+		c, err := GitLogCommit(commitHash)
+		if err != nil {
+			return commits, err
+		}
+		commits[i] = *c
+	}
+	return commits, nil
+}
+
+func GitFetchHeadCommit() (string, error) {
```
wking: This doesn't seem to be used anywhere. Maybe leave it out until we need it?

.tools/validate.go
vbatts: perhaps. maybe i'll just make this tool less involved in travis, and the .travis.yml to pass it's args  to flags.

.tools/validate.go
vbatts: perhaps. Though there is only the one test for now.

.tools/validate.go
vbatts: because subject and body can have symbols that --pretty does not escape. :-)

.tools/validate.go
wking: On Wed, Sep 09, 2015 at 12:49:38PM -0700, Vincent Batts wrote:
> perhaps. maybe i'll just make this tool less involved in travis, and
> the .travis.yml to pass it's args to flags.

That sounds easier to me.

.tools/validate.go
wking: On Wed, Sep 09, 2015 at 12:50:47PM -0700, Vincent Batts wrote:
> perhaps. Though there is only the one test for now.

Sure.  I'm ok punting until we add another test, although I'm not sure
what my odds are of remembering to comment again then ;).  An easier
approach would be to just log the error messages directly (maybe with
TAP-friendly output?) from this helper, and then return a boolean for
“there were (or weren't) errors with this commit”.

.tools/validate.go
wking: On Wed, Sep 09, 2015 at 12:52:00PM -0700, Vincent Batts wrote:
> because subject and body can have symbols that --pretty does not
> escape. :-)

So can author names:

  $ git init
  $ git config user.name 'Mua"haha'
  $ touch README
  $ git add README
  $ git commit -m 'Testing crazy author name'
  $ git log --format='{"author": "%aN"}'
  {"author": "Mua"haha"}

which is why I'm suggesting “a loop over all user-supplied fields with
a single log call per-field/commit” [1] ;).

[1]: https://github.com/opencontainers/specs/pull/167/files#r39085317

.tools/validate.go
vbatts: not sure about conforming to TAP for now.
Also, I've just accommodated having whole result sets.

.tools/validate.go
vbatts: done

.tools/validate.go
vbatts: fixed this as well.

vbatts:
yay. updated, and now that travis is enabled you can see that it also validated the DCO in this PR. :-)

vbatts:
@tianon review?

tianon:
Adorbs. :+1:

tianon:
(LGTM)

.tools/validate.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+		if !hasValid {
+			vr.Pass = false
+			vr.Msg = "does not have a valid DCO"
+		} else {
+			vr.Pass = true
+			vr.Msg = "has a valid DCO"
+		}
+
+		return vr, nil
```
wking: If we never return anything other than `nil` for `err`, I'd just make the signature:

    func(c CommitEntry) (vr ValidateResult)

.tools/validate.go
```diff
@@ -0,0 +1,269 @@
...
+var DefaultRules = []ValidateRule{
+	func(c CommitEntry) (vr ValidateResult, err error) {
+		vr.CommitEntry = c
+		hasValid := false
+		for _, line := range strings.Split(c.Body, "\n") {
+			if validDCO.MatchString(line) {
+				hasValid = true
+			}
+		}
+		if !hasValid {
```
wking: `<nit>`To avoid a double-negative (`!hasValid {} else {}`), maybe drop the `!` and flip the if/else clauses?`</nit>`

.tools/validate.go
vbatts: fair...

.tools/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	c, err := GitCommits(commitrange)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	results := ValidateResults{}
+	for _, commit := range c {
+		fmt.Printf(" * %s %s ... ", commit.AbbreviatedCommit, commit.Subject)
+		if vr, err := ValidateCommit(commit, DefaultRules); err != nil {
+			log.Fatalf("%s: %s", commit.Commit, err)
```
wking: Here's another case where I'd prefer we log the error (and eventually return a non-zero exit code) but still continue on to process the other commits.

wking: Ah, nevermind.  If we [drop the error-code][1], this block will go away entirely.

[1]: https://github.com/opencontainers/specs/pull/167/files#r39100899

.tools/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	Name  string `json:"name"`
+	Email string `json:"email"`
+	Date  string `json:"date"` // this could maybe be an actual time.Time
+}
+
+var (
+	prettyLogSubject       = `--pretty=format:%s`
+	prettyLogBody          = `--pretty=format:%b`
+	prettyLogCommit        = `--pretty=format:%H`
+	prettyLogAuthorName    = `--pretty=format:%aN`
```
wking: I'm not sure how old Travis' Git is (and I haven't looked up `--format` in Git's release notes), but at least with v2.1 you can use `--format=…` instead of `--pretty=format:…`.

I also think it makes sense to drop the leading `--pretty=format:` (or whatever) here and just list `%s`, `%b`, etc.  We can inject the option prefix once in the consuming code.

.tools/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	prettyLogAuthorEmail   = `--pretty=format:%aE`
+	prettyLogCommiterName  = `--pretty=format:%cN`
+	prettyLogCommiterEmail = `--pretty=format:%cE`
+	prettyLogFormat        = `--pretty=format:{"commit": "%H", "abbreviated_commit": "%h", "tree": "%T", "abbreviated_tree": "%t", "parent": "%P", "abbreviated_parent": "%p", "refs": "%D", "encoding": "%e", "sanitized_subject_line": "%f", "commit_notes": "%N", "verification_flag": "%G?", "signer": "%GS", "signer_key": "%GK", "author": { "date": "%aD" }, "commiter": { "date": "%cD" }}`
```
wking: I'd drop `"commit_notes": "%N"`, since it seems like more user-generated content that could contain double quotes.  But I've never seen anyone actually use Git notes, so I'm not 100% sure that `%N` is the same stuff as `git notes add …`.  `"signer": "%GS"` should also be dropped from the JSON as user-generated content.

.tools/validate.go
vbatts: done and push. PTAL

.tools/validate.go
vbatts: ohman

.tools/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return nil, err
+	}
+	c.Commiter.Name = strings.TrimSpace(string(output))
+
+	output, err = exec.Command("git", "log", "-1", prettyLogCommiterEmail, commit).Output()
+	if err != nil {
+		return nil, err
+	}
+	c.Commiter.Email = strings.TrimSpace(string(output))
```
wking: Does Go not have an equivalent of Python's `setattr`?  In Python, I'd make this:

    for format, keys in [
            ('%s', ['subject']),
            …
            ('%cE', ['committer', 'email']),
            ]:
        value = …
        obj = commit
        for key in keys[:-1]:
            obj = getattr(obj, key)
        setattr(obj, keys[-1], value)

.tools/validate.go
```diff
@@ -0,0 +1,260 @@
...
+
+	return &c, nil
+}
+
+// GitCommits returns a set of commits. If commitrange is a git still range 12345...54321, the it will be isolated set of commits. If commitrange is a single commit, the set will be beginning commit up to the hash provided.
```
wking: “the it will” → “then it will”.

And “the set will be beginning commit up to the hash provided” → “all ancestor commits up through the hash provided” (because there may be more than one parent-less commit in a given commit's ancestry, e.g. if you'd merged an orphan branch).

.tools/validate.go
vbatts: Not without reflect, which is overkill for doing a simple loop.

.tools/validate.go
wking: On Wed, Sep 09, 2015 at 03:02:14PM -0700, Vincent Batts wrote:
> ohman

https://xkcd.com/327/ ;)

.tools/validate.go
wking: On Wed, Sep 09, 2015 at 03:08:43PM -0700, Vincent Batts wrote:
> Not without reflect, which is overkill for doing a simple loop.

Ah bummer.  And I have a friend who would be waving his Lisp-y
parenthesis at this too ;).

vbatts:
updated ... PTAL!

wking:
GitHub seems to have eaten my emailed response, so I'll repeat it here:

On Wed, Sep 09, 2015 at 03:15:53PM -0700, Vincent Batts wrote:
> updated ... PTAL!

8b55acffb looks good enough to me.  I'd still recommend the --format
updates suggested in [1], but if you'd rather maintain it the way it
is, then I'm happy enough ;).

[1]: https://github.com/opencontainers/specs/pull/167/files#r39101906

vbatts:
@wking i'm not going to worry about depending on their git version.

philips:
nit: not a huge fan of hiding code. tools without the .tools would be better IMHO

philips:
good start, better than the nothing we have now. LGTM for merge.

vbatts:
@philips i only hid it because it's like project housekeeping. If it were in `./tools/` or just ignored but sitting with a name like `validate.go` it may be expected for specs tools or validation. right?

vbatts:
we can move it out later, or just have that kind of git validation be `go get`able for travis use. For now, I'm going to merge this for DCO enforcement of PRs

vbatts:
lol. feature needed!

wking:
On Wed, Sep 09, 2015 at 08:17:27PM -0700, Vincent Batts wrote:
> we can move it out later, or just have that kind of git validation
> be `go get`able for travis use…

This sounds good to me, especially since “I want to validate PRs
against our project's commit policy” seems like something that other
repositories might want to use too.

vbatts:
Yup

.tools/validate.go
laijs: Using the map instead of the structure can avoid reflect. Even using reflect, it is much simpler than current code IMO.

PR #170, opened 2015-09-10
travis: fix DCO validation for merges

vbatts:
merged PRs do not have DCO, so do not check for them

https://github.com/opencontainers/specs/issues/169

### DIFF
--- a/.tools/validate.go
+++ b/.tools/validate.go
@@ -16,14 +16,14 @@ import (
 var DefaultRules = []ValidateRule{
 	func(c CommitEntry) (vr ValidateResult) {
 		vr.CommitEntry = c
-		if len(strings.Split(c.Parent, " ")) > 1 {
+		if len(strings.Split(c["parent"], " ")) > 1 {
 			vr.Pass = true
 			vr.Msg = "merge commits do not require DCO"
 			return vr
 		}
 
 		hasValid := false
-		for _, line := range strings.Split(c.Body, "\n") {
+		for _, line := range strings.Split(c["body"], "\n") {
 			if validDCO.MatchString(line) {
 				hasValid = true
 			}
@@ -69,7 +69,7 @@ func main() {
 
 	results := ValidateResults{}
 	for _, commit := range c {
-		fmt.Printf(" * %s %s ... ", commit.AbbreviatedCommit, commit.Subject)
+		fmt.Printf(" * %s %s ... ", commit["abbreviated_commit"], commit["subject"])
 		vr := ValidateCommit(commit, DefaultRules)
 		results = append(results, vr...)
 		if _, fail := vr.PassFail(); fail == 0 {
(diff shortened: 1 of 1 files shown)


.tools/validate.go
```diff
@@ -16,6 +16,12 @@ import (
 var DefaultRules = []ValidateRule{
 	func(c CommitEntry) (vr ValidateResult) {
 		vr.CommitEntry = c
+		if len(strings.Split(c.Parent, " ")) > 1 {
```
wking: Instead of splitting at consume-time, it seems cleaner to make `CommitEntry.Parent` a `CommitEntry.Parents` array and parse at commit-lookup time.

.tools/validate.go
```diff
@@ -70,7 +76,7 @@ func main() {
 			fmt.Println("PASS")
 			if *flVerbose {
 				for _, r := range vr {
-					if !r.Pass {
+					if r.Pass {
```
wking: Heh, sorry I missed this while reviewing #167.  But we can probably drop this check entirely, since we only hit this code when `fail == 0`.

vbatts: not true. this is so that you can see the messages for passed tests as well.

.tools/validate.go
vbatts: meh. that would be another separated shell-out, and would then also have to be done for abbreviated parent as well.

vbatts:
@laijs per https://github.com/opencontainers/specs/pull/167/files#r39153774 see the cleanup commit.

.tools/validate.go
wking: On Thu, Sep 10, 2015 at 06:25:33AM -0700, Vincent Batts wrote:
> > -					if !r.Pass {
> > +					if r.Pass {
> 
> not true. this is so that you can see the messages for passed tests
> as well.

You have (from #167):

    if _, fail := vr.PassFail(); fail == 0 {
      fmt.Println("PASS")
      if *flVerbose {
        for _, r := range vr {
          if !r.Pass {
            fmt.Printf("  - %s\n", r.Msg)
          }
        }
      }
    } else {
      fmt.Println("FAIL")
      // default, only print out failed validations
      for _, r := range vr {
        if !r.Pass {
          fmt.Printf("  - %s\n", r.Msg)
        }
      }
    }

The fix you're proposing here would be the same as:

    if _, fail := vr.PassFail(); fail == 0 {
      fmt.Println("PASS")
      if *flVerbose {
        for _, r := range vr {
          fmt.Printf("  - %s\n", r.Msg)
        }
      }
    } else {
      fmt.Println("FAIL")
      // default, only print out failed validations
      for _, r := range vr {
        if !r.Pass {
          fmt.Printf("  - %s\n", r.Msg)
        }
      }
    }

So keep the Printf, but drop the r.Pass check from the ‘fail== 0’
branch (because we know that none of the checks failed).

If you want to see passes in all cases, I think you want:

    if _, fail := vr.PassFail(); fail == 0 {
      fmt.Println("PASS")
    } else {
      fmt.Println("FAIL")
    }
    if *flVerbose {
      for _, r := range vr {
        if r.Pass {
          fmt.Printf("ok")
        }
        if r.Pass {
          fmt.Printf("not ok")
        }
        fmt.Println(" %s", r.Msg)
      }
    }

.tools/validate.go
wking: On Thu, Sep 10, 2015 at 06:26:27AM -0700, Vincent Batts wrote:
> > @@ -16,6 +16,12 @@ import (
> >  var DefaultRules = []ValidateRule{
> >  	func(c CommitEntry) (vr ValidateResult) {
> >  		vr.CommitEntry = c
> > +		if len(strings.Split(c.Parent, " ")) > 1 {
> 
> meh. that would be another separated shell-out…

Why?  Can't you just have:

  … "parents": ["%P"], "abbreviated_parents": ["%p"], …,

in prettyLogFormat, and then:

  c.Parents = strings.Split(c.Parents, " ")
  c.AbbreviatedParents = strings.Split(c.AbbreviatedParents, " ")

after your Unmarshal?

.tools/validate.go
vbatts: not. test the code. ...

crosbymichael:
LGTM

vbatts:
@mrunalp review?

mrunalp:
@vbatts Looking

.tools/validate.go
wking: On Thu, Sep 10, 2015 at 11:16:35AM -0700, Vincent Batts wrote:
> > -					if !r.Pass {
> > +					if r.Pass {
> 
> not. test the code. ...

I spun off my suggestion into #174.

mrunalp:
LGTM

PR #141, opened 2015-09-03
runtime_config: Replace 'MountPoints in Spec' with 'Spec.Mounts'

wking:
I think Go's attribute syntax reads more clearly here, especially
since there is no 'Spec.MountPoints' after c18c283a (Change layout of
mountpoints and mounts, 2015-09-02, #136).

As I [suggested][1] for #136.

[1]: https://github.com/opencontainers/specs/pull/136/files#r38583510

PR #171, opened 2015-09-10
move the description of user ns mapping and default files to proper file

gao-feng:
They should stay in runtime not config.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -16,30 +16,6 @@ Valid values are the strings for capabilities defined in [the man page](http://m
     ]
 ```
 
-## User namespace mappings
-
-```json
-    "uidMappings": [
-        {
-            "hostID": 1000,
-            "containerID": 0,
-            "size": 10
-        }
-    ],
-    "gidMappings": [
-        {
-            "hostID": 1000,
-            "containerID": 0,
-            "size": 10
-        }
-    ]
-```
-
-uid/gid mappings describe the user namespace mappings from the host to the container.
-The mappings represent how the bundle `rootfs` expects the user namespace to be setup and the runtime SHOULD NOT modify the permissions on the rootfs to realize the mapping.
-*hostID* is the starting uid/gid on the host to be mapped to *containerID* which is the starting uid/gid in the container and *size* refers to the number of ids to be mapped.
-There is a limit of 5 mappings which is the Linux kernel hard limit.
-
 ## Default Devices and File Systems
 
 The Linux ABI includes both syscalls and several special file paths.
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -152,6 +152,8 @@ The Spec does not include naming schema for cgroups.
 The Spec does not support [split hierarchy](https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt).
 The cgroups will be created if they don't exist.
 
+###### Example
+
 ```json
    "cgroupsPath": "/myRuntime/myContainer"
 ```
@@ -181,6 +183,8 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 #### Set oom_score_adj
 
 More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+
+###### Example
 
 ```json
     "oomScoreAdj": 0
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -49,6 +49,32 @@ Also, when a path is specified, a runtime MUST assume that the setup for that pa
     ]
 ```
 
+## User namespace mappings
+
+###### Example
+
+```json
+    "uidMappings": [
+        {
+            "hostID": 1000,
+            "containerID": 0,
+            "size": 10
+        }
+    ],
+    "gidMappings": [
+        {
+            "hostID": 1000,
+            "containerID": 0,
+            "size": 10
+        }
+    ]
+```
+
+uid/gid mappings describe the user namespace mappings from the host to the container.
+The mappings represent how the bundle `rootfs` expects the user namespace to be setup and the runtime SHOULD NOT modify the permissions on the rootfs to realize the mapping.
+*hostID* is the starting uid/gid on the host to be mapped to *containerID* which is the starting uid/gid in the container and *size* refers to the number of ids to be mapped.
+There is a limit of 5 mappings which is the Linux kernel hard limit.
+
 ## Devices
 
 `devices` is an array specifying the list of devices to be created in the container.


runtime-config.md
```diff
@@ -36,6 +36,15 @@ Only [mounts from the portable config](config.md#mount-points) will be mounted.
     }
 }
 ```
+Note: The following filesystems MUST be made available in each Linux application's filesystem
```
wking: This is Linux-specific stuff, so I think it should go in the Linux-specific docs too.

wking:
On Thu, Sep 10, 2015 at 03:10:53AM -0700, Gao feng wrote:
> They should stay in runtime not config.

I'd rather land #126 (which needs a rebase now, but I'm waiting until
there's enough interest in merging it before doing that ;) or at least
some of its earlier commits.  With an easier mapping between JSON
files and their associated .md and .go, it's more obvious when docs
end up in the wrong place.

runtime-config.md
gao-feng: @wking I thinks describe this after the mounts Exmaple for linux will exactly tell people to make sure these filesystems related mounts should be configured in runtime.json

wking: On Thu, Sep 10, 2015 at 08:46:15PM -0700, Gao feng wrote:
> > +Note: The following filesystems MUST be made available in each Linux application's filesystem
> 
> @wking I thinks describe this after the mounts Exmaple for linux…

That's a good point.  I guess my preferred approach would be to move
the Linux-specific mount example into the Linux-specific
documentation, and just link to the per-platform examples from this
platform-agnostic documentation.  Otherwise this is a reasonable chunk
of noise for folks working on other platforms.

gao-feng: Good idea, will update, thanks!

gao-feng:
@wking Updated, please check, thanks!

runtime-config-linux.md
```diff
@@ -45,6 +45,30 @@ container via system level IPC.
...
 within the container.
 
+## User namespace mappings
+
+uid/gid mappings describe the user namespace mappings from the host to the container.
+The mappings represent how the bundle `rootfs` expects the user namespace to be setup and the runtime SHOULD NOT modify the permissions on the rootfs to realize the mapping.
+*hostID* is the starting uid/gid on the host to be mapped to *containerID* which is the starting uid/gid in the container and *size* refers to the number of ids to be mapped.
+There is a limit of 5 mappings which is the Linux kernel hard limit.
+
+```json
```
wking: I know you're just copy-pasting here, but since you reorder, it might be worth adding the `*Example*` psuedo-header that seems to be our convention for setting off examples.

runtime-config-linux.md
```diff
@@ -126,6 +150,55 @@ Next parameters can be specified:
...
+| /dev/null    | [device](http://man7.org/linux/man-pages/man4/null.4.html)                 | |
+| /dev/zero    | [device](http://man7.org/linux/man-pages/man4/zero.4.html)                 | |
+| /dev/full    | [device](http://man7.org/linux/man-pages/man4/full.4.html)                 | |
+| /dev/random  | [device](http://man7.org/linux/man-pages/man4/random.4.html)               | |
+| /dev/urandom | [device](http://man7.org/linux/man-pages/man4/random.4.html)               | |
+| /dev/tty     | [device](http://man7.org/linux/man-pages/man4/tty.4.html)                  | |
+| /dev/console | [device](http://man7.org/linux/man-pages/man4/console.4.html)              | |
+| /dev/ptmx    | [device](https://www.kernel.org/doc/Documentation/filesystems/devpts.txt)  | Bind-mount or symlink of /dev/pts/ptmx |
+
+## Mounts in Linux
```
wking: This whole file is “in Linux”.  I'd just make this header “Mounts”.  And probably add a one-line intro linking back to the platform-agnostic mounts docs for the generic information.

runtime-config-linux.md
```diff
@@ -126,6 +150,55 @@ Next parameters can be specified:
...
+    },
+    "data": {
+        "type": "bind",
+        "source": "/volumes/testing",
+        "options": ["rbind","rw"]
+    }
+}
+```
+
+Note: The following filesystems MUST be made available in each Linux application's filesystem
```
wking: These sections (and also for devices) are normative, so I'd put them above the informative examples.

runtime-config.md
```diff
@@ -12,30 +12,7 @@ Only [mounts from the portable config](config.md#mount-points) will be mounted.
...
-    "data": {
-        "type": "bind",
-        "source": "/volumes/testing",
-        "options": ["rbind","rw"]
-    }
-}
-```
+See Mounts [example](runtime-config-linux.md#mounts-in-linux) in Linux
 
 *Example (Windows)*
```
wking: Is it worth pulling the Windows (and hypervisor?  See #133) examples out into platform-specific files as well.  `runtime-config-windows.md`, etc., would match our current Linux pattern, but that would currently be a fairly sparse file.  Still, I expect it will grow in time, and breaking it out now will set the stage for cleaner future expansion.

dqminh: I think this makes sense. but probably a different PR.

wking: On Fri, Sep 11, 2015 at 09:25:56AM -0700, Daniel, Dao Quang Minh wrote:
> >  *Example (Windows)*
> 
> I think this makes sense. but probably a different PR.

Fair enough.  I'm fine with this PR landing as it stands with 5ab8432
and filing subsequent PRs to clean things up more (but that approach
has been slow before, e.g. see #141 ;).  Of my outstanding comments,
the overspecified header seems the lowest-hanging fruit [1], although
[2,3] seem easy and uncontroversial as well.

[1]: https://github.com/opencontainers/specs/pull/171/files#r39280224
[2]: https://github.com/opencontainers/specs/pull/171/files#r39280300
[3]: https://github.com/opencontainers/specs/pull/171/files#r39280066

runtime-config-linux.md
gao-feng: @wking You mean we should add `Example` before all of json examples in this file?

wking: On Sun, Sep 13, 2015 at 07:34:11PM -0700, Gao feng wrote:
> @wking You mean we should add `Example` before all of json examples
> in this file?

Yes, although we can leave examples you aren't touching for follow-up
PRs (although it looks like you got a lot of them in 107a454a0, and
I'm fine with that too).

gao-feng: Thanks, append another commit :)

wking: 463042f looks good to me.

config-linux.md
```diff
@@ -15,49 +15,3 @@ Valid values are the strings for capabilities defined in [the man page](http://m
...
-        }
-    ]
-```
-
-uid/gid mappings describe the user namespace mappings from the host to the container.
-The mappings represent how the bundle `rootfs` expects the user namespace to be setup and the runtime SHOULD NOT modify the permissions on the rootfs to realize the mapping.
-*hostID* is the starting uid/gid on the host to be mapped to *containerID* which is the starting uid/gid in the container and *size* refers to the number of ids to be mapped.
-There is a limit of 5 mappings which is the Linux kernel hard limit.
-
-## Default Devices and File Systems
```
mrunalp: This section could remain here as it is talking about the mounts and not the runtime details.

config-linux.md
```diff
@@ -15,49 +15,3 @@ Valid values are the strings for capabilities defined in [the man page](http://m
         "CAP_NET_BIND_SERVICE"
     ]
 ```
-
-## User namespace mappings
```
mrunalp: This change sounds good.

mrunalp:
@gao-feng Could you update and rebase the PR?

config-linux.md
wking: On Mon, Oct 05, 2015 at 01:01:17PM -0700, Mrunal Patel wrote:
> > -## Default Devices and File Systems
> 
> This section could remain here as it is talking about the mounts and
> not the runtime details.

Currently devices are a purely runtime.json issue (although see [1],
where I suggest moving the mknod bits back into config.json).  So I'm
ok keeping the default mounts here, but the default devices seem like
they should be moved into the runtime.json docs.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/y_Fsa2_jJaM

gao-feng:
@mrunalp updated, please check.

vishh:
LGTM

vbatts:
rebase and LGTM

gao-feng:
@wking @mrunalp @vishh @vbatts rebased, please check :)

hqhq:
LGTM

PR #173, opened 2015-09-10
version: more explicit version for comparison

vbatts:
For deterministic comparison and still providing a Version string for
representation and reuse.

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,4 +1,15 @@
 package specs
 
+import "fmt"
+
+const (
+	// VersionMajor is for an API incompatible changes
+	VersionMajor = 0
+	// VersionMinor is for functionality in a backwards-compatible manner
+	VersionMinor = 1
+	// VersionPatch is for backwards-compatible bug fixes
+	VersionPatch = 0
+)
+
 // Version is the specification version that the package types support.
-const Version = "pre-draft"
+var Version = fmt.Sprintf("%d.%d.%d", VersionMajor, VersionMinor, VersionPatch)
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 This project is where the [Open Container Initiative](http://www.opencontainers.org/) Specifications are written.
 This is a work in progress.
-We should have a first draft by end of July 2015.
 
 Table of Contents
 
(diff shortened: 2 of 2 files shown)


version.go
```diff
@@ -1,4 +1,15 @@
 package specs
 
+import "fmt"
+
+const (
+	// VersionMajor is for an API incompatible changes
+	VersionMajor = 0
+	// VersionMinor is for functionality in a backwards-compatible manner
```
wking: Maybe “VersionMinor is for backwards-compatible API changes”?  And then mirror that for `VersionMajor` with “VersionMajor is for backwards-incompatible API changes”.

version_test.go
```diff
@@ -0,0 +1,14 @@
+package specs
+
+import (
+	"regexp"
+	"testing"
+)
+
+func TestVersion(t *testing.T) {
+	re := regexp.MustCompile(`\d+\.\d+\.\d+`)
```
wking: SemVer 2.0 also allows for pre-release versions and build metadata.  See [sections 9 and 10][1].  Surely someone else has already written a SemVer validator in Go?

version.go
vbatts: this is leaning on verbiage from SemVer

version_test.go
vbatts: This _test.go can be dropped entirely. It's just a smoke test and not intended to be a SemVer complete assertion.

version.go
wking: On Thu, Sep 10, 2015 at 12:32:25PM -0700, Vincent Batts wrote:
> > +	// VersionMinor is for functionality in a backwards-compatible manner
> 
> this is leaning on verbiage from SemVer

Ah, consistency with semver.org is probably better than consistency
between the entries in this case.  So suggestion retracted ;).

version_test.go
wking: On Thu, Sep 10, 2015 at 12:33:20PM -0700, Vincent Batts wrote:
> > +	re := regexp.MustCompile(`\d+\.\d+\.\d+`)
> 
> This _test.go can be dropped entirely…

That's what I'd recommend.

philips: @vbatts in the future we can do this: https://github.com/coreos/go-semver

philips:
I don't know if this helps anything at this point. We should just make a note in the README that says:

"During the 0.* series of OCI releases we make no backwards compatibility guarantees and intend to break the schema during this series."

vbatts:
On Sep 10, 2015 3:48 PM, "Brandon Philips" <notifications@github.com> wrote:
>
> I don't know if this helps anything at this point. We should just make a
note in the README that says:
>
> "During the 0.* series of OCI releases we make no backwards compatibility
guarantees and intend to break the schema during this series."
>

Totally agree and was just fielding that response on irc

version_test.go
vbatts: @philips right on.

vbatts:
updated

removed the test and added release verbiage

wking:
On Thu, Sep 10, 2015 at 01:18:47PM -0700, Vincent Batts wrote:
> updated

Through d335ecd looks good to me.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #174, opened 2015-09-10
.tools: Fix redundant r.Pass check when 'fail == 0'

wking:
The previous version of this script would never print the success
messages, because when `fail == 0` was true, none of the messages
would satisfy `!r.Pass`.  This commit shifts the logic around to handle
the following cases:

a. If the verbose flag is set, print messages for all checks, using
   [TAP][1]'s "ok" and "not ok" to distinguish between per-check pass/fail.
b. If the verbose flag is not set, only print the failed messages (but
   still keep the "not ok" prefix).

[Spun off][2] from #170.

[1]: https://en.wikipedia.org/wiki/Test_Anything_Protocol
[2]: https://github.com/opencontainers/specs/pull/170/files#r39122874

### DIFF
--- a/.tools/validate.go
+++ b/.tools/validate.go
@@ -74,20 +74,19 @@ func main() {
 		results = append(results, vr...)
 		if _, fail := vr.PassFail(); fail == 0 {
 			fmt.Println("PASS")
-			if *flVerbose {
-				for _, r := range vr {
-					if r.Pass {
-						fmt.Printf("  - %s\n", r.Msg)
-					}
-				}
-			}
 		} else {
 			fmt.Println("FAIL")
-			// default, only print out failed validations
-			for _, r := range vr {
-				if !r.Pass {
-					fmt.Printf("  - %s\n", r.Msg)
+		}
+		for _, r := range vr {
+			if *flVerbose {
+				if r.Pass {
+					fmt.Printf("ok")
+				} else {
+					fmt.Printf("not ok")
 				}
+				fmt.Printf(" %s\n", r.Msg)
+			} else if !r.Pass {
+				fmt.Printf("not ok %s\n", r.Msg)
 			}
 		}
 	}
--- a/.tools/validate.go
+++ b/.tools/validate.go
@@ -68,8 +68,13 @@ func main() {
 	}
 
 	results := ValidateResults{}
-	for _, commit := range c {
-		fmt.Printf(" * %s %s ... ", commit["abbreviated_commit"], commit["subject"])
+
+	if *flVerbose {
+		fmt.Println("TAP version 13")
+		fmt.Printf("1..%d\n", len(c)*len(DefaultRules))
+	}
+	for i, commit := range c {
+		fmt.Printf("# %s %s ... ", commit["abbreviated_commit"], commit["subject"])
 		vr := ValidateCommit(commit, DefaultRules)
 		results = append(results, vr...)
 		if _, fail := vr.PassFail(); fail == 0 {
@@ -77,16 +82,16 @@ func main() {
 		} else {
 			fmt.Println("FAIL")
 		}
-		for _, r := range vr {
+		for j, r := range vr {
 			if *flVerbose {
 				if r.Pass {
 					fmt.Printf("ok")
 				} else {
 					fmt.Printf("not ok")
 				}
-				fmt.Printf(" %s\n", r.Msg)
+				fmt.Printf(" %d - %s\n", i*len(DefaultRules)+j+1, r.Msg)
 			} else if !r.Pass {
-				fmt.Printf("not ok %s\n", r.Msg)
+				fmt.Printf("not ok - %s\n", r.Msg)
 			}
 		}
 	}


vbatts:
i'm not seeing the brokenness?
```bash
vbatts@valse ~/src/opencontainers/specs (fix-dco-validate-on-merges) $ gr .tools/validate.go -range 92a7451...5a654b9
 * 5a654b9 .tools: cleanup the commit entry ... PASS
 * 041eb73 travis: fix DCO validation for merges ... PASS
 * 3f62423 Merge remote-tracking branch 'origin/pr/159' ... PASS
 * 712a746 Merge remote-tracking branch 'origin/pr/163' ... PASS
 * b1510a7 Merge pull request #172 from laijs/committer ... PASS
 * d3a6069 Change the rlimit type to string instead of int ... PASS
 * 339e038 Deduplicate the field of RootfsPropagation ... PASS
vbatts@valse ~/src/opencontainers/specs (fix-dco-validate-on-merges) $ gr .tools/validate.go -v -range 92a7451...5a654b9
 * 5a654b9 .tools: cleanup the commit entry ... PASS
  - has a valid DCO
 * 041eb73 travis: fix DCO validation for merges ... PASS
  - has a valid DCO
 * 3f62423 Merge remote-tracking branch 'origin/pr/159' ... PASS
  - merge commits do not require DCO
 * 712a746 Merge remote-tracking branch 'origin/pr/163' ... PASS
  - merge commits do not require DCO
 * b1510a7 Merge pull request #172 from laijs/committer ... PASS
  - merge commits do not require DCO
 * d3a6069 Change the rlimit type to string instead of int ... PASS
  - has a valid DCO
 * 339e038 Deduplicate the field of RootfsPropagation ... PASS
  - has a valid DCO
```

wking:
On Thu, Sep 10, 2015 at 12:02:39PM -0700, Vincent Batts wrote:
> i'm not seeing the brokenness?

Your post-#170 phrasing isn't broken, it's just redundant.  Details in
my commit message (5282b96, rebased onto #170), but the redundancy is
(eliding some nesting):

  if _, fail := vr.PassFail(); fail == 0 {
    …
    if r.Pass {
      fmt.Printf("  - %s\n", r.Msg)
    }
  }

How would the internal if check ever *not* fire?  That's why my
initial suggestion was to just drop the internal if, and use [1]:

  if _, fail := vr.PassFail(); fail == 0 {
    …
    fmt.Printf("  - %s\n", r.Msg)
  }

This PR does more than just drop the check, but that's what I was
pushing back on initially.

[1]: https://github.com/opencontainers/specs/pull/170/files#r39122874

vbatts:
not. because that `if r.Pass {` is _inside_ a loop for that validate _set_, not the single rule that was tested. A validation set may have failures _and_ passes.

wking:
On Fri, Sep 11, 2015 at 07:08:31AM -0700, Vincent Batts wrote:
> not. because that `if r.Pass {` is _inside_ a loop for that validate
> _set_, not the single rule that was tested. A validation set may
> have failures _and_ passes.

In general, a set may have failures and passes, but your entire loop
is inside the `if _, fail := vr.PassFail(); fail == 0` check, which
means that none of the checks in that set failed.

vbatts:
i see. I'm wrong here. If pass count is 0 then there is no need for `if
r.Pass`.

Let's remove this beginings of TAP output. It makes the layout not logical
and is not even proper TAP

wking:
On Fri, Sep 11, 2015 at 08:28:29AM -0700, Vincent Batts wrote:
> Let's remove this beginings of TAP output. It makes the layout not
> logical and is not even proper TAP

Do you have a different preference for distinguishing passing checks
from fails (maybe + vs - where you always use -)?  I'm fine just
converting the whole thing to valid TAP, but I'm also happy to punt
that to a follow-up PR.

wking:
On Fri, Sep 11, 2015 at 08:38:14AM -0700, W. Trevor King wrote:
> Fri, Sep 11, 2015 at 08:28:29AM -0700, Vincent Batts:
> > Let's remove this beginings of TAP output. It makes the layout not
> > logical and is not even proper TAP
> 
> Do you have a different preference for distinguishing passing checks
> from fails (maybe + vs - where you always use -)?  I'm fine just
> converting the whole thing to valid TAP, but I'm also happy to punt
> that to a follow-up PR.

In the absence of feedback here, I've added a follow-up commit
(2d70cd9) that makes the verbose output valid TAP and avoids the
de-indenting for non-verbose output.  I'm still happy to pull that
commit off into a follow-up PR and reroll my initial commit if you
tell me how you'd rather distingish betwee passing checks and fails.

wking:
On Fri, Sep 11, 2015 at 09:10:13PM -0700, W. Trevor King wrote:
… I've added a follow-up commit (2d70cd9)…

Looks like that commit was missing a DCO (yay, CI :).  Added with
2d70cd9 → 6beba35.

wking:
Fixed 'go fmt' with 6beba35 → 650828c.

crosbymichael:
This was moved to a separate project out of this repo.

PR #176, opened 2015-09-10
Docs overhaul

vbatts:
I'm sure this will cause a lot to need rebasing, but the layout was killing me.

I'm ready to move the source into a `./ref/go/` directory and start framing up protobuf structs or json-schema for a more canonical use as well, but that can wait for another change.

cc @mrunalp @philips @crosbymichael @LK4D4 @tianon @vishh

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 
 Table of Contents
 
+- [Container Principles](principles.md)
 - [Filesystem Bundle](bundle.md)
 - [Container Configuration](config.md)
   - [Linux Specific Configuration](config-linux.md)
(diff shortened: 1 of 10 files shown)


vbatts:
[View it](https://github.com/vbatts/specs/blob/docs-overhaul/README.md)

vbatts:
so much more is needed, this is just a first pass.
Ideally, the *.md should be all someone needs, and the source is just an example.

wking:
These seem like useful changes, but they're going to be a lot of
conflicts with #126.  What do I need to do to get some eyeballs on
that?

jonboulle:
:+1:

vbatts:
@wking hrm. there is overlap, but you've got some file renaming too that is not idiomatic.

wking:
On Thu, Sep 10, 2015 at 12:31:12PM -0700, Vincent Batts wrote:
> @wking hrm. there is overlap, but you've got some file renaming too
> that is not idiomatic.

I tried my best to be idiomatic, but I'm happy to make any adjustments
that folks ask for.

vishh:
LGTM

crosbymichael:
LGTM

@vbatts can you rebase plz?

vbatts:
rebased. PTAL.

README.md
```diff
@@ -1,17 +1,21 @@
 # Open Container Specifications
 
-This project is where the [Open Container Initiative](http://www.opencontainers.org/) Specifications are written.
-This is a work in progress.
+[Open Container Initiative](http://www.opencontainers.org/) Specifications for standards on Operating System process and application containers.
```
wking: What are “Operating System process … containers”?  If I'm parsing that right (“for standards on ((Operating System process) and application) containers”?).

bundle.md
```diff
@@ -14,7 +16,7 @@ A standard container bundle is made of the following 3 parts:
 - One or more content directories
 - A configuration file
 
-# Directory layout
+## Directory layout
```
wking: This seems like a subsection of the previous section (e.g. the previous section has the list of the three allowed parts).  So maybe remove the `## Container Format` header?

README.md
```diff
@@ -1,17 +1,21 @@
...
+
 
 Table of Contents
 
+- [Container Principles](principles.md)
 - [Filesystem Bundle](bundle.md)
 - [Container Configuration](config.md)
+  - [Runtime Configuration](runtime-config.md)
+  - [Linux Specific Runtime Configuration](runtime-config-linux.md)
```
wking: In #126, I went with “Linux-specific”, since Linux qualifies the specificity.

bundle.md
```diff
@@ -22,7 +24,10 @@ The `config.json` file contains settings that are host independent and applicati
 The `runtime.json` file contains settings that are host specific such as memory limits, local device access and mount points.
 The goal is that the bundle can be moved as a unit to another machine and run the same application if `runtime.json` is removed or reconfigured.
 
-The syntax and semantics for `config.json` are described in [this specification](config.md).
+Configuration file syntax and semantics:
+
+* [`config.json`](config.md) (immutable, host independent configuration)
+* [`runtime.json`](runtime-config.md) (mutable, host dependent configuration)
```
wking: In #126, I went with “host-independent”, since host qualifies the independence.

runtime-config-linux.md
```diff
@@ -1,3 +1,5 @@
+# Linux Runtime Configuration
```
wking: We should echo whatever we use in the TOC here.  With your current 159a3eac9fc, that's “Linux Specific Runtime Configuration” (although see [my suggested alteration][1]).

[1]: https://github.com/opencontainers/specs/pull/176/files#r39865436

runtime-config.md
```diff
@@ -1,3 +1,14 @@
+# Runtime Configuration
+
+## Hooks
+
+Lifecycle of a container's runtime, for setup and teardown.
```
wking: This doesn't parse for me.  In #126 I moved the hooks section from the current `runtime.md` (which is about the lifecycle) over to this section (which is about `runtime.json` and configuring hooks).

wking: … and moving over that text gets you the existing “Hooks allow one to run code…”.

runtime-config.md
```diff
@@ -1,3 +1,14 @@
...
+Presently there are `Prestart` and `Poststop`.
+
+* `Prestart` is a list of hooks to be run before the container process is executed
+* `Poststop` is a list of hooks to be run after the container process exits
```
wking: The hooks aren't capitalized in the spec JSON, so I'd prefer `prestart` and `poststop`.  We're also not particularly consistent in how we markup JSON keys in the spec, but I think the recent trend has been to use **prestart** instead of `prestart`.

runtime-config.md
```diff
@@ -1,3 +1,14 @@
...
+	
+
```
wking: You removed a doubled blank line from `code-of-conduct.md` in this branch, so probably don't add one here ;).

wking:
On Fri, Sep 18, 2015 at 07:40:45AM -0700, Vincent Batts wrote:
> rebased. PTAL.

I left a few copy-edit suggestions with the branch at 159a3ea.  The
only involved change would be mirroring something like my 4cfa5841
(Move hooks from lifecycle.md to runtime.md, 2015-08-28, #126) to
avoid having duplicate hook documentation in both your runtime.md and
runtime-config.md.

vbatts:
I guess the hook duplication needs cleanup too. Was hoping this would mostly be reorg.

vbatts:
alrighty. I deduped the hook's docs.

bundle.md
vbatts: this is a subsection of the bundle

README.md
vbatts: "open container" while it makes sense to us looking at projects like this and docker, are aware that these containers run on an operating system (https://en.wikipedia.org/wiki/Operating_system).
Those that are searching for "container standards" may get https://en.wikipedia.org/wiki/Intermodal_container or https://en.wikipedia.org/wiki/Containerization.
Further "open container" also has overlap with "Open Container and Consumption Statutes". Neither of these things has to do with operating systems.

tianon:
LGTM

crosbymichael:
LGTM

README.md
wking: On Fri, Sep 25, 2015 at 08:15:03AM -0700, Vincent Batts wrote:
> > -This project is where the [Open Container Initiative](http://www.opencontainers.org/) Specifications are written.
> > -This is a work in progress.
> > +[Open Container Initiative](http://www.opencontainers.org/) Specifications for standards on Operating System process and application containers.
> 
> … Neither of these things has to do with operating systems.

How about:

> [Open Container Initiative](http://www.opencontainers.org/) Specifications for application containers.
> A framework for defining containers on several operating systems and containerization technologies.

To cover both Linux / Windows / … and Linux-containers /
Windows-containers / hypervisor-virtual-hosts/….

wking: On Fri, Sep 25, 2015 at 12:46:43PM -0700, W. Trevor King wrote:
> How about…

Ah, I see I'm post-merge ;).  I'll file a follow-up PR.

vbatts: Those are all operating systems...

PR #179, opened 2015-09-11
proto: initial protobuf for config

vbatts:
To have a less source based, and more consumable example of structures,
this is an initial pass at protobuf structures for the structures in config.go

There is some exercise needed in platform specific structures.

For the sake of example, the Makefile defaults to outputing golang
source, but has a 'c' target (`make c`) for C source as well.

### DIFF
--- a/proto/Makefile
+++ b/proto/Makefile
@@ -4,19 +4,28 @@ PROTO_FILES := $(wildcard *.proto)
 GO_FILES    := $(patsubst %.proto,%.pb.go,$(PROTO_FILES))
 C_FILES     := $(patsubst %.proto,%.pb-c.c,$(PROTO_FILES))
 C_HDR_FILES := $(patsubst %.proto,%.pb-c.h,$(PROTO_FILES))
+PY_FILES    := $(patsubst %.proto,%_pb2.py,$(PROTO_FILES))
 
 default: go
+
+all: go py c
 
 go: $(GO_FILES)
 
 %.pb.go: %.proto
-	protoc --go_out=$(DESTDIR) $<
+	protoc --go_out=$(DESTDIR) $^
 
 c: $(C_FILES)
 
 %.pb-c.c: %.proto
-	protoc-c --c_out=$(DESTDIR) $<
+	protoc-c --c_out=$(DESTDIR) $^
+
+py: $(PY_FILES)
+
+%_pb2.py: %.proto
+	protoc --python_out=$(DESTDIR) $^
+
 
 clean:
-	rm -rf *~ $(GO_FILES) $(C_FILES) $(C_HDR_FILES)
+	rm -rf *~ $(GO_FILES) $(C_FILES) $(C_HDR_FILES) $(PY_FILES)
 
--- a/proto/Makefile
+++ b/proto/Makefile
@@ -1,0 +1,22 @@
+
+DESTDIR     ?= .
+PROTO_FILES := $(wildcard *.proto)
+GO_FILES    := $(patsubst %.proto,%.pb.go,$(PROTO_FILES))
+C_FILES     := $(patsubst %.proto,%.pb-c.c,$(PROTO_FILES))
+C_HDR_FILES := $(patsubst %.proto,%.pb-c.h,$(PROTO_FILES))
+
+default: go
+
+go: $(GO_FILES)
+
+%.pb.go: %.proto
+	protoc --go_out=$(DESTDIR) $<
+
+c: $(C_FILES)
+
+%.pb-c.c: %.proto
+	protoc-c --c_out=$(DESTDIR) $<
+
+clean:
+	rm -rf *~ $(GO_FILES) $(C_FILES) $(C_HDR_FILES)
+
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -7,7 +7,7 @@ message Spec {
 	// Version is the version of the specification that is supported.
 	optional string version = 1;
 	// Platform is the host information for OS and Arch.
-	optional Platform platform = 2; // [default=77];
+	optional Platform platform = 2;
 	// Process is the container's main process.
 	optional Process process = 3;
 	// Root is the root information for the container's filesystem.
@@ -69,7 +69,7 @@ enum PlatformType {
 message User {
 	// Type so that receivers of this message can `switch` for the fields
 	// expected
-	optional PlatformType type = 1 [default = LINUX];
+	optional PlatformType type = 1;
 
 	optional LinuxUser linux_type = 2;
 }
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -72,17 +72,8 @@ message User {
 	// expected
 	optional PlatformType type = 1;
 
-	optional LinuxUser linux_type = 2;
-}
-
-// LinuxUser specifies linux specific user and group information for the
-// container's main process.
-message LinuxUser {
-	// Uid is the user id.
-	optional int32 uid = 1;
-	// Gid is the group id.
-	optional int32 gid = 2;
-	repeated int32 additional_gids = 3;
+	//optional LinuxUser linux_type = 2;
+	extensions 100 to 499;
 }
 
 // Root contains information about the container's root filesystem on the host.
--- a/proto/config.proto
+++ b/proto/config.proto
@@ -1,95 +1,105 @@
 //package oci.config.bundle;
 package oci;
 
-// Spec is the base configuration for the container.  It specifies platform
+// Spec is the base configuration for the container. It specifies platform
 // independent configuration.
 message Spec {
 	// Version is the version of the specification that is supported.
-  optional string Version = 1;
+	optional string version = 1;
 	// Platform is the host information for OS and Arch.
-  optional Platform Platform = 2; // [default=77];
+	optional Platform platform = 2; // [default=77];
 	// Process is the container's main process.
-  optional Process Process = 3;
+	optional Process process = 3;
 	// Root is the root information for the container's filesystem.
-  optional Root Root = 4;
+	optional Root root = 4;
 	// Hostname is the container's host name.
-  optional string Hostname = 5;
-	// Mounts profile configuration for adding mounts to the container's filesystem.
-  repeated MountPoint Mounts = 6;
+	optional string hostname = 5;
+	// Mounts profile configuration for adding mounts to the container's
+	// filesystem.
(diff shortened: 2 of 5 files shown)


vbatts:
cc @philips @LK4D4 @vishh @crosbymichael

mrunalp:
:+1: We should probably create a branch to flesh this out and see if it meets all our requirements.

vbatts:
fair. I'm still working through it, but figured to get the conversation
started.

proto/config.proto
```diff
@@ -0,0 +1,59 @@
+package oci;
+	
+// Spec is the base configuration for the container.  It specifies platform
+// independent configuration.
+message Spec {
+	// Version is the version of the specification that is supported.
+  required string Version = 1;
```
vishh: One thing to consider with `required` fields is that it might be almost impossible to make it optional.

vbatts: are you saying to start with preferring `optional`?

vishh: Yes.

crosbymichael:
Ya, we could do a branch on the repo to figure this out if that makes it easier.

vbatts:
@vishh updated the field rules

vbatts:
So, I've stubbed out the [platform dependent] `User`, in the best pattern I can derive from the guidelines and references. Thoughts?

proto/config.proto
```diff
@@ -0,0 +1,82 @@
...
+enum PlatformType {
+  LINUX = 1;
+}
+
+// User specifies user information for the process.
+message User {
+  // Type so that receivers of this message can `switch` for the fields expected
+  optional PlatformType Type = 1 [default = LINUX];
+  // LinuxUserFields are to be used when Type is LINUX
+  optional linuxUserFields LinuxUserFields = 2;
```
wking: Is this going to give us per-platform namespacing inside User?  E.g.

```json
"user": {
  "linux": {
    "uid": 0,
    "gid": 0,
   }
}
```

That's the pattern we use in other sections of the existing config (e.g. for `linux.capabilities`), but it seems excessive here.  However, the benefits of a protobuf-based schema may warrant a few warts in the associated JSON structure.

vbatts: not namespacing, just platform dependent user attributes.

vbatts:
Alrighty, y'all. A couple of concessions, but is pretty much lined up for an initial review.

proto/Makefile
```diff
@@ -0,0 +1,31 @@
+
+DESTDIR     ?= .
+PROTO_FILES := $(wildcard *.proto)
+GO_FILES    := $(patsubst %.proto,%.pb.go,$(PROTO_FILES))
+C_FILES     := $(patsubst %.proto,%.pb-c.c,$(PROTO_FILES))
+C_HDR_FILES := $(patsubst %.proto,%.pb-c.h,$(PROTO_FILES))
+PY_FILES    := $(patsubst %.proto,%_pb2.py,$(PROTO_FILES))
+
+defualt: go
```
wking: ‘defualt’ → ‘default’ ;)

proto/config.proto
wking: On Fri, Sep 11, 2015 at 12:58:22PM -0700, Vincent Batts wrote:
> > +  // LinuxUserFields are to be used when Type is LINUX
> > +  optional linuxUserFields LinuxUserFields = 2;
> 
> not namespacing, just platform dependent user attributes.

Generated Go (from b5433ce3) has:

  // User specifies user information for the process.
  type User struct {
    // Type so that receivers of this message can `switch` for the fields expected
    Type *PlatformType `protobuf:"varint,1,opt,enum=oci.PlatformType,def=1" json:"Type,omitempty"`
    // LinuxUserFields are to be used when Type is LINUX
    LinuxUserFields  *LinuxUserFields `protobuf:"bytes,2,opt" json:"LinuxUserFields,omitempty"`
    XXX_unrecognized []byte           `json:"-"`
  }

Which looks like it's going to lead to a an explicit “LinuxUserFields”
level to me.  And again, I'm not against a bit of wordiness in support
of spec that can autogenerate bindings in a number of languages, but
that explicit platform level is not how we handle user entries now.

proto/config.proto
```diff
@@ -0,0 +1,95 @@
+//package oci.config.bundle;
+package oci;
+
+// Spec is the base configuration for the container.  It specifies platform
+// independent configuration.
+message Spec {
+	// Version is the version of the specification that is supported.
+  optional string Version = 1;
+	// Platform is the host information for OS and Arch.
+  optional Platform Platform = 2; // [default=77];
```
vishh: If any of these fields are required to be set, can we explicitly state that in the comment?

proto/config.proto
```diff
@@ -0,0 +1,95 @@
...
+	// Args specifies the binary and arguments for the application to execute.
+	repeated string Args = 3;
+	// Env populates the process environment for the process.
+	repeated string Env = 4;
+	// Cwd is the current working directory for the process and must be
+	// relative to the container's root.
+	optional string Cwd = 5;
+}
+
+enum PlatformType {
```
vishh: nit: This can be defined as part of the `User` message

proto/config.proto
```diff
@@ -0,0 +1,95 @@
...
+  LINUX = 1;
```
vishh: Enums that will ever be expanded must include a default `Unknown` value as the first value in the declaration. 
When new values are added to the enum, old clients will see the field as unset and the getter will return the default value or the first-declared value if no default is set.

proto/config.proto
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// User specifies user information for the process.
+message User {
+  // Type so that receivers of this message can `switch` for the fields expected
+  optional PlatformType Type = 1 [default = LINUX];
```
vishh: I'd recommend not setting the default to `Linux`. Let the user specify it.

proto/config.proto
vishh: [Any](https://developers.google.com/protocol-buffers/docs/proto3?hl=en#any) primitive might be useful here.

proto/runtime_config_linux.proto
```diff
@@ -0,0 +1,213 @@
...
+// LinuxStateDirectory holds the container's state information
+message DefaultState {
+  // TODO(vbatts) not as elegant in some ways, but there is not a concept of const here
+  optional string Directory = 1 [default =  "/run/opencontainer/containers"];
+}
+
+// LinuxRuntime hosts the Linux-only runtime information
+message LinuxRuntime {
+	// UIDMapping specifies user mappings for supporting user namespaces on linux.
+  repeated IDMapping UIDMapping = 1;
```
vishh: nit: Let's follow the [style guide](https://developers.google.com/protocol-buffers/docs/style)

proto/runtime_config_linux.proto
```diff
@@ -0,0 +1,213 @@
...
+// Syscall is used to match a syscall in Seccomp
+message Syscall {
+	optional string Name = 1;
+	optional string Action = 2;
+	repeated Arg Args = 3;
+}
+
+// Arg used for matching specific syscall arguments in Seccomp
+message Arg {
+	optional uint32 Index = 1;
```
vishh: nit: Add descriptions for these fields.

proto/config.proto
vbatts: @vishh ah that sounds like what I'm looking for, but that's in proto3. I'm dealing with proto2 :-\

wking: On Mon, Sep 14, 2015 at 07:41:37AM -0700, Vincent Batts wrote:
> @vishh ah that sounds like what I'm looking for, but that's in
> proto3. I'm dealing with proto2 :-\

If v3 is out, why write our schema in v2?

vishh: We can use extensions in proto2.

proto/runtime_config_linux.proto
vbatts: i've got a commit staged, and it will change up a lot of the go style.
Like `GID` -> `gid` -> `Gid`

proto/config.proto
vbatts: fair

proto/config.proto
vbatts: k

proto/config.proto
vbatts: I started to do that, but left it out as there may be uses in other messages as we go along.

proto/config.proto
vbatts: Yeah. That relationship is a bit cloudied by the `required`, `optional`, `repeated` field rules.

proto/config.proto
vbatts: I'm playing with permutations of extensions right now

vbatts: @wking general adoption

proto/runtime_config_linux.proto
vbatts: This was a copy from the *.go source. Not entirely clear on this struct. Must be a Seccomp specific syscall mapping. @mrunalp @crosbymichael thoughts?

vbatts:
updated for comments.

Also, I'll have to close this PR and open another to put these changes against a branch besides master. Not sure where/what we want that branch to be. `dev`? `protobuf`?

mrunalp:
@vbatts Maybe just call it protobuf.

vbatts:
@vishh so, i just pushed an iteration of `User` as an extension. The godoc is [here](https://gist.github.com/vbatts/6f4893578c4b139d7fb6). LMKWYT

vbatts:
moving this PR over to https://github.com/opencontainers/specs/pull/185

proto/config.proto
wking: On Mon, Sep 14, 2015 at 10:06:25AM -0700, Vincent Batts wrote:
> @wking general adoption

Does protobuf-version adoption really matter?  If we're just using it
for code-generation, it seems like you'd only need a few folks running
the protobuf-to-$X translators.  Everyone else can just consume that
generated code, without interacting with protobuf at all.

vbatts: I don't intend on commiting the generated code. So yes, it would depend on
the version of protobuf compiler available to the user

wking: On Mon, Sep 14, 2015 at 12:04:48PM -0700, Vincent Batts wrote:
> I don't intend on commiting the generated code.

No need to commit it, just ship it with release tarballs.

PR #182, opened 2015-09-12
.tools: Add per-commit Go checks (vet, fmt, and lint)

wking:
Instead of just checking the tip, run the checks on each commit in the
series.  This makes sure we catch errors with earlier commits even if
they were fixed by subsequent commits.

There are some details about low-level choices in the commit messages
themselves, but nothing that seems generic enough to be worth copying
out to the PR level.

This PR is based on #174, because as the output gets more verbose
(e.g. YAML blocks for golint's stdout) you'll likely want to feed it
into a TAP harness for compaction.  The non-verbose output is still
pretty short, although there are now more tests which you can fail ;).

If the protobuf branch lands (#179), we can swap out the Go checkers
and swap in [protoc-gen-lint][1].  I'm not aware of any protobuf
equivalent to 'go fmt' though, which is unfortunate.

[1]: https://github.com/ckaznocha/protoc-gen-lint

### DIFF
--- a/.tools/validate.go
+++ b/.tools/validate.go
@@ -74,20 +74,19 @@ func main() {
 		results = append(results, vr...)
 		if _, fail := vr.PassFail(); fail == 0 {
 			fmt.Println("PASS")
-			if *flVerbose {
-				for _, r := range vr {
-					if r.Pass {
-						fmt.Printf("  - %s\n", r.Msg)
-					}
-				}
-			}
 		} else {
 			fmt.Println("FAIL")
-			// default, only print out failed validations
-			for _, r := range vr {
-				if !r.Pass {
-					fmt.Printf("  - %s\n", r.Msg)
+		}
+		for _, r := range vr {
+			if *flVerbose {
+				if r.Pass {
+					fmt.Printf("ok")
+				} else {
+					fmt.Printf("not ok")
 				}
+				fmt.Printf(" %s\n", r.Msg)
+			} else if !r.Pass {
+				fmt.Printf("not ok %s\n", r.Msg)
 			}
 		}
 	}
--- a/.tools/validate.go
+++ b/.tools/validate.go
@@ -44,6 +44,22 @@ var DefaultRules = []ValidateRule{
 	},
 	// TODO add something for the cleanliness of the c.Subject
 	func(c CommitEntry) (vr ValidateResult) {
+		vr.CommitEntry = c
+		buf := bytes.NewBuffer([]byte{})
+		args := []string{"git", "show", "--check", c["commit"]}
+		vr.Msg = strings.Join(args, " ")
+		cmd := exec.Command(args[0], args[1:]...)
+		cmd.Stdout = buf
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			vr.Pass = false
+			vr.Detail = string(buf.Bytes())
+			return vr
+		}
+		vr.Pass = true
+		return vr
+	},
+	func(c CommitEntry) (vr ValidateResult) {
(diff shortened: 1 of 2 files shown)


wking:
I added a `git show --check …` too and some `TRAVIS_COMMIT_RANGE` debugging.  It turns out that the extra commits are because `TRAVIS_COMMIT_RANGE` is `{target-tip}...{pr-tip}` when it should be `{target-tip}..{pr-tip}`.  There shouldn't be a need to test commits that are only reachable from `{target-tip}`.  I'll see if I can find a Travis bug for this…

wking:
> `TRAVIS_COMMIT_RANGE` is `{target-tip}...{pr-tip}` when it should be `{target-tip}..{pr-tip}`.

This is travis-ci/travis-ci#4596.  Until that gets fixed, I've pushed a workaround with 8831b20.

vbatts:
honestly. Let's keep the TAP out for now.

wking:
On Mon, Oct 05, 2015 at 12:16:37PM -0700, Vincent Batts wrote:
> honestly. Let's keep the TAP out for now. 

Ok (although I don't see a need to invent a different output syntax).
What syntax would you like for pass/fail (where TAP uses "ok" and "not
ok")?

.tools/validate.go
```diff
@@ -39,6 +43,35 @@ var DefaultRules = []ValidateRule{
...
+		return vr
+	},
+	func(c CommitEntry) (vr ValidateResult) {
+		return ExecTree(c, "go", "vet", "./...")
+	},
+	func(c CommitEntry) (vr ValidateResult) {
+		return ExecTree(c, "go", "fmt", "./...")
+	},
+	func(c CommitEntry) (vr ValidateResult) {
+		vr = ExecTree(c, os.ExpandEnv("$HOME/gopath/bin/golint"), "./...")
```
vbatts: but this is what we're doing in travis, and seems cleaner there.

vbatts:
Honestly, that is not something I see a big need for a change and would be
fine with `$?` being not zero for failure, but it already prints out a
pass/fail as well.

.tools/validate.go
wking: On Mon, Oct 05, 2015 at 12:20:58PM -0700, Vincent Batts wrote:
> > +		vr = ExecTree(c, os.ExpandEnv("$HOME/gopath/bin/golint"), "./...")
> 
> but this is what we're doing in travis, and seems cleaner there.

You're not doing it for every commit in Travis, so you'd miss cases
where one commit breaks the format and a subsequent commit fixed it.

vbatts: fair, but not completely a concern right now.

wking:
On Mon, Oct 05, 2015 at 12:26:10PM -0700, Vincent Batts wrote:
> Honestly, that is not something I see a big need for a change and
> would be fine with `$?` being not zero for failure, but it already
> prints out a pass/fail as well.

It doesn't currently print pass/fail for the individual checks, it
just prints all the checks that passed [1] or all the checks that
failed [2].  There's no code currently for generating output like:

  FAIL
  ok 1 - has a valid DCO
  not ok 2 - git show --check 8831b203606aa74d2bf5274108816dde7629fa6f
  …

where a single commit has both passing and failing checks, and we
still see the passing checks.

[1]: https://github.com/opencontainers/specs/blob/f75f23fe4e9ad3cf1324530ee8876b1b3d2c8237/.tools/validate.go#L80
[2]: https://github.com/opencontainers/specs/blob/f75f23fe4e9ad3cf1324530ee8876b1b3d2c8237/.tools/validate.go#L89

crosbymichael:
This was moved to a separate project out of this repo.

Issue #108, opened 2015-08-11
Labels and extension meta data in containers


klizhentas:
I would like to discuss ability to attach application-specific meta data to container images. Here are a couple of examples that come to mind:

* bookeeping of versions of libraries included in the image that can be later inspected by security audit tools to update containers with obsolete/vulnerable libraries
* adding application-specific information that can be used by implementation to execute custom actions, e.g. host restrictions (this container can only run on a host with > 60 GB or RAM)
* adding custom signatures to the app-specific content in the image

For example, app container spec adresses this use-case by introducing labels:

> labels (list of objects, optional) used during image discovery and dependency resolution. The listed objects must have two key-value pairs: name is restricted to the AC Identifier formatting and value is an arbitrary string. Label names must be unique within the list, and (to avoid confusion with the image's name) cannot be "name". Several well-known labels are defined:
    version when combined with "name", this SHOULD be unique for every build of an app (on a given "os"/"arch" combination).
    os, arch can together be considered to describe the syscall ABI this image requires. arch is meaningful only if os is provided. If one or both values are not provided, the image is assumed to be OS- and/or architecture-independent. Currently supported combinations are listed in the types.ValidOSArch variable, which can be updated by an implementation that supports other combinations. The combinations whitelisted by default are (in format os/arch): linux/amd64, linux/i386, freebsd/amd64, freebsd/i386, freebsd/arm, darwin/x86_64, darwin/i386. See the Operating System spec for the environment apps can expect to run in given a known os label.

After trying to use labels in practice, I can say that ACI's labels have the following limitations:

* Value of the label is limited to string, what makes it hard to include structured data in the value
* Label names are unique, but not enforced in the data structure that is presented as a list.

The useful part about labels is namespaces, that helps to identify the purpose of meta data and avoid collisions.

It would be very helpful to include the mechanism of adding meta data to the open container images that addresses some of the limitations listed above.

klizhentas:
Some additional discussion can be found here:

* https://groups.google.com/a/opencontainers.org/d/msg/dev/RF6qm-75IJ0/Jcz-CC_2DwAJ
* https://groups.google.com/a/opencontainers.org/d/msg/dev/RF6qm-75IJ0/5zZ7VM73DwAJ

wking:
I second Solomon [1] in feeling like this is a useful feature, but
separate enough from “launching a container” to be worth keeping in a
separate place.  Can't folks who want labels just put them in
label.json (or wherever they like) in the bundle?

[...]

PR #188, opened 2015-09-15
Add labels to the state

gao-feng:
`Labels` is used to store some runtime specified informations,
these informations may be meaningless to other runtime, but it's
userful for the same runtime to operate this container.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -23,6 +23,9 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
 
+* **`annotations`**: (interface) holds runtime-specified information that is not structured by this specification.
+The runtime could store experimental information here before it's standardized in this specification.
+
 *Example*
 
 ```json
--- a/state.go
+++ b/state.go
@@ -13,4 +13,6 @@ type State struct {
 	Pid int `json:"pid"`
 	// BundlePath is the path to the container's bundle directory.
 	BundlePath string `json:"bundlePath"`
+	// Annotations holds runtime-specified information that is not structured by this specification.
+	Annotations interface{} `json:"annotations"`
 }


wking:
On Mon, Sep 14, 2015 at 08:28:06PM -0700, Gao feng wrote:
> 1, clearly get the runtime of container, it's useful for statistic.

Hmm, what sort of statistics are you interested in?  I think it would
be better to instrument your ‘start’ invocation (maybe for each
runtime on your system) to get this sort of thing.  I'd like to have
different container-based runtimes be inter-operable (e.g. launch with
bocker but checkpoint with runC, etc.).  There's really nothing
runtime-specific about the created container.

> 2, different kinds of runtime will not try to operate incompatible
>    container, such as, runc should not execute(runc exec) program on
>    runv container.

The current state layout doesn't really map to runV-launched
containers / applications anyway.  For example, what is runV going to
put in the ‘pid’ field?  I mentioned this in the 2015-08-26 minutes
[1], and have fuzzy memories about the decision being “we're just
punting on that for now”.

[1]: https://github.com/opencontainers/specs/wiki/MeetingMinutes:-2015-08-26#minutes

gao-feng:
@wking 
Such as how many containers which type of runtime runs. 
How about add two fields 'runtime-type" and 'runtime-name'? the same type of runtime can operate with each other.

Runv needs to store an unix socket path in state.json, we can name this socket as <pid>.sock. but this socket is totally meaningless for container based runtimes, they should not try to get useful information through the pid stored in state.json.

wking:
On Mon, Sep 14, 2015 at 11:48:02PM -0700, Gao feng wrote:
> Such as how many containers which type of runtime runs. 

If you're interested in how many containers a given runtime *starts*,
I think your best bet is to instrument your start command [1].  But
the start command seems like a small part of the container (or
hypervisor virtual machine, or whatever) lifecycle.  Once your
container is started, who cares which runtime launched it?  You should
be able to use other runtimes (assuming a compatible state) to perform
other lifecycle events.

> Runv needs to store an unix socket path in state.json, we can name
> this socket as <pid>.sock. but this socket is totally meaningless
> for container based runtimes, they should not try to get useful
> information through the pid stored in state.json.

We should address situations where there is a functional difference
(e.g. we don't have a useful application PID) directly (e.g. by
setting a ‘socket’ field and skipping the ‘pid’ field), and not
indirectly by putting the runtime name/type in the state file and
requiring consumers to map those names/types to the functional
differences.

[1]: https://github.com/opencontainers/specs/pull/188#issuecomment-140274378

config.go
```diff
@@ -67,4 +67,6 @@ type State struct {
 	Pid int `json:"pid"`
 	// Root is the path to the container's bundle directory.
 	Root string `json:"root"`
+	// Runtime is the runtime of OCI container
+	Runtime string `json:"runtime"`
```
vishh: I am not in favor of adding explicit Runtime specific information to the Spec or State. How about using runtime `labels` or `annotations`, which are arbitrary strings, for this purpose?

vbatts: Ah yeah

laijs:
@gao-feng using `labels` instead is an excellent idea, allows the implementation to add the corresponding label.  Could you update patch?

gao-feng:
@vishh @vbatts @laijs @wking patch updated, please check

wking:
Why restrict labels to the runtime itself?  For example, see container
labels in #108.  I'd recommend we add an optional “labels” to
config.json holding a list of arbitrary bundle-author-specified JSON.
Then in the state JSON, we would have:

  "labels": {
    "runtime": …stuff specified by the runtime…,
    "bundle": …stuff from config.json's “labels”…
  }

gao-feng:
@wking 
I can not find out others which need to store informations in state.
>>    "bundle": …stuff from config.json's “labels”…
it is already in config.json's "labels", right? why we add it to state.json again? the `BundlePath` in state.json will help us to access the config.json

config.go
```diff
@@ -67,4 +67,6 @@ type State struct {
 	Pid int `json:"pid"`
 	// BundlePath is the path to the container's bundle directory.
 	BundlePath string `json:"bundlePath"`
+	// Lables is the runtime itself specified information.
```
wking: Maybe “Labels holds runtime-specified information that is not structured by this specification.”?

runtime.md
```diff
@@ -22,6 +22,8 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
+* **`Labels`**: (map) Labels is the runtime itself specified information.
+This allows runtime to store some informations which are usefull for itself but meaningless for other runtime.
```
wking: I'd keep the first line of this the same as the `config.go` comment (whatever that ends up being).  For the second line, maybe “The runtime could store experimental information here before it's standardized in this specification (e.g. the socket path used to control runV virtual machines)”.

wking:
On Fri, Oct 09, 2015 at 08:03:29PM -0700, Gao feng wrote:
> >>    "bundle": …stuff from config.json's “labels”…
>
> it is already in config.json's "labels", right? why we add it to
> state.json again? the `BundlePath` in state.json will help us to
> access the config.json

Ah, fair enough.  I guess I'm still convinced that putting the bundle
or rootfs path in the state JSON is a bad idea, but that's a different
discussion [1].

Sticking with the current spec's position that bundlePath is worth the
requirements it places on the stability and accessibility of the
bundle directory and its contents, the substance of a7eda6f looks
reasonable to me.  I'd recommend we use json.RawMessage as the type,
to keep the values as opaque as possible, and I left some minor inline
copy-edits.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Zogdyb_hrsg

gao-feng:
@wking Thanks for your suggestion, updated, please check. :)

config.go
```diff
@@ -67,4 +67,6 @@ type State struct {
 	Pid int `json:"pid"`
 	// BundlePath is the path to the container's bundle directory.
 	BundlePath string `json:"bundlePath"`
+	// Labels holds runtime-specified information that is not structured by this specification
+	Labels map[string]json.RawMessage `json:"labels"`
```
vbatts: sorry, but why `json.RawMessage`? I presume for holding structures instead of just a `string`, but couldn't that be accomplished with a `map[string]interface{}` ?

wking: On Fri, Oct 16, 2015 at 06:51:30AM -0700, Vincent Batts wrote:
> > @@ -67,4 +67,6 @@ type State struct {
> >  	Pid int `json:"pid"`
> >  	// BundlePath is the path to the container's bundle directory.
> >  	BundlePath string `json:"bundlePath"`
> > +	// Labels holds runtime-specified information that is not structured by this specification
> > +	Labels map[string]json.RawMessage `json:"labels"`
>
> sorry, but why `json.RawMessage`? I presume for holding structures
> instead of just a `string`, but couldn't that be accomplished with a
> `map[string]interface{}` ?

+1 to map[string]interface{}.  Notes on using that from Go in [1].

[1]: http://blog.golang.org/json-and-go#TOC_4.

vishh: Labels are typically used to grouping objects. See in docker and kubernetes. 
I'm weary of overloading that concept to hold arbitrary data. 
I'd say lets introduce a separate `Annotation` field that users can use to pass arbitrary data. Is this information expected to be persisted in the `State`?
Also if we switch to protobufs, extensions should let us pass arbitrary data and avoid having to include `Annotations` in the Spec.
Any thoughts @vbatts @philips ?

vbatts: There is value in labels or annotations in the mutable content IMO.

Though, even in it's current form, there is nothing stopping users from provided non-spec required fields in the config.json. i.e. the config is not _invalid_ by having additional fields. And I feel that it should stay this way.
The spec will define behavior on the known fields (regardless required or optional). Should there be other fields added, it is undefined behavior, but not prohibited.

So rather than giving up to allow folks to shit all over the place, better to give a space like labels/annotations for them to dump content that is mutable and a part of the bundle's digest.

Now if there were to be an actual section for arbitrary _runtime_ annotations, would that be separated from arbitrary data? I've not got a concrete example of this besides runtime hints of asking for 2g memory vs. some projects internal version numbering / git commit reference / etc

vbatts: Can we change this `json.RawMessage` to `interface{}`?

gao-feng: @vbatts Sure, updated, thanks for your comments.

vbatts:
LGTM

wking:
a8785586 looks good to me too.

vishh:
Labels are usually key value pairs elsewhere in docker and kubernetes. They are generally used for grouping. In this PR it looks like we are attempting to store metadata that is opaque to the spec. If that is the case can we store the metadata as a string? The users of the spec can then store anything in the string, including maps serialized to json format. 
-1 on overloading the term 'labels' for storing arbitrary metadata.

vbatts:
@vishh is it the field name of "labels" that feels like a collision to you? I am not attached to that name, but there is frustration with having to unmarshal content just to map in a string typed field.

wking:
On Fri, Oct 30, 2015 at 08:37:29AM -0700, Vish Kannan wrote:
> In this PR it looks like we are attempting to store metadata that is
> opaque to the spec.

That's my preference, yeah.  Why require a particular structure for
information that has no specified use?  This is just a playground for
runtime-specific information, and I see no benefit to restricting its
content.

> If that is the case can we store the metadata as a string? The
> users of the spec can then store anything in the string, including
> maps serialized to json format.

That seems like extra (de)serialization work for runtimes that use
this field.  And I don't see a benefit.  What do you gain by requiring
a string here?
 
> -1 on overloading the term 'labels' for storing arbitrary metadata. 

I don't really care what the key associated with this unstructured
runtime playground is, so +1 to switching to whatever key you propose
instead.

vishh:
@vbatts: Got it. +1 on not using the keyword `labels`. How about `Annotations`. 

I'm curious why we are choosing `interface` which is golang specific instead of `map[string]string]`, which can be represented in other languages as well.

wking:
On Fri, Oct 30, 2015 at 09:37:45AM -0700, Vish Kannan wrote:
> @vbatts: Got it. +1 on not using the keyword `labels`. How about
> `Annotations`.

Works for me, although I think all our existing JSON keys start with
lowercase.

> I'm curious why we are choosing `interface` which is golang specific
> instead of `map[string]string]`, which can be represented in other
> languages as well.

Because that doesn't allow deeper nesting if folks want to put more
structure into their field.  interface{} is basically the C equivalent
of void*, and that's what I think we want here.

I don't think we have to worry about language support.  Many languages
can parse arbitrary JSON, and anyone that's actually going to *use*
the data in this field will want to get a schema for the part they'll
use.  We're leaving it unspecified here, but runtimes that populate
this information will likely provide their own specs for the data they
insert (if they expect third parties to use it).

crosbymichael:
since it's json can you not just store whatever you want in there without having to add a field like this ?

wking:
On Mon, Nov 16, 2015 at 01:59:02PM -0800, Michael Crosby wrote:
> since it's json can you not just store whatever you want in there
> without having to add a field like this ?

This PR allocates an approved namespace for that, otherwise you risk
conflicting with a spec-defined `annotations` (or whatever) down the
line.

crosbymichael:
and approved namespaces for putting unapproved things, makes sense ...

runtime.md
```diff
@@ -22,6 +22,8 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
+* **`Labels`**: (map) Labels holds runtime-specified information that is not structured by this specification
+The runtime could store experimental information here before it's standardized in this specification.
```
vishh: Is there a recommended format for the values? Is it byte encoded or string or something else?

config.go
```diff
@@ -67,4 +67,6 @@ type State struct {
 	Pid int `json:"pid"`
 	// BundlePath is the path to the container's bundle directory.
 	BundlePath string `json:"bundlePath"`
+	// Labels holds runtime-specified information that is not structured by this specification
+	Labels map[string]interface{} `json:"labels"`
```
vishh: I'd like to reserve the keyword `Labels` for recognized metadata which can be used for grouping.
Can we change this to be `Annotations`?

wking: On Mon, Nov 16, 2015 at 02:12:19PM -0800, Vish Kannan wrote:
> > +	// Labels holds runtime-specified information that is not structured by this specification
> > +	Labels map[string]interface{} `json:"labels"`

We're not structuring it, so it should just probably be ‘interface{}’,
not ‘map[string]interface{}’ (and sorry about the slow path of
suggestions from json.RawMessage [1] to map[string]interface{} [2] to
just interface{} :/).

[1]: https://github.com/opencontainers/specs/pull/188#issuecomment-147030855
[2]: https://github.com/opencontainers/specs/pull/188#discussion_r42244153

gao-feng: Yep, don't need array, just one interface{} is enough. thanks

gao-feng:
@vishh @crosbymichael @wking Updated, thanks you guys.

runtime.md
```diff
@@ -23,6 +23,9 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
 
+* **`annotations`**: (interface) holds runtime-specified information that is not structured by this specification.
```
vishh: What does an `interface` mean outside of golang?

runtime.md
```diff
@@ -23,6 +23,9 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+The runtime could store experimental information here before it's standardized in this specification.
```
vishh: This is one of the use-cases. Clarify that. In the future, we can add experimental API versions, which might handle this use-case.

state.go
```diff
@@ -13,4 +13,6 @@ type State struct {
 	Pid int `json:"pid"`
 	// BundlePath is the path to the container's bundle directory.
 	BundlePath string `json:"bundlePath"`
+	// Annotations holds runtime-specified information that is not structured by this specification.
```
vishh: Why are we mentioning `runtime` here? Are we precluding humans from ever trying to use this Spec :smiley: ?

wking: On Mon, Nov 16, 2015 at 08:01:42PM -0800, Vish Kannan wrote:
> > + // Annotations holds runtime-specified information that is not
> >      structured by this specification.
> 
> Why are we mentioning `runtime` here? Are we precluding humans from
> ever trying to use this Spec :smiley: ?

No, we're reserving this namespace *for the runtime to define*.  So
anyone is welcome to read it, but the runtime is the ultimate
authority on any specification for the content, and the only one who
should be writing to it.  Otherwise we'd need sub-namespaces to keep
different players separate (e.g. see [1], but Gao pointed out that we
didn't need a state JSON entry for the bundle author [2]).

[1]: https://github.com/opencontainers/specs/pull/188#issuecomment-146912525
[2]: https://github.com/opencontainers/specs/pull/188#issuecomment-147026995

runtime.md
wking: On Mon, Nov 16, 2015 at 07:59:57PM -0800, Vish Kannan wrote:
> > +* **`annotations`**: (interface) holds runtime-specified
> >    information that is not structured by this specification.
> 
> What does an `interface` mean outside of golang?

Data with an unspecified schema.  E.g. void* in C [1].

[1]: https://github.com/opencontainers/specs/pull/188#issuecomment-152603357

wking:
161c98e9 looks good to me.

state.go
gao-feng: I can't find out other users of annotations right now, can you show me the use case? absolutely you can read it, but right now only runtime stores information in it.

vishh: Can we rename `runtime` to `users` instead?

vishh:
Just one nit. Otherwise LGTM

state.go
wking: On Wed, Dec 16, 2015 at 10:36:16AM -0800, Vish Kannan wrote:
> > +	// Annotations holds runtime-specified information that is not structured by this specification.
> 
> Can we rename `runtime` to `users` instead?

This *is* runtime-specified information.  User information just goes
somewhere in the bundle directory [1].

[1]: https://github.com/opencontainers/specs/pull/188#issuecomment-147026995

runtime.md
jonboulle: Eh, this is weird. We are already saying this is [JSON encoded](https://github.com/opencontainers/specs/pull/188/files#diff-b84a8d65d8ed53f4794cd2db7e8ea731L11), so we should just say the type here is an arbitrary JSON type (either an object or a value). "interface" is unhelpful.

vishh: +1

wking: On Wed, Dec 16, 2015 at 10:53:41AM -0800, Jonathan Boulle wrote:
> > +* **`annotations`**: (interface) holds runtime-specified information that is not structured by this specification.
>
> Eh, this is weird. We are already saying this is [JSON
> encoded](https://github.com/opencontainers/specs/pull/188/files#diff-b84a8d65d8ed53f4794cd2db7e8ea731L11),
> so we should just say the type here is an arbitrary JSON type
> (either an object or a value). "interface" is unhelpful.

“interface” is a Go-ism.  So I'm fine with interface in the Go type
and “JSON” in these Markdown docs.

philips:
If we do this I really really think we need to tell people to namespace their json fields. If the goal of the state.json is to encourage standard operations then having extensions here seems really dangerous without namespacing.

```
"annotations": {
  "example.com/funruntime": { "O_PONIES": true }
}
```

wking:
On Tue, Dec 22, 2015 at 10:53:33PM -0800, Brandon Philips wrote:
> If we do this I really really think we need to tell people to
> namespace their json fields. If the goal of the state.json is to
> encourage standard operations then having extensions here seems
> really dangerous without namespacing.
> 
> "annotations": {
>   "example.com/funruntime": { "O_PONIES": true }
> }

There is only one author (the runtime) that is allowed to write to
this field, so I don't see any change for collisions on *writing*.  It
would be good to avoid ambiguity on *reading* though (e.g. two
different runtimes define annotations with overlapping keys but
distinct semantics.  This field is (and should be) completely opaque
to the spec, but I'm open to an informative suggestion that runtime
authors insert a field so consumers can have good confidence that the
data they read comes from the runtime they expect.  For example:

  "annotations": {
    "runtime": "funruntime",
    "version": "0.1.0",
    "O_PONIES": true
  }

philips:
I am not saying don't make it opaque to the spec, but we really should
encourage namespacing one way or the other as the readers will get confused.

vbatts:
@philips I agree on the encouragement. Perhaps that verbiage ought to start for annotations in the bundled config.json, since that will be from bundle authors (not just the runtime)

wking:
On Mon, Jan 04, 2016 at 10:38:07AM -0800, Vincent Batts wrote:
> Perhaps that verbiage ought to start for annotations in the bundled
> config.json, since that will be from bundle authors (not just the
> runtime)

I think this PR is just about the state JSON and runtime annotations.
Config annotations by bundle authors are [1] (spun off from #108).

For this PR, I'd suggest phrasing like [2]:

  Runtimes are encouraged to mark and version any annotations so users
  can consume them safely.  For example:

    "annotations": {"runtime": "fun-runtime", "version": "0.1.0", "O_PONIES": true}

  Which will allow consumers to determine with high confidence that
  those annotations should be interpreted according to v0.1.0 of a
  specification published by the fun-runtime authors.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/p_u2xps6jvo/G3YZ1OL8DwAJ
     Subject: Labels and extension meta data in containers #108
     Date: Tue, 11 Aug 2015 14:14:04 -0700
     Message-ID: <CAEMTkLbqUhyY97iPZ3_4axJ3yUwX9oOVOnQQr_Ks_aRr8R-1mg@mail.gmail.com>
[2]: https://github.com/opencontainers/specs/pull/188#issuecomment-166953349

state.go
vishh: But what prevents users from adding their own metadata here? At the Spec level, why do we have to differentiate between runtimes and other users?

wking: On Mon, Jan 04, 2016 at 11:45:11AM -0800, Vish Kannan wrote:
> > +	// Annotations holds runtime-specified information that is not structured by this specification.
> 
> But what prevents users from adding their own metadata here?

There are no technical limits to something like [1], but with a
required BundlePath there is no need [2].  My personal preference
would be to drop the BundlePath requirement [3], but addressing that
seems like it's outside the scope for this PR.

[1]: https://github.com/opencontainers/specs/pull/188#issuecomment-146912525
[2]: https://github.com/opencontainers/specs/pull/188#issuecomment-147026995
[3]: https://github.com/opencontainers/specs/pull/188#issuecomment-147030855

vishh:
Closing this in favor of #331. Kindly re-open if you find #331 to not satisfy all the requirements @gao-feng...

wking:
On Wed, Mar 09, 2016 at 11:18:12AM -0800, Vish Kannan wrote:
> Closing this in favor of #331.

I think we should re-open this PR or land something similar.  #331
ended up dropping it's state JSON additions [1], and we'll want a
place for the runtime to store unspecified information.  For example,
runV wants to store a socket path [2], and you'd also need a socket
path for proposal 3 in the create/start split discussion [3].

[1]: https://github.com/opencontainers/specs/pull/331#issuecomment-194457970
[2]: https://github.com/opencontainers/specs/pull/188#issuecomment-140296423
[3]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/k55FQrBzBgAJ
     Subject: Re: Splitting Start into create() and run()
     Date: Thu, 24 Mar 2016 15:04:14 -0700
     Message-ID: <20160324220414.GC23066@odin.tremily.us>

PR #189, opened 2015-09-15
introduce mount type 'null'

gao-feng:
bundle.md said:
> The goal is that the bundle can be moved as a unit to another machine and run the same application if `runtime.json` is removed or reconfigured.

If we move the bundle from Linux to Windows. there is no linux filesystem on Windows,
and config.md said:
>Each record in this array must have configuration in [runtime config](runtime-config.md#mount-configuration).

So removing the `runtime.json` will not help to sucessfully run oci container with mounts configured.

This patch introduces a new mount type 'null', it means runtime should ignore this mount entry.
So we can change the os specified filesystem type to 'null' to make sure container can run sucessfully
on other os.

### DIFF
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -6,7 +6,7 @@ Values are objects with configuration of mount points.
 The parameters are similar to the ones in [the Linux mount system call](http://man7.org/linux/man-pages/man2/mount.2.html).
 Only [mounts from the portable config](config.md#mount-points) will be mounted.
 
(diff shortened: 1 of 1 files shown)


vbatts:
Is moving a bundle to windows an actual use-case?

dqminh:
@vbatts iiuc in hypervisor case, a bundle can be shared between different operation systems.

vbatts:
Ah. Hmm.

gao-feng:
Think about this again, for hypervisor case, the guest os should be Linux too, so the filesystems still available.
Let's forget this pr temporarily, I need to reconsider about running bundle on different platform.
Thanks.

mrunalp:
Let's close this for now and can re-open later if required.

PR #190, opened 2015-09-16
Remove the unneeded build tag from the config_linux.go

laijs:
config_linux.go already has the "_linux" for the go build,
so the build tag in the file is redundant.


vbatts:
LGTM

mrunalp:
LGTM

Issue #166, opened 2015-09-09
[Tracker] Platform/Compiler agnostic schema


vbatts:
like #135 began to fix, this needs to track our movement from golang compiler constrained schema, to something like json-schema.
At that point, we can have the provided `*.go` (if at all) as a reference for valid structs for the schema.

wking:
On Wed, Sep 09, 2015 at 10:30:24AM -0700, Vincent Batts wrote:
> like #135 began to fix…

How is this related to #135?

> … this needs to track our movement from golang compiler constrained
> schema, to something like json-schema.

JSON Schema is fine with me (although their docs can be hard to browse
;).  Are any other schema languages under consideration?  If we're all
on-board with JSON Schema, I'd just start working on it :p.  If we
don't think everybody's on board, I'd suggest floating alternatives on
the mailing list, and linking to the list thread from this issue.

dqminh:
I think a richer schema is definitely beneficial.

Does json schema support things like one field can have different types (like uid/gid can be string on windows iirc ), and `oneOf/anyOf` relationship for things like `Resources/CgroupPaths` etc. ?

wking:
On Wed, Sep 09, 2015 at 11:25:47AM -0700, Daniel, Dao Quang Minh wrote:
> Does json schema support things like one field can have different
> types (like uid/gid can be string on windows iirc ),

That sounds like oneOf.

> … and `oneOf/anyOf` relationship for things like
> `Resources/CgroupPaths` etc. ?

Yes, see [1].

Maybe the approach should be “start translating the spec to JSON
Schema (or any other proposed schema language), and then compare the
partial translations to see which feels most expressive”?

[1]: http://spacetelescope.github.io/understanding-json-schema/reference/combining.html

vbatts:
@wking it's related to #135 because the schema is presently tied up in source that will compile differently on various platforms. So to have a way, for instance, to validate a linux config, from an OSX host, is not straight forward. This is what we were talking about on the call today.

wking:
On Wed, Sep 09, 2015 at 07:53:07PM -0700, Vincent Batts wrote:
> @wking it's related to #135 because the schema is presently tied up
> in source that will compile differently on various platforms…

#135 is a step towards a platform-agnostic Go spec.  Is that a valid
end-goal?  It may be better than a platform-specific Go spec, but it
seems like dead-end work if this issue gets traction for a non-Go spec
language.  On the other hand, if we expect a non-Go spec to take a
[...]

Issue #38, opened 2015-06-30
linux: username: how to figure out uid reliably


philips:
Right now the spec says you need to specify an OS relevant user id for the process to exec on behalf of. Many people don't think about this low-level primitive and rely on user databases like `/etc/passwd`. In order to support a user saying `apache` in the open container configuration we would need to do hacks on Linux:  Sadly this requires hacks:

- parse `/etc/passwd` (if it exists!)
- call `getent passwd` inside of the filesystem

This spec likely should make a recommendation on what needs to be done here and in what order if we are to support a "username".

This issues replaces #10 and is being refiled since we made a decision to be more explicit and simple for the initial draft milestone.

wking:
On Tue, Jun 30, 2015 at 04:14:17PM -0700, Brandon Philips wrote:
> Right now the spec says you need to specify an OS relevant user id
> for the process to exec on behalf of. Many people don't think about
> this low-level primitive and rely on user databases like
> `/etc/passwd`. In order to support a user saying `apache` in the
> open container configuration…

I don't think we need to do this.  Using UID/GID is easy for the
runtime [1], and ‘id -u apache’, etc., is easy for bundle authors.

[1]: https://github.com/opencontainers/specs/issues/10#issuecomment-117369515

wking:
On Sun, Sep 06, 2015 at 02:01:42PM -0700, W. Trevor King wrote:
> I don't think we need to do this.  Using UID/GID is easy for the
> runtime [1], and ‘id -u apache’, etc., is easy for bundle authors.

Of course, they'd have to enter the bundle for that ‘id’ call to work
;).  So maybe not super-trivial, but I still think we should be
offloading the UID/GID lookup to bundle-author tooling.

julz:
bumping this since we're getting past the first milestone and starting to talk about usability now. I can't think of a good way to implement a pretty common use case like `docker run -u myuser /bin/bash` without mapping usernames to uids. I think this becomes particularly important where you're trying to support 'runc exec' or having multiple bundles that share a namespace. If you consider a case where the second bundle wants to log in as a user added/modified by the first bundle, it doesn't make sense to try to create the bundle with a pre-mapped uid: the user literally may not exist (or may exist with the wrong uid) when the bundle is created. The only sane way I can think of to handle this is for the runtime to do the mapping before executing the second bundle.

[...]

Issue #10, opened 2015-06-25
linux: figuring out the uid/gid


philips:
A config.json may specify a string "username" on Linux as the user to execute the process as. The trouble is we need to know what uid/gid this maps to inside of the container. Sadly this requires hacks:

- parse /etc/passwd (if it exists!)
- call `getent passwd` inside of the filesystem

This spec likely should make a recommendation on what needs to be done here and in what order.

jonboulle:
Also important to note that the latter is a potential [attack vector](https://github.com/appc/spec/issues/340#issue-72915581).

mrunalp:
A viable option may be to punt this to higher level code and just allow numeric uid/gid/groups in the spec for Linux.

LK4D4:
I like idea of numerical representation. Some implementations can provide extension with names, but numbers are easier for spec I think.

taylorb-microsoft:
The windows implementation of a user number would be a SID's (https://msdn.microsoft.com/en-us/library/windows/desktop/aa379571(v=vs.85).aspx).  There are a set of 'well known SIDs' as well as per-user SIDs.

philips:
@jonboulle If you execute the "passwd probe process" with the same level of permissions as the process will it seems equivalent.

mrunalp:
There is also a question of how to represent this in the spec. Make it OS specific or jam all fields into the object and the json only sets what it needs for that OS?

taylorb-microsoft:
At the OCF level I'd prefer to find an OS agnostic options when ever possible (it will only strengthen the spec) - then upper levels can, at there option, use other more specific and even OS specific terms.

*For Example:*
If in the OCF JSON we could find a common string - i.e. SID for windows and UID for Linux.  Then the upper layer could for Windows say "Administrator" (which maps to a well known SID) or for Linux "root" or even define an OS common option say "god" which maps to "Administrator" on Windows and "root" on Linux.

<tangent> speaking to finding OS agnostic options - as we go though this process it might be worth identifying a method of calling out 'this is OS specific with strong intent to unify' - while there are parts of Windows I may never be able to change, there are parts that can and should be changed and while it might take a release or to before we can integrate some changes that should not remove the intent. </tangent>

philips:
After @taylorb-microsoft's feedback I think we are into the zone where this becomes an OS specific field. There is nothing common between the parsing or interpretation of the user field between OSes.

mrunalp:
[...]

PR #191, opened 2015-09-16
use /etc/passwd in place of the User struct

laijs:
The type of `user` becomes `string` instead of platform-specific
structure, so the Spec becomes a platform-agnostic structure.

It makes we can parse the config.json as a Spec before knowing the
os. It removes the User structure confliction for different platforms.

For linux-based container, parsing the /etc/passwd gives us a consistent
and more reliable way to assign the uid/gids to the process.

### DIFF
--- a/config.md
+++ b/config.md
@@ -71,24 +71,14 @@ Each record in this array must have configuration in [runtime config](runtime-co
 * **cwd** (string, optional) is the working directory that will be set for the executable.
 * **env** (array of strings, optional) contains a list of variables that will be set in the process's environment prior to execution. Elements in the array are specified as Strings in the form "KEY=value". The left hand side must consist solely of letters, digits, and underscores `_` as outlined in [IEEE Std 1003.1-2001](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html).
 * **args** (string, required) executable to launch and any flags as an array. The executable is the first element and must be available at the given path inside of the rootfs. If the executable path is not an absolute path then the search $PATH is interpreted to find the executable.
-
-The user for the process is a platform-specific structure that allows specific control over which user the process runs as.
-For Linux-based systems the user structure has the following fields:
-
-* **uid** (int, required) specifies the user id.
-* **gid** (int, required) specifies the group id.
-* **additionalGids** (array of ints, optional) specifies additional group ids to be added to the process.
+* **user** (string, required) is the user name for the process. The runtime should resolve the required information for the process by its platform-specific ways. For Linux-based containers, the runtime could parse the /etc/passwd and /etc/group inside the rootfs.
 
 *Example (Linux)*
 
 ```json
 "process": {
     "terminal": true,
-    "user": {
-        "uid": 1,
-        "gid": 1,
-        "additionalGids": [5, 6]
-    },
+    "user": "root",
     "env": [
         "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
         "TERM=xterm"
--- a/config.go
+++ b/config.go
@@ -21,8 +21,8 @@ type Spec struct {
 type Process struct {
 	// Terminal creates an interactive terminal for the container.
 	Terminal bool `json:"terminal"`
-	// User specifies user information for the process.
-	User User `json:"user"`
+	// User specifies user name for the process.
+	User string `json:"user"`
 	// Args specifies the binary and arguments for the application to execute.
 	Args []string `json:"args"`
 	// Env populates the process environment for the process.
--- a/config_linux.go
+++ b/config_linux.go
@@ -14,14 +14,3 @@ type Linux struct {
 	// Capabilities are linux capabilities that are kept for the container.
 	Capabilities []string `json:"capabilities"`
 }
-
-// User specifies linux specific user and group information for the container's
-// main process.
-type User struct {
-	// UID is the user id.
-	UID int32 `json:"uid"`
-	// GID is the group id.
-	GID int32 `json:"gid"`
-	// AdditionalGids are additional group ids set for the container's process.
-	AdditionalGids []int32 `json:"additionalGids"`
-}


julz:
yes, please :-)

see also https://github.com/opencontainers/specs/issues/38

laijs:
#38 #135 #166 

@philips @totherme @glyn @julz @vbatts @wking

glyn:
Seems reasonable to me.

mrunalp:
We can discuss this in today's call.

config.md
```diff
@@ -71,24 +71,14 @@ Each record in this array must have configuration in [runtime config](runtime-co
...
 * **args** (string, required) executable to launch and any flags as an array. The executable is the first element and must be available at the given path inside of the rootfs. If the executable path is not an absolute path then the search $PATH is interpreted to find the executable.
-
-The user for the process is a platform-specific structure that allows specific control over which user the process runs as.
-For Linux-based systems the user structure has the following fields:
-
-* **uid** (int, required) specifies the user id.
-* **gid** (int, required) specifies the group id.
-* **additionalGids** (array of ints, optional) specifies additional group ids to be added to the process.
+* **user** (string, required) is the user name for the process. The runtime should resolve the required information for the process by its platform-specific ways. For Linux-based containers, the runtime could parse the /etc/passwd and /etc/group inside the rootfs.
```
wking: “could parse /etc/passwd and /etc/group” is to wishy.  A bundle author will need to know exactly how this lookup will be performed, so they can setup their bundle to support it.  How about using the appc wording referenced from #38?

wking:
On Wed, Sep 16, 2015 at 02:29:49AM -0700, Lai Jiangshan wrote:
> The type of `user` becomes `string` instead of platform-specific
> structure, so the Spec becomes a platform-agnostic structure.

appc uses strings but keeps separate fields for user, group, and
additional-gids (although they still use integers for this last one,
see #38).  I'm ok with strings, but if we use a single string we'll
want to define the syntax for packing each of these fields in
(‘<user>[:<group>[:<additional-group>…]]’?) and make sure we're
explicit about how we expect the lookup to work (see appc's text
linked from #38 for an example).

laijs:
I just noticed it just reverts the 8af3eb64b22da540acf8b67700c46a0dd72d1ec3. @crosbymichael 

From the #10, I learned something bad about the pure string name.

laijs:
@wking suggested use the structured-and-rich-information-encoded name as AppC used.
But a structured-and-rich-information-encoded-string is not better than json.RawMessage IMO.  (see  the 44b4415b169efd439f424495bf42fb6a4a82549f in #135), I don't want to reinvent the wheel.

Could any maintainer directly give a approach?

I consider that mapping the User struct from the username is a good idea. Both the username and the map are only valid in the config file, so there is no security problem. And we can benefit from the User struct as current code.

wking:
On Thu, Sep 17, 2015 at 01:01:25AM -0700, Lai Jiangshan wrote:
> @wking suggested use the structured-and-rich-information-encoded
> name as AppC used.

I think there are two things going on here:

1. Do we want to support name → ID lookup?
2. Do we want structured fields for user, group, and additional-group
   information?

I think most folks are on board with (1).  In yesterday's meeting
(minutes in [1], but not yet posted to the wiki), I think the
consensus was that we want to keep explicit JSON fields (2).  And we
want to move a step beyond appc's overloaded string (“If no valid
matches are found, then if the string is all numerical, it shall be
converted to an integer and used as the UID/GID” and they have a path
fallback in there too) and have explicit fields for the different
lookup types.  So something like:

* **uid** (int, optional) specifies the user ID.
* **username** (string, optional) specifies the user name.
  After entering the container, the runtime process will parse `/etc/passwd` to find the associated ID before exec'ing the application.
  If the name is not in `/etc/passwd`, the runtime process will print a message to its stderr, initiate the teardown procedures, and return a non-zero exit code.
  A configuration is invalid if both **uid** and **username** are set.
  If neither are set, the runtime must use UID 0.
* **gid** (int, optional) specifies the group ID.
* **groupname** (string, optional) specifies the group name.
  The logic is the same as for **uid** and **username**, except the **groupname** lookup will use `/etc/group`.
* **additionalGids** (array of ints, optional) specifies additional group IDs to be added to the process.
* **additionalGroupNames** (array of strings, optional) specifies additional group names to be added to the process.
  The logic is the same as for **gid** and **groupname**, except that specifying both **additionalGids** and **additionalGroupNames** is valid.
  If both are specified, the union of group IDs will be used for the application.

We can add similar fields for userPath and groupPath if we want later
on.

> Could any maintainer directly give a approach?

Waiting for approval from a maintainer is a good idea.  Sometime what
I think I heard in a meeting doesn't turn out to be what they heard
;).  That's probably unavoidable when we're agreeing without some
explicit text to make sure we're talking about the same thing.

> Both the username and the map are only valid in the config file, so
> there is no security problem.

The security problem is if you call getent in the container before
dropping privileges [3], since that would execute bundle-supplied
code.  Using runtime-supplied code to read /etc/passwd and /etc/groups
to perform the lookup is safe (and after entering the container you've
already setup blkio rate limiting etc., so a huge /etc/passwd won't
even hog your disk bandwidth).

[1]: https://docs.google.com/document/d/1ZdUPSQvg-KKA-S4I67GXSgOfGXtjIgEPk36FtnGI_gM/edit
[2]: https://github.com/appc/spec/blob/4fba3b37e64b8ed325a22e12f9f10e3acd59e012/spec/aci.md
[3]: https://github.com/appc/spec/issues/340

crosbymichael:
@philips can you check this out and see if it's inline with what you were thinking on your original issue?

julz:
@philips @crosbymichael @vbatts  looks like this issue stalled a few months ago, we're still maintaining a fork so we can run our tests on mac and would love to be able to stop doing that. There's also https://github.com/opencontainers/specs/issues/166 which seems like a more generic approach to making the spec importable on multiple platforms, but it also hasn't been updated since September. In the meantime just making User a string as in this PR (and documenting the semantics) looks like it solves the problem. Any new thoughts about this?

laijs:
I don't think #166 is feasible in short time.
I didn't expect it has been consuming so much time without any progress.
The runv and the hyper have to use a simple trick (rename) to handle it on mac currently.

https://github.com/hyperhq/runv/blob/master/Makefile.am#L19-L23
https://github.com/hyperhq/hyper/blob/master/Makefile.am#L26-L30

hqhq:
@laijs Haven't read the whole story, but I see the problem is you can still see `User` struct on MAC which defined in `config_linux.go`? Is adding a build flag in `config_linux.go` gonna fix your problem?

Ignore it if I'm missing something big and total wrong :)

wking:
On Fri, Dec 04, 2015 at 01:35:34AM -0800, Qiang Huang wrote:
> @laijs Haven't read the whole story, but I see the problem is you
> can still see `User` struct on MAC which defined in
> `config_linux.go`? Is adding a build flag in `config_linux.go` gonna
> fix your problem?

Hmm, this^ sounds more complicated to me.  In the absence of native
containerization (e.g. with the runV approach), there's not a clear
link between the host OS (which sets the build flags) and the launched
container (where process.user will be applied, although how that
information is passed to the guest machine is beyond me).  So whatever
structure you build into the runtime will be wrong for some bundles.
I think the only solution is build a runtime that can support all
bundle (e.g. #185, which is in the roadmap from #230, so it's
hopefully not too far out).

laijs:
@hqhq the go build ignores the xxxx_linux.go in the mac. (the build tag in _linux.go is redundant if you add the tag)

the runv supports linux containers on the mac, so it requires the linux specified User structure. so we prefer the file name xxx-linux.go, or put then in the linux/ directory without _linux suffix.

see #135

config.md
philips: getent is really the only correct way to handle this. Perhaps we just say if you want a UID map you can put that into your config.json similar to the mount points? https://github.com/opencontainers/specs/blob/master/config.md#mount-points

philips:
I think we should probably just retire ourselves to the fact that the concept of user is so platform specific that trying to put it into the shared part of the spec is impossible.

More and more I don't think there is anything that can safely be shared between platforms besides environment variables.

config.md
wking: On Tue, Dec 22, 2015 at 11:07:42PM -0800, Brandon Philips wrote:
> > +* **user** (string, required) is the user name for the
> > process. The runtime should resolve the required information for
> > the process by its platform-specific ways. For Linux-based
> > containers, the runtime could parse the /etc/passwd and /etc/group
> > inside the rootfs.
>
> getent is really the only correct way to handle this.

I don't buy that ;).  For example, I don't see much difference
between:

> Perhaps we just say if you want a UID map you can put that into your
> config.json similar to the mount points?
> https://github.com/opencontainers/specs/blob/master/config.md#mount-points

and defining the mappings in /etc/passwd and /etc/group.  I think you
cover a useful subset of NSS implementations by supporting /etc/passwd
and /etc/group, and you allow runtimes to avoid running container-side
code to perform the lookup (for example, they can use a host-side NSS
implementation to parse the bundle's /etc/passwd and /etc/group).

wking:
On Tue, Dec 22, 2015 at 11:09:06PM -0800, Brandon Philips wrote:
> I think we should probably just retire ourselves to the fact that
> the concept of user is so platform specific that trying to put it
> into the shared part of the spec is impossible.

I don't see a platform-compatibility problem with the schema outlined
in [1].  Platforms that don't support UIDs can error out if ‘uid’ is
set, and we can add platform-specific lookup logic.  E.g. for
‘username’ we can require /etc/passwd for Linux lookups, some registry
(or whatever they use) for Windows lookups, etc.  Although we'd
probably want separate fields for “host platform” and “container
platform” so the hypervisor folks can figure out what rules to apply.

> More and more I don't think there is anything that can safely be
> shared between platforms besides environment variables.

That is a pretty pessimistic view, and argues in favor of just
splitting the spec into opencontainers/specs-linux,
opencontainers/specs-windows, … (this is what I did with the
Linux-only ccon [2]).  Cross-platform specs may be a bit more
complicated (see all the platform caveats here [3]), but I think it's
a worthy goal for areas where there is sufficient overlap to avoid it
being *too* complicated ;).

[1]: https://github.com/opencontainers/specs/pull/191#issuecomment-141150387
[2]: https://github.com/wking/ccon
[3]: https://docs.python.org/dev/library/subprocess.html#subprocess.Popen

philips:
On Wed, Dec 23, 2015 at 9:57 AM W. Trevor King <notifications@github.com>
wrote:

> On Tue, Dec 22, 2015 at 11:09:06PM -0800, Brandon Philips wrote:
> > I think we should probably just retire ourselves to the fact that
> > the concept of user is so platform specific that trying to put it
> > into the shared part of the spec is impossible.
>
> I don't see a platform-compatibility problem with the schema outlined
> in [1]. Platforms that don't support UIDs can error out if ‘uid’ is
> set, and we can add platform-specific lookup logic. E.g. for
> ‘username’ we can require /etc/passwd for Linux lookups, some registry
> (or whatever they use) for Windows lookups, etc. Although we'd
> probably want separate fields for “host platform” and “container
> platform” so the hypervisor folks can figure out what rules to apply.
>

With such a complex set of rules on each platform we might as well just
make different schemas. What is the point of a shared schema with different
semantics based on context?

And FWIW, I am not saying we should give up on shared language and shared
spec. Just that we are deriving very little value from having a shared
schema.

config.md
philips: I don't think we should let random files inside of the container affect behavior of the runtime. And this is what /etc/passwd parsing does. It lets the contents inside of the container control the UID the runtime executes processes as.

We could trivially create a tool that extracts the /etc/passwd into a map and puts it into the JSON schema. Thoughts @jonboulle @vbatts ?

wking: On Wed, Dec 23, 2015 at 11:19:16AM -0800, Brandon Philips wrote:
> > +* **user** (string, required) is the user name for the
> >  process. The runtime should resolve the required information for
> >  the process by its platform-specific ways. For Linux-based
> >  containers, the runtime could parse the /etc/passwd and
> >  /etc/group inside the rootfs.
>
> I don't think we should let random files inside of the container
> affect behavior of the runtime. And this is what /etc/passwd parsing
> does. It lets the contents inside of the container control the UID
> the runtime executes processes as.

I don't see a point in distinguishing between “inside the container”
(just /etc/passwd) and “inside the bundle” (both /etc/passwd and the
OCI configs).  What do you gain by making such a distinction?

philips: The /etc/passwd may be modified at runtime by processes executing inside of that filesystem while the manifest is inaccessible. So, if we have operations like "launch another process" the behavior may change based on the new state of that file.

wking:
On Wed, Dec 23, 2015 at 11:17:10AM -0800, Brandon Philips wrote:
> On Wed, Dec 23, 2015 at 9:57 AM W. Trevor King wrote:
> > On Tue, Dec 22, 2015 at 11:09:06PM -0800, Brandon Philips wrote:
> > > I think we should probably just retire ourselves to the fact
> > > that the concept of user is so platform specific that trying to
> > > put it into the shared part of the spec is impossible.
> >
> > I don't see a platform-compatibility problem with the schema
> > outlined in [1]. Platforms that don't support UIDs can error out
> > if ‘uid’ is set, and we can add platform-specific lookup
> > logic. E.g. for ‘username’ we can require /etc/passwd for Linux
> > lookups, some registry (or whatever they use) for Windows lookups,
> > etc. Although we'd probably want separate fields for “host
> > platform” and “container platform” so the hypervisor folks can
> > figure out what rules to apply.
> 
> With such a complex set of rules on each platform we might as well
> just make different schemas. What is the point of a shared schema
> with different semantics based on context?

Because it lets you agree where you can.  For example, all of our
platforms support the same semantics for fields backed by Go's Cmd.  I
don't see a need to duplicate all of that in per-platform schemas
because the platforms divergence on user modeling.

> And FWIW, I am not saying we should give up on shared language and
> shared spec. Just that we are deriving very little value from having
> a shared schema.

Currently the spec doesn't have much normative content outside the
config schema.  There's the state schema (which also has
cross-platform issues, e.g. PIDs are non-sensical for hypervisors),
some generic lifecycle wording, and… what?  So I'm not sure what value
there is to “shared language” in the absence of at least some schema
overlap.

philips:
On Wed, Dec 23, 2015 at 11:44 AM W. Trevor King <notifications@github.com>
wrote:

> On Wed, Dec 23, 2015 at 11:17:10AM -0800, Brandon Philips wrote:
> > On Wed, Dec 23, 2015 at 9:57 AM W. Trevor King wrote:
> > > On Tue, Dec 22, 2015 at 11:09:06PM -0800, Brandon Philips wrote:
> > > > I think we should probably just retire ourselves to the fact
> > > > that the concept of user is so platform specific that trying to
> > > > put it into the shared part of the spec is impossible.
> > >
> > > I don't see a platform-compatibility problem with the schema
> > > outlined in [1]. Platforms that don't support UIDs can error out
> > > if ‘uid’ is set, and we can add platform-specific lookup
> > > logic. E.g. for ‘username’ we can require /etc/passwd for Linux
> > > lookups, some registry (or whatever they use) for Windows lookups,
> > > etc. Although we'd probably want separate fields for “host
> > > platform” and “container platform” so the hypervisor folks can
> > > figure out what rules to apply.
> >
> > With such a complex set of rules on each platform we might as well
> > just make different schemas. What is the point of a shared schema
> > with different semantics based on context?
>
> Because it lets you agree where you can. For example, all of our
> platforms support the same semantics for fields backed by Go's Cmd. I
> don't see a need to duplicate all of that in per-platform schemas
> because the platforms divergence on user modeling.
>

Sure, lets all use the same fields for the Cmd stuff and share that.

But, for User it feels we should just make it explicit and pull it into the
Linux, Windows, etc schemas.

>

config.md
wking: On Wed, Dec 23, 2015 at 11:42:49AM -0800, Brandon Philips wrote:
> > +* **user** (string, required) is the user name for the
> >  process. The runtime should resolve the required information for
> >  the process by its platform-specific ways. For Linux-based
> >  containers, the runtime could parse the /etc/passwd and
> >  /etc/group inside the rootfs.
> 
> The /etc/passwd may be modified at runtime by processes executing
> inside of that filesystem while the manifest is inaccessible. So, if
> we have operations like "launch another process" the behavior may
> change based on the new state of that file.

Yeah, this lookup should happen in the container process after joining
the namespace but before exec'ing the user-configured code.  So if
/etc/passwd changes and you launch a new process, you get the new UID.
I don't have a problem with that (it seems like what the user intended
when they modified /etc/passwd).

wking:
On Wed, Dec 23, 2015 at 11:49:35AM -0800, Brandon Philips wrote:
> Sure, lets all use the same fields for the Cmd stuff and share that.
> 
> But, for User it feels we should just make it explicit and pull it
> into the Linux, Windows, etc schemas.

I have no problem with that.  And once we actually have *something*
for non-Linux user schemas we can revisit whether the split makes
sense.  I'm also fine just rolling ahead with a Linux-centric user
schema, and revisiting whether or not we need a split after a
non-Linux OS speaks up with “this won't work for us, we need
$ALTERNATIVE”.

I don't see the need to namespace this (like we do now with
linux.namespaces, etc.), because the platform entry in config.json is
sufficient to specify which per-platform user schema we're using (and
I'd recommend dropping our existing ‘linux’ namespacing for the same
reasons).  But perhaps that's too awkward to handle in Go, and we're
too committed to single-pass-into-Go-types JSON-parsing approach.

philips:
On Wed, Dec 23, 2015 at 11:57 AM W. Trevor King <notifications@github.com>
wrote:

> On Wed, Dec 23, 2015 at 11:49:35AM -0800, Brandon Philips wrote:
> > Sure, lets all use the same fields for the Cmd stuff and share that.
> >
> > But, for User it feels we should just make it explicit and pull it
> > into the Linux, Windows, etc schemas.
>
> I have no problem with that. And once we actually have *something*
> for non-Linux user schemas we can revisit whether the split makes
> sense. I'm also fine just rolling ahead with a Linux-centric user
> schema, and revisiting whether or not we need a split after a
> non-Linux OS speaks up with “this won't work for us, we need
> $ALTERNATIVE”.
>

Ack.

I don't see the need to namespace this (like we do now with
> linux.namespaces, etc.), because the platform entry in config.json is
> sufficient to specify which per-platform user schema we're using (and
> I'd recommend dropping our existing ‘linux’ namespacing for the same
> reasons). But perhaps that's too awkward to handle in Go, and we're
> too committed to single-pass-into-Go-types JSON-parsing approach


I don't think we are committed but there just isn't great tooling for an
alternative. We should keep that discussion on
https://github.com/opencontainers/specs/pull/185 though.

crosbymichael:
We are currently working on fixing this for cross platform concerns but as far as linux goes the passwd lookup should be done before config generation.  There are many different problems defering this lookup to the runtime and it also removes the ability of using things like nss to looking the uid/gid for the system. 

We reviewed this together in the f2f oci meeting and are going to keep uid/gid as required for linux and consumers should do lookups with either nss or etc/passwd before populating the config for the container.

PR #193, opened 2015-09-16
runtime: config: linux: Edit BlockIO struct

runcom:
`WeightDevice`, `ThrottleReadBpsDevice`, `ThrottleWriteBpsDevice`, `ThrottleReadIOpsDevice`, `ThrottleWriteIOpsDevice` are now slices to well defined structs to allow setting multiple devices in their respective blkio file. By using a string to represents those values it wasn't possible to set correct values when multiple devices were passed in the config (either newline separated or comma separated).

I don't know how this worked before to define multiple devices for these cgroups resources. Writing something like `8:0 500\n8:16 300` isn't working (there's a Docker PR https://github.com/docker/docker/pull/13959 which passes that format to libcontainer and by passing that format to current master runc isn't working as well (neither using comma) because it just writes it to cgroup file which doesn't accept newlines/commas).
`8:0 500\n8:16 300` will only set the first device before the `\n` when written to `blkio.weight_device` for instance.

I'm not fond of `blockIODevice` embedded struct and specifying major:minor for this. I think it will suffice specifying just a `Path` and do the syscall.Stat to retrieve major:minor where it needs to (I'm thinking about libcontainer as it's currently doing for Rlimits for instance).
By just specifing a `Path` this will help also for example to just receive a string like `/dev/sda` and do the major:minor convert in the implementation. It's useful to have the plain string like for checking  which scheduler the device is using (`/sys/block/sda/queue/scheduler` w/o converting back major:minor to path in implementation).

I'll open an issue in `runc` to track this and update `runc/libcontainer` if this patch is correct and lands.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -50,16 +50,16 @@ within the container.
 Devices is an array specifying the list of devices to be created in the container.
 Next parameters can be specified:
 
-* type - type of device: 'c', 'b', 'u' or 'p'. More info in `man mknod`
-* path - full path to device inside container
-* major, minor - major, minor numbers for device. More info in `man mknod`.
+* **type** - type of device: `c`, `b`, `u` or `p`. More info in `man mknod`
+* **path** - full path to device inside container
+* **major, minor** - major, minor numbers for device. More info in `man mknod`.
                  There is special value: `-1`, which means `*` for `device`
                  cgroup setup.
-* permissions - cgroup permissions for device. A composition of 'r'
-                (read), 'w' (write), and 'm' (mknod).
-* fileMode - file mode for device file
-* uid - uid of device owner
-* gid - gid of device owner
+* **permissions** - cgroup permissions for device. A composition of `r`
+                (read), `w` (write), and `m` (mknod).
+* **fileMode** - file mode for device file
+* **uid** - uid of device owner
+* **gid** - gid of device owner
 
 ```json
    "devices": [
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
 
 `cgroupsPath` can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
 
-Optionally, cgroups limits can be specified via `resources`.
+You can configure a container's cgroups via the `resources` field of the Linux configuration.
+Do not specify `resources` unless limits have to be updated.
+For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
+
+#### Disable out-of-memory killer
 
 ```json
-    "resources": {
-        "disableOOMKiller": false,
-        "memory": {
-            "limit": 0,
-            "reservation": 0,
-            "swap": 0,
-            "kernel": 0,
-            "swappiness": -1
-        },
-        "cpu": {
-            "shares": 0,
-            "quota": 0,
-            "period": 0,
-            "realtimeRuntime": 0,
-            "realtimePeriod": 0,
-            "cpus": "",
-            "mems": ""
-        },
-        "blockIO": {
-            "blkioWeight": 0,
-            "blkioWeightDevice": "",
-            "blkioThrottleReadBpsDevice": "",
-            "blkioThrottleWriteBpsDevice": "",
-            "blkioThrottleReadIopsDevice": "",
-            "blkioThrottleWriteIopsDevice": ""
-        },
-        "hugepageLimits": null,
-        "network": {
-            "classId": "",
-            "priorities": null
-        }
+    "disableOOMKiller": false
+```
+
+#### Memory
+
+```json
+    "memory": {
+        "limit": 0,
+        "reservation": 0,
+        "swap": 0,
+        "kernel": 0,
+        "swappiness": -1
     }
 ```
 
-Do not specify `resources` unless limits have to be updated.
-For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
+#### CPU
+
+```json
+    "cpu": {
+        "shares": 0,
+        "quota": 0,
+        "period": 0,
+        "realtimeRuntime": 0,
+        "realtimePeriod": 0,
+        "cpus": "",
+        "mems": ""
+    }
+```
+
+#### Block IO Controller
+
+`blockIO` represents the cgroup subsystem `blkio` which implements the block io controller.
+For more information, see the [kernel cgroups documentation about `blkio`](https://www.kernel.org/doc/Documentation/cgroups/blkio-controller.txt).
+
+The following parameters can be specified to setup the block io controller:
+
+* **`blkioWeight`** *(uint16, optional)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules. The range is from 10 to 1000.
+
+* **`blkioLeafWeight`** *(uint16, optional)* - equivalents of `blkioWeight` for the purpose of deciding how much weight tasks in the given cgroup has while competing with the cgroup's child cgroups. The range is from 10 to 1000.
+
+* **`blkioWeightDevice`** *(array, optional)* - specifies the list of devices which will be bandwidth rate limited. The following parameters can be specified per-device:
+    * **`major, minor`** *(int64, required)* - major, minor numbers for device. More info in `man mknod`.
+    * **`weight`** *(uint16, optional)* - bandwidth rate for the device, range is from 10 to 1000.
+    * **`leafWeight`** *(uint16, optional)* - bandwidth rate for the device while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only.
+
+    You must specify at least one of `weight` or `leafWeight` in a given entry, and can specify both.
+
+* **`blkioThrottleReadBpsDevice`**, **`blkioThrottleWriteBpsDevice`**, **`blkioThrottleReadIOPSDevice`**, **`blkioThrottleWriteIOPSDevice`** *(array, optional)* - specify the list of devices which will be IO rate limited. The following parameters can be specified per-device:
+    * **`major, minor`** *(int64, required)* - major, minor numbers for device. More info in `man mknod`.
+    * **`rate`** *(uint64, required)* - IO rate limit for the device
+
+###### Example
+
+```json
+    "blockIO": {
+        "blkioWeight": 0,
+        "blkioLeafWeight": 0,
+        "blkioWeightDevice": [
+            {
+                "major": 8,
+                "minor": 0,
+                "weight": 500,
+                "leafWeight": 300
+            },
+            {
+                "major": 8,
+                "minor": 16,
+                "weight": 500
+            }
+        ],
+        "blkioThrottleReadBpsDevice": [
+            {
+                "major": 8,
+                "minor": 0,
+                "rate": 600
+            }
+        ],
+        "blkioThrottleWriteIOPSDevice": [
+            {
+                "major": 8,
+                "minor": 16,
+                "rate": 300
+            }
+        ]
+    }
+```
+
+#### Huge page limits
+
+```json
+   "hugepageLimits": [
+        {
+            "pageSize": "2MB",
+            "limit": 9223372036854771712
+        }
+   ]
+```
+
+#### Network
+
+```json
+   "network": {
+        "classId": "ClassId",
+        "priorities": [
+            {
+                "name": "eth0",
+                "priority": 500
+            },
+            {
+                "name": "eth1",
+                "priority": 1000
+            }
+        ]
+   }
+```
 
 ## Sysctl
 
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -104,20 +104,46 @@ type InterfacePriority struct {
 	Priority int64 `json:"priority"`
 }
 
+// blockIODevice holds major:minor format supported in blkio cgroup
+type blockIODevice struct {
+	// Major is the device's major number.
+	Major int64 `json:"major"`
+	// Minor is the device's minor number.
+	Minor int64 `json:"minor"`
+}
+
+// WeightDevice struct holds a `major:minor weight` pair for blkioWeightDevice
+type WeightDevice struct {
+	blockIODevice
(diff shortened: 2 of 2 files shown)


runtime-config-linux.md
```diff
@@ -169,11 +169,11 @@ Optionally, cgroups limits can be specified via `resources`.
...
-            "blkioWeightDevice": "",
-            "blkioThrottleReadBpsDevice": "",
-            "blkioThrottleWriteBpsDevice": "",
-            "blkioThrottleReadIopsDevice": "",
-            "blkioThrottleWriteIopsDevice": ""
+            "blkioWeightDevice": null,
+            "blkioThrottleReadBpsDevice": null,
+            "blkioThrottleWriteBpsDevice": null,
+            "blkioThrottleReadIopsDevice": null,
+            "blkioThrottleWriteIopsDevice": null
```
wking: There doesn't seem to be much point in having these examples if they're all `null`.  Maybe drop some of them completely, and then set the remaining ones to something more useful to show the new syntax?

runcom: I left it as it was but yes makes sense

wking:
On Wed, Sep 16, 2015 at 07:51:53AM -0700, Antonio Murdaca wrote:
> I'm not fond of `blockIODevice` embedded struct and specifying
> major:minor for this. I think it will suffice specifying just a
> `Path` and do the syscall.Stat to retrieve major:minor where it
> needs to (I'm thinking about libcontainer as it's currently doing
> for Rlimits for instance).

I'm fine with blockIODevice.  One benefit is that we don't need to
specify ordering between mounting, mknod calls, and these resource
limits, because we can set major:minor limits before the path exists.

runcom:
> On Wed, Sep 16, 2015 at 07:51:53AM -0700, Antonio Murdaca wrote:
>  I'm not fond of `blockIODevice` embedded struct and specifying
>  major:minor for this. I think it will suffice specifying just a
>  `Path` and do the syscall.Stat to retrieve major:minor where it
>  needs to (I'm thinking about libcontainer as it's currently doing
>  for Rlimits for instance).
> > I'm fine with blockIODevice.  One benefit is that we don't need to
> > specify ordering between mounting, mknod calls, and these resource
> > limits, because we can set major:minor limits before the path exists.

fine, I guess `/proc/partitions` or something could be used to map device numbers back to path but shouldn't be a problem now

runtime-config-linux.md
runcom: Fixed

wking:
89bed76 looks good to me, although I'm a bit surprised that all we
have to update are the Go types and examples.  Can we add a longer
example linking to [1] and sketching out the intended mapping between
our docs and that interface?  A list of options for each type with
their intended type, the fact that they're required, and a one-line
summary of their interpretation would help me anyway.  For example:

  * *major* (int64, required) [number][devices] for the devices that will be limited.
  * *minor* (int64, required) [number][devices] for the devices that will be limited.
  * *weight* (int16, required) in [CFQ IO scheduling][cfq]

  [devices]: https://www.kernel.org/doc/Documentation/devices.txt
  [cfq]: https://www.kernel.org/doc/Documentation/block/cfq-iosched.txt

where I've tried to use the “whole thing reads like a sentence”
phrasing [2].

[1]: https://www.kernel.org/doc/Documentation/cgroups/blkio-controller.txt
[2]: https://github.com/opencontainers/specs/pull/145#issuecomment-138971339

runcom:
> 89bed76 looks good to me, although I'm a bit surprised that all we
> have to update are the Go types and examples.  Can we add a longer
> example linking to [1] and sketching out the intended mapping between
> our docs and that interface?  A list of options for each type with
> their intended type, the fact that they're required, and a one-line
> summary of their interpretation would help me anyway.  For example:
> 
>   * *major* (int64, required) [number][devices] for the devices that will be limited.
>   * *minor* (int64, required) [number][devices] for the devices that will be limited.
>   * *weight* (int16, required) in [CFQ IO scheduling][cfq]
> 
>   [devices]: https://www.kernel.org/doc/Documentation/devices.txt
>   [cfq]: https://www.kernel.org/doc/Documentation/block/cfq-iosched.txt
> 
> where I've tried to use the “whole thing reads like a sentence”
> phrasing [2].
> 
> [1]: https://www.kernel.org/doc/Documentation/cgroups/blkio-controller.txt
> [2]: https://github.com/opencontainers/specs/pull/145#issuecomment-138971339

sounds good to have these docs around, I'll update

wking:
On Wed, Sep 16, 2015 at 12:51:13PM -0700, W. Trevor King wrote:
>   [cfq]: https://www.kernel.org/doc/Documentation/block/cfq-iosched.txt
> …
> [1]: https://www.kernel.org/doc/Documentation/cgroups/blkio-controller.txt

These also talk about leaf_weight.  While we're formalizing these
entries, do we want to expose that too?

Maybe with:

  * *weight* (int16, optional) when competing within this cgroup in
     [CFQ IO scheduling][cfq]
  * *leafWeight* (int16, optional) when competing with cgroup children
     in [CFQ IO scheduling][cfq]

  You must specify at least one of `weight` or `leafWeight` in a given
  entry, and can specify both.

Or we can have separate entries (like you're currently doing for
weight and throttle) for handling weight and leaf_weight.  Or we can
punt for now, and come back and look at leaf_weight when someone asks
for it with a particular use-case in mind.

wking:
On Wed, Sep 16, 2015 at 12:56:08PM -0700, Antonio Murdaca wrote:
> sounds good to have these docs around, I'll update

I had some other cgroups docs stubbing in #99 if you want to borrow
from there (although there hasn't been discussion about explicitly
handling device-access cgroups, so probably leave that section out ;).

runcom:
> Or we can have separate entries (like you're currently doing for
> weight and throttle) for handling weight and leaf_weight.

these are separate cause one is dealing with *weight* and the other with *rate*

> Or we can
> punt for now, and come back and look at leaf_weight when someone asks
> for it with a particular use-case in mind.

It shoudln't be so much hard to put leaf_weight into account. And I like your proposed struct with both weight and leaf_weight.

> I had some other cgroups docs stubbing in #99 if you want to borrow
> from there (although there hasn't been discussion about explicitly
> handling device-access cgroups, so probably leave that section out ;).

Thanks!

hqhq:
> I'm not fond of blockIODevice embedded struct and specifying major:minor for this. I think it will suffice specifying just a Path and do the syscall.Stat to retrieve major:minor where it needs to (I'm thinking about libcontainer as it's currently doing for Rlimits for instance).

Though systemd takes device path instead of major minor number, but for struct, I think is't better to store just major minor numbers. 

And +1 for this change.

runcom:
> Though systemd takes device path instead of major minor number, but for struct, I think is't better to store just major minor numbers.

systemd was one of the reason I thought having `Path` would have been good. But whatever, I agree having device numbers is better.

runcom:
ping @vbatts @mrunalp @LK4D4 @wking @hqhq 

I've updated the structure of the cgroups section here by splitting the example into per resource section and filled the bllkio one as per current PR topic. If you're happy with this path/format I can do a follow up PR updating the rest of the resources (I'd leave this PR as is now w/o going any further with other modifications if you're ok with this)

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+	"disableOOMKiller": false
 ```
 
+#### Memory
+
+```json
+	"memory": {
+		"limit": 0,
```
wking: I think our other examples use four-space indents for JSON, so it's probably better to stick with that and skip the tabs.

runcom: Sure, missed that

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+			{
+				"major": 8,
+				"minor": 16,
+				"rate": 300
+			}
+		]
+	}
+```
+
+* **blkioWeight** (uint16, optional) - specifies per cgroup weight, this is default weight of the group on all the devices until and unless overridden by per device rule, range is from 10 to 1000
```
wking: I prefer putting the normative attribute listing and semantic descriptions (i.e. the list starting with the line I'm commenting on) before the informative example section.  See also #197.

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+		],
+		"blkioThrottleReadBpsDevice": [
+			{
+				"major": 8,
+				"minor": 0,
+				"rate": 600
+			}
+		],
+		"blkioThrottleWriteBpsDevice": null,
+		"blkioThrottleReadIOPSDevice": null,
```
wking: I'd  rather leave these out of the example instead of explicitly declaring them `null`.  You document them in the list of available attributes, and give other examples of the rate-setting structure.

runcom: thought the same, I'll remove them

runtime-config-linux.md
wking: And probably:

* “per cgroup” → “per-cgroup”
* “, this is default” → “.  This is the default”
* “all the devices” → “all devices”
* “per device rule” → “per-device rules”.
* “, range is from” → “.  The range is from”

I'd consider leaving the range description out (and just expecting folks to get it from the kernel docs) so we don't go stale in the unlikely event that the kernel changes the range.

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+			}
+		]
+	}
+```
+
+* **blkioWeight** (uint16, optional) - specifies per cgroup weight, this is default weight of the group on all the devices until and unless overridden by per device rule, range is from 10 to 1000
+
+* **blkioLeafWeight** (uint16, optional) - equivalents of `blkioWeight` for the purpose of deciding how much weight tasks in the given cgroup has while competing with the cgroup's child cgroups, range is from 10 to 1000
+
+* **blkioWeightDevice** (array, optional) - specifies the list of devices on which bandwidth rate limit will be applied. Next parameters can be specified per device:
```
wking: “Next parameters can be specified per device” → “The following parameters can be specified per-device”

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+	* **major** (int64, required) - major number of the device that will be limited
+	* **minor** (int64, required) - minor number of the device that will be limited
```
wking: Maybe worth recycling the **major, minor** combination you touch in the devices docs above?

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+	* **weight** (uint16, optional) - bandwidth rate for the device, range is from 10 to 1000
+	* **leafWeight** (uint16, optional) - bandwidth rate for the device while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only
+
+	You must specify at least one of `weight` or `leafWeight` in a given entry, and can specify both.
+
+* **blkioThrottleReadBpsDevice**, **blkioThrottleWriteBpsDevice**, **blkioThrottleReadIOPSDevice**, **blkioThrottleWriteIOPSDevice** (array, optional) - specify the list of devices on which IO rate limit will be applied. Next parameters can be specified per device:
```
wking: “on which IO rate limit will be applied” → “which will be I/O rate limited”.  Or something like that.  And I'd suggest [the same following-parameter phrasing][1].

I don't have strong feelings about I/O vs IO, but [Wikipedia seems to prefer the former][2].  The kernel docs go both ways, but with v4.1:

    $ git grep ' IO ' Documentation/ | wc -l
    331
    $ git grep ' I/O ' Documentation/ | wc -l
    1067

[1]: https://github.com/opencontainers/specs/pull/193/files#r39772244
[2]: https://en.wikipedia.org/wiki/Input/output

wking:
On Thu, Sep 17, 2015 at 05:55:38AM -0700, Antonio Murdaca wrote:
> I've updated the structure of the cgroups section here by splitting
> the example into per resource section and filled the bllkio one as
> per current PR topic.

Looks good to me :).  I've left a few copy-edit suggestions, but I'm
on board with all the significant changes in a89565b2.  I'd be fine if
it landed as-is, with follow up PRs to accept/reject my copy-edit
suggestions (but I'd rather they got rerolled in or rejected here ;).

crosbymichael:
LGTM

runtime-config-linux.md
```diff
@@ -146,45 +146,128 @@ The cgroups will be created if they don't exist.
...
+* **`blkioLeafWeight`** *(uint16, optional)* - equivalents of `blkioWeight` for the purpose of deciding how much weight tasks in the given cgroup has while competing with the cgroup's child cgroups. The range is from 10 to 1000.
+
+* **`blkioWeightDevice`** *(array, optional)* - specifies the list of devices which will be bandwidth rate limited. The following parameters can be specified per-device:
+    * **`major, minor`** *(int64, required)* - major, minor numbers for device. More info in `man mknod`.
+    * **`weight`** *(uint16, optional)* - bandwidth rate for the device, range is from 10 to 1000.
+    * **`leafWeight`** *(uint16, optional)* - bandwidth rate for the device while competing with the cgroup's child cgroups, range is from 10 to 1000, cfq scheduler only.
+
+    You must specify at least one of `weight` or `leafWeight` in a given entry, and can specify both.
+
+* **`blkioThrottleReadBpsDevice`**, **`blkioThrottleWriteBpsDevice`**, **`blkioThrottleReadIOPSDevicei`**, **`blkioThrottleWriteIOPSDevice`** *(array, optional)* - specify the list of devices which will be IO rate limited. The following parameters can be specified per-device:
```
mrunalp: typo: blkioThrottleReadIOPSDevicei

runcom: fixed

mrunalp:
LGTM

runcom:
Thanks, I'll create a PR in runc to update libcontainer

mrunalp:
@runcom Sounds good!

PR #198, opened 2015-09-17
Change HugepageLimit.Limit type to uint64

runcom:
As in kernel

### DIFF
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -92,8 +92,10 @@ type Rlimit struct {
 
 // HugepageLimit structure corresponds to limiting kernel hugepages
 type HugepageLimit struct {
+	// Pagesize is the hugepage size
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Sep 17, 2015 at 12:19:44PM -0700, Antonio Murdaca wrote:
> As in kernel

I didn't see anything about uint64 in a quick skim over [1], but it's
pretty clear from [2] (I think that's the right setting ;), so looks
good to me.

[1]: https://www.kernel.org/doc/Documentation/cgroups/hugetlb.txt
[2]: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/mm/memcontrol.c?id=v4.2#n4330

crosbymichael:
LGTM

vbatts:
LGTM

PR #199, opened 2015-09-23
runtime: config: linux: add cgroups informations

runcom:
- add information to cgroup resources controllers with examples
- add pids cgroup information and example
- reflect kernel types

I'm submitting this as a first iteration of specifying structures here in the specs so any help/ideas are well accepted for subsequent PRs

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -6,10 +6,23 @@ A namespace wraps a global system resource in an abstraction that makes it appea
 Changes to the global resource are visible to other processes that are members of the namespace, but are invisible to other processes.
 For more information, see [the man page](http://man7.org/linux/man-pages/man7/namespaces.7.html).
 
-Namespaces are specified in the spec as an array of entries.
-Each entry has a type field with possible values described below and an optional path element.
+Namespaces are specified as an array of entries inside the `namespaces` root field.
+The following parameters can be specified to setup namespaces:
+
+* **`type`** *(string, required)* - namespace type. The following namespaces types are supported:
+    * **`pid`** processes inside the container will only be able to see other processes inside the same container
+    * **`network`** the container will have its own network stack
+    * **`mount`** the container will have an isolated mount table
+    * **`ipc`** processes inside the container will only be able to communicate to other processes inside the same container via system level IPC
+    * **`uts`** the container will be able to have its own hostname and domain name
+    * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container
+
+* **`path`** *(string, optional)* - path to namespace file
+
 If a path is specified, that particular file is used to join that type of namespace.
 Also, when a path is specified, a runtime MUST assume that the setup for that particular namespace has already been done and error out if the config specifies anything else related to that namespace.
+
+###### Example
 
 ```json
     "namespaces": [
@@ -36,32 +49,29 @@ Also, when a path is specified, a runtime MUST assume that the setup for that pa
     ]
 ```
 
-#### Namespace types
-
-* **`pid`** processes inside the container will only be able to see other processes inside the same container.
-* **`network`** the container will have its own network stack.
-* **`mount`** the container will have an isolated mount table.
-* **`ipc`** processes inside the container will only be able to communicate to other processes inside the same
-container via system level IPC.
-* **`uts`** the container will be able to have its own hostname and domain name.
-* **`user`** the container will be able to remap user and group IDs from the host to local users and groups
-within the container.
-
 ## Devices
 
-Devices is an array specifying the list of devices to be created in the container.
-Next parameters can be specified:
+`devices` is an array specifying the list of devices to be created in the container.
 
-* **`type`** - type of device: `c`, `b`, `u` or `p`. More info in `man mknod`
-* **`path`** - full path to device inside container
-* **`major, minor`** - major, minor numbers for device. More info in `man mknod`.
-                 There is special value: `-1`, which means `*` for `device`
-                 cgroup setup.
-* **`permissions`** - cgroup permissions for device. A composition of `r`
-                (read), `w` (write), and `m` (mknod).
-* **`fileMode`** - file mode for device file
-* **`uid`** - uid of device owner
-* **`gid`** - gid of device owner
+The following parameters can be specified:
+
+* **`type`** *(char, required)* - type of device: `c`, `b`, `u` or `p`. More info in `man mknod`.
+
+* **`path`** *(string, optional)* - full path to device inside container
+
+* **`major, minor`** *(int64, required)* - major, minor numbers for device. More info in `man mknod`. There is a special value: `-1`, which means `*` for `device` cgroup setup.
+
+* **`permissions`** *(string, optional)* - cgroup permissions for device. A composition of `r` (*read*), `w` (*write*), and `m` (*mknod*).
+
+* **`fileMode`** *(uint32, optional)* - file mode for device file
+
+* **`uid`** *(uint32, optional)* - uid of device owner
+
+* **`gid`** *(uint32, optional)* - gid of device owner
+
+**`fileMode`**, **`uid`** and **`gid`** are required if **`path`** is given and are otherwise not allowed.
+
+###### Example
 
 ```json
    "devices": [
@@ -154,11 +164,38 @@ For example, to run a new process in an existing container without updating limi
 
 #### Disable out-of-memory killer
 
+`disableOOMKiller` contains a boolean (`true` or `false`) that enables or disables the Out of Memory killer for a cgroup.
+If enabled (`false`), tasks that attempt to consume more memory than they are allowed are immediately killed by the OOM killer.
+The OOM killer is enabled by default in every cgroup using the `memory` subsystem.
+To disable it, specify a value of `true`.
+For more information, see [the memory cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/memory.txt).
+
+* **`disableOOMKiller`** *(bool, optional)* - enables or disables the OOM killer
+
+###### Example
+
 ```json
     "disableOOMKiller": false
 ```
 
 #### Memory
+
+`memory` represents the cgroup subsystem `memory` and it's used to set limits on the container's memory usage.
+For more information, see [the memory cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/memory.txt).
+
+The following parameters can be specified to setup the controller:
+
+* **`limit`** *(uint64, optional)* - sets limit of memory usage
+
+* **`reservation`** *(uint64, optional)* - sets soft limit of memory usage
+
+* **`swap`** *(uint64, optional)* - sets limit of memory+Swap usage
+
+* **`kernel`** *(uint64, optional)* - sets hard limit for kernel memory
+
+* **`swappiness`** *(uint64, optional)* - sets swappiness parameter of vmscan (See sysctl's vm.swappiness)
+
+###### Example
 
 ```json
     "memory": {
@@ -171,6 +208,27 @@ For example, to run a new process in an existing container without updating limi
 ```
 
 #### CPU
+
+`cpu` represents the cgroup subsystems `cpu` and `cpusets`.
+For more information, see [the cpusets cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/cpusets.txt).
+
+The following parameters can be specified to setup the controller:
+
+* **`shares`** *(uint64, optional)* - specifies a relative share of CPU time available to the tasks in a cgroup
+
+* **`quota`** *(uint64, optional)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+
+* **`period`** *(uint64, optional)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only)
+
+* **`realtimeRuntime`** *(uint64, optional)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources
+
+* **`realtimePeriod`** *(uint64, optional)* - same as **`period`** but applies to realtime scheduler only
+
+* **`cpus`** *(cpus, optional)* - list of CPUs the container will run in
+
+* **`mems`** *(mems, optional)* - list of Memory Nodes the container will run in
+
+###### Example
 
 ```json
     "cpu": {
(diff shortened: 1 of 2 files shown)


runtime-config-linux.md
```diff
@@ -4,11 +4,24 @@ A namespace wraps a global system resource in an abstraction that makes it appea
...
+Namespaces are specified as an array of entries inside the `namespaces` root field.
+The following parameters can be specified to setup namespaces:
+
+* **`type`** *(string, required)* - namespace type. The following namespaces types are supported:
+    * **`pid`** processes inside the container will only be able to see other processes inside the same container
+    * **`network`** the container will have its own network stack
+    * **`mount`** the container will have an isolated mount table
+    * **`ipc`** processes inside the container will only be able to communicate to other processes inside the same container via system level IPC
+    * **`uts`** the container will be able to have its own hostname and domain name
+    * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container
```
wking: I'd consider just echoing the “isolates” column of namespaces(7) for these entries.  For example:

> * **ipc** isolates System V IPC, POSIX message queues

wking: Ah, I see this is cut/pasted/reformatted from below.  In that case, it's probably best to leave the text as it stands ;).

runtime-config-linux.md
```diff
@@ -34,32 +47,27 @@ Also, when a path is specified, a runtime MUST assume that the setup for that pa
...
-* **fileMode** - file mode for device file
-* **uid** - uid of device owner
-* **gid** - gid of device owner
+* **`permissions`** *(string, optional)* - cgroup permissions for device. A composition of `r` (*read*), `w` (*write*), and `m` (*mknod*).
+
+* **`fileMode`** *(uint32, optional)* - file mode for device file
+
+* **`uid`** *(uint32, optional)* - uid of device owner
+
+* **`gid`** *(uint32, optional)* - gid of device owner
```
wking: These docs don't actually match my understanding of the intended semantics (see #101.  For example, I think `fileMode` is required iff `path` is given), so I'd be cautious about adding `optional` and `required`.  Adding type information and reformatting should be fine though.

runtime-config-linux.md
```diff
@@ -152,12 +160,38 @@ For example, to run a new process in an existing container without updating limi
 
 #### Disable out-of-memory killer
 
+`disableOOMKiller` contains a flag (`true` or `false`) that enables or disables
+the Out of Memory killer for a cgroup. If enabled (`false`), tasks that attempt
+to consume more memory than they are allowed are immediately killed by the OOM killer.
```
wking: This sentence [should be on one line][1].  And I'd recommend “flag” → “Boolean”.

[1]: https://github.com/opencontainers/specs/tree/v0.1.1#markdown-style

runtime-config-linux.md
runcom: I'll fix these inconsistencies by documenting them as per #101 rather than not adding required|optional, sounds good?

runtime-config-linux.md
wking: And since we have a default (`false`), I'd suggest leading in with the usual pattern:

> **`disableOOMKiller`** (bool, optional) contains…

Or whatever we decide our pattern for introducing JSON fields is going to be.

runtime-config-linux.md
```diff
@@ -152,12 +160,38 @@ For example, to run a new process in an existing container without updating limi
...
+
+###### Example
+
 ```json
     "disableOOMKiller": false
 ```
 
 #### Memory
 
+`memory` represents the cgroup subsystem `memory` and it's used to set limits on memory use of the container.
```
wking: Maybe end with “… limits on the container's memory use.”?

wking: And since all the fields inside are optional, I'd like an explicit lead-in like:

> **`memory`** (object, optional) represents…

runtime-config-linux.md
```diff
@@ -152,12 +160,38 @@ For example, to run a new process in an existing container without updating limi
...
+For more information, see [the memory cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/memory.txt).
+
+The following parameters can be specified to setup the memory controller:
+
+* **`limit`** *(uint64, optional)* - set limit of memory usage
```
wking: [Sentence phrasing][1] would be something like:

> **`limit`** *(uint64, optional)* sets the limit of memory usage.

[1]: https://github.com/opencontainers/specs/pull/145#issuecomment-138971339

runtime-config-linux.md
```diff
@@ -170,6 +204,27 @@ For example, to run a new process in an existing container without updating limi
...
+`cpu` represents the cgroup subsystems `cpu` and `cpusets`.
+For more information, see [the cpusets cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/cpusets.txt).
+
+The following parameters can be specified to setup the cpu controller:
+
+* **`shares`** *(uint64, optional)* - specifies a relative share of CPU time available to the tasks in a cgroup
+
+* **`quota`** *(uint64, optional)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+
+* **`period`** *(uint64, optional)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (cfs scheduler only)
```
wking: The [CFS docs][1] use caps (“CFS”, not “cfs”).  And I think we should link those scheduler docs here, so maybe:

> … ([CFS scheduler][CFS] only).
>
> [CFS]: https://www.kernel.org/doc/Documentation/scheduler/sched-design-CFS.txt

[1]: https://www.kernel.org/doc/Documentation/scheduler/sched-design-CFS.txt

runtime-config-linux.md
```diff
@@ -170,6 +204,27 @@ For example, to run a new process in an existing container without updating limi
...
+
+* **`realtimeRuntime`** *(uint64, optional)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources
+
+* **`realtimePeriod`** *(uint64, optional)* - same as **`period`** but applies on realtime scheduler only
```
wking: “applies on” → “applies to”.  And I'd like to link upstream docs where possible.  In this case, maybe that's https://www.kernel.org/doc/Documentation/scheduler/sched-rt-group.txt?

runtime-config-linux.md
```diff
@@ -242,6 +297,18 @@ The following parameters can be specified to setup the block io controller:
...
 
+`hugepageLimits` represents the `hugetlb` controller which allows to limit the
+HugeTLB usage per control group and enforces the controller limit during page fault.
+For more information, see the [kernel cgroups documentation about HugeTLB](https://www.kernel.org/doc/Documentation/cgroups/hugetlb.txt).
+
+`hugepageLimits` is an array of entries, each having the following structure:
+
+* **`pageSize`** *(string, required)* - hugepage size
+
+* **`limit`** *(uint64, required)* - limit of *hugepagesize* hugetlb usage
```
wking: I think we want “… limit in bytes of…” here.  And probably best to just use “HugeTLB” in all your text references (you currently have “`hugetbl`”, “HugeTBL”, and this “hugetbl”).

runtime-config-linux.md
```diff
@@ -253,9 +320,23 @@ The following parameters can be specified to setup the block io controller:
 
 #### Network
 
+`network` represents the cgroup subsystems `net_cls` and `net_prio`.
+For more information, see [the net\_cls cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/net_cls.txt) and [the net\_prio cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/net_prio.txt).
+
+The following parameters can be specified to setup these cgroup controllers:
+
+* **`classID`** *(string, optional)* - network class identifier the cgroup's network packets will be tagged with
```
wking: [Sentence phrasing][1] would give something like:

> * **`classID`** *(hex-string, optional)* is the tag attached to the network packets to target filtering, Traffic Controller prioritization, etc.

[1]: https://github.com/opencontainers/specs/pull/145#issuecomment-138971339

runtime-config-linux.md
```diff
@@ -253,9 +320,23 @@ The following parameters can be specified to setup the block io controller:
...
+
+* **`priorities`** *(array, optional)* - specifies a list of map of the priorities assigned to traffic originating from
```
wking: JSON calls maps [“objects”][1], so I'd stick with their terminology here.  Maybe:

> * **`priorities`** *(array of priorities, optional)* contains priority assignments for traffic originating from processes in the cgroup and exiting via various interfaces.
>   Priorities are JSON objects with the following members:
>     * **`name`** …

[1]: http://json.org/

runtime-config-linux.md
```diff
@@ -269,11 +350,31 @@ The following parameters can be specified to setup the block io controller:
    }
 ```
 
+#### Pids
```
wking: ps(1) uses “PID” (not “pid”) in textual references.

runtime-config-linux.md
```diff
@@ -269,11 +350,31 @@ The following parameters can be specified to setup the block io controller:
...
+
+`pids` represents the cgroup subsystem `pids`.
+For more information, see [the pids cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/pids.txt
```
wking: The first sentence doesn't help me very much ;).  And we have lots of “For more information …” sentences that seem wordy for setting up a link (although that pattern pre-dates this PR).  How about:

> **`pids`** *(object, optional)* configures the [process number controller][pids].
>
> [pids]: https://www.kernel.org/doc/Documentation/cgroups/pids.txt

runtime-config-linux.md
```diff
@@ -269,11 +350,31 @@ The following parameters can be specified to setup the block io controller:
...
+).
+
+The following paramters can be specified to setup the pids controller:
```
wking: “the pids controller” reads awkwardly to me.  Since we're in the PIDs section, how about just “the controller”?

runtime-config-linux.md
wking: On Wed, Sep 23, 2015 at 09:57:22AM -0700, Antonio Murdaca wrote:
> I'll fix these inconsistencies by documenting them as per #101
> rather than not adding required|optional, sounds good?

It's probably best to leave the #101 docs in that PR to avoid stalling
this one for the same reasons that have stalled #101 (I'm not sure
what those reasons are).  So I'd recommend making as few changes to
the existing text in this section as possible to avoid controversy.

wking:
I left a number of minor copy-edit suggestion, but overall this looks
great to me :).  I'm especially fond of the linkable examples
(e.g. [1]).

On Wed, Sep 23, 2015 at 06:01:17AM -0700, Antonio Murdaca wrote:
> - reflect kernel types

This would be good to get right ;).  I haven't dug through the kernel
docs/code to double-check your changes (which are mostly int64 →
uint64), but they all sound reasonable to me.  I'm suspicious about
int64 → int for Pids.Limit.  Does the kernel really use an
implementation-specific size there?

[1]: https://github.com/runcom/specs/blob/7ca9693616c570da659e0512831ee3c5896f1907/runtime-config-linux.md#example-1

runcom:
> I'm suspicious about
> int64 → int for Pids.Limit.  Does the kernel really use an
> implementation-specific size there?

it's wrong, mistake

runcom:
ping @vbatts @crosbymichael @mrunalp

runtime-config-linux.md
```diff
@@ -154,12 +164,40 @@ For example, to run a new process in an existing container without updating limi
...
+
+###### Example
+
 ```json
     "disableOOMKiller": false
 ```
 
 #### Memory
 
+`memory` represents the cgroup subsystem `memory` and it's used to set limits on the container's memory use.
```
mrunalp: s/use/usage ?

runcom: fixed

mrunalp:
LGTM

mrunalp:
@vishh you want to take a look?

runtime-config-linux.md
```diff
@@ -36,32 +49,29 @@ Also, when a path is specified, a runtime MUST assume that the setup for that pa
...
-* **`fileMode`** - file mode for device file
-* **`uid`** - uid of device owner
-* **`gid`** - gid of device owner
+* **`fileMode`** *(uint32, optional)* - file mode for device file
+
+* **`uid`** *(uint32, optional)* - uid of device owner
+
+* **`gid`** *(uint32, optional)* - gid of device owner
+
+**`fileMode`**, **`uid`** and **`gid`** are required if **`path`** is given, otherwise not allowed
```
wking: This line needs at least a trailing period, and maybe an end like “and are otherwise not allowed.”

runtime-config-linux.md
```diff
@@ -154,12 +164,40 @@ For example, to run a new process in an existing container without updating limi
 
 #### Disable out-of-memory killer
 
+`disableOOMKiller` contains a boolean (`true` or `false`) that enables or disables
+the Out of Memory killer for a cgroup.
```
wking: This sentence needs to be collapsed to a single line.

runtime-config-linux.md
```diff
@@ -255,9 +326,23 @@ The following parameters can be specified to setup the block io controller:
...
 
+`network` represents the cgroup subsystems `net_cls` and `net_prio`.
+For more information, see [the net\_cls cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/net_cls.txt) and [the net\_prio cgroup man page](https://www.kernel.org/doc/Documentation/cgroups/net_prio.txt).
+
+The following parameters can be specified to setup these cgroup controllers:
+
+* **`classID`** *(string, optional)* - is the network class identifier the cgroup's network packets will be tagged with
+
+* **`priorities`** *(array, optional)* - specifies a list of objects of the priorities assigned to traffic originating from
+processes in the group and egressing the system on various interfaces. The following parameters can be specified per-priority:
```
wking: These two sentences need to be one-per-line.

wking:
There are a few ‘cfq’ → ‘CFS’ changes, but those aren't the same thing
[1,2].  I expect block limits should really be CFQ, while CPU limits
should be CFS.

[1]: https://www.kernel.org/doc/Documentation/block/cfq-iosched.txt
[2]: https://www.kernel.org/doc/Documentation/scheduler/sched-design-CFS.txt

mrunalp:
Yeah, for block it should be CFQ.

runcom:
> Yeah, for block it should be CFQ.

sorry I mixed both valules, will fix

runcom:
Updated PTAL

wking:
2ce2c86 looks good to me.

runcom:
is there anything missing here?

vbatts:
LGTM

runtime_config_linux.go
```diff
@@ -151,29 +151,29 @@ type BlockIO struct {
...
 type Memory struct {
 	// Memory limit (in bytes)
-	Limit int64 `json:"limit"`
+	Limit uint64 `json:"limit"`
 	// Memory reservation or soft_limit (in bytes)
-	Reservation int64 `json:"reservation"`
+	Reservation uint64 `json:"reservation"`
 	// Total memory usage (memory + swap); set `-1' to disable swap
-	Swap int64 `json:"swap"`
+	Swap uint64 `json:"swap"`
```
vishh: Why uint64 here? How can we set it to `-1`? I'd much rather retain int64 since negative values have special meaning in the kernel for some of these fields.

vishh: I will send a PR to clean this up.

runcom: @vishh Pretty sure I went thou the kernel source and checked this can't be negative, but maybe I made some mistakes checking, thanks

runcom: definitely made a mistake based on the comment :)

Issue #15, opened 2015-06-25
Version compatibility with an evolving spec


wking:
The current docs [say][1]:

> If the container is compliant with multiple versions, it SHOULD advertise the most recent known version to be supported.

That's going to make backward compatibility with old tools hard.  I'd rather take major and minor numbers from [semantic versioning][2], so a v1.3 config would be compatible with a v1.3 launcher, or a v1.4 launcher, etc., etc., but not a v1.2 launcher or a v2.0 launcher.  That gives you some granularity for specifying which of several compatible features you need (e.g. UID-mapping for #10  was only added in v1.3, but v1.3 launchers can still handle the v1.0 containers).

One limitation to the sem-ver appoach is that feature addition needs to be serialized.  If you have to orthogonal extensions, and A lands in v1.3 while B lands in v1.4, implementations that want to support feature B also need to support feature A.  For example, Notmuch uses [feature tags][3] to mark supported optional features (which is what sem-ver's minor releases are for).

So depending on how flexible you want to make life for spec implementors, I'd recommend chosing either sem-ver (simple, some cross compatibility, serialized features) or feature tags (more complicated, lots of flexibility, parallel features).  Having a single integer that spec authors are supposed to bump (the current recommendation here) gives you essentially zero cross-compatibility without building the list of changing features into the implementation itself.

[1]: https://github.com/opencontainers/specs/blob/master/config.md#manifest-version
[2]:  http://semver.org/
[3]: http://git.notmuchmail.org/git/notmuch/blob/d22df9503b0fcf06b41b8b6bcb75207b549f08d7:/NEWS#l197

philips:
For now I am OK with a linearized release process. If we find a need later we can re-evaluate.

wking:
On Thu, Jun 25, 2015 at 09:52:13AM -0700, Brandon Philips wrote:
> For now I am OK with a linearized release process.

Sounds good.  Can we use sem-ver then to provide cross-compatibility
between a given config and implementations at different minor-versions
of the spec?

jonboulle:
+1, specifically [SemVer 2.0.0](http://semver.org/spec/v2.0.0.html)

crosbymichael:
+1

philips:
Fixed via https://github.com/opencontainers/specs/commit/9c90ccdb93ec820070ce2e039ee342147f9994e4

wking:
> For now I am OK with a linearized release process. If we find a need later we can re-evaluate.

[...]

PR #201, opened 2015-09-23
runtime-config-linux: Require libseccomp constants

wking:
Don't punt readers to libseccomp, because this spec is supposed to be
[a self-sufficient contract between the bundle-author and the
runtime][1].  The bundle-author shouldn't care if the runtime uses
libseccomp, or which version of libseccomp it uses.  But they can rely
on the listed constants being supported (and we may want to bring over
brief descriptions to go along with the constants themselves).

Runtime authors, on the other hand, may want to know that v2.2.3 of
libseccomp provides all the required constants.  I've added links to
the libseccomp 2.2.3 header from which these constants were extracted,
and that should be enough of a hint for bundle authors to use a
compatible libseccomp version (if they want to use libseccomp at all).

Spun off from #200, which has since landed.

[1]: https://github.com/opencontainers/specs/pull/200/files#r40257217

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -319,10 +319,9 @@ For more information about Apparmor, see [Apparmor documentation](https://wiki.u
 Seccomp provides application sandboxing mechanism in the Linux kernel.
 Seccomp configuration allows one to configure actions to take for matched syscalls and furthermore also allows matching on values passed as arguments to syscalls.
 For more information about Seccomp, see [Seccomp kernel documentation](https://www.kernel.org/doc/Documentation/prctl/seccomp_filter.txt)
-The actions, architectures, and operators are strings that match the definitions in seccomp.h from [libseccomp](https://github.com/seccomp/libseccomp) and are translated to corresponding values.
-A valid list of constants as of Libseccomp v2.2.3 is contained below.
+The set of valid constants MUST include the following (extracted from [libseccomp][]):
 
-Architecture Constants
+[Architecture Constants][libseccomp-architecture]:
 * `SCMP_ARCH_X86`
 * `SCMP_ARCH_X86_64`
 * `SCMP_ARCH_X32`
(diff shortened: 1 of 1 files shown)


mheon:
I don't know if we should make the set of valid constants listed here a hard requirement for implementation (mostly with regards to architectures). I can see some implementations of the spec only supporting x86-64, and thus would never require the MIPS architecture constants, for example.

Perhaps there should be a requirement that all constants be recognized, but a note that there is no guarantee an implementation supports all architectures?

wking:
On Wed, Sep 23, 2015 at 02:36:36PM -0700, Matthew Heon wrote:
> Perhaps there should be a requirement that all constants be
> recognized, but a note that there is no guarantee an implementation
> supports all architectures?

That sounds reasonable.  What about actions?  I see runC doesn't
support trace [1].  On the other hand, that means that there will be
an [out-of-spec contract between the runtime and bundle author][2],
and I don't think we have good tooling now to figure out if a
particular bundle is compatible with a particular runtime (short of
passing the config to the runtime and checking for errors ;).  This
sort of optional-feature stuff is what I floated a more flexible
feature-flags scheme in #15, and if we want to commit to an
optional-spec-support path, I'd like to see us using something like
that to explicitly tag the optional feature blocks.

[1]: https://github.com/opencontainers/runc/blob/v0.0.4/libcontainer/seccomp/config.go#L36
[2]: https://github.com/opencontainers/specs/pull/200#discussion_r40257217

mheon:
On actions: I think it's reasonable to make implementing all of them mandatory. They currently encompass the entirety of what the Linux kernel can do on a Seccomp rule match, and I don't believe there are plans to add more any time soon, so this should remain a fairly stable part of the API. As you mentioned, this would require that runC also implement trace, but that should be simple to do.

Seccomp as a whole would be a good candidate for being an optional feature, if a set of feature flags were to emerge. We already have runC builds with no Seccomp integration present, and not all are going to link against the latest libseccomp (which mostly limits the architectures which can be used with the library).

wking:
Looking back at #200, I realize now that libseccomp architecture is
not going to be a bundle-author decision, and will likely be filled in
by the runtime-caller before invoking the runtime.  Maybe this should
be a more generic “kernel architecture” field (and not
seccomp-specific?).  Then we don't have to worry about requiring
support, since the runtime is clearly independent of the kernel
architecture (and matching bundles with compatible kernels is the job
of some out-of-scope external tooling, see #73).

crosbymichael:
NOTLGTM

We were talking about defining our own const and this goes the opposite way and introduces wording like **MUST** that I don't think should be there.

mrunalp:
Agree.

PR #204, opened 2015-09-29
convert **name** to **`name`**

laijs:
convert ```**name**``` to ```**`name`**```

we have both styles in the `*.md`, we should use only one of the styles.
And **`name`** is much prettier than **name** in the result,
so we use **`name`**


vbatts:
heh *pretty* :art: :smile_cat: 

LGTM

mrunalp:
LGTM

Issue #115, opened 2015-08-26
Setting oom_score_adj while launching applications


wking:
This spins off a more tightly-scoped version of #114.  The [`oom_score_adj`][1] is more of a host-side and/or multi-container-orchestration issue, and less of a bundle issue, which means it probably should be in `runtime.json` if/once #88 lands.  Possible approaches for setting this include:

1. The runtime writes to `/proc/<pid>/oom_score_adj`, which would need a config-side setting.
2. The host injects a pre-start hook to write to `/proc/<pid>/oom_score_adj`.  This would require a hook with sufficient permissions for the write.
3. The application has a startup phase where it handles this sort of thing before execing the main process.  This would require an application have sufficient permissions for the write, although it could drop them after writing and before execing the “real” application.

A number of attributes where you could use (2) currently have explicit, (1)-style configs for via hooks (e.g. setting up networking and creating cgroups and namespaces).  I'd guess the balance involves “how easy is it to handle without (1)?” and “how frequently will folks be tweaking this attribute?”, with high-cost or high-frequency attributes being handled via (1).  So which way do we think makes the most sense for this particular setting?

Is it easy to handle via (2) or (3)?  It seems like (2) would be easy assuming sufficient hook permissions, but (3) is probably too annoying to be worth the trouble.

How frequently do we expect folks will use this?  I can't weigh in here, since I haven't set this.  And I expect most runtime managers that set this will be doing it automatically, so in that case it's a wash between (1) and (2) for difficulty.

If those assumptions are correct, then I think we should go with (2), since that is the least work on the spec/runtime-implementation side.  If nobody chimes in with anti-(2) thoughts in the next few days, I'll merge opencontainers/runc#160 locally see whether I can get it working ;).

[1]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt

wking:
I merged opencontainers/runc#160 (at opencontainers/runc@4f7ff04) into runc's master (at opencontainers/runc@0f85e4e), reverted opencontainers/runc@cc232c47 (part of opencontainers/runc#232) to avoid clobbering my adjustment, wrote a host-side hook to handle the adjustment:

    $ cat /tmp/oom-score-adj.sh
    #!/bin/busybox sh
    ADJ="$1"
    #PID=$(jq -r .pid)
    PID=$(jq -r .init_process_pid)
    echo "${ADJ}" > "/proc/${PID}/oom_score_adj"

[...]

PR #207, opened 2015-09-30
runtime: Add more steps to the lifecycle docs

wking:
And explain how the handling differs when an application joins an
existing container vs. creating a new container.  This is spun off
from the [mailing list thread][1] now that that's settled down.

This write-up is just about Linux containers, since I don't understand
the other systems well enough to do them justice.  I'm happy to roll
in notes for other platforms if folks know what to write ;).

I've pulled the hook-environment change I made in [this post][2] out
into a separate commit, and I'm happy to push that as a separate PR if
it makes review easier.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/frUXLljXy8Y
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/frUXLljXy8Y/_sg-BidnAwAJ

### DIFF
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -18,11 +18,11 @@ Hook paths are absolute and are executed from the host's filesystem.
 
 ### Pre-start
 
-The pre-start hooks are called after the container process is spawned, but before the user supplied command is executed.
+The pre-start hooks are called [after the container process is spawned, but before the user supplied command is executed](runtime.md#typical-lifecycle).
 They are called after the container namespaces are created on Linux, so they provide an opportunity to customize the container.
 In Linux, for e.g., the network namespace could be configured in this hook.
 
-If a hook returns a non-zero exit code, then an error including the exit code and the stderr is returned to the caller and the container is torn down.
+If a hook returns a non-zero exit code, [then an error including the exit code and the stderr is returned to the caller and the container is torn down](runtime.md#typical-lifecycle).
 
 ### Post-start
 
@@ -33,7 +33,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 
 ### Post-stop
 
-The post-stop hooks are called after the container process is stopped.
+The post-stop hooks are called [after the container process is stopped](runtime.md#typical-lifecycle).
 Cleanup or debugging could be performed in such a hook.
 If a hook returns a non-zero exit code, then an error is logged and the remaining hooks are executed.
 
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -26,7 +26,7 @@ If a hook returns a non-zero exit code, [then an error including the exit code a
 
 ### Post-start
 
-The post-start hooks are called after the user process is started.
+The post-start hooks are called [after the user process is started](runtime.md#typical-lifecycle).
 For example this hook can notify user that real process is spawned.
 
 If a hook returns a non-zero exit code, then an error is logged and the remaining hooks are executed.
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -12,6 +12,7 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
 Hooks allow one to run code before/after various lifecycle events of the container.
 Hooks MUST be called in the listed order.
 The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+All hooks execute in the host environment (e.g. the same namespace, cgroups, etc. that apply to the host process).
 
 Hook paths are absolute and are executed from the host's filesystem.
 
--- a/runtime.md
+++ b/runtime.md
@@ -39,17 +39,18 @@ A typical lifecyle progresses like this:
 4. The runtime executes any [pre-start hooks](runtime-config.md#pre-start)
 5. The runtime [executes the container process](#start-process)
 6. The container process is running
-7. A user tells the runtime to send a termination signal to the container process
-8. The runtime [sends a termination signal to the container process](#stop-process)
-9. The container process exits
-10. The runtime [terminates any other processes in the container](#stop-process)
-11. The runtime executes any [post-stop hooks](runtime-config.md#post-stop)
-12. The runtime [removes the container](#cleanup)
+7. The runtime executes any [post-start hooks](runtime-config.md#post-start)
+8. A user tells the runtime to send a termination signal to the container process
+9. The runtime [sends a termination signal to the container process](#stop-process)
+10. The container process exits
+11. The runtime [terminates any other processes in the container](#stop-process)
+12. The runtime executes any [post-stop hooks](runtime-config.md#post-stop)
+13. The runtime [removes the container](#cleanup)
 
 With steps 7 and 8, the user is explicitly stopping the container process (via the runtime), but it's also possible that the container process could exit for other reasons.
-In that case we skip directly from 6 to [10](#stop-process).
+In that case we skip directly from 6 to [10](#stop-process), skipping any post-start hooks that hadn't been launched and terminating any in-progress post-start hook.
 
-Failure in a pre-start hook or other setup task can cause a jump straight to [11](runtime-config.md#post-stop).
+Failure in a pre-start hook or other setup task can cause a jump straight to [12](runtime-config.md#post-stop).
 
 ### Create
 
--- a/runtime.md
+++ b/runtime.md
@@ -29,24 +29,59 @@ The root directory to the bundle is provided in the state so that consumers can 
 }
 ```
 
-## Lifecycle
+## Typical lifecycle
+
+A typical lifecyle progresses like this:
+
+1. There is no container
+2. A user tells the runtime to start a container and launch a process inside it
+3. The runtime [creates the container](#create)
+4. The runtime executes any [pre-start hooks](runtime-config.md#pre-start)
+5. The runtime [executes the container process](#start-process)
+6. The container process is running
+7. A user tells the runtime to send a termination signal to the container process
+8. The runtime [sends a termination signal to the container process](#stop-process)
+9. The container process exits
+10. The runtime [terminates any other processes in the container](#stop-process)
+11. The runtime executes any [post-stop hooks](runtime-config.md#post-stop)
+12. The runtime [removes the container](#cleanup)
+
+With steps 7 and 8, the user is explicitly stopping the container process (via the runtime), but it's also possible that the container process could exit for other reasons.
+In that case we skip directly from 6 to [10](#stop-process).
+
+Failure in a pre-start hook or other setup task can cause a jump straight to [11](runtime-config.md#post-stop).
 
 ### Create
 
-Creates the container: file system, namespaces, cgroups, capabilities.
+Create the container: file system, namespaces, cgroups, capabilities, etc.
+The invoked process forks, with one branch that stays in the host namespace and another that enters the container.
+The host process carries out all container setup actions, and continues running for the life of the container so it can perform teardown after the container process exits.
+The container process changes users and drops privileges in preparation for the container process start.
+At this point, the host process writes the [`state.json`](#state) file with the host-side version of the container-process's PID (the container process may be in a PID namespace).
 
 ### Start (process)
 
-Runs a process in a container.
-Can be invoked several times.
+After the pre-start hooks complete, the host process signals the container process to execute the runtime.
+The runtime execs the process defined in `config.json`'s [**`process`** attribute](config.md#process-configuration).
+On Linux hosts, some information for this execution may come from outside the `config.json` and `runtime.json` specifications.
+See the [Linux-specific notes for details](runtime-linux.md#file-descriptors).
 
 ### Stop (process)
 
-Not sure we need that from runc cli.
-Process is killed from the outside.
+Send a termination signal to the container process (can optionally send other signals to the container process, e.g. a kill signal).
+When the process exits, the host process collects it's exit status to return as its own exit status.
+If there are any remaining processes in the container's cgroup (and [we only support unified-hierarchies](runtime-config-linux.md#control-groups)), the host process kills and reaps them.
 
-This event needs to be captured by runc to run onstop event handlers.
+### Cleanup
 
-## Hooks
+The host process removes the [`state.json`](#state) file and the container: unmounting file systems, removing namespaces, etc.
+This is the inverse of create.
+The host process then exits with the container processes's exit status.
 
-See [runtime configuration for hooks](./runtime-config.md)
+## Joining existing containers
+
+Joining an existing container looks just like the usual workflow, except that the container process [joins the target container](runtime-config-linux.md#control-groups) at the beginning of step 3.
+It can then, depending on its configuration, continue to create an additional child cgroup underneath the one it joined.
+
+When exiting, the reaping logic in the [stop phase](#stop-process) is the same.
+If the container process created a child cgroup, all other processes in that child cgroup are reaped, but no other processes in the joined cgroup (which the container process did not create) are reaped.


runtime.md
```diff
@@ -29,24 +29,59 @@ The root directory to the bundle is provided in the state so that consumers can
 }
 ```
 
-## Lifecycle
+## Typical lifecycle
+
+A typical lifecyle progresses like this:
+
+1. There is no container or running application
+2. A user tells the runtime to start a container+application
```
mrunalp: start a container for an application.. ?

crosbymichael: `A user tells the runtime to start a container`  Don't use the word application, its just a container

mrunalp: Yeah, I think it makes sense to leave out application altogether.

wking: On Wed, Sep 30, 2015 at 11:52:37AM -0700, Mrunal Patel wrote:
> > +2. A user tells the runtime to start a container+application
> 
> start a container for an application.. ?

Without explicit create and start steps, it's “to create a container
and launch an application inside that container”, but that seemed to
long.  Should I just replace “container+application” with “container
and application”?

wking:
I've rebased on the new master and added 80c1729 since #202 landed.  I'm not sure what the intended behavior isif the application dies while we're running post-start hooks.  In 80c1729 I've gone with “we terminate any in-progress post-start hook and skip any that hadn't been called yet”, but “terminate” is vague (TERM?  KILL?  TERM and KILL after 5 seconds if necessary?).  @LK4D4, do you want something different than the 80c1729 phrasing?

justincormack:
Is this supposed to be descriptive or normative? It might be clearer to insert normative language where useful.

Destroying the "container process" (might be a better term, its usually going to be container pid 1 but I suppose you might not have a pid namespace) will often implicitly destroy the container, so I would allow for that to happen, so hooks after that may not be able to see container any more.

wking:
On Thu, Oct 01, 2015 at 02:23:34AM -0700, Justin Cormack wrote:
> Is this supposed to be descriptive or normative?

I'd rather have the summaries in the enumerated list be descriptive
(so you can cut a few corners there), but the sections they link to be
normative (so everyone is on the same page for what's really
happening).

> Destroying the "container process" (might be a better term, its
> usually going to be container pid 1 but I suppose you might not have
> a pid namespace) will often implicitly destroy the container, so I
> would allow for that to happen, so hooks after that may not be able
> to see container any more.

I don't see “destroy” in the language through e0ec87b.  Are you
suggesting that language for:

  11. The runtime [terminates any other processes in the container](#stop-process)

This termination is qualified by the “Joining existing containers”
section, that clarifies the “other processes in the container” means
“other processes in any cgroup you created” (at least on Linux).  I'm
happy to roll in changes that make that more clear (maybe in [2]
itself?), if you have suggestions for the text you'd like to see.

This would also be easier if we landed a glossary like that proposed
in #107, although that doesn't currently equate “container processes”
with “processes in a unified-hierarchy cgroup” like we do in this PR.

[1]: https://github.com/wking/opencontainer-specs/blob/lifecycle/runtime.md#joining-existing-containers
[2]: https://github.com/wking/opencontainer-specs/blob/lifecycle/runtime.md#stop-process

wking:
I did a bit more digging in runC's cleanup implementation yesterday,
and here's what it's currently doing:

1. If the container process was the first in a PID namespace, it's
   death will automatically cleanup any other processes in that
   namespace.  Otherwise, the current runC code kills the other
   processes in the cgroup [1,2].  The ‘Freeze’ function requires a
   freezer cgroup [3], but the killer carries on if it fails [4].  The
   ‘GetPids’ function looks at the device cgroup [5].  The ‘Kill’ call
   sends a SIGKILL [6,7,8,9].  Wait for the killed processes to exit
   [10].
2. Remove any created cgroups [11]
3. Remove /run/opencontainer/containers/{container-id} [12].
4. Run the post-stop hooks [13].

I'm not sure when persistent namespaces (e.g. network namespaces) are
removed.

That's a bit different from the cleanup logic currently proposed here:

a. I suggest running the post-stop hooks before cleaning up the
   container.
b. I suggest removing the state.json file first (with the just-pushed
   e0ec87b → 717a999) to be the inverse of the container-creation
   process, where state.json is written last.
c. I require a unified hierarchy, which I thought was our intention
   after 429f936 (Adding cgroups path to the Spec, 2015-09-02, #137)
   [14].  The unified hierarchy is still experimental [15], but it
   makes handling containers easier because we don't have to create a
   PID namespace or device cgroup before we can easily define “the
   process belonging to the container”.  With the unified hierarchy
   (‘mount -t cgroup -o __DEVEL__sane_behavior cgroup $MOUNT_POINT’),
   you can add processes to a cgroup that doesn't use any controllers,
   which is what you'd want if someone left linux.devices and
   linux.resources out of their runtime.json.  I don't know of a good
   way to do this with the split hierarchy, although the devices
   controller seems like a reasonable choice for a canonical
   hierarchy, since you can join one without configuring it and keep
   the default matching the parent [16].

I think this PR's current approach to (a) and (b) makes sense, and it
would be easy to update runC to match.  I'm less clear on (c), since I
like the unified-hierarchy approach, but am cautious about relying on
an experimental kernel interface.  Maybe we should relax our
unified-hierarchy phrasing and explicitly mention the device
controller as one that must be created and whose cgroup.procs defines
the set of container processes.

[1]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/container_linux.go#L241-L245
[2]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/init_linux.go#L304-L312
[3]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/cgroups/fs/apply_raw.go#L179-L203
[4]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/init_linux.go#L310
[5]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/cgroups/fs/apply_raw.go#L205-L217
[6]: https://golang.org/pkg/os/#Process.Kill
[7]: https://golang.org/src/os/doc.go?s=1097:1127#L25
[8]: https://golang.org/src/os/exec_posix.go#L52
[9]: https://golang.org/src/os/exec_posix.go#L18
[10]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/init_linux.go#L328-L332
[11]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/container_linux.go#L246
[12]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/container_linux.go#L247-L249
[13]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/container_linux.go#L251-L262
[14]: https://github.com/opencontainers/specs/pull/137/files/#diff-b753d6f046da0a2956b5218efd779b7fR139
[15]: https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt
[16]: https://www.kernel.org/doc/Documentation/cgroups/devices.txt

runtime-config.md
```diff
@@ -12,27 +12,28 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
 Hooks allow one to run code before/after various lifecycle events of the container.
 Hooks MUST be called in the listed order.
 The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+All hooks execute in the host environment (e.g. the same namespace, cgroups, etc. that apply to the host process).
```
crosbymichael: This is incorrect.  They are forked from the runtime process so they inherit from the runtime, not all host processes.

runtime.md
```diff
@@ -29,24 +29,60 @@ The root directory to the bundle is provided in the state so that consumers can
...
+The invoked process forks, with one branch that stays in the host namespace and another that enters the container.
+The host process carries out all container setup actions, and continues running for the life of the container so it can perform teardown after the container process exits.
+The container process changes users and drops privileges in preparation for the container process start.
+At this point, the host process writes the [`state.json`](#state) file with the host-side version of the container-process's PID (the container process may be in a PID namespace).
 
 ### Start (process)
 
-Runs a process in a container.
-Can be invoked several times.
+After the pre-start hooks complete, the host process signals the container process to execute the runtime.
```
crosbymichael: This sentence does not make sense.  ` the host process signals the container process to execute the runtime.`????

runtime-config.md
wking: On Wed, Oct 07, 2015 at 11:36:22AM -0700, Michael Crosby wrote:
> > +All hooks execute in the host environment (e.g. the same
> > namespace, cgroups, etc. that apply to the host process).
> 
> This is incorrect.  They are forked from the runtime process so they
> inherit from the runtime, not all host processes.

There are at least two runtime processes here.  There is a process
launched with (for example) ‘runC start’, which I've been referring to
as the “host process”.  That process forks and the child ends up in
the container, eventually running the config.json process binary.  I'd
been referring to the latter as the “container process”.  Do you
prefer alternative names for those two processes?

runtime.md
wking: On Wed, Oct 07, 2015 at 11:40:34AM -0700, Michael Crosby wrote:
> > +After the pre-start hooks complete, the host process signals the
> > container process to execute the runtime.
>
> This sentence does not make sense.  ` the host process signals the
> container process to execute the runtime.`????

I haven't looked into the runC implementation here, but my guess was
that:

1. ‘runC start’ would launch the host process.
2. The host process would create as much of the container as it could
   without a container process to hold open things like a PID
   namespace.
3. The host process would fork, and the child of that fork would be
   the container process.
4. The container process would join the existing (partial container)
   and finish it off (e.g. by creating a PID namespace, changing
   users, dropping priveledges, etc.)
5. The host process would launch the pre-start hooks.
6. The host process would signal the container process that the
   pre-start hooks had completed successfully.
7. The container process would catch that signal and exec the
   config.json process.

The line you're quoting was supposed to sketch out (6) and (7).  Can
you clear me up on how it actually works?

wking: On Wed, Oct 07, 2015 at 11:51:50AM -0700, W. Trevor King wrote:
> Wed, Oct 07, 2015 at 11:40:34AM -0700, Michael Crosby:
> > > +After the pre-start hooks complete, the host process signals
> > > the container process to execute the runtime.
> >
> > This sentence does not make sense.  ` the host process signals the
> > container process to execute the runtime.`????
> 
> I haven't looked into the runC implementation here, but my guess was
> that:
> 
> …
> 6. The host process would signal the container process that the
>    pre-start hooks had completed successfully.
> 7. The container process would catch that signal and exec the
>    config.json process.
> 
> The line you're quoting was supposed to sketch out (6) and (7).  Can
> you clear me up on how it actually works?

It looks like libcontainer actually has a fairly detailed spec for how
it sets up containers.  That talks about using a pipe instead of
singals for this trigger [1].  Signals sound easier to me (permissions
discussed in [2]), but the generic “signals” verb could cover all such
triggers (including piped messages) and not be restricted to POSIX
signals [3], since this is an internal runtime mechanism.

[1]: https://github.com/opencontainers/runc/blob/v0.0.4/libcontainer/SPEC.md#runtime-and-init-process
[2]: http://man7.org/linux/man-pages/man2/kill.2.html
[3]: http://man7.org/linux/man-pages/man7/signal.7.html

runtime-config.md
```diff
@@ -12,27 +12,28 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
...
 
 Hook paths are absolute and are executed from the host's filesystem.
 
 ### Pre-start
 
-The pre-start hooks are called after the container process is spawned, but before the user supplied command is executed.
+The pre-start hooks are called [after the container process is spawned, but before the user supplied command is executed](runtime.md#typical-lifecycle).
```
vishh: Do we need to mention `container process` here? These hooks are invoked once the majority of the container runtime has been setup, but before the container's init process has been spawned. We need to clearly specify what sub-section of the complete container environment is available for these hooks.

runtime-config.md
```diff
@@ -12,27 +12,28 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
...
 They are called after the container namespaces are created on Linux, so they provide an opportunity to customize the container.
 In Linux, for e.g., the network namespace could be configured in this hook.
 
-If a hook returns a non-zero exit code, then an error including the exit code and the stderr is returned to the caller and the container is torn down.
+If a hook returns a non-zero exit code, [then an error including the exit code and the stderr is returned to the caller and the container is torn down](runtime.md#typical-lifecycle).
 
 ### Post-start
 
-The post-start hooks are called after the user process is started.
+The post-start hooks are called [after the user process is started](runtime.md#typical-lifecycle).
```
vishh: How about `init process` instead of `user process`?

runtime-config.md
wking: On Mon, Oct 12, 2015 at 12:47:00PM -0700, Vish Kannan wrote:
> > -The pre-start hooks are called after the container process is spawned, but before the user supplied command is executed.
> > +The pre-start hooks are called [after the container process is spawned, but before the user supplied command is executed](runtime.md#typical-lifecycle).
> 
> Do we need to mention `container process` here? These hooks are
> invoked once the majority of the container runtime has been setup,
> but before the container's init process has been spawned. We need to
> clearly specify what sub-section of the complete container
> environment is available for these hooks.

The container is complete when the pre-start hooks run.  The container
process has been spawned (it's just running runtime-supplied code).
The only change that happens *after* the pre-start hooks is that that
container process execs the user-supplied process, swapping in the
user-selected code.  For a bit of runC-side docs on this, see the
synchronization pipe in [1].

[1]: https://github.com/opencontainers/runc/blob/v0.0.4/libcontainer/SPEC.md#runtime-and-init-process

runtime-config.md
wking: On Mon, Oct 12, 2015 at 12:48:09PM -0700, Vish Kannan wrote:
> >  
> >  ### Post-start
> >  
> > -The post-start hooks are called after the user process is started.
> > +The post-start hooks are called [after the user process is started](runtime.md#typical-lifecycle).
> 
> How about `init process` instead of `user process`?

The container process (initially running runtime-supplied code) is
also the init process.  Here's how I see the initial communication:

  Host process                Container process
  ----------------------      ------------------------------------
  start
  launches child          ->
  blocks                      joins namespaces
                              unshares namespaces
                              configures unshared namespaces
                              drops permissions, changes user, ...
                          <-  signals that container is complete
  runs pre-start hooks        blocks
  signals hooks complete  ->
                              closes pipe
                              executes user process

So there's a lot going on in the container process before it turns
itself into the “user process”.

wking: On Mon, Oct 12, 2015 at 12:48:09PM -0700, Vish Kannan wrote:
> How about `init process` instead of `user process`?

Also “init process” implies a PID namespace, and those are optional
with this spec.

wking: On Mon, Oct 12, 2015 at 12:56:13PM -0700, W. Trevor King wrote:
>   Host process                Container process
>   ----------------------      ------------------------------------
>   start
>   launches child          ->
>   blocks                      joins namespaces

After launching and before blocking, the host process should also be
adding the container process to its cgroup (if you're using any
cgroups).

runtime-config.md
vishh: You are stating runC's behavior. Why do we have to require that behavior in the Spec?

wking: On Mon, Oct 12, 2015 at 01:20:20PM -0700, Vish Kannan wrote:
> You are stating runC's behavior. Why do we have to require that
> behavior in the Spec?

This commit is mostly about describing the runC approach so we have a
non-controversial starting point for future evolution (“is that what
runC does” is something that is harder to disagree on than “should
that be how the spec works” ;).

That being said, I *do* think all runtimes will need to execute
runtime-specified code in the container process [1].

[1]: https://github.com/opencontainers/specs/issues/115#issuecomment-147512709

runtime-config.md
vishh: I get the runC behavior @wking. 

One alternative to the runC behavior is that of unsharing namespaces (excepting pid), bind mounting the namespaces, and letting the hooks run before starting a new process in a new pid namespace (if requested by the user). 
In this scenario, the `container process` isn't required.

Similarly, I'm wondering if the new semantics you are introducing are valid in other OSes.
I'd like to solicit feedback from the folks who are working on Spec implementations on other other OS'es like Windows and Solaris.

vishh:
I find this PR introducing concepts to the Spec that are implementation details of runC. Input from folks working on implementing the Spec in non-Linux OSes will definitely help.

wking:
On Mon, Oct 12, 2015 at 03:50:19PM -0700, Vish Kannan wrote:
> I find this PR introducing concepts to the Spec that are
> implementation details of runC. Input from folks working on
> implementing the Spec in non-Linux OSes will definitely help.

The initial comment for this PR points out that it's just about Linux
containers.  And as I said there, I'm happy to roll in notes for other
platforms if folks know what to write.

And the goal on Linux is to document what runC does now [1], since
that's something that is easier to agree on.

So +1 to feedback like “on Windows, container creation will need
language like …” or “in commit abc, runC is actually cleaning up
containers by doing …”.  But I think feedback like “I think we should
use the unified-hierarchy [2]” should be spun off to the list (it's
mentioned as an option in [3]) and not handled here.

[1]: https://github.com/opencontainers/specs/pull/207#discussion_r41798699
[2]: https://github.com/opencontainers/specs/pull/207#issuecomment-145934135
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/n7vxlhn6Q7g

runtime-config.md
wking: On Mon, Oct 12, 2015 at 03:48:06PM -0700, Vish Kannan wrote:
> One alternative to the runC behavior is that of unsharing namespaces
> (excepting pid), bind mounting the namespaces, and letting the hooks
> run before starting a new process in a new pid namespace (if
> requested by the user).  In this scenario, the `container process`
> isn't required.

In the “user requested a PID namespace” case, that sounds like “the
container isn't fully setup” (since you don't have the requested PID
namespace at pre-start time).

In either case, there is *still* a container process with
runtime-specified code at pre-start-hook time, because it's holding
open the unshared namespaces.

vishh: Not really. If you bind-mount the ns files, you don't need a process to
hold the namespaces.

wking: On Mon, Oct 12, 2015 at 04:03:30PM -0700, Vish Kannan wrote:
> Not really. If you bind-mount the ns files, you don't need a process
> to hold the namespaces.

Oh, that will make my life a lot easier :).  Is that:

Host process                    Temporary child
--------------------------      -----------------------------------------
Launch a child process      ->
                                Call clone(2) with CLONE_NEW*
                                Bind mount /proc/self/ns/* somewhere else
                                pivot_root, etc.
                            <-  Exit
Use the bind-mounted files

In that case, excepting the PID namespace from the pre-start
requirements seems like a reasonable prices to pay.  And it will make
explicit ‘create’ and ‘start’ commands possible [1].

On the other hand, we'll probably need to revisit the state JSON that
landed in #87 now that the ‘pid’ entry doesn't make sense for *either*
pre-start or post-stop.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/frUXLljXy8Y
     Message-ID: <20150915034717.GO18018@odin.tremily.us>

wking: On Mon, Oct 12, 2015 at 04:14:38PM -0700, W. Trevor King wrote:
> Mon, Oct 12, 2015 at 04:03:30PM -0700, Vish Kannan:
> > Not really. If you bind-mount the ns files, you don't need a process
> > to hold the namespaces.
> 
> Oh, that will make my life a lot easier :).  Is that…

And it looks like it is :).  Language supporting this in namespaces(7):

  Bind mounting (see mount(2)) one of the files in this directory to
  somewhere else in the filesystem keeps the corresponding namespace
  of the process specified by pid alive even if all processes
  currently in the namespace terminate.

For example:

  tty1$ unshare --user --uts --map-root-user
  tty1$ hostname alice
  tty1$ echo $$
  25477

  tty2# touch /tmp/uts-alice
  tty2# mount --bind /proc/25477/ns/uts /tmp/uts-alice

  tty1$ exit

  tty2# ps aux | grep 25477 | grep -v grep
  … no hits …
  tty2# nsenter --uts=/tmp/uts-alice hostname
  alice

vishh:
@wking: I suggest closing this PR and starting an email thread to solicit feedback from other members.
I hope the existing comments gives you an idea as to what is required to come up with a working proposal.

wking:
On Mon, Oct 12, 2015 at 04:57:21PM -0700, Vish Kannan wrote:
> @wking: I suggest closing this PR and starting an email thread to
> solicit feedback from other members.  I hope the existing comments
> gives you an idea as to what is required to come up with a working
> proposal.

Email thread exists [1] (also linked in this PR's original post).  I
moved it to a PR after that thread settled down and I got clearance in
a meeting (see action items in [2]).  I'm happy to continue discussion
in the thread if that's easier for everyone else, but I'm still not
sure what changes you're looking for.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/frUXLljXy8Y
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2015/opencontainers.2015-09-30-17.00.html

crosbymichael:
I think there is still a lot of discussion needed on the lifecycle but this is not an accurate base to start from and will end up being more work if we iterate on this PR instead of discussing first and starting fresh. 

We need to make this detailed but having some of the people who work on runtimes comment as wells as abstract where this works cross platform.  We also need to figure out why we are defining the lifecycle and the types of issues it solves by having this defined or else it will be targeted at a something that is not useful for a consumer.

wking:
On Mon, Oct 12, 2015 at 05:16:15PM -0700, Michael Crosby wrote:
> I think there is still a lot of discussion needed on the lifecycle
> but this is not an accurate base to start from and will end up being
> more work if we iterate on this PR instead of discussing first and
> starting fresh.

Fair enough.  @mrunalp volunteered to write one up [1], and this
attempt was because he hadn't gotten around it yet [2].  I'd even be
fine just copying over runC's libcontainer/SPEC.md and then iterating
on that in more tightly-focused PRs.

> We also need to figure out why we are defining the lifecycle and the
> types of issues it solves by having this defined or else it will be
> targeted at a something that is not useful for a consumer.

I think *this* is pretty clear.  For example, it's not currently clear
whether we can rely on the ‘pid’ entry being there in the pre-start
and/or post-stop hooks, what portions of the existing container are
available for pre-start hooks to work with (see #115), what gets
cleaned up when a container exits, what “container processes” even
means [3], what “standard operations” means [4], or what a runtime API
should look like [5].  All of these are difficult discussions to have
without specs for what we think a container lifecycle looks like.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/frUXLljXy8Y/8pH-81moGgAJ
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/frUXLljXy8Y/aKFRX56qAAAJ
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/n7vxlhn6Q7g
[4]: https://github.com/opencontainers/specs/pull/209#discussion_r41047185
[5]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/BIxya5eSNLo

PR #210, opened 2015-10-02
Cleanup bundle.md

duglin:
Mainly just moved stuff around, but also tried to add some clarity around
what is required w.r.t. naming and location of files/dirs.

But to highlight some things that I want to make sure we ALL agree on:
- config.json and runtime.json MUST use those exact names
- name of rootfs dir is user defined - but nudge them towards `rootfs`
- all 3 (config.json, runtime.json and rootfs dir) MUST all be in the same location/dir
- that 'location' is NOT part of the bundle itself (this may matter when someone defines the on-the-wire format of a bundle)

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -1,39 +1,27 @@
-# Bundle
+# Filesystem Bundle
 
 ## Container Format
 
-This section defines a format for encoding a container as a *bundle* - a directory organized in a certain way, and containing all the necessary data and metadata for any compliant runtime to perform all standard operations against it.
+This section defines a format for encoding a container as a *filesystem bundle* - a set of files organized in a certain way, and containing all the necessary data and metadata for any compliant runtime to perform all standard operations against it.
 See also [OS X application bundles](http://en.wikipedia.org/wiki/Bundle_%28OS_X%29) for a similar use of the term *bundle*.
 
-The format does not define distribution.
-In other words, it only specifies how a container must be stored on a local filesystem, for consumption by a runtime.
-It does not specify how to transfer a container between computers, how to discover containers, or assign names or versions to them.
-Any distribution method capable of preserving the original layout of a container, as specified here, is considered compliant.
+The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local file system so that it can be consumed by a compliant runtime.
 
-A standard container bundle is made of the following 3 parts:
+A Standard Container bundle contains all the information needed to load and run a container.
+This includes the following three artifacts which MUST all reside in the same directory on the local filesystem:
 
-- A top-level directory holding everything else
-- One or more content directories
-- A configuration file
+1. `config.json` : immutable, host independent configuration.
+This REQUIRED file, which MUST be named `config.json`, contains settings that are host independent and application specific such as security permissions, environment variables and arguments.
+See [`config.json`](config.md) for more details.
 
-## Directory layout
+2. `runtime.json` : mutable, host dependent configuration.
+This REQUIRED file, which MUST be named `runtime.json`, contains settings that are host specific such as memory limits, local device access and mount sources.
+The goal is that the bundle can be moved as a unit to another runtime and run the same application if `runtime.json` is reconfigured.
+See [`runtime.json`](runtime-config.md) for more details.
 
-A Standard Container bundle is a directory containing all the content needed to load and run a container.
-This includes two configuration files `config.json` and `runtime.json`, and a rootfs directory.
-The `config.json` file contains settings that are host independent and application specific such as security permissions, environment variables and arguments.
-The `runtime.json` file contains settings that are host specific such as memory limits, local device access and mount points.
-The goal is that the bundle can be moved as a unit to another machine and run the same application if `runtime.json` is removed or reconfigured.
+3. A directory representing the root filesystem of the container.
+While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
+This directory MUST be referenced from within the `config.json` file.
 
-Configuration file syntax and semantics:
-
-* [`config.json`](config.md) (immutable, host independent configuration)
-* [`runtime.json`](runtime-config.md) (mutable, host dependent configuration)
-
-A single `rootfs` directory MUST be in the same directory as the `config.json`.
-The names of the directories may be arbitrary, but users should consider using conventional names as in the example below.
-
-```
-config.json
-runtime.json
-rootfs/
-```
+While these three artifacts MUST all be present in a single directory on the local filesytem, that directory itself is not part of the bundle.
+In other words, a tar archive of a *bundle* will have these artifacts at the root of the archive, not nested within a top-level directory.


wking:
On Fri, Oct 02, 2015 at 07:56:16AM -0700, Doug Davis wrote:
> - config.json and runtime.json MUST use those exact names

Agreed, although I'm fine if runtimes support extensions to select
other names (and I've added that flexibility to my proposed
command-line API spec [1,2], but we can revisit this if/when that
lands in this repo).

> - name of rootfs dir is user defined - but nudge them towards
>   `rootfs`

I don't see the point of requiring this [3], but yeah, I think that's
how the spec reads now.

> - all 3 (config.json, runtime.json and rootfs dir) MUST all be in
>   the same location/dir

I don't think we require that for rootfs, since the path from
config.json's root could contain directories (relative to config.json?
Relative to the runtime's current working directory?).

> - that 'location' is NOT part of the bundle itself (this may matter
>   when someone defines the on-the-wire format of a bundle)

+1.

[2]: https://gist.github.com/wking/1d69118ba8b750f85bfc
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/6ZKMNWujDhU

duglin:
> > - all 3 (config.json, runtime.json and rootfs dir) MUST all be in
> > the same location/dir
> 
> I don't think we require that for rootfs, since the path from
> config.json's root could contain directories (relative to config.json?
> Relative to the runtime's current working directory?).

In bundle.md it say:
- - - - - -
A single rootfs directory MUST be in the same directory as the config.json
- - - - - - 

I don’t personally have a preference yet - was just trying to keep the current state of things and get clarity. 

-Doug

wking:
On Fri, Oct 02, 2015 at 10:50:32AM -0700, Doug Davis wrote:
> > > - all 3 (config.json, runtime.json and rootfs dir) MUST all be in
> > > the same location/dir
> >
> > I don't think we require that for rootfs, since the path from
> > config.json's root could contain directories (relative to
> > config.json?  Relative to the runtime's current working
> > directory?).
> 
> In bundle.md it say:
> - - - - - -
> A single rootfs directory MUST be in the same directory as the
> config.json

Well that's pretty unambiguous ;), so I'd guess language like that is
fine for now.  It makes the relative-path bit in config.md a bit out
of place, but polishing for consistency is probably secondary to
figuring out why we actually need to require a root directory (and
what it means when a container doesn't create a mount namespace), or
whether it can be optional.

bundle.md
```diff
@@ -1,39 +1,28 @@
...
+This section defines a format for encoding a container as a *filesystem bundle* - a set of files organized in a certain way, and containing all the necessary data and metadata for any compliant runtime to perform all standard operations against it.
 See also [OS X application bundles](http://en.wikipedia.org/wiki/Bundle_%28OS_X%29) for a similar use of the term *bundle*.
 
-The format does not define distribution.
-In other words, it only specifies how a container must be stored on a local filesystem, for consumption by a runtime.
-It does not specify how to transfer a container between computers, how to discover containers, or assign names or versions to them.
-Any distribution method capable of preserving the original layout of a container, as specified here, is considered compliant.
+The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local file system so that it can be consumed by a compliant runtime.
+Issues such as distribution, including how to transfer a container between runtimes, assigning names, versioning of bundle, or discovery of bundles are out of scope of this specification.
```
mrunalp: You mean out of scope for this section?

wking: On Mon, Oct 05, 2015 at 09:45:31AM -0700, Mrunal Patel wrote:
> > +Issues such as distribution, including how to transfer a
> > container between runtimes, assigning names, versioning of bundle,
> > or discovery of bundles are out of scope of this specification.
> 
> You mean out of scope for this section?

Maybe we should just drop this line to punt on this issue, because I
think it's out of scope for this specification.  Lots of back and
forth on this point in [1] ;).

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/OqnUp4jOacs

mrunalp: Yeah, taking the line out for now may not be a bad idea.

duglin: I'm not following. These topics are out of scope of the "specification" right now, no? I know the charter discussions are happening to decide for sure, but as of now, they're out of scope. So, I don't see the issue with saying what's out of scope.  What am I missing?

wking: On Mon, Oct 05, 2015 at 10:27:00AM -0700, Doug Davis wrote:
> > +Issues such as distribution, including how to transfer a
> > container between runtimes, assigning names, versioning of bundle,
> > or discovery of bundles are out of scope of this specification.
>
> I'm not following. These topics are out of scope of the
> "specification" right now, no? I know the charter discussions are
> happening to decide for sure, but as of now, they're out of
> scope. So, I don't see the issue with saying what's out of scope.
> What am I missing?

To me “out of scope” is explicitly “we will never cover this”, and I'd
use “unspecified” for “we will cover this at some point, but haven't
got around to it yet”.  With that phrasing, I think we all agree that
these distribution issues are currently unspecified, but we disagree
about whether they're out of scope for this spec.

But I don't have sources to cite for that distinction, so maybe it's
just me ;).

vbatts: that is the distinction. Discoverability is not explicitly out of scope.

duglin: I think what's there is more appropriate since all we can say is a reflection of the charter - and as of now its "out of scope" - we will never cover this w/o a charter change.  Using a word like "unspecified" could be interpreted as "we're choosing not to define it at this time but we could later", but that's not true w/o a charter change.

@vbatts : what's in or out of scope is still TBD. As much as I think the current scope of what's in this doc & PR says is too limited, I'm trying to go with what's agreeable for "in scope" for now and then will let the charter discussions change this doc later.

vbatts: not. the charter currently says

> d. Decentralized. Discovery of container images should be simple and facilitate a federated namespace and distributed retrieval.

So there should not be wording that something like this is "out of scope"

duglin: To be honest, I have no idea what a section titled "OCI Values" even means and I've been pushing to fix that because some people interpret it as just fluff filler text that's non-normative, while others (as you've hinted) view it as part of our "in scope list of work items".  Because if its our list of work items then when 'c' includes "image auditing" and "cryptographic primitives", where is that in our work?

But I was actually hoping to avoid charter discussions as part of this PR.  I'm ok with removing this sentence **for now** but immediately after the charter is finalized I think we need to add something like it back in so people have the right level of expectations w.r.t. what our goals are.

crosbymichael: ya, i would remove for now until the charter is finished, otherwise this pr LGTM

duglin:
questionable text removed - all comments addressed - I think

crosbymichael:
LGTM

bundle.md
```diff
@@ -1,39 +1,27 @@
...
 
-A standard container bundle is made of the following 3 parts:
+A Standard Container bundle contains all the information needed to load and run a container.
+This includes the following three artifacts which MUST all reside in the same directory on the local filesystem:
 
-- A top-level directory holding everything else
-- One or more content directories
-- A configuration file
+1. `config.json` : immutable, host independent configuration.
+This file, which MUST be named `config.json`, contains settings that are host independent and application specific such as security permissions, environment variables and arguments.
```
wking: “security permissions” seems vague.  Do you mean `linux.capabilities`?  I'm not sure how that separates out from cgroups which are also sometimes about permissions (e.g. the device cgroup is just about permissions), but I'd probably drop that entry from your list of examples.  However, after dropping it, the only remaining entries are about `process`, so I think we should maybe just stop after “application specific” ;).

bundle.md
```diff
@@ -1,39 +1,27 @@
...
+See [`config.json`](config.md) for more details.
 
-## Directory layout
+2. `runtime.json` : mutable, host dependent configuration.
+This file, which MUST be named `runtime.json`, contains settings that are host specific such as memory limits, local device access and mount points.
```
wking: Minor nit: `config.json` has a [`MountPoint`][1] type, so I'd rather refer to the `runtime.json` entries as “mount sources” or some such.

[1]: https://github.com/opencontainers/specs/blob/fe096749b4bf11905101362938ed10b2c01f20e6/config.go#L16-L17

duglin: used "mount sources".

bundle.md
duglin: Not sure who wrote that phrase originally. @vbatts @crosbymichael any thoughts? Remove examples or is there another example we should use?

crosbymichael: it's just an example, being vague is ok.

wking: On Mon, Oct 05, 2015 at 03:40:11PM -0700, Michael Crosby wrote:
> > +This file, which MUST be named `config.json`, contains settings
> > that are host independent and application specific such as
> > security permissions, environment variables and arguments.
> 
> it's just an example, being vague is ok. 

I think it's misleading, because almost all of the security stuff is
going to be handled in runtime.json.  If we mean Linux capabilities,
it seems easy enough to just say that.

But whatever, there's a link to the config specs right after each
bullet point, so folks can see what's actually in the files ;).

@duglin, ‘git log -p -G 'security permissions'’ points to 7232e4b1
(specs: introduce the concept of a runtime.json, 2015-07-30, #88).
The only inline comment on anything in its general viscinity was the
conversation I had with @philips [1] that lead to me filing #107 with
a definition for “application” (since removed from that PR [2]).

[1]: https://github.com/opencontainers/specs/pull/88#discussion_r35930309
[2]: https://github.com/opencontainers/specs/pull/107#issuecomment-145619316

bundle.md
```diff
@@ -1,39 +1,27 @@
...
+See [`runtime.json`](runtime-config.md) for more details.
 
-A Standard Container bundle is a directory containing all the content needed to load and run a container.
-This includes two configuration files `config.json` and `runtime.json`, and a rootfs directory.
-The `config.json` file contains settings that are host independent and application specific such as security permissions, environment variables and arguments.
-The `runtime.json` file contains settings that are host specific such as memory limits, local device access and mount points.
-The goal is that the bundle can be moved as a unit to another machine and run the same application if `runtime.json` is removed or reconfigured.
+3. A directory representing the root filesystem of the container.
+While the name of this directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
+This directory will be referenced from within the `config.json` file.
```
mrunalp: s/will/MUST/ ?

mrunalp:
LGTM

PR #145, opened 2015-09-03
runtime: Deduplicate phrasing in state attribute definitions

wking:
Replace:

    $ENTRY ($TYPE) $ENTRY is the ...

with:

    $ENTRY ($TYPE) The ...

It's more concise and avoids the need to argue over Pid vs PID ;).

This is my first commit from crosbymichael/specs#1, rebased to pickup
the `version` addition that landed while #87 was in-flight.

PR #211, opened 2015-10-02
Add some clarity around the state.json file

duglin:

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -2,22 +2,26 @@
 
 ## State
 
-The runtime state for a container is persisted on disk so that external tools can consume and act on this information.
-The runtime state is stored in a JSON encoded file.
-It is recommended that this file is stored in a temporary filesystem so that it can be removed on a system reboot.
-On Linux based systems the state information should be stored in `/run/opencontainer/containers`.
-The directory structure for a container is `/run/opencontainer/containers/<containerID>/state.json`.
-By providing a default location that container state is stored external applications can find all containers running on a system.
+Runtime MUST store container metadata on disk so that external tools can consume and act on this information.
+It is recommended that this data be stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux/Unix based systems the metadata MUST be stored under `/run/opencontainer/containers`.
+For non-Linux/Unix based systems the location of the root metadata directory is currently undefined.
+Within that directory there MUST be one directory for each container created, where the name of the directory MUST be the ID of the container.
+For example: for a Linux container with an ID of `173975398351`, there will be a corresponding directory: `/run/opencontainer/containers/173975398351`.
+Within each container's directory, there MUST be a JSON encoded file called `state.json` that contains the runtime state of the container.
+For example: `/run/opencontainer/containers/173975398351/state.json`.
 
-* **`version`** (string) Version of the OCI specification used when creating the container.
-* **`id`** (string) ID is the container's ID.
-* **`pid`** (int) Pid is the ID of the main process within the container.
-* **`bundlePath`** (string) BundlePath is the path to the container's bundle directory.
+The `state.json` file MUST contain all of the following properties:
 
+* **`version`**: (string) is the OCF specification version used when creating the container.
+* **`id`**: (string) is the container's ID.
+This MUST be unique across all containers on this host.
+There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
-This allows the hook to perform clean and teardown logic after the runtime destroys its own state.
-
-The root directory to the bundle is provided in the state so that consumers can find the container's configuration and rootfs where it is located on the host's filesystem.
+This allows the hooks to perform cleanup and teardown logic after the runtime destroys its own state.
+* **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
+* **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
+This is provided so that consumers can find the container's configuration and root filesystem on the host.
 
 *Example*
 


runtime.md
```diff
@@ -2,22 +2,26 @@
...
-By providing a default location that container state is stored external applications can find all containers running on a system.
-
-* **`version`** (string) Version of the OCI specification used when creating the container.
-* **`id`** (string) ID is the container's ID.
-* **`pid`** (int) Pid is the ID of the main process within the container.
-* **`bundlePath`** (string) BundlePath is the path to the container's bundle directory.
-
+Metadata for containers are stored on disk so that external tools can consume and act on this information.
+It is recommended that this data be stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the metadata MUST be stored in `/run/opencontainer/containers`.
```
vbatts: Maybe we can say Linux/Unix? Thoughts?

runtime.md
```diff
@@ -2,22 +2,26 @@
...
-On Linux based systems the state information should be stored in `/run/opencontainer/containers`.
-The directory structure for a container is `/run/opencontainer/containers/<containerID>/state.json`.
-By providing a default location that container state is stored external applications can find all containers running on a system.
-
-* **`version`** (string) Version of the OCI specification used when creating the container.
-* **`id`** (string) ID is the container's ID.
-* **`pid`** (int) Pid is the ID of the main process within the container.
-* **`bundlePath`** (string) BundlePath is the path to the container's bundle directory.
-
+Metadata for containers are stored on disk so that external tools can consume and act on this information.
```
wking: Alternative openings in case folks like them better ;):

* “Container metadata is stored…”
* “Runtimes MUST store container metadata…”

runtime.md
wking: On Fri, Oct 02, 2015 at 01:40:43PM -0700, Vincent Batts wrote:
> > +On Linux based systems the metadata MUST be stored in `/run/opencontainer/containers`.
> 
> Maybe we can say Linux/Unix? Thoughts?

“Systems with POSIX paths”?  I'm fine leaving this as “Linux” and
opting in other architectures as we add support for them.

Why don't we require a particular path on Windows?

runtime.md
```diff
@@ -2,22 +2,26 @@
...
+It is recommended that this data be stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux based systems the metadata MUST be stored in `/run/opencontainer/containers`.
+For non-Linux based systems the location of the root metadata directory is currently undefined.
+Within that directory there MUST be one directory for each container created, where the name of the directory MUST be the ID of the container.
+For example: for a container with an ID of `173975398351`, there will be a corresponding directory: `/run/opencontainer/containers/173975398351`.
```
wking: While we're tightening up this text, it might be time to revisit my [container-ID charset suggestions][1].  The spec still reads (to me) like you could setup a container whose ID was `../../../etc/passwd`.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/P9gZBYhiqDE

runtime.md
duglin: picked second

runtime.md
duglin: used Linux/Unix.
If someone from MSFT tells me what dir to use I can add it.

vbatts: that's where we look forward to the PoC implementations to help define that.

runtime.md
```diff
@@ -2,22 +2,26 @@
...
+Within each container's directory, there MUST be a JSON encoded file called `state.json` that contains the runtime state of the container.
+For example: `/run/opencontainer/containers/173975398351/state.json`.
+
+The `state.json` file MUST contain all of the following properties:
+
+* **`version`**: (string) Version of the OCF specification used when creating the container.
```
wking: The README for this project suggests (to me) either “OCS” or “OCIS” or “OCI Specification”.  I suggest we pick one abbreviation and have the README say:

>  The Open Container Initiative Specifications (OCIS) …

or whatever.  And then use that short name throughout the spec.  Until we do that, I'd probably just leave this as “OCI specification”.

runtime.md
```diff
@@ -2,22 +2,26 @@
...
+* **`id`**: (string) ID is the container's ID.
+This MUST be unique across all containers on this host.
+There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
-This allows the hook to perform clean and teardown logic after the runtime destroys its own state.
-
-The root directory to the bundle is provided in the state so that consumers can find the container's configuration and rootfs where it is located on the host's filesystem.
+This allows the hooks to perform clean and teardown logic after the runtime destroys its own state.
```
wking: This is just copy/pasted text, but I'd rather see “clean” changed to “cleanup”.

wking: And the “runtime destroys its own state” makes no sense to me.  We can't run hooks after the runtime destroys its state because the runtime is [passing that state to the hook on stdin][1].

[1]: https://github.com/opencontainers/specs/blob/v0.1.1/runtime.md#hooks

runtime.md
vbatts: hmmm. OCF had been mentioned, but not much defined.
Perhaps just "OpenContainers spec" or similar for now.

runtime.md
duglin: done

runtime.md
```diff
@@ -2,22 +2,26 @@
...
+* **`pid`**: (int) Pid is the ID of the main process within the container, as seen by the host.
+* **`bundlePath`**: (string) BundlePath is the path to the root of container's bundle directory.
```
wking: “the container's bundle directory” seems clear enough to me without “the root of” qualifying it.

vbatts: fair. I would expect a full or absolute path.

runtime.md
duglin: I assumed that teardown stuff was trying to say that the ID might still be usable by the hooks, even though the metadata files are gone, because its possible for the ID to be some reference into a DB (or something) someplace else.

wking:
I like this PR a lot, but it's going to conflict with #145, and I
think those (minor) changes are useful.  I'm fine if you just squash
them into this branch, or if I rebase mine on this, or if you rebase
yours on mine, ….  If I can help manage that conflict in any way, let
me know ;).

runtime.md
duglin: added "absolute" and remove "the root of"

wking: On Fri, Oct 02, 2015 at 02:07:06PM -0700, Vincent Batts wrote:
> > +* **`bundlePath`**: (string) BundlePath is the path to the root of container's bundle directory.
> 
> fair. I would expect a full or absolute path.

+1 to “… the absolute path to…”.

runtime.md
wking: On Fri, Oct 02, 2015 at 02:07:26PM -0700, Doug Davis wrote:
> I assumed that teardown stuff was trying to say that the ID might
> still be usable by the hooks, even though the metadata files are
> gone, because its possible for the ID to be some reference into a DB
> (or something) someplace else.

What metadata files?  The hooks are getting this JSON on stdin, not
via /run/opencontainer/{id}/state.json.

duglin:
@wking I think I got your edits from the other one

runtime.md
```diff
@@ -2,22 +2,26 @@
...
+* **`version`**: (string) is the OCF specification version used when creating the container.
+* **`id`**: (string) is the container's ID.
+This MUST be unique across all containers on this host.
+There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
-This allows the hook to perform clean and teardown logic after the runtime destroys its own state.
-
-The root directory to the bundle is provided in the state so that consumers can find the container's configuration and rootfs where it is located on the host's filesystem.
+This allows the hooks to perform cleanup and teardown logic after the runtime destroys its own state.
+* **`pid`**: (int) is the process ID of the main process within the container, as seen by the host.
```
wking: I'd change to “is the ID of the main process” to avoid doubling up on “process”.

duglin:
I agree that we need to tighten up the format of the ID. I was going to leave that for a different PR so that this is more about just cleaning up the text, but if we can agree quickly (w/o much bike shedding) then I'll add it.

To get the ball rolling, what about ID must conform to: `[0-9a-zA-Z_]+`  ?

runtime.md
```diff
@@ -2,22 +2,26 @@
...
-* **`id`** (string) ID is the container's ID.
-* **`pid`** (int) Pid is the ID of the main process within the container.
-* **`bundlePath`** (string) BundlePath is the path to the container's bundle directory.
-
+Runtime MUST store container metadata on disk so that external tools can consume and act on this information.
+It is recommended that this data be stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux/Unix based systems the metadata MUST be stored in `/run/opencontainer/containers`.
+For non-Linux/Unix based systems the location of the root metadata directory is currently undefined.
+Within that directory there MUST be one directory for each container created, where the name of the directory MUST be the ID of the container.
+For example: for a container with an ID of `173975398351`, there will be a corresponding directory: `/run/opencontainer/containers/173975398351`.
```
wking: Maybe qualify this with “Linux/Unix” somehow?  Something like “For example, a Linux directory `/run/opencontainer/containers/173975398351` corresponds to a container with an ID of `173975398351`.”

wking:
On Fri, Oct 02, 2015 at 02:22:21PM -0700, Doug Davis wrote:
> @wking I think I got your edits from the other one

Looks good, closing #145 in favor of this.

wking:
On Fri, Oct 02, 2015 at 02:27:12PM -0700, Doug Davis wrote:
> I agree that we need to tighten up the format of the ID. I was going
> to leave that for a different PR so that this is more about just
> cleaning up the text, but if we can agree quickly (w/o much bike
> shedding) then I'll add it.

Fair.  So far, it's just been silence, so I don't think bikeshedding
will be an issue ;).

> To get the ball rolling, what about ID must conform to:
> `[0-9a-zA-Z_]+` ?

I think you want ‘-’ too, so you can use UUIDs (links in my thread).

vbatts:
>> To get the ball rolling, what about ID must conform to:
>> `[0-9a-zA-Z_]+` ?
> I think you want ‘-’ too, so you can use UUIDs (links in my thread).

While there may be the need for rigidity, I'm not sure that needs to be the case right now. UUID needs to be allowed. But possibly colons too.

duglin:
Let's let the ID format for another PR.

vbatts:
sounds good

runtime.md
```diff
@@ -2,22 +2,26 @@
...
-By providing a default location that container state is stored external applications can find all containers running on a system.
-
-* **`version`** (string) Version of the OCI specification used when creating the container.
-* **`id`** (string) ID is the container's ID.
-* **`pid`** (int) Pid is the ID of the main process within the container.
-* **`bundlePath`** (string) BundlePath is the path to the container's bundle directory.
-
+Runtime MUST store container metadata on disk so that external tools can consume and act on this information.
+It is recommended that this data be stored in a temporary filesystem so that it can be removed on a system reboot.
+On Linux/Unix based systems the metadata MUST be stored in `/run/opencontainer/containers`.
```
mrunalp: s/in/under ?

mrunalp:
LGTM

runtime.md
duglin: changed to "under"

crosbymichael:
LGTM

PR #213, opened 2015-10-06
runtime: Change 'destroys state' to 'removes container'

wking:
The runtime hasn't destroyed any of it's internal state before calling
the post-stop hooks (it still knows the state JSON it's going to pipe
to the hooks, and could still have the original configs in memory if
it wanted to keep that around).  What's changed by the time we have
the post-stop hooks (and the reason to keep an explicit container ID
around) is that the *kernel* state (namespaces and cgroups) might be
gone by the time the post-stop hooks are run (any PID namespace that
was created by the container process will certainly be gone by then).
This commit updates the wording to more clearly mean "after removing
namespaces and cgroups" and to avoid folks interpreting it as "after
releasing some internal-to-the-runtime-process memory".

This line landed as part of #87, and the only review so far seems to
have been an [“it's” → “its” typo fix][1].

If the “removes the container” phrasing bothers people for some
reason, I'm happy to reroll.  I'm also happy to put the “destroys its
own state” phrasing back in if someone can clarify what state it was
talking about and it turns out that there is some important
internal-to-the-runtime entity that is being destroyed by post-stop
time (in which case I'd like to add additional wording to clarify that
entity in the spec).

[1]: https://github.com/opencontainers/specs/pull/87#discussion_r38236155

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -18,7 +18,7 @@ The `state.json` file MUST contain all of the following properties:
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
-This allows the hooks to perform cleanup and teardown logic after the runtime destroys its own state.
+This allows the hooks to perform cleanup and teardown logic after the runtime removes the container.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.


crosbymichael:
The runtime does not remove the container.

wking:
On Wed, Oct 07, 2015 at 11:57:24AM -0700, Michael Crosby wrote:
> The runtime does not remove the container.

Who removes the container?  With container defined as “the device
cgroup” (see [1]), it looks like runC is removing it here [2,3].

[1]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/cgroups/fs/apply_raw.go#L205-L217
[2]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/container_linux.go#L246
[3]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/cgroups/fs/apply_raw.go#L134

crosbymichael:
Not one person said a container is defined as the device cgroup.  You keep make up your own glossary and terms and try to insert them into the spec and this just adds confusion for everyone reading it.  Cgroups are just one of the features in the linux kernel used to create a container.

wking:
On Wed, Oct 07, 2015 at 01:08:40PM -0700, Michael Crosby wrote:
> Not one person said a container is defined as the device cgroup.
> You keep make up your own glossary and terms and try to insert them
> into the spec and this just adds confusion for everyone reading it.

I agree that landing a glossary (#107 or otherwise) would help a lot.

> Cgroups are just one of the features in the linux kernel used to
> create a container.

And the device cgroup is how runC currently decides what processes to
kill during cleanup [1,2,3] in the absence of a PID namespace.  So the
union of PID namespace and device cgroup seems to define “container
processes”.

The runtime doesn't remove the PID namespace (that happens
automatically on container-process death), but it does remove the
device cgroup and other persistent portions of the container.  Besides
the kernel's cleanup (PID namespace, etc.) and the runtime's cleanup
(device cgroup, etc.), is anyone else removing portions of the
container?

[1]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/container_linux.go#L241-L245
[2]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/init_linux.go#L304-L334
[3]: https://github.com/opencontainers/runc/blob/cc84f2cc9b4e66ab85ae83546d50a74ae87fe034/libcontainer/cgroups/fs/apply_raw.go#L205-L217

PR #214, opened 2015-10-06
config: Require a new UTS namespace for config.json's hostname

wking:
The UTS namespace is for [hostnames and NIS domain names][1].  Without
a new namespace, the hostname entry would clobber the host
environment's hostname.

Clobbering the host's hostname or a joined-namespace's hostname might
be acceptable for folks who trust their bundles, but it's not allowed
by the “error out if the config specifies anything else related to
that namespace” language that landed in 02b456e9 (Clarify behavior
around namespaces paths, 2015-09-08, #158).

[1]: http://man7.org/linux/man-pages/man7/namespaces.7.html

### DIFF
--- a/config.md
+++ b/config.md
@@ -104,7 +104,7 @@ For Linux-based systems the user structure has the following fields:
 
 ## Hostname
 
-* **`hostname`** (string, optional) as it is accessible to processes running inside.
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM

mrunalp:
@crosbymichael @LK4D4 PTAL

LK4D4:
LGTM

PR #220, opened 2015-10-07
Remove extra space from VersionMinor

mrunalp:


wking:
On Wed, Oct 07, 2015 at 01:41:25PM -0700, Mrunal Patel wrote:
>   * Fix an extra space in VersionMinor

Looks good to me.

This would have been caught before landing if we had a check like
(bdaf4467e, .tools: Run 'git show --check ...' on each commit, #182)
in place.

vbatts:
LGTM

PR #222, opened 2015-10-12
Add oom_score_adj to the runtime Spec.

### LINKED ISSUE #114: Support for Quality of Service

vishh:
Fixes #114

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -158,6 +158,14 @@ For example, to run a new process in an existing container without updating limi
     "disableOOMKiller": false
 ```
 
+#### Set oom_score_adj
+
+More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+
+```json
+    "oomScoreAdj": 0
+```
+
 #### Memory
 
 ```json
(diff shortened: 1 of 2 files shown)


vishh:
cc @mrunalp

wking:
Are we not happy with the pre-start hook approach to this?  See #115
and [1].  If not, can someone give more details about where my #115
decision tree breaks down?  Is the explicit field approach somehow
easier to use than a pre-start hook?  If so, how?

[1]: https://github.com/opencontainers/specs/issues/114#issuecomment-137170642

mrunalp:
@vishh Missing Signed-off-by (causing the build to fail).

vishh:
@wking: https://github.com/opencontainers/specs/issues/114#issuecomment-147500395

mrunalp:
LGTM @LK4D4 @crosbymichael ping

vbatts:
LGTM for now. (the cgroups structures had a good deal of discussion)

PR #225, opened 2015-10-13
Expand on the definition of our ops

duglin:
There are a few "OPEN" question in the doc that I'd like to brainstorm on.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -1,19 +1,15 @@
 # Runtime and Lifecycle
 
+## Scope of a Container
+
+Barring access control concerns, the entity using a runtime to create a container MUST be able to use the operations defined in this specification against that same container.
+Whether other entities using the same, or other, instance of the runtime can see that container is out of scope of this specification.
+
 ## State
 
-Runtime MUST store container metadata on disk so that external tools can consume and act on this information.
-It is recommended that this data be stored in a temporary filesystem so that it can be removed on a system reboot.
-On Linux/Unix based systems the metadata MUST be stored under `/run/opencontainer/containers`.
-For non-Linux/Unix based systems the location of the root metadata directory is currently undefined.
-Within that directory there MUST be one directory for each container created, where the name of the directory MUST be the ID of the container.
-For example: for a Linux container with an ID of `173975398351`, there will be a corresponding directory: `/run/opencontainer/containers/173975398351`.
-Within each container's directory, there MUST be a JSON encoded file called `state.json` that contains the runtime state of the container.
-For example: `/run/opencontainer/containers/173975398351/state.json`.
+The state of a container MUST include, at least, the following propeties:
 
-The `state.json` file MUST contain all of the following properties:
-
-* **`version`**: (string) is the OCF specification version used when creating the container.
+* **`ociVersion`**: (string) is the OCI specification version used when creating the container.
 * **`id`**: (string) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
@@ -23,37 +19,111 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
 
-*Example*
-
+When serialized in JSON, the format MUST adhere to the following pattern:
 ```json
 {
-    "version": "0.2.0",
-    "id": "oc-container",
+    "ociVersion": "0.2.0",
+    "id": "oci-container1",
     "pid": 4422,
     "bundlePath": "/containers/redis"
 }
 ```
 
+See [Query State](#query-state) for information on retrieving the state of a container.
+
 ## Lifecycle
 The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
 
-1. OCI compliant runtime is invoked by passing the bundle path as argument.
-2. The container's runtime environment is created according to the configuration in [`config.json`](config.md).
-   Any updates to `config.json` after container is running do not affect the container.
-3. The container's state.json file is written to the filesystem.
-4. The prestart hooks are invoked by the runtime.
-   If any prestart hook fails, then the container is stopped and the lifecycle continues at step 8.
-5. The user specified process is executed in the container.
-6. The poststart hooks are invoked by the runtime.
-   If any poststart hook fails, then the container is stopped and the lifecycle continues at step 8.
-7. Additional actions such as pausing the container, resuming the container or signaling the container may be performed using the runtime interface.
-   The container could also error out or crash.
-8. The container is destroyed by undoing the steps performed during create phase (step 2).
-9. The poststop hooks are invoked by the runtime and errors, if any, are logged.
-10. The state.json file associated with the container is removed and the return code of the container's user specified process is returned or logged.
+1. OCI compliant runtime is invoked with a reference to the location of the bundle.
+   How this reference is passed to the runtime is an implementation detail.
+2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
+   Any updates to `config.json` after container is running MUST not affect the container.
+3. The prestart hooks MUST be invoked by the runtime.
+   If any prestart hook fails, then the container MUST be stopped and the lifecycle continues at step 8.
+4. The user specified process MUST be executed in the container.
+5. The poststart hooks MUST be invoked by the runtime.
+   If any poststart hook fails, then the container MUST be stopped and the lifecycle continues at step 8.
+6. Additional actions such as pausing the container, resuming the container or signaling the container MAY be performed using the runtime interface.
+   The container MAY also error out, exit or crash.
+7. The container MUST be destroyed by undoing the steps performed during create phase (step 2).
+8. The poststop hooks MUST be invoked by the runtime and errors, if any, MAY be logged.
 
 Note: The lifecycle is a WIP and it will evolve as we have more use cases and more information on the viability of a separate create phase.
 
+## Operations
+
+OCI compliant runtimes MUST support the following operations, unless the operation is not supported by the base operating system.
+
+### Errors
+In cases where the specified operation generates an error, this specification does not mandate how, or even if, that error is returned or exposed to the user of an implementation.
+Unless otherwise stated, generating an error MUST leave the state of the environment as if the operation were never attempted - modulo any possible trivial ancillary changes such as logging.
+
+### Query State
+
+`state <container-id>`
+
+This operation MUST generate an error if it is not provided the ID of a container.
+This operation MUST return the state of a container as specified in the [State](#state) section.
+In particular, the state MUST be serialized as JSON.
+
+
+### Start
+
+`start <container-id> <path-to-bundle>`
+
+This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
+If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error.
+Using the data in `config.json`, that are in the bundle's directory, this operation MUST create a new container.
+This includes creating the relevant namespaces, resource limits, etc and configuring the appropriate capabilities for the container.
+A new process within the scope of the container MUST be created as specified by the `config.json` file otherwise an error MUST be generated.
+
+Attempting to start an already running container MUST have no effect on the container and MUST generate an error.
+
+### Stop
+
+`stop <container-id>`
+
+This operation MUST generate an error if it is not provided the container ID.
+This operation MUST stop and delete a running container.
+Stopping a container MUST stop all of the processes running within the scope of the container.
+Deleting a container MUST delete the associated namespaces and resources associated with the container.
+Once a container is deleted, its `id` MAY be used by subsequent containers.
+Attempting to stop a container that is not running MUST have no effect on the container and MUST generate an error.
+
+### Exec
+
+`exec <container-id> <path-to-json>`
+
+This operation MUST generate an error if it is not provided the container ID and a path to the JSON describing the process to start.
+The JSON describing the new process MUST adhere to the [Process configuration](config.md#process-configuration) definition.
+This operation MUST create a new process within the scope of the container.
+If the container is not running then this operation MUST have no effect on the container and MUST generate an error.
+Executing this operation multiple times MUST result in a new process each time.
+Example:
+```
+{
+    "terminal": true,
+    "user": {
+        "uid": 0,
+        "gid": 0,
+        "additionalGids": null
+    },
+    "args": [
+        "/bin/sleep",
+        "60"
+    ],
+    "env": [
+        "version=1.0"
+    ],
+    "cwd": "...",
+}
+```
+This specification does not manadate the name of this JSON file.
+See the specification of the `config.json` file for the definition of these fields.
+The stopping, or exiting, of these secondary process MUST have no effect on the state of the container.
+In other words, a container (and its PID 1 process) MUST NOT be stopped due to the exiting of a secondary process.
+
 ## Hooks
 
-See [runtime configuration for hooks](./config.md#hooks)
+Many of the operations specified in this specification have "hooks" that allow for additional actions to be taken before or after each operation.
+See [runtime configuration for hooks](./config.md#hooks) for more information.


runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
 }
 ```
 
-## Lifecycle
+## Operations
+
+OCI compliant runtimes MUST support the following operations, unless support for the operation can not be supported by the base operating system (kernel) itself.
+
+OPEN: we should consider specifying the options/args for each operation.
+For example, checkpoint (for runc) seems to have a lot of params.
```
wking: There's a [mailing list thread][1] for this command-line specification, and initial work [in this Gist][2].

[2]: https://gist.github.com/wking/1d69118ba8b750f85bfc

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
 
 ### Create
 
-Creates the container: file system, namespaces, cgroups, capabilities.
+Using the `config.json` and `runtime.json` files, along with the file system path referenced in the `config.json`, this operation MUST create a new container.
+This includes creating the relevant namespaces, cgroups and configuring the appropriate capabilities for the container.
+
+### Start
+
+This operation MUST start a previously created container.
```
wking: I'm personally in favor of having distinct “create” and “start” steps, but [we don't have that in runC now][1].  I'm in favor of [just echoing runC's approach now to get some lifecycle docs landed][2], after which we can have more focused discussions about things like this.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/frUXLljXy8Y/8pH-81moGgAJ
[2]: https://github.com/opencontainers/specs/pull/207#discussion_r41798699

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+Starting a container MUST create a new process, as specified by the `config.json` file within the scope of the container.
+Attempting to start an already running container MUST have no effect on the container and MUST generate an error.
+Starting a stopped container, one that had been previously be running, MUST create a new process, as specified by the `config.json` file, within the scope of the container.
```
wking: Separating “launch a process” from “create a container” is not possible when a Linux container includes a PID namespace (that requires a process to hold it open).  See [here][1] for more details.

[1]: https://github.com/opencontainers/specs/issues/115#issuecomment-147553272

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+OPEN: the entire "generate an error" stuff in here is something we should discuss.
+My initial thought was that it would be better to generate an error so that the call would know exactly what happened - meaning, did something new take place or was the container already running. If we didn't return an error (or something) then the caller wouldn't know.  And they can always ignore errors if they want/need to.
+
+### Stop
+
+This operation MUST stop a running container.
+Stopping a container MUST stop all of the processes running within the scope of the container.
```
wking: On Linux do you mean “[freeze][1]”, “`SIGSTOP`”, “`SIGTERM`”, or something else?  SIGTERM will remove the PID namespace which will partially destroy the container (e.g. you'll have to set that up again if you want to launch new processes in a new container recycling the first container's persistent portions).

[1]: https://www.kernel.org/doc/Documentation/cgroups/freezer-subsystem.txt

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+This operation MUST delete a container.
+Deleting a container MUST delete the associated namespaces and cgroups for the container.
+Deleting a container that is not stopped MUST first stop the container as specified by the `stop` operation.
+If the implied `stop` operations fails then the `delete` operation MUST fail.
+
+### Pause
+
+This operation MUST suspend all processes that are running within the container.
+If the container is not running, either stopped or aleady paused, then this operation MUST have no effect on the container and MUST generate an error.
```
wking: I'd rather have pause be idempotent.  If you wanted a paused container and it's already paused, hooray, success ;).  But I'm ok with anything that gives folks a way to automatically distinguish between “we couldn't pause that container” and “it was already paused”.

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+### Unpause
+
+This operation MUST resume all processes that were previously paused via the `pause` operation.
+If the container is not paused then this operation MUST have no effect on the container and MUST generate an error.
```
wking: I think this should be idempotent [too][1].  See also the [freezer docs][2], which make it sound like writes to `freezer.state` are idempotent.

[1]: https://github.com/opencontainers/specs/pull/225/files#r41900419
[2]: https://www.kernel.org/doc/Documentation/cgroups/freezer-subsystem.txt

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+### Exec
+
+This operation MUST create a new process within the scope of the container.
+If the container is not running then this operation MUST have no effect on the container and MUST generate an error.
+Executing this operation multiple times MUST result in a new process each time.
+Unlike the container's main process which is specified in the `config.json` file, this specification does not define how the `exec` process is defined.
```
wking: Once you've separated “create” from “start”, I don't think you need distinct “exec” command.  Just “start” with a more join-focused config instead of a more creation-focused config.

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+### Checkpoint
+
+This operation MUST create a checkpoint of a running container.
+If the container is not running then this operation MUST have no effect on the container and MUST generate an error.
+This specification does not specify how, or to where, the checkpoint data is persisted.
```
wking: I think we have enough to digest without getting this deep (and pause/unpause are maybe too deep as well ;).  Can we focus on the create/start/cleanup lifecycle first, and then talk about these more sophisticated actions after those have landed?

wking:
On Tue, Oct 13, 2015 at 10:34:04AM -0700, Doug Davis wrote:
> There are a few "OPEN" question in the doc that I'd like to
> brainstorm on.

I think this belongs in the lifecycle thread [1,2], but I'm in favor
of anything that ends up in some kind of lifecycle docs landing :).  I
left a few comments inline with more detailed suggestions.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/frUXLljXy8Y
[2]: https://github.com/opencontainers/specs/pull/207#issuecomment-147554207

crosbymichael:
@duglin do you think adding things like pause/resume and checkpoint/restore would hurt other platforms that do not have this type of feature set?  Maybe these are somethings that we can leave out for broader adoption and spec out the main usecase of create, start, stop, delete?

duglin:
@crosbymichael that's why I worded the intro paragraph the way I did. I tried to make it say that these ops are only required to be supported if the platform itself supports it. Perhaps some better wording is needed, but what do you think about the general idea?

crosbymichael:
I think it's good overall.  Just don't want to add wording that could prohibit runtimes being compliant for some of these "extra" features.

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
 }
 ```
 
-## Lifecycle
+## Operations
+
+OCI compliant runtimes MUST support the following operations, unless support for the operation can not be supported by the base operating system (kernel) itself.
```
mrunalp: Maybe simplify a bit to "unless the operation is not supported by the base operating system" ?

wking: On Wed, Oct 14, 2015 at 10:55:52AM -0700, Mrunal Patel wrote:
> > +OCI compliant runtimes MUST support the following operations,
> > unless support for the operation can not be supported by the base
> > operating system (kernel) itself.
> 
> Maybe simplify a bit to "unless the operation is not supported by
> the base operating system" ?

+1 to Mrunal's wording, but I'd switch “base” → “host” as well.

runtime.md
philips: The semantics of this get super weird if you allow arbitrary PIDs to join another PID namespace too.

runtime.md
philips: Is freezer idempotent across random tasks that join the namespace later? It seems like it operates per pid so I am not sure how this works if something just joins the namespaces.

wking: On Wed, Oct 14, 2015 at 08:09:09PM -0700, Brandon Philips wrote:
> Is freezer idempotent across random tasks that join the namespace
> later? It seems like it operates per pid so I am not sure how this
> works if something just joins the namespaces.

I'm not sure what namespace you're talking about here, but it is
random processes that join the freezer cgroup later are frozen on
join:

  term1# mkdir /sys/fs/cgroup/freezer/abc

  term2# echo $$ >/sys/fs/cgroup/freezer/abc/cgroup.procs

  term1# echo FROZEN >/sys/fs/cgroup/freezer/abc/freezer.state
  term1# cat /sys/fs/cgroup/freezer/abc/freezer.state 
  FROZEN

  term3# echo $$ >/sys/fs/cgroup/freezer/abc/cgroup.procs

  term1# cat /sys/fs/cgroup/freezer/abc/cgroup.procs 
  15929
  16039
  term1# cat /sys/fs/cgroup/freezer/abc/freezer.state 
  FROZEN
  # ps aux | grep '15929\|16039' | grep -v grep
  root     15929  0.0  0.0  20788  4568 pts/11   D+   20:40   0:00 bash
  root     16039  0.0  0.0  20656  4352 pts/12   D+   20:43   0:00 bash

philips: So, in order for this to work, a process that joins a PID namespace must join the freezer cgroup too?

wking: On Wed, Oct 14, 2015 at 08:53:11PM -0700, Brandon Philips wrote:
> So, in order for this to work, a process that joins a PID namespace
> must join the freezer cgroup too?

I think the current “pause” description in this PR isn't clear enough
to decide whether it's going to work or not in a given situation ;).
If the definition is “pause all the processes in that container's
freezer cgroup”, then yes, it will work ;).  If you want to involve
PID namespaces somehow, then no, it probably won't work.  Personally,
I'm in favor of unified-cgroups, and a cgroup-based definition of
“container processes” [1], in which case the freezer approach will
work.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/n7vxlhn6Q7g

runtime.md
wking: On Wed, Oct 14, 2015 at 08:05:21PM -0700, Brandon Philips wrote:
> > +Stopping a container MUST stop all of the processes running
> >  within the scope of the container.
> 
> The semantics of this get super weird if you allow arbitrary PIDs to
> join another PID namespace too.

If stop is “freeze the process's cgroup, SIGKILL everything in it and
its descendents, and unfreeze the cgroup” it seems to not be
particularly weird.  What do you see going wrong?

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+Starting a stopped container, one that had been previously be running, MUST create a new process, as specified by the `config.json` file, within the scope of the container.
+
+OPEN: the entire "generate an error" stuff in here is something we should discuss.
+My initial thought was that it would be better to generate an error so that the call would know exactly what happened - meaning, did something new take place or was the container already running. If we didn't return an error (or something) then the caller wouldn't know.  And they can always ignore errors if they want/need to.
+
+### Stop
+
+This operation MUST stop a running container.
+Stopping a container MUST stop all of the processes running within the scope of the container.
+Stopping a container MUST NOT delete the associated namespaces or cgroups for the container.
```
crosbymichael: I would not say this today because its the opposite of how it actually works.

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+Attempting to stop a container that is not running MUST have no effect on the container and MUST generate an error.
+
+### Delete
+
+This operation MUST delete a container.
+Deleting a container MUST delete the associated namespaces and cgroups for the container.
+Deleting a container that is not stopped MUST first stop the container as specified by the `stop` operation.
```
crosbymichael: Would we rather error here and make the user stop the container ?

crosbymichael:
@duglin overall I think it looks good.  I would still suggest removing pause/resume and checkpoint/restore until we get more information about other platforms and how they are going to handle some of the actions.  It's always easier to add more later than take somethings away.

runtime.md
dqminh: +1 for this. I dont think delete should be coupled with how to stop an application.

runtime.md
```diff
@@ -33,24 +33,79 @@ This is provided so that consumers can find the container's configuration and ro
...
+This includes creating the relevant namespaces, cgroups and configuring the appropriate capabilities for the container.
+
+### Start
+
+This operation MUST start a previously created container.
+Starting a container MUST create a new process, as specified by the `config.json` file within the scope of the container.
+Attempting to start an already running container MUST have no effect on the container and MUST generate an error.
+Starting a stopped container, one that had been previously be running, MUST create a new process, as specified by the `config.json` file, within the scope of the container.
+
+OPEN: the entire "generate an error" stuff in here is something we should discuss.
```
mrunalp: Generating an error makes sense.

wking: On Tue, Nov 17, 2015 at 04:21:56PM -0800, Mrunal Patel wrote:
> > +OPEN: the entire "generate an error" stuff in here is something we should discuss.
> 
> Generating an error makes sense.

I think we all agree that we want error reporting, but we don't have a
clear picture for how to report errors to the caller.  With both the
standard streams and error code already assigned to the container
process, what channel is the runtime using to report internal errors?

runtime.md
mrunalp: I think depends on whether paths were passed for the namespaces/cgroups or the runtime created those for the container.

wking: On Tue, Nov 17, 2015 at 04:38:22PM -0800, Mrunal Patel wrote:
> > +Stopping a container MUST stop all of the processes running within the scope of the container.
> > +Stopping a container MUST NOT delete the associated namespaces or cgroups for the container.
> 
> I think depends on whether paths were passed for the
> namespaces/cgroups or the runtime created those for the container.

+1.  We should only be removing namespaces and cgroups (and killing
any processes that are still in them or their ancestors) if we're
stopping/deleting the container that created them.  I think everyone's
on board with that (although it's not how the “delete” phrasing
currently reads here).

However, this PR is splitting runC's current cleanup into distinct
“stop” and “delete” actions, and I think that's what @crosbymichael is
objecting to (because runC doesn't currently have a way to
stop-but-not-delete a container).

runtime.md
philips: Weirdness: I put a process into an existing containers namespace but not put it into the same cgroup? What is the behavior?

Maybe I am missing something about freezer.

wking: On Tue, Dec 22, 2015 at 11:13:44PM -0800, Brandon Philips wrote:
> > +Stopping a container MUST stop all of the processes running within the scope of the container.
> 
> Weirdness: I put a process into an existing containers namespace but
> not put it into the same cgroup? What is the behavior?

It doesn't get killed.  Unless one of the namespaces it joined was a
PID namespace, and the init for that PID namespace or one of its
ancestors belonged to the cgroup, in which case the kernel's usual
PID-namespace would SIGKILL your process.

That's not weird though, we were clear that stop would only SIGKILL
processes in the given freezer cgroup, and you didn't join the cgroup.
What would you expect to happen?  It's only weird if you don't explain
the stop algorithm, everybody forms conflicting mental images of what
that algorithm is, and then they are surprised when those images don't
match up.

runtime.md
duglin: @wking can you remove the comments that aren't asking for a specific change? Not sure how many others are like me, but I like to have a clean PR (no outstanding comments that need resolving) before they are merged, and if a PR has a lot of comments that are just "FYI" in nature then its hard to pick out the ones that need action from the ones we can ignore.

runtime.md
duglin: What specific wording change are you suggesting here?

runtime.md
duglin: I purposely used the word "generate" instead of "report" or something else that implies the end user sees it. I think that gets into implementation details. For example, some impl might just log everything and their "user" never sees any of it and I think that's valid from this specs perspective.  The point is that we all agree its an error condition and the action did not happen but how the impl deal with the error is up to it.

runtime.md
wking: On Mon, Jan 04, 2016 at 10:44:32AM -0800, Doug Davis wrote:
> @wking can you remove the comments that aren't asking for a specific
> change?

Will do.  Now that we have a tagged set of open list issues [1,2]
(including the command-line API [3]), I think there's less need for
either my FYI comments or entries like this.  So my actionable
suggestion here is “drop this line in favor of [1] for listing open
high-level discussion”.

[1]: https://wking.github.io/nmbug-oci/
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/18TA4iOf48M
[3]: https://wking.github.io/nmbug-oci/#Features-with-external-work

runtime.md
wking: On Mon, Jan 04, 2016 at 10:47:19AM -0800, Doug Davis wrote:
> >  
> >  ### Create
> >  
> > -Creates the container: file system, namespaces, cgroups, capabilities.
> > +Using the `config.json` and `runtime.json` files, along with the file system path referenced in the `config.json`, this operation MUST create a new container.
> > +This includes creating the relevant namespaces, cgroups and configuring the appropriate capabilities for the container.
> > +
> > +### Start
> > +
> > +This operation MUST start a previously created container.
> > +Starting a container MUST create a new process, as specified by the `config.json` file within the scope of the container.
> > +Attempting to start an already running container MUST have no effect on the container and MUST generate an error.
> > +Starting a stopped container, one that had been previously be running, MUST create a new process, as specified by the `config.json` file, within the scope of the container.
> 
> What specific wording change are you suggesting here?

That the “Create” and “Start” sections are rolled into a single
“Start” entry until we have more clarity on how separate steps would
work.  For my take at unified start/create wording, see [1].  Although
I also point out a number of open issues there (which I suggest we
don't land in this repository until we've sorted them out [2]).

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/frUXLljXy8Y/aKFRX56qAAAJ
     Subject: Re: Documenting container lifecycles
     Date: Sat, 19 Sep 2015 14:01:06 -0700
     Message-ID: <20150919210106.GA30120@odin.tremily.us>
[2]: https://github.com/opencontainers/specs/pull/225/files#r48766534

runtime.md
wking: On Mon, Jan 04, 2016 at 10:49:48AM -0800, Doug Davis wrote:
> > +OPEN: the entire "generate an error" stuff in here is something we should discuss.
> 
> I purposely used the word "generate" instead of "report" or
> something else that implies the end user sees it. I think that gets
> into implementation details. For example, some impl might just log
> everything and their "user" never sees any of it and I think that's
> valid from this specs perspective.  The point is that we all agree
> its an error condition and the action did not happen but how the
> impl deal with the error is up to it.

That sounds like a useful distinction to me, and I think its worth
spelling it in this PR.  How about adding an error-handling section
somewhere (the glossary?  #107) that translates your generate
vs. report disctinction into spec language, and then just use “MUST
generate an error” like you already do in this section.

Once we do that, we can drop the OPEN issue here, and open a mailing
list thread if/when we want something in the spec about the reporting
aspect.

runtime.md
wking: To turn this line sub-thread into an actionable comment [1], I'd
suggest we drop the stop section from this PR until we have
clarification on what processes count as “within the scope of the
container”.  The list thread for that is [2], and it's tagged in [3].

[1]: https://github.com/opencontainers/specs/pull/225/files#r48766534
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/n7vxlhn6Q7g/NPWpAMlxBgAJ
     Subject: Definining "container" and "container processes" on Linux
     Date: Thu, 8 Oct 2015 09:47:08 -0700
     Message-ID: <20151008164708.GM28418@odin.tremily.us>
[3]: https://wking.github.io/nmbug-oci/#Features-in-discussion

runtime.md
wking: To turn this into an [actionable suggestion][1], I'd recommend dropping the exec section from this PR and focusing on mailing-list work to build consensus on splitting start from create ([existing][1] [threads][2]).

[1]: https://github.com/opencontainers/specs/pull/225/files#r48766534
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/frUXLljXy8Y
[3]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/GmdaZ2TtDAAJ

runtime.md
wking: To turn this into an [actionable suggestion][1], I'd recommend dropping this section (and the restore section) until we land clarity around start/stop ;). 

[1]: https://github.com/opencontainers/specs/pull/225/files#r48766534

runtime.md
duglin: up will add a section that talks about generating errors.

runtime.md
duglin: What do other people think?  Given runc just went thru the work to add support for this I'm inclined to say we should include it in v1.0.

runtime.md
duglin: ok - removing the implicit delete() call.

runtime.md
wking: On Mon, Jan 04, 2016 at 12:32:05PM -0800, Doug Davis wrote:
> Given runc just went thru the work to add support for this I'm
> inclined to say we should include it in v1.0.

I'm +1 to landing this by v1, assuming it comes with some testable
implementation (otherwise how do we certify runtime compliance with
the v1 spec?).  I'm just -1 to landing this in this PR before we even
have clarity on what “stop” means ;).

runtime.md
duglin: @crosbymichael ignoring for a moment what the code does today... should stopping a container delete the namespaces and cgroups? Shouldn't delete and stop be distinct actions?

runtime.md
wking: On Mon, Jan 04, 2016 at 12:12:49PM -0800, Doug Davis wrote:
> > +OPEN: the entire "generate an error" stuff in here is something we should discuss.
> 
> up will add a section that talks about generating errors.

The “Errors” section that landed with b669866 → 7a5b1f9 looks good to me.

runtime.md
```diff
@@ -54,6 +54,76 @@ The lifecycle describes the timeline of events that happen from when a container
 
 Note: The lifecycle is a WIP and it will evolve as we have more use cases and more information on the viability of a separate create phase.
 
+## Operations
+
+OCI compliant runtimes MUST support the following operations, unless support for the operation can not be supported by the base operating system (kernel) itself.
```
wking: I still think @mrunalp's [earlier suggestion][1] for simplifying this sentence would be a good change.

[1]: https://github.com/opencontainers/specs/pull/225#discussion_r42028244

runtime.md
wking: Another potential place for the “should we separate create and start?” discussion is #226, although I'd recommend moving that discussion to the mailing list.

runtime.md
crosbymichael: Its kinda a hard question because it goes into how you define a container.  Technically I think its a bad idea to keep namespaces around because its buggy and racy just to make it fit someones made up definition of a container.  We would be jumping through hoops in code and on the system for no real use case other than making the technical reality match something that one of us typed. 

We dont keep around all the memory of a VM when its stopped just because we made a UI with a stop and delete button so I dont think it makes sense to do this.

Just because we want a stop and delete phase does not mean we need to go looking for something that we can keep laying around to clean up later just to support the thought.  Maybe we don't need a stop and delete at all so maybe we should discuss some reasons why people think it should be added.

duglin: Still coming up to speed on the details of this but let's say I have a container that gets stopped and then started again later. If we delete the namespaces/cgroup during the stop, is there anything that can not be reliably recreated upon the next time the container is started?  Any worry about some other namespace/cgroup grabbing something that we'd like to hold on to?  If not, then I think you're probably right.

crosbymichael: What you described is how it works today and this has never come up as an issue except in "wouldn't it be cool if..." discussions.

duglin: In that case, are you suggesting we go all the way and just merge delete() and stop()?

crosbymichael: Yes, for the looks of what delete is defined as right now.  I dont see any advantages right now from having it other than more operations and more complexity for implementer

wking: On Mon, Jan 04, 2016 at 01:42:12PM -0800, Michael Crosby wrote:
> What you described is how it works today and this has never come up
> as an issue except in "wouldn't it be cool if..." discussions.

“Kill everything in the container, removing its namespaces and
cgroups” sounds like “delete” to me, so I don't see much reason to
distinguish between “stop” and “delete” if that's also what “stop”
means.

For a situation in which preserving namespaces is how it works today,
see my create/exec/destroy example [1,2].  One place where this moves
beyond “wouldn't it be cool if…” is that a spec based on this kind of
create/exec/delete separation lets us get out of the hook business and
its ordering issues [3]; folks can use a shell script (or whatever)
for both the runtime actions and the currently-hook actions.

[1]: https://github.com/opencontainers/specs/issues/226#issuecomment-166202135
[2]: https://github.com/wking/ccon/tree/master/examples/good/create-exec-delete-root
[3]: https://github.com/opencontainers/specs/pull/265#discussion_r47372360

runtime.md
duglin: re: idempotent - this might apply to other ops as well, and we can probably go either way - but my thinking was that an OCI impl should be able to tell its user whether any particular action actually had some effect on the container, or if it was ignored - like its already paused.  If the caller decides to (happily) ignore "already done" types of errors then that's its choice, but it may also want to be able to report that condition to its users.  Granted, if so we could just mandate that they check the state of the container first themselves and if its already paused then don't even bother calling runc - if we want ignore the racey aspect of it all.

I'd like to know what others think.

duglin:
Updated.  Merged create/start and stop/delete.
A couple of other questions/comments for folks to weigh in on too.

runtime.md
wking: On Tue, Jan 05, 2016 at 12:47:47PM -0800, Doug Davis wrote:
> re: idempotent…

Can we move this discussion to the list [1]?  It seems like a
high-level discussion [2].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/clXqcULznFU
[2]: https://github.com/opencontainers/specs/#discuss-your-design

runtime.md
```diff
@@ -37,23 +37,66 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+This operation MUST resume all processes that were previously paused via the `pause` operation.
+If the container is not paused then this operation MUST have no effect on the container and MUST generate an error.
+
+### Exec
+
+This operation MUST create a new process within the scope of the container.
+If the container is not running then this operation MUST have no effect on the container and MUST generate an error.
+Executing this operation multiple times MUST result in a new process each time.
+Unlike the container's main process which is specified in the `config.json` file, this specification does not define how the `exec` process is defined.
```
wking: nit: trailing whitespace.

And I don't see any point to declaring an action but not specifying how to configure/trigger it.  How would you test runtimes for conformance with this action?  What do bundle-authors or runtime-callers do to use this feature?

duglin: I'll add something about exec's config file.

runtime.md
```diff
@@ -37,23 +37,66 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+OCI compliant runtimes MUST support the following operations, unless the operation is not supported by the base operating system.
+
+### Errors
+In cases where the specified operation generates an error, this specification does not mandate how, or even if, that error is returned or exposed to the user of an implementation.
+Unless otherwise stated, generating an error MUST leave the state of the environment as if the operation were never attempted - modulo any possible trivial ancillary changes such as logging.
+
+### Start
+
+Using the `config.json` and `runtime.json` files, along with the filesystem path referenced in the `config.json`, this operation MUST create a new container.
```
wking: Which filesystem path?  Is that `root.path`?  I'd rather just link to the config docs (e.g. “Using the [`config.json`](config.md) and [`runtime.json`](runtime-config.md) files, this operation MUST create a new container”).  Why call out a particular setting, since `root` is already discussed in bundle.md and config.md?

runtime.md
```diff
@@ -37,23 +37,66 @@ This is provided so that consumers can find the container's configuration and ro
...
+This includes creating the relevant namespaces, cgroups and configuring the appropriate capabilities for the container.
```
wking: `cgroups` is Linux specific, but these operations should not be.  I'd recommend we go with something more generic like “relevant namespaces, resource limits, etc.” and expect folks to interpret “namespaces” generically here and not as “the Linux namespace implementation”.  The landed glossary (#107) has accepted wording we might want to borrow.

runtime.md
```diff
@@ -37,23 +37,66 @@ This is provided so that consumers can find the container's configuration and ro
...
+A new process in the container's PID namespace MUST be created as specified by the `config.json` file.
```
wking: Why call out the PID namespace?  We've already referenced `config.json` and `runtime.json`.  All of those settings are already documented, and we just need to say that “start” includes applying *all* of the settings in those files.

runtime.md
```diff
@@ -37,23 +37,66 @@ This is provided so that consumers can find the container's configuration and ro
...
+Attempting to start an already running container MUST have no effect on the container and MUST generate an error.
+
+### Stop
+
+This operation MUST stop and delete a running container.
+Stopping a container MUST stop all of the processes running within the scope of the container.
+Deleting a container MUST delete the associated namespaces and cgroups for the container.
```
wking: Another cgroups refernence that I'd remove.  I suggest we follow #231's step 8 and say something like:

> Deleting a container undoes the steps performed by the [start](.#start) action.

runtime.md
wking: On Tue, Jan 05, 2016 at 02:12:40PM -0800, Doug Davis wrote:
> > +Unlike the container's main process which is specified in the `config.json` file, this specification does not define how the `exec` process is defined. 
> 
> I'll add something about exec's config file.

To get this PR landed faster, I suggest pulling the exec action out
into a follow-up PR.  But maybe everyone else thinks we're closer to
an exec-spec consensus than I do ;).

runtime.md
duglin: Well, since the definition of a "container" varies between people I wanted to be clear that as part of starting a container we start a new process, and that process isn't just running **anywhere** (like on the host) its running in the container's PID namespace.  Might be anal to say it, but unless its wrong I think its better to be clear.

wking: On Thu, Jan 07, 2016 at 10:40:43AM -0800, Doug Davis wrote:
> > +A new process in the container's PID namespace MUST be created as specified by the `config.json` file.
>
> Well, since the definition of a "container" varies between people I
> wanted to be clear that as part of starting a container we start a
> new process, and that process isn't just running **anywhere** (like
> on the host) its running in the container's PID namespace.  Might be
> anal to say it, but unless its wrong I think its better to be clear.

I agree that dipping into informative concrete examples is useful, but
I'd rather avoid normative language here when the process and PID
namespace settings are already fully spec'ed in other places (and if
the namespace for config.json's process is not clear, that seems like
something that should be addressed in [1], [2], or both).  How about:

  All of the settings defined in [`config.json`](config.md) and
  [`runtime.json`](runtime.md) are applied in this operation.  For
  example, creating namespaces, setting resource limits, mounting
  filesystems, creating devices, launching a container process,
  setting capabilities for that process, and executing the configured
  command all happen here.

[1]: https://github.com/opencontainers/specs/blob/6a6ba6775567e726c405456e74b0e52350104eda/config.md#process-configuration
[2]: https://github.com/opencontainers/specs/blob/6a6ba6775567e726c405456e74b0e52350104eda/runtime-config-linux.md#namespaces

duglin:
Another round of edits have been made based on the feedback.

LK4D4:
@duglin Am I understand correctly, that this PR will solve "Actions" issue from 0.3.0 milestone?

duglin:
@LK4D4 between this one and Vish's create/start split work, yes I think it'll cover it.

duglin:
Updated this based on last week's f2f.
I couldn't remember if we agreed to keep or drop 'exec' for now I kept it since I think we'll need a variant of it for the create/start split - if we agreed to remove it let me know and I'll drop it.

wking:
On Mon, Jan 18, 2016 at 07:35:05AM -0800, Doug Davis wrote:
> Updated this based on last week's f2f.

Is there more of a write up (or will there be) on the shift to:

* Specifying something that looks like a command-line API in this
  spec.  Is [1] deprecated?  Should I PR this repository with the
  additional command-line API details contained in [1]?

* Taking a container ID instead of a path to state JSON (e.g. --state
  [2,3]).  I think I've outlined the reasons why I don't want to
  require a runtime-maintained registry in those references.

[1]: https://github.com/wking/oci-command-line-api
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/q6TYqVZOcX8/W1RVyCXCCQAJ
     Subject: Re: removal of /run/opencontainer/containers, add --state?
     Date: Tue, 8 Dec 2015 15:49:57 -0800
     Message-ID: <20151208234957.GK2767@odin.tremily.us>
[3]: https://github.com/wking/oci-command-line-api/pull/14

duglin:
@wking there's no agreement yet on a standard cmd line.  What we did agree to was to add some text to the description of the ops to make it look like we were specifying a cmd line, when really we're just providing additional text to help people understand what's needed to execute that op.  That extra text might lead to standardizing the cmd line but not yet - kind of weasel wording but that's where we landed.

runtime.md
```diff
@@ -23,37 +14,127 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+
+`start <path-to-bundle> <container-id>`
+
+This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
+If the ID provided is not unique across all containers running on this host, or is not valid in any other way, the implementation MUST generate an error.
+Using the data in `config.json` and `runtime.json` files, that are in the bundle's directory, this operation MUST create a new container.
+This includes creating the relevant namespaces, resource limits, etc and configuring the appropriate capabilities for the container.
+A new process within the scope of the container  MUST be created as specified by the `config.json` file.
+The current working directory of the process MUST be set to the `cwd` field of the config.
+If the path specified by the `cwd` fields does not exist then an error MUST be generated.
```
wking: I think these **`cwd`** requirements belong in the [**`cwd`** section][1].

[1]: https://github.com/opencontainers/specs/blob/v0.2.0/config.md#process-configuration

wking:
On Mon, Jan 18, 2016 at 11:15:01AM -0800, Doug Davis wrote:
> @wking there's no agreement yet on a standard cmd line.  What we did
> agree to was to add some text to the description of the ops to make
> it look like we were specifying a cmd line, when really we're just
> providing additional text to help people understand what's needed to
> execute that op.

Ok, that makes sense.  I think a list of arguments might be more
in-line with the JSON portion of the specs.  For example, you could
have generic wording that the runtime MUST generate an error and exit
before attempting the action if a required argument was missing.  And
then have just:

  ### Start

  Arguments:

  * **`path-to-bundle`** (required) The path to the bundle containing
    the configuration files for the new container.
  * **`container-id`** (required) The ID to use for the new container.
    If the ID provided is not unique across all containers running on
    this host, or is not valid in any other way, the runtime MUST
    generate an error and exit without attempting the action.

so folks don't confuse the list of arguments with a required
command-line syntax.

On the container-ID-scoping front, the unique-to-host rule you're
suggesting will require the executing-runtime be aware of other
runtimes (e.g. runC erroring out because there is already a runV
container with ID ‘oci-container1’).  If that's too strong a
requirement (and I think it is), here are some other scope options:

* host/runtime: (e.g. runC only has to error out if there's already a
  runC container with ID ‘oci-container1’).  This avoids requiring
  inter-runtime communication.
* host/runtime/user: (e.g. runC called by wking only has to error out
  if there's already a runC container registered by wking with ID
  ‘oci-container1’).  This avoids ID-squatting denial of service,
  where one user can deny another their preferred container IDs by
  starting dummy containers on them (e.g. your ‘nginx-1’ container
  doesn't start because I already have an ‘nginx-1’ container).
* host/runtime/user/namespace: (e.g. runC called by wking in the
  PID/mount/IPC/… or mount namespace with inode 123 only has to error
  out if there's already a runC container registered by wking from the
  PID/mount/IPC/… namespace with inode 123 with ID ‘oci-container1’).
  Pick any namespaces that impact the runtime's chosen state-sharing
  mechanism.
* Don't bother enforcing uniqueness at this level.  I like this choice
  [1,2].

If we do end up requiring a runtime-maintained state registery, I
think we want to be very clear about which level of “host” scoping
we're requiring.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/q6TYqVZOcX8/W1RVyCXCCQAJ
     Subject: Re: removal of /run/opencontainer/containers, add --state?
     Date: Tue, 8 Dec 2015 15:49:57 -0800
     Message-ID: <20151208234957.GK2767@odin.tremily.us>
[2]: wking/oci-command-line-api#14

runtime.md
```diff
@@ -23,37 +14,127 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+8. The poststop hooks MUST be invoked by the runtime and errors, if any, MAY be logged.
 
 Note: The lifecycle is a WIP and it will evolve as we have more use cases and more information on the viability of a separate create phase.
 
+## Operations
+
+OCI compliant runtimes MUST support the following operations, unless the operation is not supported by the base operating system.
+
+### Errors
+In cases where the specified operation generates an error, this specification does not mandate how, or even if, that error is returned or exposed to the user of an implementation.
```
vishh: There is value in standardizing errors or at least error codes, if we aim to achieve interop in the near future.

runtime.md
```diff
@@ -23,37 +14,127 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+Unless otherwise stated, generating an error MUST leave the state of the environment as if the operation were never attempted - modulo any possible trivial ancillary changes such as logging.
+
+### Query State
+
+`state <container-id>`
+
+This operation MUST generate an error if it is not provided the ID of a container.
+This operation MUST return the state of a container as specified in the [State](#state) section.
```
vishh: Should we specify that the output should be `json serialized`?

runtime.md
```diff
@@ -23,37 +14,127 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+`stop <container-id>`
+
+This operation MUST generate an error if it is not provided the container ID.
+This operation MUST stop and delete a running container.
+Stopping a container MUST stop all of the processes running within the scope of the container.
+Deleting a container MUST delete the associated namespaces and resources associated with the container.
+Once a container is deleted, its `id` MAY be used by subsequent containers.
+Attempting to stop a container that is not running MUST have no effect on the container and MUST generate an error.
+
+### Pause
```
vishh: Why is Pause required? Can we not include it in the Spec? Pause requires freezer cgroup to be mounted.

runtime.md
```diff
@@ -23,37 +14,127 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+
+`pause <container-id>`
+
+This operation MUST generate an error if it is not provided the container ID.
+This operation MUST suspend all processes that are running within the container.
+If the container is not running, either stopped or aleady paused, then this operation MUST have no effect on the container and MUST generate an error.
+
+### Unpause
```
vishh: ditto

runtime.md
```diff
@@ -23,37 +14,127 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+
+This operation MUST generate an error if it is not provided the container ID.
+This operation MUST resume all processes that were previously paused via the `pause` operation.
+If the container is not paused then this operation MUST have no effect on the container and MUST generate an error.
+
+### Exec
+
+`exec <container-id> <path-to-json>`
+
+This operation MUST generate an error if it is not provided the container ID and a path to the JSON describing the process to start.
```
vishh: Where is the process specification defined?

runtime.md
wking: On Wed, Jan 20, 2016 at 10:59:19AM -0800, Vish Kannan wrote:
> > +This operation MUST return the state of a container as specified in the [State](#state) section.
> 
> Should we specify that the output should be `json serialized`?

+1, since this PR currently relaxes the State section's wording to a
conditional “When serialized in JSON…” [1].

[1]: https://github.com/duglin/specs/commit/f9064ec953b5dfc040c4c76fb20130710761c139#diff-b84a8d65d8ed53f4794cd2db7e8ea731R17

runtime.md
wking: On Wed, Jan 20, 2016 at 11:00:34AM -0800, Vish Kannan wrote:
> > +### Pause
> 
> Why is Pause required?

I think the plan was to leave it out for now [1,2].

[1]: https://github.com/opencontainers/specs/pull/225#issuecomment-147885232
[2]: https://github.com/opencontainers/specs/pull/225#issuecomment-155245857

runtime.md
wking: On Wed, Jan 20, 2016 at 10:39:48AM -0800, Vish Kannan wrote:
> > +In cases where the specified operation generates an error, this
> > specification does not mandate how, or even if, that error is
> > returned or exposed to the user of an implementation.
>
> There is value in standardizing errors or at least error codes, if
> we aim to achieve interop in the near future.

Yeah, but the less we have to bite off in this PR, the better ;).  We
can always have followup discussion to build consensus around error
reporting.

duglin: I say that the state must be serialize as JSON up above in the "state" section.  But I'll make it explicit in here too.

runtime.md
duglin: I say that the state must be serialize as JSON up above in the "state" section. But I'll make it explicit in here too.

runtime.md
vishh: This PR is acting as a base for lots of discussions already. If we were to go with your logic, we should split this PR into several small ones.

wking: On Thu, Jan 21, 2016 at 10:16:09AM -0800, Vish Kannan wrote:
> This PR is acting as a base for lots of discussions already. If we
> were to go with your logic, we should split this PR into several
> small ones.

Sounds good to me :).

runtime.md
wking: On Thu, Jan 21, 2016 at 10:13:28AM -0800, Doug Davis wrote:
> > +This operation MUST return the state of a container as specified in the [State](#state) section.
> 
> I say that the state must be serialize as JSON up above in the
> "state" section. But I'll make it explicit in here too.

The State wording (as of f9064ec9) says “When serialized as JSON …”,
which implies (to me) that we're ok with other serializations where
appropriate.  I'm fine with that distinction for the state in general
(I'd be ok with runtime extention like --format=… that selected a
different state serialization), but think we need an explicit
requirement here for returning JSON when the action is invoked with
whatever the OCI-specified invocation method turns out to be ;).

runtime.md
duglin: I think its important to say it here because this is where people might think its ok to create the dir for them.

runtime.md
duglin: oops forgot that we agreed to remove it - will remove.

runtime.md
duglin: doesn't the test below cover it?

runtime.md
duglin: I added some text mandating it.

runtime.md
```diff
@@ -23,37 +14,112 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+In cases where the specified operation generates an error, this specification does not mandate how, or even if, that error is returned or exposed to the user of an implementation.
+Unless otherwise stated, generating an error MUST leave the state of the environment as if the operation were never attempted - modulo any possible trivial ancillary changes such as logging.
+
+### Query State
+
+`state <container-id>`
+
+This operation MUST generate an error if it is not provided the ID of a container.
+This operation MUST return the state of a container as specified in the [State](#state) section.
+In particular, the state MUST be serialize in JSON.
```
wking: “serialize in” → “serialized as”?

runtime.md
wking: On Thu, Jan 21, 2016 at 10:35:22AM -0800, Doug Davis wrote:
> > +A new process within the scope of the container  MUST be created as specified by the `config.json` file.
> > +The current working directory of the process MUST be set to the `cwd` field of the config.
> > +If the path specified by the `cwd` fields does not exist then an error MUST be generated.
> 
> I think its important to say it here because this is where people
> might think its ok to create the dir for them.

Hopefully runtime-authors implementing ‘cwd’ read both this section
and the ‘cwd’ section [1] ;).  And if we don't mention ‘cwd’ here, the
only place bundle-authors will hear about it is [1].  So I'd rather
leave it out here and avoid splitting the normative ‘cwd’ spec between
two sections.

[1]: https://github.com/opencontainers/specs/blob/v0.2.0/config.md#process-configuration

runtime.md
duglin: fixed

runtime.md
duglin: The problem is, the config.md file just specifies what's in the file not, necessarily, how you process each field during a create.  I think those are two distinct conversations otherwise there will be dup text.

runtime.md
wking: On Thu, Jan 21, 2016 at 10:36:34AM -0800, Doug Davis wrote:
> > +This operation MUST generate an error if it is not provided the
> > container ID and a path to the JSON describing the process to
> > start.
>
> doesn't the test below cover it?

It's enough to work from, especially since it looks a lot like [1].
But I'd rather explicitly reference [1] and say “the config JSON MUST
contain a ‘process’ entry as it's root object” and then soften your
“MUST adhere to the following format” to an informative example.

See also [2], where I argue for a single process schema that we use
wherever we need process information (vs. a custom process schema for
each location).

[1]: https://github.com/opencontainers/specs/blob/v0.2.0/config.md#process-configuration
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Cn5L6g1prgA
     Subject: Unify container-process and hook-process structures
     Date: Mon, 4 Jan 2016 21:56:37 -0800
     Message-ID: <20160105055637.GG6362@odin.tremily.us>

runtime.md
wking: On Thu, Jan 21, 2016 at 10:47:38AM -0800, Doug Davis wrote:
> The problem is, the config.md file just specifies what's in the file
> not, necessarily, how you process each field during a create.  I
> think those are two distinct conversations otherwise there will be
> dup text.

I think config.md should specify both, which gets you testable clarity
on what a field means all collected in a single place.  Putting it
another way, any information about create-time field handling
(e.g. “it is an error to specify a directory that doesn't exist by
exec-time”) is important to anyone interacting with that field.

wking:
On Mon, Jan 18, 2016 at 11:47:12AM -0800, W. Trevor King wrote:
> * host/runtime/user/namespace: (e.g. runC called by wking in the
>   PID/mount/IPC/… or mount namespace with inode 123 only has to
>   error out if there's already a runC container registered by wking
>   from the PID/mount/IPC/… namespace with inode 123 with ID
>   ‘oci-container1’).  Pick any namespaces that impact the runtime's
>   chosen state-sharing mechanism.

We didn't settle on suggested spec wording in today's meeting, but it
sounded like this was the option we would require in the spec.  It's
basically the most narrowly scoped state registry you can have without
decoupling the registry from the runtime (my preferred path) [1].

And this is the minimum required scoping; runtimes are free to return
state in other cases (e.g. when asked by a different user, or from a
different namespace) if they want to.

[1]: https://github.com/opencontainers/specs/pull/225#issuecomment-172633075

duglin:
rebased

wking:
GitHub seems to have eaten a comment I mailed 15 minutes ago, so reposting via the web UI:

On Thu, Jan 28, 2016 at 10:00:33AM -0800, Doug Davis wrote:
> rebased

      $ git diff --word-diff=color 2614bce..324866d -- runtime.md

has the rebase looking like it's just adapting to #284.  I think we're
still missing:

* Explicit docs around the exec JSON (1).
* Wording around the state-registry's scoping (2).

And I'd like to see, but am ok with punting from this PR:

* A more detailed, less command-line-like argument display (3).
* No required state registry in the runtime (3).
* Consolidating the process.cwd specs in config.md (4).
* More clarity on how stopping works (5), what processes are “within
  the scope of the container” (6), etc.

You probably also want to remove the brainstorm comment from the
commit message, now that this is getting close to landing.

1: https://github.com/opencontainers/specs/pull/225/files#r50444323
2: https://github.com/opencontainers/specs/pull/225#issuecomment-175790463
3: https://github.com/opencontainers/specs/pull/225#issuecomment-172633075
4: https://github.com/opencontainers/specs/pull/225#discussion_r50444816
5: https://github.com/opencontainers/specs/pull/225#discussion_r41900150
6: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/n7vxlhn6Q7g
     Subject: Definining "container" and "container processes" on Linux
     Date: Thu, 8 Oct 2015 09:47:08 -0700
     Message-ID: <20151008164708.GM28418@odin.tremily.us>

duglin:
updated with @wking's 2 must haves

wking:
GitHub does not like my email today.  Reposting a comment I sent half an hour ago:

On Thu, Jan 28, 2016 at 12:31:50PM -0800, Doug Davis wrote:
> updated with @wking's 2 must haves

Thanks :).  324866d → 1862670 has wording that looks good to me for
the exec JSON.

I think the registry-scoping text needs some tweaking though.  For one
thing, isn't exec supposed to be optional (dependant on platform
support) (1)?

1: https://github.com/opencontainers/specs/blob/v0.2.0/ROADMAP.md#define-standard-container-actions

runtime.md
```diff
@@ -23,37 +19,113 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
...
+### Start
+
+`start <container-id> <path-to-bundle>`
+
+This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
+If the ID provided is not unique across all containers running on this host, or is not valid in any other way, the implementation MUST generate an error.
+Using the data in `config.json`, that are in the bundle's directory, this operation MUST create a new container.
+This includes creating the relevant namespaces, resource limits, etc and configuring the appropriate capabilities for the container.
+A new process within the scope of the container  MUST be created as specified by the `config.json` file.
+The current working directory of the process MUST be set to the `cwd` field of the config.
```
crosbymichael: It is weird that the `cwd` is called out here.  Why pick out one field to document in this section?  Can we just remove these two sentences?

crosbymichael:
Except for the `cwd` sentences this:

LGTM

runtime.md
wking: On Thu, Jan 28, 2016 at 03:11:12PM -0800, Michael Crosby wrote:
> > +The current working directory of the process MUST be set to the `cwd` field of the config.
>
> It is weird that the `cwd` is called out here.

I agree.  There is some previous discussion starting at [1][1] (to avoid
rehashing that here).

[1]: https://github.com/opencontainers/specs/pull/225#discussion_r50035796

duglin:
ok cwd sentences are gone.

crosbymichael:
Thanks

LGTM

duglin:
I just remember why I added that sentence about the cwd - I think we need to be clear that runtimes are not allowed to create that dir for people. If it doesn't exit then start() needs to generate an error.  I modified config.md to say this but if you guys would prefer i can move that to a different PR.

crosbymichael:
we can do that in another place.  I think that type of requirement is valid for any things.  Such as if the process does not exist that the user specifies an error is returned.  

We can probably take a pass at structured error codes for things in like a .4/.5 release because it will be another large undertaking .

duglin:
ok

wking:
I've filed suggestions addressing [my concerns with the recent state-scope wording][1] as a fix-up commit against this branch in duglin/specs#1.

[1]: https://github.com/opencontainers/specs/pull/225#issuecomment-176420592

mrunalp:
LGTM

mrunalp:
@philips @vbatts @hqhq @LK4D4 PTAL

hqhq:
LGTM

duglin:
Modified the sentence on line 75 ( https://github.com/opencontainers/specs/pull/225/files#diff-b84a8d65d8ed53f4794cd2db7e8ea731R75 ) concerning name uniqueness per a comment from @wking.

LK4D4:
LGTM

runtime.md
```diff
@@ -23,37 +19,111 @@ This allows the hooks to perform cleanup and teardown logic after the runtime de
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
 
-*Example*
-
+When serialized in JSON, the format MUST adhere to the following pattern:
```
AkihiroSuda, line 22: What was the intent of this sentence? Does "the following pattern" include the indentation pattern too?
- https://github.com/opencontainers/runtime-spec/issues/1177

duglin: commented in the issue

PR #229, opened 2015-10-22
add 'mandatory tag' and implement bundle test by using tag-reflect

liangchenye:


mrunalp:
@liangchenye Thanks for the PR. I am planning to create a PR for runtime validation shortly. We should probably discuss merging the two.

config.go
```diff
@@ -4,67 +4,67 @@ package specs
 // independent configuration.
 type Spec struct {
 	// Version is the version of the specification that is supported.
-	Version string `json:"version"`
+	Version string `json:"version" mandatory:"required"`
```
vbatts: i really like this usage of golang's tags.

vbatts:
@philips have you pushed a branch from the session we had yet?

liangchenye:
Since we are using golang, I suggest  we can use 'pointer' to refer the 'optional' structure.

And can we make the how directory like this?
specs/
  |___  spec (all *md)
  |___  schema  (all *go)

ChengTiesheng:
Quite agree with @liangchenye, specs can focus on drafting standards, around the specs, the representation implementations can be arbitrarily, such as Golang or C.

vbatts:
@liangchenye yes. This is something I would like to see done. And grouping of specs as well, since there are aspects of the bundle that will be distinct from the runtime of the configuration file. Also as the lifecycle of the runtime is written up, it would be separate from the bundle. These individual pieces will live under the `./specs/` directory.

config.go
liangchenye: Since the similar 'tag' will be generated automaticly by the protobuf golang compiler, and proto.Marshal makes thing easier. I'll write another implementation based on that.

liangchenye:
Close, and will submit to Mrunal's repo

PR #231, opened 2015-10-22
Add lifecycle for containers

mrunalp:
The lifecycle described is generic and should apply to all platforms.
It provides leeway for the runtimes to be flexible in how they
implement it.

Thanks to @crosbymichael for the feedback!

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
 ```
 
 ## Lifecycle
+The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
 
-### Create
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created according to the configuration in config.json.
+   Any updates to config.json after container is running do not affect the container.
+3. The container's state.json file is written to the filesystem under /run/opencontainer/<runtime>/containers/<id>/.
+4. The prestart hooks are invoked by the runtime.
+   If any prestart hook fails, then the container is stopped and the lifecycle continues at step 8.
+5. The user specified process is executed in the container.
+6. The poststart hooks are invoked by the runtime.
+   If any poststart hook fails, then the container is stopped and the lifecycle continues at step 8.
+7. Additional actions such as pausing the container, resuming the container or signaling the container may be performed using the runtime interface.
+   The container could also error out or crash.
+8. The container is destroyed by undoing the steps performed during create phase (step 2).
+9. The poststop hooks are invoked by the runtime and errors, if any, are logged.
+10. The state.json file associated with the container is removed and the return code of the container's user specified process is returned or logged.
 
-Creates the container: file system, namespaces, cgroups, capabilities.
-
-### Start (process)
-
-Runs a process in a container.
-Can be invoked several times.
-
-### Stop (process)
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+Note: The lifecycle is a WIP and it will evolve as we have more use cases and more information on the viability of a separate create phase.
 
 ## Hooks
 


runtime.md
```diff
@@ -34,22 +34,20 @@ This is provided so that consumers can find the container's configuration and ro
...
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is started to when it is stopped or crashes unexpectedly.
+
+* OCI compliant runtime is invoked by passing the bundle path as argument.
+* The container's runtime environment is created.
+* The prestart hooks are invoked in the host environment by the runtime.
```
vishh: @mrunalp: If we split `creation` from `start`, then why can't users directly invoke the hooks?

runtime.md
```diff
@@ -34,22 +34,20 @@ This is provided so that consumers can find the container's configuration and ro
...
-### Stop (process)
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is started to when it is stopped or crashes unexpectedly.
+
+* OCI compliant runtime is invoked by passing the bundle path as argument.
+* The container's runtime environment is created.
```
wking: We probably want an enumerated list here, instead of an unordered list.  That makes it easier to explain jumping between entries (e.g. if the container process dies while the pre-start hooks are executing).

runtime.md
```diff
@@ -34,22 +34,20 @@ This is provided so that consumers can find the container's configuration and ro
...
+* The prestart hooks are invoked in the host environment by the runtime.
+  If any prestart hook fails, then the container is killed and an error code is returned.
+* The user specified process is executed in the container.
+* The poststart hooks are invoked in the host environment by the runtime.
+  If any poststart hook fails, then the container is killed and an error code is returned.
+* Additional actions such as pausing the container, checkpointing the container, resuming the container or signaling the container may be performed using the runtime interface.
+  The container could also error out or crash.
+* The container is destroyed by undoing the steps performed during create phase if required.
+* The poststop hooks are invoked in the host environment and errors, if any, are logged.
```
wking: Since all hooks are invoked in the host environment, I'd just add that documentation to the hooks section (e.g. [my attempt here][1]).  Documenting that in *both* locations can't hurt ;), but having that documented only in the lifecycle section seems odd.

[1]: https://github.com/opencontainers/specs/pull/207/files#diff-695bac0ce528e5c6765a6666fcccc425R15

runtime.md
```diff
@@ -34,22 +34,20 @@ This is provided so that consumers can find the container's configuration and ro
...
+* The state associated with the container is removed and the return code of the container is returned or logged.
```
wking: It's not clear (as of 72fa3086) when the state file is written.  We should probably add an explicit step for that if we're going to tell people when to clean it up.

runtime.md
mrunalp: @vishh This is just a timeline of container lifecycle. For some runtimes like runc, this would be an internal implementation detail of what happens during start. I think we can tackle whether we want a separate create when we discuss the actions/API/interface.

runtime.md
mrunalp: We will just add that to hooks section if required. Taking out unnecessary details from here.

mrunalp:
Updated.

runtime.md
vishh: @mrunalp: But isn't that part of the lifecycle? When I think of a container in the OCI context, I assume this PR will describe its lifecycle.
 Can you explain why you think this is an implementation detail?

runtime.md
```diff
@@ -34,22 +34,21 @@ This is provided so that consumers can find the container's configuration and ro
...
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is started to when it is stopped or crashes unexpectedly.
+
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created.
+3. The container's state is written to the filesystem.
+4. The prestart hooks are invoked by the runtime.
+   If any prestart hook fails, then the container is killed (step 8) and an error code is returned.
```
wking: I think this is just “the container is killed and the lifecycle continues at step 8”.  Actually killing the container would be part of step 7 (where you talk about signaling the container), and the rest of the activity after step 8 would happen in the “failed pre-start hook” case too.

mrunalp:
@philips @vishh @vbatts PTAL

mrunalp:
@LK4D4 PTAL

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
+4. The prestart hooks are invoked by the runtime.
+   If any prestart hook fails, then the container is killed and the lifecycle continues at step 8.
+5. The user specified process is executed in the container.
+6. The poststart hooks are invoked by the runtime.
+   If any poststart hook fails, then the container is killed and the lifecycle continues at step 8.
+7. Additional actions such as pausing the container, checkpointing the container, resuming the container or signaling the container may be performed using the runtime interface.
+   The container could also error out or crash.
+8. The container is destroyed by undoing the steps performed during create phase if required.
+9. The poststop hooks are invoked by the runtime and errors, if any, are logged.
+10. The state associated with the container is removed and the return code of the container is returned or logged.
```
duglin: s/of the container/of the container's user specific process/

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
-### Stop (process)
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is started to when it is stopped or crashes unexpectedly.
+
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created.
```
duglin: Should we add?
```
...is create according to the configuration specified in the 'config.json' file.
This file is only processed (in a read-only fashion) during the container's creation processing, and any
external changes made to this file once the container is created will not have any effect on the container.
```

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
+3. The container's state is written to the filesystem.
```
duglin: Let's be specific and say where the `state.json` file resides - e.g. `/run/opencontainers/runc/<containerID>/state.json`

runtime.md
duglin: s/state/state file (`state.json`)/

duglin:
In order to avoid things getting out of sync, and duplicating information, I think it would make sense to combine the lifecycle information here with the definition of our ops, from `runtime.md`.  Unless there's a reason to keep them separated? But they seem pretty tightly linked to me, at least until we get to the point where "ops" are exposed via some API/cmd-line spec, which we're not doing yet.

duglin:
In particular, I think it would make sense to pull in the lower-level details mentioned in #225.  So each step could have a one-liner summary followed but a more detailed list of the technical details.  Seeing it all in one spot keeps it simple, aligned and easier to follow.

runtime.md
wking: On Wed, Oct 28, 2015 at 07:36:30AM -0700, Doug Davis wrote:
> > +2. The container's runtime environment is created.
> 
> Should we add?
> ```
> ...is create according to the configuration specified in the
> 'config.json' file.  This file is only processed (in a read-only
> fashion) during the container's creation processing, and any
> external changes made to this file once the container is created
> will not have any effect on the container.
> ```

It also depends on runtime.json (since #88).  But the rest of that
statement still holds after “the 'config.json' file” → “`config.json`
and `runtime.json`”).

runtime.md
wking: On Wed, Oct 28, 2015 at 07:37:11AM -0700, Doug Davis wrote:
> > +3. The container's state is written to the filesystem.
> 
> Let's be specific and say where the `state.json` file resides -
> e.g. `/run/opencontainers/runc/<containerID>/state.json`

There is no specific location for non-Linux OSes at the moment.  See
[1] (in an outdated diff on #211), where @vbatts says we're waiting
for PoC implementations on other OSes before specifying paths for
them.  And there's been recent talk about making this write optional
as well, to support users who don't have permissions to write to a
system-wide directory.

[1]: https://github.com/opencontainers/specs/pull/211#discussion_r41067134

wking:
On Wed, Oct 28, 2015 at 07:52:11AM -0700, Doug Davis wrote:
> In particular, I think it would make sense to pull in the
> lower-level details mentioned in #225.  So each step could have a
> one-liner summary followed but a more detailed list of the technical
> details.  Seeing it all in one spot keeps it simple, aligned and
> easier to follow.

I like the approach with a one-liner overview linking out to more
details for each step (which is how I setup #207 [1]).  That makes it
easy to get the big picture and still drill down into the individual
entries for more detail.

[1]: https://github.com/wking/opencontainer-specs/blob/717a99992aa675a4f50563154b3ba83b0ec2ff3b/runtime.md#typical-lifecycle

crosbymichael:
LGTM

mrunalp:
@vishh @LK4D4 PTAL. As discussed in the call last week, we can start with something simple here that we can refine.

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
-Runs a process in a container.
-Can be invoked several times.
-
-### Stop (process)
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is started to when it is stopped or crashes unexpectedly.
```
vishh: nit: `to when it ceases to exist` maybe?

vishh: nit: How about `container is created` instead of `started`?

wking: On Tue, Nov 17, 2015 at 04:21:21PM -0800, Vish Kannan wrote:
> > -This event needs to be captured by runc to run onstop event handlers.
> > +The lifecycle describes the timeline of events that happen from when a container is started to when it is stopped or crashes unexpectedly.
> 
> nit: How about `container is created` instead of `started`?

Both of Vish's suggestions sound like improvements to me.

mrunalp: Updated.

runtime.md
wking: While we're tweaking the text, both of @duglin's suggestions look good to me too.

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
+
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created.
+3. The container's state is written to the filesystem.
+4. The prestart hooks are invoked by the runtime.
```
vishh: This step is subject to change based on namespaces pinning POC right? Shall we explicitly mention that?

vishh: Why are we using the keyword `runtime` here? Are we precluding the Spec from ever being used by humans :smirk: ?

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
+   If any prestart hook fails, then the container is killed and the lifecycle continues at step 8.
+5. The user specified process is executed in the container.
+6. The poststart hooks are invoked by the runtime.
```
vishh: What do you mean by `runtime`? Can we not use that keyword?

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
+   If any poststart hook fails, then the container is killed and the lifecycle continues at step 8.
+7. Additional actions such as pausing the container, checkpointing the container, resuming the container or signaling the container may be performed using the runtime interface.
```
vishh: Checkpointing and restoring is essentially a new phase and it might require additional clarification. Shall we exclude those features from the initial version?

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
+   The container could also error out or crash.
+8. The container is destroyed by undoing the steps performed during create phase if required.
```
vishh: What do you mean by `if required`?

runtime.md
```diff
@@ -34,22 +34,23 @@ This is provided so that consumers can find the container's configuration and ro
...
-
-### Stop (process)
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
+
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
```
vishh: What is a `runtime`? To me, runtime seems to indicate higher level systems.

runtime.md
vishh: By `state` I assume you are only referring to the internal `state`. Any changes that the spawned processed introduced to the root filesystem will not be cleaned up right?

runtime.md
wking: On Tue, Nov 17, 2015 at 05:33:53PM -0800, Vish Kannan wrote:
> > +7. Additional actions such as pausing the container,
> >     checkpointing the container, resuming the container or
> >     signaling the container may be performed using the runtime
> >     interface.
> 
> Checkpointing and restoring is essentially a new phase and it might
> require additional clarification. Shall we exclude those features
> from the initial version?

+1

runtime.md
wking: On Tue, Nov 17, 2015 at 05:34:15PM -0800, Vish Kannan wrote:
> > +8. The container is destroyed by undoing the steps performed
> >     during create phase if required.
> 
> What do you mean by `if required`?

For example, the create phase might have created a PID namespace, but
the kernel has already cleaned it up after the container process died.
In that case, there is nothing for the runtime to do for cleanup.

runtime.md
wking: On Tue, Nov 17, 2015 at 05:37:56PM -0800, Vish Kannan wrote:
> > +1. OCI compliant runtime is invoked by passing the bundle path as
> >     argument.
> 
> What is a `runtime`? To me, runtime seems to indicate higher level
> systems.

It's “tools like runC”, see “compliant runtime” in [1] and #107.  What
do you call those?

[1]: https://github.com/opencontainers/specs/blob/v0.1.1/bundle.md

runtime.md
wking: On Tue, Nov 17, 2015 at 05:41:08PM -0800, Vish Kannan wrote:
> > +10. The state associated with the container is removed and the
> >      return code of the container is returned or logged.
> 
> By `state` I assume you are only referring to the internal
> `state`. Any changes that the spawned processed introduced to the
> root filesystem will not be cleaned up right?

+1 to not cleaning up bundle-side filesystem changes, and another
reason why @duglin's state.json clarification is useful.

runtime.md
mrunalp: There is a note at the bottom to this effect.

runtime.md
mrunalp: okay, taking that out for now.

runtime.md
mrunalp: Taking out 'if required' phasing. The intent was that if a namespace is passed in by path then it shouldn't be cleaned up. However, that is a bit low level for the lifecycle.

runtime.md
mrunalp: Yep, runtime means a OCI compliant runtime.

runtime.md
mrunalp: Updated state to state.json

runtime.md
mrunalp: Updated

runtime.md
mrunalp: Updated. We can update it again once we reach decision around not using system wide directory.

mrunalp:
Pushed updates to address comments.

runtime.md
```diff
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
...
+2. The container's runtime environment is created according to the configuration in config.json.
+   Any updates to config.json after container is running do not affect the container.
+3. The container's state.json file is written to the filesystem under /run/opencontainer/<runtime>/containers/<id>/.
```
wking: We should not specify the path here (just stop after “to the filesystem”).  See @vbatts [here][1] in response to a comment on an outdated diff.

[1]: https://github.com/opencontainers/specs/pull/211#discussion_r41067134

runtime.md
```diff
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
...
+4. The prestart hooks are invoked by the runtime.
+   If any prestart hook fails, then the container is killed and the lifecycle continues at step 8.
+5. The user specified process is executed in the container.
+6. The poststart hooks are invoked by the runtime.
+   If any poststart hook fails, then the container is killed and the lifecycle continues at step 8.
+7. Additional actions such as pausing the container, resuming the container or signaling the container may be performed using the runtime interface.
```
wking: With ec85489 → bd549a2 you dropped “checkpointing the container”, but left in pause/resume/signalling.  @vishh didn't call those out in his [original comment][1], but they also seem like something we'd want to address after landing a simpler start/cleanup lifecycle.  As it stands, this is the only reference to pausing in the spec, and it doesn't look detailed enough to be useful.

[1]: https://github.com/opencontainers/specs/pull/231#discussion_r45148935

runtime.md
```diff
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
...
+
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created according to the configuration in config.json.
+   Any updates to config.json after container is running do not affect the container.
+3. The container's state.json file is written to the filesystem under /run/opencontainer/<runtime>/containers/<id>/.
+4. The prestart hooks are invoked by the runtime.
+   If any prestart hook fails, then the container is killed and the lifecycle continues at step 8.
+5. The user specified process is executed in the container.
+6. The poststart hooks are invoked by the runtime.
+   If any poststart hook fails, then the container is killed and the lifecycle continues at step 8.
```
philips: what is the difference between killed and destroyed below?

mrunalp: I meant sending a kill signal but that is too linux specific. How about we just modify it to just say jump to step 8?

wking: On Tue, Dec 01, 2015 at 01:46:40PM -0800, Mrunal Patel wrote:
> > +   If any poststart hook fails, then the container is killed and the lifecycle continues at step 8.
>
> I meant sending a kill signal but that is too linux specific. How
> about we just modify it to just say jump to step 8?

I think some form of “intentionally stopped” is appropriate here
(reversing step 5), so we can leave step 8 for reversing step 2.  That
seems like a clear, cross-platform distinction to me.

runtime.md
```diff
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
...
+7. Additional actions such as pausing the container, resuming the container or signaling the container may be performed using the runtime interface.
+   The container could also error out or crash.
+8. The container is destroyed by undoing the steps performed during create phase.
```
dqminh: create phase are not too clear to me i.e what are the steps in the create phase. Seems like 2->6 but it's nicer to be sure.

Also undo the steps doesnt seem very clear either ( like how would we want to undo the start of hooks etc. ). Destroy to me seems like:
- stop the application's process
- clean up the state created by the runtime during the creation phase ( i.e. new cgroups / new namespaces )

wking: On Wed, Dec 02, 2015 at 10:48:14AM -0800, Daniel, Dao Quang Minh wrote:
> > +2. The container's runtime environment is created according to the configuration in config.json.
> > +   Any updates to config.json after container is running do not affect the container.
>
> create phase are not too clear to me i.e what are the steps in the
> create phase. Seems like 2->6 but it's nicer to be sure.

I'd assumed it was just step 2, since that's the only one that
mentions “create” except for step 8's reference to undoing the
creation.  But yeah, adding explicit step numbers to any step
references seems like an easy way to clarify.

runtime.md
hqhq: `runtime interface` --> `runtime interfaces`?

runtime.md
hqhq: Yeah step 8 is not clear and a bit of confusing. I agree it should be just step 2. What about:
The container is destroyed by undoing the container runtime environment specified by step 2.

But meanwhile, creation and cleanup should be FILO(first in last out), so if we remove state.json after destroying container environment, we should write state.json before create container environment. So I suggest we move step 3 before step 2. Technically, I think it makes sense that we mark a state file as a start and cleanup it last.

wking: On Thu, Dec 03, 2015 at 05:58:17AM -0800, Qiang Huang wrote:
> But meanwhile, creation and cleanup should be FILO(first in last
> out), so if we remove state.json after destroying container
> environment, we should write state.json before create container
> environment.

Writing state.json before creating the container is not possible,
because we don't have a container PID before creating the container
environment.  But FILO makes sense, so I'd remove state.json at the
start of cleanup.  That's a bit awkward, because we'd still be passing
the state JSON to the post-stop hooks.  The PID in the state JSON
wouldn't make sense at that point [1], but I think that's a separate
issue.

[1]: https://github.com/wking/oci-command-line-api/pull/3#issuecomment-161512521

hqhq: Right, better we keep the time line as it is.

runtime.md
mrunalp: @hqhq interface would support multiple operations so I think singular form is fine here.

runtime.md
mrunalp: I changed killed -> stopped.

runtime.md
mrunalp: Added a reference to step 2 on step 8.

runtime.md
```diff
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
...
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created according to the configuration in config.json.
+   Any updates to config.json after container is running do not affect the container.
+3. The container's state.json file is written to the filesystem under /run/opencontainer/<runtime>/containers/<id>/.
+4. The prestart hooks are invoked by the runtime.
+   If any prestart hook fails, then the container is stopped and the lifecycle continues at step 8.
+5. The user specified process is executed in the container.
+6. The poststart hooks are invoked by the runtime.
+   If any poststart hook fails, then the container is stopped and the lifecycle continues at step 8.
+7. Additional actions such as pausing the container, resuming the container or signaling the container may be performed using the runtime interface.
```
wking: I [still][1] think we want to drop pause/resume/signal until we have more clarity on how they will work.  For more background here, especially around complications with shared state on Linux, see wking/oci-command-line-api#3.  Alternatively, if we already have a clear idea of how these will work [in the absence of /run/opencontainer][2], I'd appreciate some comments in the CLI-spec PR explaining what the CLI should look like.

[1]: https://github.com/opencontainers/specs/pull/231#discussion_r45532412
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/q6TYqVZOcX8

runtime.md
```diff
@@ -34,22 +34,24 @@ This is provided so that consumers can find the container's configuration and ro
...
-
-Not sure we need that from runc cli.
-Process is killed from the outside.
-
-This event needs to be captured by runc to run onstop event handlers.
+The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
+
+1. OCI compliant runtime is invoked by passing the bundle path as argument.
+2. The container's runtime environment is created according to the configuration in config.json.
+   Any updates to config.json after container is running do not affect the container.
```
wking: We probably need to mention runtime.json here too.  I'm [in favor of rolling it back into config.json][1], but we [aren't there yet][2].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/0QbyJDM9fWY
[2]: https://github.com/opencontainers/specs/blob/4a63e81a807edec3d67ed2b6bd99a6c2b288676f/bundle.md#container-format

wking: Previous discussion [starting here][1] (and I really miss the email-based workflow where previous discussions like that are linkable without having to find the right “Show outdated diff” button :p).

[1]: https://github.com/opencontainers/specs/pull/231#discussion_r43262848

dqminh:
LGTM

I think this is a great first step, we can further discuss the details for each individual actions, but the overall flow and structure looks right.

vbatts:
LGTM

PR #232, opened 2015-10-25
change the name of struct `blockIODevice` to `BlockIODevice`

wangkirin:
the struct `blockIODevice` is begin with lower case letters, and we cannot export it and instantiate it from other packages.This PR fixed it 

Signed-off-by: Wang Qilin <wangqilin2@huawei.com>

### DIFF
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -106,8 +106,8 @@ type InterfacePriority struct {
 	Priority int64 `json:"priority"`
 }
 
-// blockIODevice holds major:minor format supported in blkio cgroup
-type blockIODevice struct {
+// BlockIODevice holds major:minor format supported in blkio cgroup
+type BlockIODevice struct {
 	// Major is the device's major number.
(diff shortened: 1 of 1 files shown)


runcom:
IIRC I did this way so the inner struct was hidden and you just had to instantiate the upper struct with only the public fields (I explicitly did this way cause I didn't want another struct just for Major, Minor and you can always use the other struct without instantiating the private one)

vbatts:
This is intentional. It'll stay not exported for now.

wking:
On Sun, Oct 25, 2015 at 08:06:35AM -0700, Antonio Murdaca wrote:
> IIRC I did this way so the inner struct was hidden and you just had
> to instantiate the upper struct with only the public fields…

This got unpacked in more detail on IRC today, starting with [1].

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-03-02.log.html#t2016-03-02T00:06:46

PR #233, opened 2015-10-27
Fix cgroups value types in the spec.

vishh:
Follow up of [this comment](https://github.com/opencontainers/specs/pull/199/files#r43151805)

### DIFF
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -24,11 +24,11 @@ type LinuxRuntime struct {
 	Sysctl map[string]string `json:"sysctl"`
 	// Resources contain cgroup information for handling resource constraints
 	// for the container
-	Resources *Resources `json:"resources"`
+	Resources *Resources `json:"resources,omitempty"`
 	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
 	// The path is expected to be relative to the cgroups mountpoint.
 	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
-	CgroupsPath string `json:"cgroupsPath"`
+	CgroupsPath *string `json:"cgroupsPath,omitempty"`
 	// Namespaces contains the namespaces that are created and/or joined by the container
 	Namespaces []Namespace `json:"namespaces"`
 	// Devices are a list of device nodes that are created and enabled for the container
@@ -93,9 +93,9 @@ type Rlimit struct {
 // HugepageLimit structure corresponds to limiting kernel hugepages
 type HugepageLimit struct {
 	// Pagesize is the hugepage size
-	Pagesize string `json:"pageSize"`
+	Pagesize *string `json:"pageSize,omitempty"`
 	// Limit is the limit of "hugepagesize" hugetlb usage
-	Limit uint64 `json:"limit"`
+	Limit *uint64 `json:"limit,omitempty"`
 }
 
 // InterfacePriority for network interfaces
@@ -118,9 +118,9 @@ type blockIODevice struct {
 type WeightDevice struct {
 	blockIODevice
 	// Weight is the bandwidth rate for the device, range is from 10 to 1000
-	Weight uint16 `json:"weight"`
+	Weight *uint16 `json:"weight,omitempty"`
 	// LeafWeight is the bandwidth rate for the device while competing with the cgroup's child cgroups, range is from 10 to 1000, CFQ scheduler only
-	LeafWeight uint16 `json:"leafWeight"`
+	LeafWeight *uint16 `json:"leafWeight,omitempty"`
 }
 
 // ThrottleDevice struct holds a `major:minor rate_per_second` pair
@@ -133,59 +133,59 @@ type ThrottleDevice struct {
 // BlockIO for Linux cgroup 'blkio' resource management
 type BlockIO struct {
 	// Specifies per cgroup weight, range is from 10 to 1000
-	Weight uint16 `json:"blkioWeight"`
+	Weight *uint16 `json:"blkioWeight,omitempty"`
 	// Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, range is from 10 to 1000, CFQ scheduler only
-	LeafWeight uint16 `json:"blkioLeafWeight"`
+	LeafWeight *uint16 `json:"blkioLeafWeight,omitempty"`
 	// Weight per cgroup per device, can override BlkioWeight
-	WeightDevice []*WeightDevice `json:"blkioWeightDevice"`
+	WeightDevice []*WeightDevice `json:"blkioWeightDevice,omitempty"`
 	// IO read rate limit per cgroup per device, bytes per second
-	ThrottleReadBpsDevice []*ThrottleDevice `json:"blkioThrottleReadBpsDevice"`
+	ThrottleReadBpsDevice []*ThrottleDevice `json:"blkioThrottleReadBpsDevice,omitempty"`
 	// IO write rate limit per cgroup per device, bytes per second
-	ThrottleWriteBpsDevice []*ThrottleDevice `json:"blkioThrottleWriteBpsDevice"`
+	ThrottleWriteBpsDevice []*ThrottleDevice `json:"blkioThrottleWriteBpsDevice,omitempty"`
 	// IO read rate limit per cgroup per device, IO per second
-	ThrottleReadIOPSDevice []*ThrottleDevice `json:"blkioThrottleReadIOPSDevice"`
+	ThrottleReadIOPSDevice []*ThrottleDevice `json:"blkioThrottleReadIOPSDevice,omitempty"`
 	// IO write rate limit per cgroup per device, IO per second
-	ThrottleWriteIOPSDevice []*ThrottleDevice `json:"blkioThrottleWriteIOPSDevice"`
+	ThrottleWriteIOPSDevice []*ThrottleDevice `json:"blkioThrottleWriteIOPSDevice,omitempty"`
 }
 
 // Memory for Linux cgroup 'memory' resource management
 type Memory struct {
-	// Memory limit (in bytes)
-	Limit uint64 `json:"limit"`
-	// Memory reservation or soft_limit (in bytes)
-	Reservation uint64 `json:"reservation"`
-	// Total memory usage (memory + swap); set `-1' to disable swap
-	Swap uint64 `json:"swap"`
-	// Kernel memory limit (in bytes)
-	Kernel uint64 `json:"kernel"`
+	// Memory limit (in bytes).
+	Limit *uint64 `json:"limit,omitempty"`
+	// Memory reservation or soft_limit (in bytes).
+	Reservation *uint64 `json:"reservation,omitempty"`
+	// Total memory limit (memory + swap).
+	Swap *uint64 `json:"swap,omitempty"`
+	// Kernel memory limit (in bytes).
+	Kernel *uint64 `json:"kernel,omitempty"`
 	// Kernel memory limit for tcp (in bytes)
-	KernelTCP uint64 `json:"kernelTCP"`
-	// How aggressive the kernel will swap memory pages. Range from 0 to 100. Set -1 to use system default
-	Swappiness uint64 `json:"swappiness"`
+	KernelTCP *uint64 `json:"kernelTCP"`
+	// How aggressive the kernel will swap memory pages. Range from 0 to 100.
+	Swappiness *uint64 `json:"swappiness,omitempty"`
 }
 
 // CPU for Linux cgroup 'cpu' resource management
 type CPU struct {
-	// CPU shares (relative weight vs. other cgroups with cpu shares)
-	Shares uint64 `json:"shares"`
-	// CPU hardcap limit (in usecs). Allowed cpu time in a given period
-	Quota uint64 `json:"quota"`
-	// CPU period to be used for hardcapping (in usecs). 0 to use system default
-	Period uint64 `json:"period"`
-	// How many time CPU will use in realtime scheduling (in usecs)
-	RealtimeRuntime uint64 `json:"realtimeRuntime"`
-	// CPU period to be used for realtime scheduling (in usecs)
-	RealtimePeriod uint64 `json:"realtimePeriod"`
-	// CPU to use within the cpuset
-	Cpus string `json:"cpus"`
-	// MEM to use within the cpuset
-	Mems string `json:"mems"`
+	// CPU shares (relative weight (ratio) vs. other cgroups with cpu shares).
+	Shares *uint64 `json:"shares,omitempty"`
+	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
+	Quota *uint64 `json:"quota,omitempty"`
+	// CPU period to be used for hardcapping (in usecs).
+	Period *uint64 `json:"period,omitempty"`
+	// How much time realtime scheduling may use (in usecs).
+	RealtimeRuntime *uint64 `json:"realtimeRuntime,omitempty"`
+	// CPU period to be used for realtime scheduling (in usecs).
+	RealtimePeriod *uint64 `json:"realtimePeriod,omitempty"`
+	// CPUs to use within the cpuset. Default is to use any CPU available.
+	Cpus *string `json:"cpus,omitempty"`
+	// List of memory nodes in the cpuset. Default is to use any available memory node.
+	Mems *string `json:"mems,omitempty"`
 }
 
 // Pids for Linux cgroup 'pids' resource management (Linux 4.3)
 type Pids struct {
-	// Maximum number of PIDs. A value <= 0 indicates "no limit".
-	Limit int64 `json:"limit"`
+	// Maximum number of PIDs. Default is "no limit".
+	Limit *int64 `json:"limit,omitempty"`
 }
 
 // Network identification and priority configuration
@@ -201,21 +201,21 @@ type Network struct {
 // Resources has container runtime resource constraints
 type Resources struct {
 	// DisableOOMKiller disables the OOM killer for out of memory conditions
-	DisableOOMKiller bool `json:"disableOOMKiller"`
-	// Specify an oom_score_adj for the container. Optional.
-	OOMScoreAdj int `json:"oomScoreAdj"`
+	DisableOOMKiller *bool `json:"disableOOMKiller,omitempty"`
+	// Specify an oom_score_adj for the container.
+	OOMScoreAdj *int `json:"oomScoreAdj,omitempty"`
 	// Memory restriction configuration
-	Memory Memory `json:"memory"`
+	Memory *Memory `json:"memory,omitempty"`
 	// CPU resource restriction configuration
-	CPU CPU `json:"cpu"`
+	CPU *CPU `json:"cpu,omitempty"`
 	// Task resource restriction configuration.
-	Pids Pids `json:"pids"`
+	Pids *Pids `json:"pids,omitempty"`
 	// BlockIO restriction configuration
-	BlockIO BlockIO `json:"blockIO"`
+	BlockIO *BlockIO `json:"blockIO,omitempty"`
 	// Hugetlb limit (in bytes)
-	HugepageLimits []HugepageLimit `json:"hugepageLimits"`
+	HugepageLimits []HugepageLimit `json:"hugepageLimits,omitempty"`
 	// Network restriction configuration
-	Network Network `json:"network"`
+	Network *Network `json:"network,omitempty"`
 }
 
 // Device represents the information on a Linux special device file


wking:
On Tue, Oct 27, 2015 at 10:59:48AM -0700, Vish Kannan wrote:
>   * Fix cgroups value types in the spec.

This unwinds a lot of type changes from #199.  Are we sure we want to
use “-1 means skip this” instead of “just don't set this field in the
JSON to skip it”?  And then use pointers or protobuf with optional
fields for the Go?

vishh:
Unless these field become pointers, I do not see how I can skip these
fields in json.

wking:
On Tue, Oct 27, 2015 at 11:21:02AM -0700, Vish Kannan wrote:
> Unless these field become pointers, I do not see how I can skip
> these fields in json.

I'm fine with pointers (and I expect Go generated from a protobuf
schema will use something like that too).

vbatts:
ah no. int64 makes more sense here. Pointers is a bit too much complication here.

LGTM

wking:
On Tue, Oct 27, 2015 at 01:36:50PM -0700, Vincent Batts wrote:
> ah no. int64 makes more sense here. Pointers is a bit too much
> complication here.

I'm fine punting this to the protobuf transition (at which point these
fields will become optional as far as the parser is concerned), but I
think a bit of extra complication on the bindings side is a fair price
to pay for cleaner semantics on the JSON/spec side.

hqhq:
I don't understand why we take -1 as unlimited? What's the intention here?
With this change, we need to change some codes in runc as well, I just didn't see the necessity.
@vishh Could you elaborate?

vishh:
@hqhq: Sometimes, we want to use the system defaults for certain cgroups values. The default value of `0` in go cannot be used to express that intent, because `0` can be an explicit value that the user has set. Using `-1` to use the system defaults is just one possible way to address this. Other options are, using a complex type that can understand the notion of `unspecified`, or convert cgroup fields into key value pairs in a map, etc.

hqhq:
@vishh I understand and agree with this intention, but so far there is only `swappiness` would take `0` as an explicit value that user has set, and I think this would be the only one with the latest kernel support. So can we just fix `Swappiness` type to `int64` and leave the others? Otherwise there would be backward compatibility problems for runc users like Docker.

vishh:
Without this change, how do you propose that a user will express their intent of using the system defaults?

hqhq:
@vishh Except exception like `Swappiness`, all other system defaults are invalid, using them will be taken as equally as not using them, they are expressed in docs.

cyphar:
Requiring to explicitly set `-1` to mean "I want the default" breaks the backwards-compatibility of the spec. If a user wishes to update the spec version they have to merge their old version with the new version with `-1` set. Surely it makes more sense to set the JSON defaults value (`0`) to be the system default? I understand `swappiness` is an exception, but why change the rest? Also, I'm fairly sure that some of those settings actually accept `uint64` values, so what are we going to do in that case?

If we *had* to do this, surely a better way would be to set them to `null` (so we use pointers -- AND we get the choice of having the JSON default producing the system default value) and then in consumers of the spec, they could then translate the spec to their own `struct` which uses `uint64` with the defaults set for values which were `null`.

vishh:
Hmm. Considering that the Spec is still not v1.0, do we really care about
backwards compat? There are extensive changes being proposed which will
definitely break backwards compat.
I feel using different values as defaults will add cognitive burden to
users of the Spec.
That said, if all other maintainers and users feel that the cognitive
burden isn't a big deal, I can rework this PR.

The general issue here is the fact that the Spec is wrapping the linux APIs
without adding much value.

wking:
On Fri, Oct 30, 2015 at 09:42:51AM -0700, Vish Kannan wrote:
> I feel using different values as defaults will add cognitive burden
> to users of the Spec.

Using pointers in the Go types so ‘null’ and missing are valid JSON
for “don't touch this” is consistent and has a low cognitive burden.
It just has a slightly higher *implementation* burden, but I don't
think the increase there is too big an ask.

cyphar:
@vishh

> Hmm. Considering that the Spec is still not v1.0, do we really care about
backwards compat? There are extensive changes being proposed which will
definitely break backwards compat.

Ah, fair enough. But there are other reasons why I don't agree with using `-1` to mean "use the default".

> I feel using different values as defaults will add cognitive burden to
users of the Spec.

TBH, I feel like `-1` is more of a cognitive burden, because omitting an option in the spec would mean "use the actual value `0`, not the default". Explicitly setting "I want the default" IMHO makes more intuitive sense if it's a clearly non-integer value like `null`.

wking:
On Fri, Oct 30, 2015 at 05:44:34PM -0700, Aleksa Sarai wrote:
> TBH, I feel like `-1` is more of a cognitive burden, because
> omitting an option in the spec would mean "use the actual value `0`,
> not the default". Explicitly setting "I want the default" IMHO makes
> more intuitive sense if it's a clearly non-integer value like
> `null`.

I may be misunderstanding, but it sounds like you're saying:

  {
    "linux": {
      "resources": {
         "memory": {
           "limit": null,
           …
         }
      }
    }
  }

would mean “don't touch memory.limit_in_bytes”.  That makes sense to
me.  It also sounds like you're saying:

  {
    "linux": {
      "resources": {
         "memory": {
         }
      }
    }
  }

through an empty/missing runtime.json would mean “set
memory.limit_in_bytes to zero”.  I think all the cases that don't set
a .linux.resources.memory.limit key should also be “don't touch
memory.limit_in_bytes”.

For example, say the 1.0 spec defines cgroups that are pretty much
like we have now, but the 1.1 minor bump adds support for
cpuset.cpu_exclusive.  With “unset means use Go's default value for
that type”, every bundle using a 1.0 config will then start disabling
cpuset.cpu_exclusive.  With “unset means don't touch it”, bundles with
1.0 configs would keep not touching file, just like they had with 1.0
runtimes.

cyphar:
@wking No, I agree with the proposal of using pointers so that using `null` *and* not setting it will both result in the semantic of "use the default". When I said 

> TBH, I feel like -1 is more of a cognitive burden, because omitting an option in the spec would mean "use the actual value 0, not the default". 

i was referring specifically to the current state of this PR (using `-1` as the "please use the default" value). Because if you omit an option, Go's default JSON semantic is such that it would mean "use the value 0". 

I **don't** agree this is how it should work, and I'm all for making **both** setting a value to `null` and omitting it from the config meaning "use the default".

crosbymichael:
I'm ok with nil for values where the zero value means something for our settings.  @vishh what do you think?

vishh:
SGTM.

cyphar:
@crosbymichael Weren't we talking about making *all* settings have `nil` mean "use the default".

crosbymichael:
All or only the ones that make sense.  Its so we don't end up with things like memory swap where you have to set it to -1 if your system does not support it or your containers will fail

wking:
On Mon, Nov 09, 2015 at 05:07:36PM -0800, Michael Crosby wrote:
> All or only the ones that make sense.

I think “don't set this value” or “don't create this cgroup” are
things that make sense for all of our settings under runtime
linux.resources.  Is there anything under there you feel should not be
nullable?  It may be easier to have this conversation on a per-setting
basis, so I'm happy to do that (on IRC?  On the mailing list?) if it's
easier than trying to decide on large blocks of settings en masse.

cyphar:
@crosbymichael setting memory swap as `null` means "use the default" which should be read as "use the system default". If it's not implemented, the system default is to not use the limit.

mrunalp:
@cyphar Yes, that's what we agreed to. @vishh Could you update the PR?

vishh:
@mrunalp @crosbymichael PR updated. PTAL

wking:
With b684645 both resources and cgroupsPath are omitempty, so I think
we'll want to define what happens (or explicitly say it's undefined)
if both are unset.  More discussion on that in #237, #247, and #251.

You also change Memory.Swappiness and similar from uint64 to *int64.
Maybe you meant *uint64, or are you trying to unwind #199?

You also define a few defaults (e.g. “Default is to use any CPU
available”).  Instead of doing that, I think we just want to document
a blanket “if a given resource is unset, the runtime MUST not touch
the associated cgroups configuration”.  That way the container will
just inherit whatever restrictions apply to its parent cgroup (if
we're creating a new cgroup) or we don't modify existing values that
aren't explicitly configured in runtime.json (if we're joining an
existing cgroup, see #247).

runtime_config_linux.go
```diff
@@ -200,20 +200,20 @@ type Network struct {
 type Resources struct {
 	// DisableOOMKiller disables the OOM killer for out of memory conditions
 	DisableOOMKiller bool `json:"disableOOMKiller"`
```
mrunalp: Should this be a pointer too?

runtime_config_linux.go
```diff
@@ -133,57 +133,57 @@ type ThrottleDevice struct {
...
+	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
+	Quota *int64 `json:"quota,omitempty"`
+	// CPU period to be used for hardcapping (in usecs).
+	Period *int64 `json:"period,omitempty"`
+	// How much time realtime scheduling may use (in usecs).
+	RealtimeRuntime *int64 `json:"realtimeRuntime,omitempty"`
+	// CPU period to be used for realtime scheduling (in usecs).
+	RealtimePeriod *int64 `json:"realtimePeriod,omitempty"`
+	// CPUs to use within the cpuset. Default is to use any CPU available.
+	Cpus string `json:"cpus,omitempty"`
```
mrunalp: Convert these to pointers too? That way we have the semantics that if a value is nil then don't touch it. If it isn't nil then set it.

mrunalp: I guess for string, the empty string could substitute for nil but might be worth making a pointer for consistency.

cyphar: I think it should be a pointer, and we can say that empty strings are undefined or will fail (unless it's a valid value, like all other invalid values).

runtime_config_linux.go
vishh: isn't the default false?

runtime_config_linux.go
vishh: Strings are easy to use as-is right? I see the value in semantics, but a
user can just ignore these fields instead of mentioning `nil` or `""` too.

vishh: How about not including this field at all? if the user input is json, then
this field can be ommitted after this change.

runtime_config_linux.go
mrunalp: This is more for the runtime implementation. nil means don't touch it. Even if the default is false, the runtime will have to go and overwrite the file if it isn't nil.

runtime_config_linux.go
mrunalp: @cyphar I agree

mrunalp:
If both Resources and cgroupsPath are unset, then we could support the don't touch cgroups use case.

cyphar:
@mrunalp Don't we require `devices` for security reasons, won't we always have to do so?

wking:
On Wed, Nov 18, 2015 at 04:59:14PM -0800, Mrunal Patel wrote:
> If both Resources and cgroupsPath are unset, then we could support
> the don't touch cgroups use case.

That's what I thought (#237), but there is apparently not a consensus
around it yet [1].

[1]: https://github.com/opencontainers/specs/pull/237#issuecomment-152294240

wking:
On Wed, Nov 18, 2015 at 05:08:29PM -0800, Aleksa Sarai wrote:
> @mrunalp Don't we require `devices` for security reasons, won't we
> always have to do so?

Do we have to define security policy in the spec?  Can't we leave that
to bundle authors and runtime callers to configure whatever resource
restrictions they deem appropriate?

mrunalp:
There are use cases such as sandboxed desktop applications where this isn't required. If necessary one could add the configuration as one sees fit.

cyphar:
Ah okay, fair enough. Just checking :P. What happens if just `Resources` is `null` but `cgroupPath` is set?

runtime_config_linux.go
hqhq: Yeah, I think this should be pointer, the default false would be a problem when you want to update the resource of a existed container.

runtime_config_linux.go
hqhq: I tend to agree with we use pointer as well, for consistency is one reason.

And the reason we made this PR is because default value has meaning for some integer entries, technically this can happen on string too. Maybe not now, but for standard, we can consider further, make it pointer would be a good choice.

runtime_config_linux.go
```diff
@@ -133,57 +133,57 @@ type ThrottleDevice struct {
...
+	// List of memory nodes in the cpuset. Default is to use any memory node.
```
hqhq: I'm afraid this'll cause confusion, because the default would really depend on the parent cgroup. And we offered the ability `cgroupsPath` to let user choose parent.

Maybe we just drop this part, and leave it as it was. So is Cpus.

wking:
On Thu, Nov 19, 2015 at 12:14:54AM -0800, Aleksa Sarai wrote:
> What happens if just `Resources` is `null` but `cgroupPath` is set?

You join/create those cgroups but don't alter their configuration
(#247).

runtime_config_linux.go
mrunalp: +1

cyphar:
/ping @vishh. Can you attend to the comments and wrap up this PR? Just make all configuration options pointers. Cheers.

hqhq:
@vishh Can you address these new comments? Then I think it's good to go :)

runtime_config_linux.go
vishh: With integer types we don't have the option to differentiate between `null` and `0`. But with string types, that is not an issue right?

runtime_config_linux.go
vishh: I guess the intent was to say that if the field was not set, any available memory node will be used.
I don't see why that is incorrect? Same applies to Cpus.

runtime_config_linux.go
vishh: Done

vishh:
@mrunalp: Updated this PR. PTAL

runtime_config_linux.go
wking: On Wed, Dec 16, 2015 at 09:59:24AM -0800, Vish Kannan wrote:
> > +	// CPUs to use within the cpuset. Default is to use any CPU available.
> > +	Cpus string `json:"cpus,omitempty"`
> 
> With integer types we don't have the option to differentiate between
> `null` and `0`. But with string types, that is not an issue right?

That assumes that the empty string will never be semantically
meaningful (i.e. you'll never want to actually write the empty string
to a cgroups pseudo-file).  That may be a valid assumption, but I see
no point in relying on it when a string pointer here would be easy to
add.

vishh: Its software. Everything is easy :) But does that mean we should do it?

wking: On Wed, Dec 16, 2015 at 11:50:51AM -0800, Vish Kannan wrote:
> > +	// CPUs to use within the cpuset. Default is to use any CPU available.
> > +	Cpus string `json:"cpus,omitempty"`
> 
> Its software. Everything is easy :) But does that mean we should do
> it?

Doing string pointers seems a lot easier than imposing restrictions on
future semantically meaningful cgroup config values (especially since
we're not in control of what lands in kernel cgroup settings).

vishh: I really don't see how an empty string would mean something specific in the
future in these cases.

mrunalp: I think having them as pointers will be good from a consistency point of view. That way the code knows that irrespective of the type it always has to check for nil and only apply settings if it isn't nil.

wking: On Wed, Dec 16, 2015 at 12:44:11PM -0800, Vish Kannan wrote:
> > +	// CPUs to use within the cpuset. Default is to use any CPU available.
> > +	Cpus string `json:"cpus,omitempty"`
> 
> I really don't see how an empty string would mean something specific
> in the future in these cases.

Me neither, but *string is cheap, and my crystal ball is occasionally
cloudy ;).

runtime_config_linux.go
hqhq: I guess there is difference between `any memory node` and `any available memory node`?

vishh: Sure. Updated

runtime_config_linux.go
```diff
@@ -133,59 +133,59 @@ type ThrottleDevice struct {
...
+	Swap *int64 `json:"swap,omitempty"`
+	// Kernel memory limit (in bytes).
+	Kernel *int64 `json:"kernel,omitempty"`
 	// Kernel memory limit for tcp (in bytes)
-	KernelTCP uint64 `json:"kernelTCP"`
-	// How aggressive the kernel will swap memory pages. Range from 0 to 100. Set -1 to use system default
-	Swappiness uint64 `json:"swappiness"`
+	KernelTCP *int64 `json:"kernelTCP"`
+	// How aggressive the kernel will swap memory pages. Range from 0 to 100.
+	Swappiness *int64 `json:"swappiness,omitempty"`
```
hqhq: I understand `Swappiness` can be negative so it should be `*int64`, but why the others are also changed from `uint64` to `*int64` not `*uint64`?

runtime_config_linux.go
```diff
@@ -133,59 +133,59 @@ type ThrottleDevice struct {
...
+	// CPU shares (relative weight (ratio) vs. other cgroups with cpu shares).
+	Shares *int64 `json:"shares,omitempty"`
+	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
+	Quota *int64 `json:"quota,omitempty"`
+	// CPU period to be used for hardcapping (in usecs).
+	Period *int64 `json:"period,omitempty"`
+	// How much time realtime scheduling may use (in usecs).
+	RealtimeRuntime *int64 `json:"realtimeRuntime,omitempty"`
+	// CPU period to be used for realtime scheduling (in usecs).
+	RealtimePeriod *int64 `json:"realtimePeriod,omitempty"`
```
hqhq: So are these.

runtime_config_linux.go
wking: On Wed, Dec 16, 2015 at 06:35:05PM -0800, Qiang Huang wrote:
> > +	// How aggressive the kernel will swap memory pages. Range from 0 to 100.
> > +	Swappiness *int64 `json:"swappiness,omitempty"`
> 
> I understand `Swappiness` can be negative so it should be `*int64`…

If the range is from 0 to 100 ([1] and the comment here), then why use
a signed integer (although the kernel uses a signed integer ;)?  Using
an unsigned integer lets you offload one more constraint-check to the
JSON validator.

[1]: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/mm/vmscan.c?h=v4.3#n139

cyphar: I'd also like to point out that using `-1` would've been the wrong type given the diff (the old code used `uint64` and told you to use `-1` for "default").

wking: On Wed, Dec 16, 2015 at 09:03:47PM -0800, Aleksa Sarai wrote:
> >  	// Kernel memory limit for tcp (in bytes)
> > -	KernelTCP uint64 `json:"kernelTCP"`
> > -	// How aggressive the kernel will swap memory pages. Range from 0 to 100. Set -1 to use system default
> > -	Swappiness uint64 `json:"swappiness"`
> > +	KernelTCP *int64 `json:"kernelTCP"`
> > +	// How aggressive the kernel will swap memory pages. Range from 0 to 100.
> > +	Swappiness *int64 `json:"swappiness,omitempty"`
>
> I'd also like to point out that using `-1` would've been the wrong
> type given the diff (the old code used `uint64` and told you to use
> `-1` for "default").

Those have just been out of sync since #199 landed.  An earlier
version of this PR (0d03c3d4) was turning it back to a signed integer
to fix this, but since we're not consistently using pointers we don't
need to restore signedness.

hqhq: OK so we should use `*uint64` for `Swappiness` as well, and default would be using system default. Then my question is why all these fields (Limit, Swap etc) are changed to `*int64` instead of `*uint64`?

crosbymichael:
LGTM

runtime_config_linux.go
```diff
@@ -133,59 +133,59 @@ type ThrottleDevice struct {
...
-	// How many time CPU will use in realtime scheduling (in usecs)
-	RealtimeRuntime uint64 `json:"realtimeRuntime"`
-	// CPU period to be used for realtime scheduling (in usecs)
-	RealtimePeriod uint64 `json:"realtimePeriod"`
-	// CPU to use within the cpuset
-	Cpus string `json:"cpus"`
-	// MEM to use within the cpuset
-	Mems string `json:"mems"`
+	// CPU shares (relative weight (ratio) vs. other cgroups with cpu shares).
+	Shares *int64 `json:"shares,omitempty"`
```
mrunalp: Since these are pointers now, shouldn't we match the kernel data types and have these be unsigned?
If it is nil, it means that we don't touch the value. There is no need for setting these to negative values.

cyphar:
I'm not a maintainer, but NACK. There are still some issues with types that should be unsigned.

vishh:
Comments addressed

mrunalp:
LGTM

philips:
thanks for doing this @vishh

PR #234, opened 2015-10-28
config: linux: update description of PidsLimit

cyphar:
Fix a misleading comment for how PidsLimit works when given a limit of
0. A limit of 0 is identical to a limit of 1, and so there is no reason
to distinguish the two. Since the default value for integers in Go is 0,
it makes more sense to make the default value mean that there is no
limit.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com> (github: cyphar)

### DIFF
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -182,7 +182,7 @@ type CPU struct {
 
 // Pids for Linux cgroup 'pids' resource management (Linux 4.3)
 type Pids struct {
-	// Maximum number of PIDs. A value < 0 implies "no limit".
+	// Maximum number of PIDs. A value <= 0 indicates "no limit".
 	Limit int64 `json:"limit"`
 }
 


vbatts:
*preciously* LGTM

runtime_config_linux.go
```diff
@@ -182,7 +182,9 @@ type CPU struct {
 
 // Pids for Linux cgroup 'pids' resource management (Linux 4.3)
 type Pids struct {
-	// Maximum number of PIDs. A value < 0 implies "no limit".
+	// Maximum number of PIDs. A value <= 0 implies "no limit".
+	// While a limit of 0 is technically valid, it is precisely identical to
+	// a limit of 1, so there is no distinction.
```
wking: Do you want both “≤ 0 means no limit” and “0 is the same as 1”?  I'd expect either “≤ 0 means no limit and 0 is the same as -1” or “< 0 means no limit and 0 is the same as 1”.  Or does 1 really mean “no limit”?

cyphar: Oh, sorry. What I meant was that "<= 0 means no limit *within the spec*, but a limit of 0 is identical to a limit of 1 in the PIDs controller". I guess I'm mixing specification and justification, based on the discussion in opencontainers/runc#58. I'll fix up the wording to only mention the actual specification and only leave the justification in the commit message.

wking: On Wed, Oct 28, 2015 at 09:28:23AM -0700, Aleksa Sarai wrote:
> Oh, sorry. What I meant was that "<= 0 means no limit *within the
> spec*, but a limit of 0 is identical to a limit of 1 in the PIDs
> controller". I guess I'm mixing specification and justification,
> based on the discussion in opencontainers/runc#58.

Oh, I see.  In that case, my personal preference would be to just use
a uint pointer with “not set in the JSON means the runtime should
ignore” (see discussion in #233).  But however this lands, any
semantic rules (like what values the runtime should ignore) should
also land in the runtime-config-linux.md docs where bundle authors
will find them.

cyphar: @wking I agree (as I posted there) that #233 should in fact use pointers so that omitting and setting null have the same meaning (while also not requiring sentinel values). I'll revive this once we decide what we're going to do on #233.

crosbymichael:
LGTM

PR #235, opened 2015-10-28
cgroup: Add support for memory.kmem.tcp.limit_in_bytes

yangdongsheng:
Signed-off-by: Dongsheng Yang <yangds.fnst@cn.fujitsu.com>

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -201,6 +201,8 @@ The following parameters can be specified to setup the controller:
 
 * **`kernel`** *(uint64, optional)* - sets hard limit for kernel memory
 
+* **`kernelTCP`** *(uint64, optional)* - sets hard limit for kernel memory in tcp using
+
 * **`swappiness`** *(uint64, optional)* - sets swappiness parameter of vmscan (See sysctl's vm.swappiness)
 
 ###### Example
@@ -211,6 +213,7 @@ The following parameters can be specified to setup the controller:
         "reservation": 0,
         "swap": 0,
         "kernel": 0,
+        "kernelTCP": 0,
         "swappiness": -1
     }
 ```
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -158,6 +158,8 @@ type Memory struct {
 	Swap uint64 `json:"swap"`
 	// Kernel memory limit (in bytes)
 	Kernel uint64 `json:"kernel"`
+	// Kernel memory limit for tcp (in bytes)
+	KernelTCP uint64 `json:"kernelTCP"`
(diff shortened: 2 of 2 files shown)


vbatts:
while i'm not opposed to this particularly, It does circle back to the conversation about requiring a field for every kernel parameter that exists.

vishh:
Yes. Frankly I don't see much value in this unless we provide a higher level abstraction like that of [lmctfy](https://github.com/google/lmctfy/blob/master/include/lmctfy.proto).

vishh:
There is an on-going discussion about cgroups support in the mailing list. Lets hold on to this PR until that thread comes to a conclusion. 
Sorry for the trouble @yangdongsheng!

yangdongsheng:
@vishh That's fine. I will read the discussion then. Thanx for your review. :)

vbatts:
@vishh thoughts on revisiting this? or proposal on better abstraction?

vishh:
@vbatts: Cgroup parameters have to be pointers. That is the current agreement. This PR requires that change, otherwise LGTM!

hqhq:
LGTM, @vishh Do you mind merge this and rebase in https://github.com/opencontainers/specs/pull/233, this can accelerate the process :)

yangdongsheng:
@hqhq Thanx, that's good point. I agree with that.

vishh:
SGTM.

runtime-config-linux.md
```diff
@@ -201,6 +201,8 @@ The following parameters can be specified to setup the controller:
 
 * **`kernel`** *(uint64, optional)* - sets hard limit for kernel memory
 
+* **`kernelTCP`** *(uint64, optional)* - sets hard limit for kernel memory in tcp using
```
wking: using… what?  We still have this unfinished sentence [in the current tip][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/c337cd2c0a425ac9a7e17a624ac3be23a02fe53f/config-linux.md#L262

vishh: Oops.. @wking Can you post a PR to fix this ?

wking: On Mon, Apr 11, 2016 at 03:54:09PM -0700, Vish Kannan wrote:
> > +* **`kernelTCP`** *(uint64, optional)* - sets hard limit for kernel memory in tcp using
> 
> Oops.. @wking Can you post a PR to fix this ?

@vbatts just pushed c1612af → d4e7326 in #370 which should cover this.

wking: Err, actually, he's adjusting docs for other settings.  I'll file a PR
for this on.

vbatts: i didn't address this comment

PR #236, opened 2015-10-29
runtime-config-linux: Add info to clarify oom_score_adj

ChengTiesheng:
Signed-off-by: ChengTiesheng <chengtiesheng@huawei.com>

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -180,7 +180,11 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 
 #### Set oom_score_adj
 
-More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+oom_score_adj works with disableOOMKiller. If out-of-memory killer enabled("disableOOMKiller": false), oom_score_adj can be set to adjust the oom-killer score, the value of oom_score_adj is added to the badness score before it is used to determine which task to kill. If out-of-memory killer disabled("disableOOMKiller": true), the value of oom_score_adj is useless and will be omitted. For more information, see [the proc filesystem man page](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+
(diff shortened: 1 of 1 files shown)


ChengTiesheng:
Add more description for oom_score_adj to make it clearer.

vishh:
disableOOMKiller is croup scoped whereas oom_score_adj is system scoped. If there is memory pressure on the system, oom_score_adj will be taken into account to choose a victim process which can be a part of a croup that has OOM kills disabled. 

Your PR send to suggest otherwise.

vishh:
Ping

hqhq:
Hmmm, the change is not accurate and could cause confusion, @ChengTiesheng can you update?

runtime-config-linux.md
```diff
@@ -180,7 +180,11 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 
 #### Set oom_score_adj
 
-More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+oom_score_adj works with disableOOMKiller. If out-of-memory killer enabled("disableOOMKiller": false), oom_score_adj can be set to adjust the oom-killer score, the value of oom_score_adj is added to the badness score before it is used to determine which task to kill. If out-of-memory killer disabled("disableOOMKiller": true), the value of oom_score_adj is useless and will be omitted. For more information, see [the proc filesystem man page](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
```
vishh: What about system OOMs? The per cgroup `disableOOMKiller` doesn't matter in such cases.

In general, let's not try to restate the Linux APIs again and instead add links to relevant Linux API documentation. It will be much easier to maintain.

vishh:
In general, let's not try to restate the Linux APIs again and instead add links to relevant Linux API documentation. It will be much easier to maintain.

vbatts:
carrying this over to #292

PR #237, opened 2015-10-29
runtime-config-linux: Add a way to disable runtime cgroup manipulation

wking:
So folks can use a different model, if the current linux.resources
approach isn't scratching their itch.

For more motivation, see the discussion [here][1].

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -162,6 +162,9 @@ You can configure a container's cgroups via the `resources` field of the Linux c
 Do not specify `resources` unless limits have to be updated.
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
+If neither `resources` nor `cgroupsPath` is specified, the runtime MUST not manipulate or require the existence of any cgroups.
+You probably only want to do this if you are handling cgroups yourself via [hooks](runtime-config.md#hooks).
+
 #### Disable out-of-memory killer
 
 `disableOOMKiller` contains a boolean (`true` or `false`) that enables or disables the Out of Memory killer for a cgroup.


crosbymichael:
NOTLGTM

We are not even done discussion things and you are already opening PRs to change functionality of runtimes implementing the spec

wking:
On Thu, Oct 29, 2015 at 11:41:25AM -0700, Michael Crosby wrote:
> We are not even done discussion things and you are already opening
> PRs to change functionality of runtimes implementing the spec

I hadn't heard any pushback on “we need a way to disable the runtime's
cgroup handling”, and there was explicit positive feedback in [1] and
implicit acceptance in [2].  Is there a reason why “don't touch my
cgroups” would be hard to handle?  Or why “neither cgroupsPath nor
resources were specified” should mean something else?

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/CBlt7nCaDAAJ
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/5MMpjU2dDAAJ

vbatts:
@wking please don't just throw up PRs when the discussion is not close to a census.

wking:
On Thu, Oct 29, 2015 at 12:26:40PM -0700, Vincent Batts wrote:
> @wking please don't just throw up PRs when the discussion is not
> close to a census.

Punted back to the list in the sub-thread starting with [1].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/c9mv6qXtDAAJ
     Message-ID: <20151029194427.GA30073@odin.tremily.us>

PR #241, opened 2015-11-12
Update ROADMAP.md

RobDolinMS:
Adds RobDolinMS to help with validation tooling


vbatts:
two things. Need the DCO signature on the commit.
Also, as we tease out what it means to be an owner, what pieces are do you see yourself signing up for?

vbatts:
Let's move participants and working docs to the wiki. https://github.com/opencontainers/specs/wiki/RoadMap:

PR #248, opened 2015-11-16
ROADMAP: Link to list thread on a single, unified config file

wking:
I expect folks with competing views will also [post to the list][1],
either advocating their desired approach (in new threads?) or poking
holes in my proposal.  So having a list of links here will make it
easier for the decision makers to keep track of the proposals while
they review their relative merits.

[1]: https://github.com/opencontainers/specs/blob/v0.1.1/README.md#contributing


crosbymichael:
No, lets not add proposals to the roadmap, its a recipe for disaster and bikesheding

PR #250, opened 2015-11-17
use the lifecycle states to define the hooks

laijs:


hqhq:
Lifecycle modification is on going in #231 , this PR needs to be hold on.

laijs:
The intention of this pr is, as its title said, using the lifecycle states to define the hooks.
it has something(changing to the lifecycle) conflicting with #231, but this part is unimportant.
I will remove this part ("remove `(process)` from lifecycle states" and cross reference), and re-summit the pr. thx.

vbatts:
This looks fine for the most part. The switch from "exits" -> "stopped" seems to imply the hooks may get triggered before the container has fully ended. Like if the pid(s) are sent signals to terminate, but they hang in a Wait state, would the hooks still get called?

philips:
This PR is too confusing. "container is started" is confusing: the only way it has meaning is if we agree on a term that isn't "container" for the application process.

vbatts:
#231 is merged. Please update.

PR #251, opened 2015-11-17
Change the behavior when cgroupsPath is absent

hqhq:
The former definition was join "/" when `cgroupsPath` is absent, it's
not clear whether to join the root cgroup or create a sub cgroup under "/".

Join root cgroup would be a bad idea cause that's not what a container
should do. And craete a sub cgroup would be missing definition about what
the cgroup name should that be.

So I think we should leave this to implementations what the default
cgroup path should be.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -146,7 +146,7 @@ For more information, see the [kernel cgroups documentation](https://www.kernel.
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
 `cgroupsPath` is expected to be relative to the cgroups mount point.
-If not specified, cgroups will be created under '/'.
+If `cgroupsPath` is not specified, implementations can define the default cgroup path.
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
 The Spec does not support [split hierarchy](https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt).


wking:
On Tue, Nov 17, 2015 at 05:59:12AM -0800, Qiang Huang wrote:
> So I think we should leave this to implementations what the default
> cgroup path should be.

I'm +1 on this PR.  Users who care about what the cgroup will be or
where it will live in their hierarchy should set cgroupsPath.
Otherwise, the runtime should have full flexibility to do whatever it
thinks is best.  That's what we do now with container IDs, where runC
has an --id flag for folks who care, and internal defaults for folks
who don't (and this PR is even better, because it specifies a standard
way for the caller to supply their preferred value).

This is related to (but does not conflict with) #247.

runtime-config-linux.md
```diff
@@ -146,7 +146,7 @@ For more information, see the [kernel cgroups documentation](https://www.kernel.
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
 `cgroupsPath` is expected to be relative to the cgroups mount point.
-If not specified, cgroups will be created under '/'.
+If not specified or specified as empty string, implementations can define the default cgroup path.
```
vishh: Isn't empty string our means of `not specifying` cgroups?

vishh: How about rephrasing as `If cgroups path is not specified, behavior is undefined`?

hqhq:
@vishh You are right, specified as empty string would be the same as not specified, and I changed the description to `behavior undefined` as you suggested, thanks.

runtime-config-linux.md
wking: On Tue, Nov 17, 2015 at 05:43:12PM -0800, Vish Kannan wrote:
> > -If not specified, cgroups will be created under '/'.
> > +If not specified or specified as empty string, implementations can define the default cgroup path.
> 
> How about rephrasing as `If cgroups path is not specified, behavior
> is undefined`?

The behavior is defined: the runtime will create a new cgroup, add the
container process to it, and perform any resource allocation specified
by ‘resources’.  The only thing that's not defined when cgroupsPath is
not specified is the name of the cgroup that the runtime uses.  So I'd
rather go back to the original phrasing, where the only unspecified
behavior is how the runtime picks the default name used when
cgroupPath is unset.

hqhq: @wking Are you suggesting we go back to my original phrasing in this PR?

I think `undefined behavior` is OK, because I think undefined means we don't restrict this behavior, doesn't mean the behavior is unknown.

wking: On Wed, Nov 18, 2015 at 03:22:54AM -0800, Qiang Huang wrote:
> > -If not specified, cgroups will be created under '/'.
> > +If not specified or specified as empty string, implementations can define the default cgroup path.
> 
> @wking Are you suggesting we go back to my original phrasing in this PR?

Yes.

> I think `undefined behavior` is OK, because I think undefined means
> we don't restrict this behavior, doesn't mean the behavior is
> unknown.

It would mean that the spec has nothing to say about what happens when
cgroupsPath is unset, but I think we *do* want to define what happens
when cgroupsPath is unset (and just have the runtime pick its favorite
cgroupsPath name and then carry on as if cgroupsPath was set).

vishh: What do you mean by *favorite*? The fact that its implementation dependent,
makes it undefined from a Specification perspective.

vishh: This is specifically for developer workflow. Honestly, I don't see why we
are spending so much energy here @wking. This PR LGTM either ways.

wking: On Wed, Nov 18, 2015 at 10:23:39AM -0800, Vish Kannan wrote:
> > -If not specified, cgroups will be created under '/'.
> > +If not specified or specified as empty string, implementations can define the default cgroup path.
> 
> This is specifically for developer workflow.

It doesn't read that way to me.  “implementations can define the
default cgroup path” sounds like “if you don't set cgroupsPath, we'll
pick one for you” (which I think makes sense).  However, “behavior is
undefined” sounds like “the runtime can do *anything* it wants,
including just exiting with an error and not launching a container at
all” (which I think makes very little sense).

laijs:
"Undefined behavior" means (it is bug)/(it can corrupt)/etc. when cgroupsPath is absent.
I don't think you want to say it.

see
https://en.wikipedia.org/wiki/Undefined_behavior
https://en.wikipedia.org/wiki/Unspecified_behavior

I think we can use
If `cgroupsPath` is not specified, the behavior is unspecified.
or
If `cgroupsPath` is not specified, the behavior is implementation-defined.

hqhq:
@vishh @wking OK, I've updated this PR, PTAL.

hqhq:
@laijs Thanks for the links, I've changed it the original phrasing.

wking:
On Wed, Nov 18, 2015 at 05:56:53PM -0800, Qiang Huang wrote:
> @laijs Thanks for the links, I've changed it the original phrasing.

d663a5b0 looks good to me (again :).

mrunalp:
LGTM @crosbymichael @vishh @philips @vbatts PTAL

vbatts:
LGTM

PR #252, opened 2015-11-17
runtime-config-linux.md: fix the type of cpus and mems

laijs:


hqhq:
Although it's likely we'll change all cgroup settings to points in favor of #233 , it's definitely a bug in spec for now if I didn't miss anything, I'm OK we let it in first, so LGTM.

wking:
On Tue, Nov 17, 2015 at 06:26:28AM -0800, Qiang Huang wrote:
> Although it's likely we'll change all cgroup settings to points in
> favor of #233 , it's definitely a bug in spec for now if I didn't
> miss anything, I'm OK we let it in first, so LGTM.

+1 to all that ;).

mrunalp:
LGTM

PR #253, opened 2015-11-18
Clarify backwards compatibility for major version 0

hqhq:
I think we agreed on this, but there is no docs for this, and it is easy to cause confusing as https://github.com/opencontainers/runc/pull/406 and https://github.com/opencontainers/runc/pull/406#issuecomment-157685174

### DIFF
--- a/config.md
+++ b/config.md
@@ -10,7 +10,7 @@ Below is a detailed description of each field defined in the configuration forma
 
 ## Manifest version
 
-* **`version`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OCF specification with which the container bundle complies. The Open Container spec follows semantic versioning and retains forward and backward compatibility within major versions. For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series.
+* **`version`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OCF specification with which the container bundle complies. The Open Container spec follows semantic versioning and retains forward and backward compatibility within major versions. For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series. NOTE that there is no guarantee for forward or backward compatibility for version 0.x.
 
 *Example*
 


duglin:
```For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series.```
is this really true?  If a new feature was added in 1.2 I wouldn't expect a 1.1 impl to support it.

marcosnils:
@duglin what this really means is that if you have something that implements 1.0.1 it will still work with any 1.x series. Of course it won't support the new features, but you'll get guarantee that it won't break the current compatibility if you upgrade versions.

duglin:
yes I know that's the intent but that's not what it says.  :-)
Would be more accurate to say something like "features in 1.N will continue to wotk in 1.N+1"

marcosnils:
@duglin seems pretty clear for me. 

> For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series.
> is this really true? If a new feature was added in 1.2 I wouldn't expect a 1.1 impl to support it.

What I understand from this sentence is that if I have an impl that runs un 1.0.1 it will still run with any 1.x series.

duglin:
does a 1.0.1 feature work in 1.0.0 ?

marcosnils:
@duglin if my runtime implements the 1.0.0 series, even though I upgrade it to 1.0.1 it will still work (without the new features of course). I won't have the new features unless I update my runtime to support them.

mikebrow:
I figured the person writing the container was to put in his container the minimum version supported. Not sure why we are having him put in a version number if the meaning of the version number is more like a schema number?

wking:
On Wed, Nov 18, 2015 at 04:00:49AM -0800, Qiang Huang wrote:
> I think we agreed on this, but there is no docs for this…

There are docs.  We link to SemVer v2.0.0 [1], which has a point 4:

  Major version zero (0.y.z) is for initial development. Anything may
  change at any time. The public API should not be considered stable.

So I don't think we need to add a note reiterating that.

I *do* think we should drop the 1.0.1 sentence though, since (a)
compatibility is well covered by the SemVer spec, and getting all the
nuances right in a comment here is difficult and WET and (b) we
probably won't be cutting patch releases for the spec (more on this in
[2]).

[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/JlHbtEfKC3c
     Subject: Versioning specs in the face of ambiguity
     Message-ID: <20150902035544.GJ32473@odin.tremily.us>

vbatts:
as this only pertains to pre-release versions, it LGTM

crosbymichael:
LGTM

PR #255, opened 2015-11-24
Clarify the meaning of hook elements

hqhq:
I'm not sure if this is the intention here, but I think it'll be better
to keep consistency with golang packages.

### DIFF
--- a/runtime-config.md
+++ b/runtime-config.md
@@ -43,7 +43,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
         "prestart": [
             {
                 "path": "/usr/bin/fix-mounts",
-                "args": ["arg1", "arg2"],
+                "args": ["fix-mounts", "arg1", "arg2"],
                 "env":  [ "key1=value1"]
             },
             {
@@ -58,7 +58,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
         "poststop": [
             {
                 "path": "/usr/sbin/cleanup.sh",
-                "args": ["-f"]
+                "args": ["cleanup.sh", "-f"]
             }
         ]
     }
@@ -66,6 +66,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 
 `path` is required for a hook.
 `args` and `env` are optional.
+The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://golang.org/pkg/os/exec/#Cmd).
 
 ## Mount Configuration
 


runtime-config.md
```diff
@@ -66,6 +66,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 
 `path` is required for a hook.
 `args` and `env` are optional.
+The meanings are the same as `Path`, `Args` and `Env` in [golang Cmd](https://golang.org/pkg/os/exec/#Cmd).
```
mrunalp: s/meanings/semantics

hqhq:
@mrunalp Updated, thanks.

wking:
On Tue, Nov 24, 2015 at 03:13:14AM -0800, Qiang Huang wrote:
> I'm not sure if this is the intention here, but I think it'll be
> better to keep consistency with golang packages.

Previous discussion in #118 and [1].  My current preference is for
semantics like Python's “require args, path is optional” [2] instead
of Go's “require path, args is optional” [3], because I expect
additional arguments to be more common than needing to override
argv[0].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/luaaorsya10
     Subject: Explicit args[0] for hooks?
     Message-ID: <20150815191523.GK21585@odin.tremily.us>
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/luaaorsya10/IdjzUYl9GQAJ
     Message-ID: <20150911201050.GZ5912@odin.tremily.us>
[3]: https://golang.org/pkg/os/exec/#Cmd

philips:
We keep flip-flopping on this one and it is a total bikeshed. I don't think it matters at the end of the day but are there real systems out there where people specify the arg0 and don't use a symlink like /bin/busybox -> /bin/ls?!

hqhq:
@wking Thanks for your links. After briefly reading #118 and opencontainers/runc#160 , I think using the Go style is intended. I'm OK with Go style or Python style, to be honest, Python style seems a bit more reasonable but I prefer using Go style, just don't want to mix too many rules, and Go style works fine.

@philips Since this PR didn't change any semantics but clarify it, I don't think we are flip-flopping here. I don't know why specs gave examples that way, but that really cause confusing as https://github.com/opencontainers/runc/pull/413.

wking:
On Wed, Nov 25, 2015 at 05:47:49AM -0800, Brandon Philips wrote:
> We keep flip-flopping on this one and it is a total bikeshed.

Of the potential APIs listed in [1], I think (c) is unambiguously
superior.  Breaking it down again:

a. ‘path’ is required, and ‘[path] + args…’ is the executed argv

   This provides no way to override ‘argv[0]’.  It's where
   opencontainers/runc#160 started, and is like Go's Command [2].

b. ‘path’ is required, and ‘args…’ is the executed argv.  If ‘args’
   isn't specified, ‘[path]’ is the executed argv.

   This is where opencontainers/runc#160 ended up, and it's what Go
   uses [2].  This lets you override ‘argv[0]’, but when you don't
   need to override it you'll end up duplicating ‘argv[0]’ in both
   path and ‘args[0]’.

c. ‘path’ is optional, ‘args…’ is the executed argv, and ‘path’ takes
   precedence (when it's set) over ‘args[0]’ when searching for the
   executable.

   This is how Python does it [4], and its lovely.  You only need to
   touch ‘path’ when you need to override ‘argv[0]’.

d. ‘args…’ is thet executed argv, and ‘args[0]’ is used when searching
   for the executable.

   This is like (c), except without the ability to override ‘argv[0]’.
   I don't know of a language where this is significantly easier to
   implement than (c), so (d) seems needlessly harsh.

I haven't heard any pushback against (c) (other than a “variadic”
reference I'm not sure I understood [5]).  And yeah, (b) gives you all
the flexibility you need.  But (c) is equally powerful, often less
wordy, and never more wordy, so I don't see a reason to not adopt it.

[1]: https://github.com/opencontainers/specs/pull/118#issuecomment-138689707
[2]: https://golang.org/pkg/os/exec/#Command
[3]: https://golang.org/pkg/os/exec/#Cmd
[4]: https://docs.python.org/3/library/subprocess.html#subprocess.Popen
[5]: https://github.com/opencontainers/specs/pull/118#issuecomment-139644209

mrunalp:
Since this is just clarifying the way it works currently, it LGTM.

vbatts:
LGTM

PR #257, opened 2015-11-29
Example lists "root' but text mentions "bundlePath"

nponeccop:

PR #144, opened 2015-09-03
runtime: Add 'version' to the state.json example

wking:
The version field was added while 180df9d (Add runtime state
configuration and structs, 2015-07-29, #87) was [in-flight][1], and it
missed getting documented in the example.

[1]: https://github.com/opencontainers/specs/pull/87#issuecomment-135117343

PR #259, opened 2015-12-01
Update runtime example

### LINKED ISSUE #258: Example doesn't match the spec doc in for state.json
In `specs/runtime.md`, the example of a `state.json` file does not match the text immediately above it in the following ways:

* there is no `version` property
* the `bundlePath` property is called `root` in the example

marcosnils:
Fixes #258

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -27,9 +27,10 @@ This is provided so that consumers can find the container's configuration and ro
 
 ```json
 {
+    "version": "0.2",
     "id": "oc-container",
     "pid": 4422,
-    "root": "/containers/redis"
+    "bundlePath": "/containers/redis"
 }
 ```
 


runtime.md
```diff
@@ -27,9 +27,10 @@ This is provided so that consumers can find the container's configuration and ro
 
 ```json
 {
+    "version": "0.2",
```
wking: This is handled by #144, which I've just rebased to reflect the current 0.2.0 version.

runtime.md
```diff
@@ -27,9 +27,10 @@ This is provided so that consumers can find the container's configuration and ro
...
     "id": "oc-container",
     "pid": 4422,
-    "root": "/containers/redis"
+    "bundlePath": "/containers/redis"
```
wking: This is handled by #257.

wking:
I think both of the changes you're making are handled by existing PRs,
so I suggest we close this as a duplicate (and get those other PRs
landed :).

marcosnils:
Duplicate of #257 and #144

Issue #256, opened 2015-11-24
 Spec should have a reference to RFC2119 


mikebrow:
For clarity purposes the words like MUST should be defined by the spec. For example:

The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.. link: http://tools.ietf.org/html/rfc2119

wking:
On Tue, Nov 24, 2015 at 02:36:55PM -0800, Mike Brown wrote:
> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
> "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
> document are to be interpreted as described in RFC 2119.. link:
> http://tools.ietf.org/html/rfc2119

+1.

philips:
+1. We should probably have some huge flag day where we agree to make these
changes.

vbatts:
:+1:

wking:
This can probably be closed now that #260 has landed.

PR #260, opened 2015-12-01
adds text describing the upper-case keywords used in the spec

mikebrow:
It may not be flag day yet... but here's the PR to address issue #256
Signed-off-by: Mike Brown <brownwm@us.ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Table of Contents
   - [Linux Specific Runtime](runtime-linux.md)
 - [Implementations](implementations.md)
 
+In the specifications in the above table of contents, the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
+
 # Use Cases
 
 To provide context for users the following section gives example use cases for each part of the spec.


keyword-guidelines.md
```diff
@@ -0,0 +1,102 @@
+# Keyword Guidelines for the Specifications
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" as used in the specification
+documents are to be interpreted as described in [RFC 2119](http://www.ietf.org/rfc/rfc2119.txt) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997.)
+      
+## Introduction
```
duglin: I think adding the reference (and link) to RFC2119 is good enough. I don't think we should repeat it here, so I'd remove the text below.  Repeating it means that if, heaven forbid, there's a typo in there we'd have to explain which one wins and that's just annoying.  Avoid the entire issue by just having a reference.

duglin: Which means we probably don't need an entire file for it, just add it to the top of each stand-alone doc.

mikebrow:
Updated to address @duglin 's comments.

wking:
On Tue, Dec 01, 2015 at 01:57:53PM -0800, Mike Brown wrote:
> It may not be flag day yet...

This is all pre-1.0, so I don't see a need for a flag day.  c32dc93d
looks good to me, but I'd also be fine without the “Bradner … 1997”
parenthetical.

README.md
```diff
@@ -16,6 +16,11 @@ Table of Contents
   - [Linux Specific Runtime](runtime-linux.md)
 - [Implementations](implementations.md)
 
+In the specifications in the above table of contents, the keywords "MUST",
+"MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in
+[RFC 2119](http://www.ietf.org/rfc/rfc2119.txt) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997.)
```
wking: I prefer the HTML RFC renderings (e.g. [RFC 2199][1]), because they have anchors for linking, links to errata, obsoleting RFCs, etc.  They also have a link to the text and PDF renderings, so folks who would rather not view the HTML are a click away from their preferred format.

[1]: http://tools.ietf.org/html/rfc2119

mikebrow:
Changed the link to point to the root html page that links to the other renderings, errata, etc.

wking:
On Thu, Dec 03, 2015 at 12:34:17PM -0800, Mike Brown wrote:
> Changed the link to point to the root html page that links to the
> other renderings, errata, etc.

8cc7832 looks good to me.

hqhq:
@mikebrow In specs, all files should be formatted one sentence per line, see https://github.com/opencontainers/specs/blob/master/README.md#markdown-style

Otherwise, LGTM. Thanks.

mikebrow:
Changed to one line per spec md style requirement.

wking:
On Fri, Dec 04, 2015 at 07:35:01AM -0800, Mike Brown wrote:
> Changed to one line per spec md style requirement.   

Ah, I'd forgotten to check for that one ;).  57a196a0 still looks good
to me :).

mrunalp:
LGTM

mrunalp:
@crosbymichael @vbatts @philips PTAL

crosbymichael:
LGTM

PR #262, opened 2015-12-03
Add hyphen for host specific

hqhq:
We have `host-specific` and `host specific` in the same doc,
add hyphen to keep consistency.


vbatts:
good catch. I think it should go the other way though, and add hyphen to where it is missing.

hqhq:
@vbatts Updated.

vbatts:
LGTM

PR #263, opened 2015-12-09
*: printable documents

vbatts:
I do not like having this build step of printable documentation
depending on pulling a container, but the pandoc+latex combo is a big
bundle. This is the minimal and cleanest approach for using these tools,
for now.

Just run `make pdf && xdg-open ./output/docs.pdf`

Per: https://github.com/opencontainers/specs/blob/master/ROADMAP.md#printablecompiled-spec

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,43 @@
+
+DOCKER ?= $(shell which docker)
+DOC_FILES := \
+	README.md \
+	code-of-conduct.md \
+	principles.md \
+	ROADMAP.md \
+	implementations.md \
+	bundle.md \
+	runtime.md \
+	runtime-linux.md \
+	config.md \
+	config-linux.md \
+	runtime-config.md \
+	runtime-config-linux.md
+
+docs: pdf html
+
+pdf:
+	@mkdir -p output/ && \
+	$(DOCKER) run \
+	-it \
+	--rm \
+	-v $(shell pwd)/:/input/:ro \
+	-v $(shell pwd)/output/:/output/ \
+	-u $(shell id -u) \
+	vbatts/pandoc -f markdown_github -t latex -o /output/docs.pdf $(patsubst %,/input/%,$(DOC_FILES)) && \
+	ls -sh $(shell readlink -f output/docs.pdf)
+
+html:
+	@mkdir -p output/ && \
+	$(DOCKER) run \
+	-it \
+	--rm \
+	-v $(shell pwd)/:/input/:ro \
+	-v $(shell pwd)/output/:/output/ \
+	-u $(shell id -u) \
+	vbatts/pandoc -f markdown_github -t html5 -o /output/docs.html $(patsubst %,/input/%,$(DOC_FILES)) && \
+	ls -sh $(shell readlink -f output/docs.html)
+
+clean:
+	rm -rf output/ *~
+


Makefile
```diff
@@ -0,0 +1,30 @@
...
+	code-of-conduct.md \
+	ROADMAP.md \
+	implementations.md \
+	bundle.md \
+	runtime.md \
+	runtime-linux.md \
+	config.md \
+	config-linux.md \
+	runtime-config.md \
+	runtime-config-linux.md
```
tianon: Is maintaining this hard-coded list going to be a hassle?  (ie, would something like `$(wildcard *.md)` also be correct without the same maintainer cost?)

vbatts: I thought that too, but order matters for how it assembles them in the pdf

wking: On Wed, Dec 09, 2015 at 09:37:07AM -0800, Tianon Gravi wrote:
> > +DOC_FILES := \
> > +	README.md \
> > +	code-of-conduct.md \
> > +	ROADMAP.md \
> > +	…
> > +	runtime-config-linux.md
> 
> Is maintaining this hard-coded list going to be a hassle?  (ie,
> would something like `$(wildcard *.md)` also be correct without the
> same maintainer cost?)

It looks like the ordering is significant, so the wildcard approach
would require numbering the filenames (e.g. 01-README.md).  I think
the hard-coded approach is fine, since filename churn has been small
so far, and will only get smaller as the repository matures (excepting
a few isolated reorgs).

wking: On Wed, Dec 09, 2015 at 09:37:07AM -0800, Tianon Gravi wrote:
> > +DOC_FILES := \
> > +	README.md \
> > +	code-of-conduct.md \
> > +	ROADMAP.md \
> > +	implementations.md \
> > +	bundle.md \
> > +	runtime.md \
> > +	runtime-linux.md \
> > +	config.md \
> > +	config-linux.md \
> > +	runtime-config.md \
> > +	runtime-config-linux.md

List is missing principles.md (insert after code-of-conduct.md?).

vbatts:
preview here --> https://vbatts.fedorapeople.org/oci.pdf

Makefile
tianon: Ahhhh, ok, makes sense. :+1:

wking:
On Wed, Dec 09, 2015 at 09:53:07AM -0800, Vincent Batts wrote:
> preview here --> https://vbatts.fedorapeople.org/oci.pdf

That seems to be preserving our “one sentence per line”, which looks
pretty bad in compiled output ;).  I also don't think PDF is worth the
cost of the pandoc/LaTeX toolchain (especially if the internal links
don't work ;).  I think we might have better luck if we use Jekyll and
include the Markdown to get a long HTML doc [1], letting folks print
that to PDF if they still want a PDF.

[1]: http://jekyllrb.com/docs/templates/#includes

vbatts:
I did consider jekyll or similar. The we might could have it displayed on https://opencontainers.github.io/specs/ or such.

vbatts:
updated. added principle.md, an `html` target, and a default `docs` target that generates pdf and html

crosbymichael:
nice!

LGTM

vishh:
LGTM :heart:

vishh:
Would it be possible to auto-generate docs based on the schema (and comments) eventually?

vbatts:
@vishh nice hope. perhaps something like swagger. This is a first effort.

vbatts:
what do y'all say about where to have these generated artifacts per release?

vishh:
+1

dqminh:
@vbatts maybe uploaded as a tar file to github for now ? Would be nice to publish them to a more accessible place as well

tianon:
Yeah, I'd be +1 to GitHub releases for now. :smile:

vbatts:
right on. It looks like just doing a git push of a tag, makes a release, but I can not attach files that way. Files can only be attached by doing the release _through_ the github UI.

vbatts:
https://vbatts.fedorapeople.org/oci/v0.1.1/ for reference of the first tag...

tianon:
IIRC, you can make a "release" on top of an existing tag you've pushed though

vbatts:
well hot damn
https://github.com/opencontainers/specs/releases/tag/v0.1.1

vbatts:
removing the files from the reference host...

philips:
lgtm

vbatts:
3 LGTMs. merging

PR #264, opened 2015-12-09
index.md: Add a single-page overview for easy printing

wking:
As an alternative to #263.  Example HTML currently [here][1], and it's
not especially beautiful (e.g. it has the same “[links point out of
the file][2]” issue.  But we can fix/mitigate some of those issues
with Makefile rules and a gh-pages release build.  The pandoc/LaTeX
approach from #263 could fix link targets with similar Makefile rules,
so I think it's a wash there.  But this approach seems better to me
because “push to gh-pages” is a smaller barrier-to-entry than “install
this pandoc/LaTeX docker image”.

The _includes directory holds copies of files to avoid:

    A file was included in index.md that is a symlink or does not exist
    in your _includes directory. For more information, see
    https://help.github.com/articles/page-build-failed-file-is-a-symlink.

[1]: https://wking.github.io/opencontainer-specs/
[2]: https://github.com/opencontainers/specs/pull/263#issuecomment-163341758

### DIFF
--- a/_includes/README.md
+++ b/_includes/README.md
@@ -1,0 +1,144 @@
+# Open Container Specifications
+
+[Open Container Initiative](http://www.opencontainers.org/) Specifications for standards on Operating System process and application containers.
+
+
+Table of Contents
+
+- [Container Principles](principles.md)
+- [Filesystem Bundle](bundle.md)
+- Configuration
+  - [Container Configuration](config.md)
+  - [Container Configuration (Linux-specific)](config-linux.md)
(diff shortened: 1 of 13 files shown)


wking:
In case it wasn't clear from my topic message, d57693a is just a proof-of-concept sketching out this approach.  If it seems like a useful direction, I can add some of the polishing I mentioned in the topic message (e.g. we'll want a rule to automatically copy `*.md` into `_includes`).

wking:
The #263 approach landed, so there's no need for this.

PR #266, opened 2015-12-11
Clarify cgroupsPath to be absolute

hqhq:
The former description is not accurate, it only says `relative to the cgroups mount point`,
if the path is absolute, it's easy to understand, but if it's relative, should it still be relative
to the cgroups mount point?

I think we can define it to be only absolute path, or we can also clarify if it's relative,
let implementation define what it would be relative to.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -145,7 +145,7 @@ cgroups provide controls to restrict cpu, memory, IO, pids and network for the c
 For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
+`cgroupsPath` is expected to be an absolute path which is relative to the cgroups mount point.
 If `cgroupsPath` is not specified, implementations can define the default cgroup path.
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.


wking:
On Thu, Dec 10, 2015 at 11:54:10PM -0800, Qiang Huang wrote:
> The former description is not accurate, it only says `relative to
> the cgroups mount point`, if the path is absolute, it's easy to
> understand, but if it's relative, should it still be relative to the
> cgroups mount point?

I'd understood it to mean “relative to the cgroups mount point
(regardless of a leading slash)”.

> I think we can define it to be only absolute path, or we can also
> clarify if it's relative, let implementation define what it would be
> relative to.

I don't think “relative to an implementation-defined thing” is
particularly useful.  How about we adjust the example to:

  "cgroupsPath": "myRuntime/myContainer"

and just error out if cgroupsPath starts with a slash?  Then we
clearly meant “relative to the cgroups mount point” for all provided
cgroupsPaths.

mrunalp:
I'd prefer it to be specified as an absolute path (internally relative to cgroups path). One reason is that it will be consistent with what one sees in /proc/<pid>/cgroup.
For e.g.
```
memory:/user.slice/user-0.slice/session-1.scope
```

wking:
On Wed, Dec 16, 2015 at 10:10:26AM -0800, Mrunal Patel wrote:
> I'd prefer it to be specified as an absolute path (internally
> relative to cgroups path).

I think that “absolute but internally relative to the cgroups mount
point” is harder to explain than “relative to the cgroups mount
point”, but…

> One reason is that it will be consistent with what one sees in
> /proc/<pid>/cgroup.
>
> For e.g.
> ```
> memory:/user.slice/user-0.slice/session-1.scope
> ```

+1 for consistency.  If this is the consistency we're shooting for
(vs. consistency with “relative paths” more generally), I'd mention it
in the spec.

hqhq:
Close in favor of https://github.com/opencontainers/specs/pull/270 
They are fixing the same problem.

PR #267, opened 2015-12-11
specs: linux: add intel_rdt cgroup support in specs

xiaochenshen:
Add support for the intel_rdt cgroup resource in Linux-specific
configuration to support config.json.

The intel_rdt cgroup subsystem will be available in Linux 4.6 (or later).

Signed-off-by: Xiaochen Shen xiaochen.shen@intel.com

This is the prerequisite of this runc proposal: https://github.com/opencontainers/runc/issues/433
For more information about intel_rdt cgroup, please refer to: https://github.com/opencontainers/runc/issues/433

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -163,7 +163,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 ## Control groups
 
 Also known as cgroups, they are used to restrict resource usage for a container and handle device access.
-cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
+cgroups provide controls to restrict cpu, memory, IO, pids, network and intel_rdt for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
@@ -451,6 +451,23 @@ The following paramters can be specified to setup the controller:
 ```json
    "pids": {
         "limit": 32771
+   }
+```
+
+#### Intel rdt 
+
+`intelRdt` represents the cgroup subsystem `intel_rdt`.
+For more information, see [the intel_rdt cgroup man page](https://lkml.org/lkml/2015/12/17/574).
+
+The following paramters can be specified to setup the controller:
+
+* **`l3Cbm`** *(uint64, optional)* - specifies L3 cache capacity bitmask (CBM) in the cgroup
+
+###### Example
+
+```json
+   "intelRdt": {
+        "l3Cbm": 4080 
    }
 ```
 
--- a/config_linux.go
+++ b/config_linux.go
@@ -211,6 +211,12 @@ type Network struct {
 	Priorities []InterfacePriority `json:"priorities,omitempty"`
 }
 
+// IntelRdt for Linux cgroup 'intel_rdt' resource management
+type IntelRdt struct {
+	// L3 cache capacity bitmask (CBM) for container
+	L3Cbm *uint64 `json:"l3Cbm,omitempty"`
+}
+
 // Resources has container runtime resource constraints
 type Resources struct {
 	// Devices are a list of device rules for the whitelist controller
@@ -231,6 +237,8 @@ type Resources struct {
 	HugepageLimits []HugepageLimit `json:"hugepageLimits,omitempty"`
 	// Network restriction configuration
 	Network *Network `json:"network,omitempty"`
+	// IntelRdt restriction configuration
+	IntelRdt *IntelRdt `json:"intelRdt,omitempty"`
 }
 
 // Device represents the mknod information for a Linux special device file


wking:
On Fri, Dec 11, 2015 at 11:48:42AM -0800, Xiaochen Shen wrote:
> Add support for the intel_rdt cgroup resource in Linux-specific
> runtime configuration to support runtime.json.

There has been discussion of factoring cgroup handling out of this
spec [1].  I'm not sure where that discussion stands (I'm still
personally in favor of using hooks for cgroups), but it's probably
worth getting some clarity on the current OCI consensus before pushing
forward with this.

If we do want to keep a thin wrapper around all the kernel cgroup APIs
in the specs, than the suggested changes look like they cover the new
controller's only resource file [2].

[2]: https://lkml.org/lkml/2015/10/2/74

mrunalp:
LGTM

hqhq:
Needs rebase and fit current resource value types.

xiaochenshen:
I have git-rebased against latest master branch. New commit: 3f61bc569d59215a2a0822f75814e13121707264

runtime_config_linux.go
```diff
@@ -216,6 +222,8 @@ type Resources struct {
 	HugepageLimits []HugepageLimit `json:"hugepageLimits,omitempty"`
 	// Network restriction configuration
 	Network *Network `json:"network,omitempty"`
+	// IntelRdt restriction configuration
+	IntelRdt IntelRdt `json:"intelRdt,omitempty"`
```
hqhq: I think it's better to use pointer to be consistent with other fields, you can see from https://github.com/opencontainers/specs/pull/233 why we are doing this.

xiaochenshen: Thanks. Changed the types of IntelRdt and L3Cbm to **pointer** according to #233.
Commit: 233311be23d559919fbe3b9890de7319ce115d54

hqhq:
Sorry didn't mention it before, can you add more description about intel_rdt cgroup in `runtime-config-linux.md` like the other subsystems?

xiaochenshen:
@hqhq Added intel_rdt cgroup description in `runtime-config-linux.md` in commit f5ca4800ee7d33449cdd2d4ad0b970616d57058a

hqhq:
I'm not sure if an on going patchset is proper for specs reference, or maybe we should wait it merge to kernel upstream before add it to specs? Other than that, the whole change LGTM.

Let's wait for one more maintainer's input before merge this.
ping @vishh @crosbymichael @LK4D4 @vbatts @dqminh

mrunalp:
I did not realize that this wasn't merged yet. In that case, we should wait for it to be merged to the kernel first.

runtime_config_linux.go
```diff
@@ -198,6 +198,12 @@ type Network struct {
 	Priorities []InterfacePriority `json:"priorities"`
 }
 
+// IntelRdt for Linux cgroup 'intel_rdt' resource management
+type IntelRdt struct {
+	// L3 cache capacity bitmask (CBM) for container
+	L3Cbm *string `json:"l3Cbm,omitempty"`
```
vishh: Why is it a string and not an integer?

vishh:
+1 for holding on to this PR until this feature is merged and becomes stable upstream.
In general, for new experimental features, the spec should be extensible and not require updating the Spec.

runtime_config_linux.go
wking: On Tue, Dec 22, 2015 at 11:19:43AM -0800, Vish Kannan wrote:
> > +	// L3 cache capacity bitmask (CBM) for container
> > +	L3Cbm *string `json:"l3Cbm,omitempty"`
> 
> Why is it a string and not an integer?

So folks can use hex [1] (which JSON doesn't support [2])?  That's
probably not a good enough reason, since convenient hand-editing is a
lower priority for us than convenient machine readability.  A rule for
handling traditionally-hex values should go into a best-practices doc
(#273) if/when we decide to do that.

[1]: https://lkml.org/lkml/2015/10/2/74
[2]: http://json.org/

xiaochenshen:
@mrunalp Sure, we would wait for the kernel patch 'upstreamed' in case of any changes in the data structures or APIs

philips:
This is another example of the spec being of questionable valuable with regards to encapsulating the entire cgroups API. I created a tag of "platforms/linux/blocked-upstream" to track these.

runtime_config_linux.go
xiaochenshen: @vishh @wking Thank you for pointing this out.
In this cgroup context, intel_rdt.l3_cbm represents the L3 cache capacity **bitmask**. In kernel, by nature, read/write with 'hexadecimal string' is the best choice for user-friendly and user-readable. Kernel handles it as 'u64' via 'write_u64' interface inside for 'machine readability'.

In oci/specs, currently I choose the type 'string' instead of 'uint64' for some reasons:
1. JSON doesn't natively support hexadecimal, as @wking mentioned. For me it is really an important reason. If JSON supports hexadecimal or something more readable for bitmask, 'uint64' type inside with '0x' prefix is perfect, we don't need discuss this issue here any more.  
2. User-readability VS machine-readability. JSON doesn't support hexadecimal means user-readability issue even though kernel treats hexadecimal (0x-prefix), octal (0-prefix) and decimal integers without discrimination. For example (l3Cbm: 0xff0, bits 4-11 are set):
(1) Type string:
```json
"intel_rdt": {
    "l3Cbm": "0xff0"
}
```
(2) Type uint64:
```json
"intel_rdt": {
    "l3Cbm": 4080
}
```
(3) If JSON supports hexadecimal, with type uint64 looks like:
```json
"intel_rdt": {
    "l3Cbm": 0xff0
}
```
Case (3) should be perfect, but it doesn't work. Case (2), '4080' is very confusing as a bitmask value for end user. As a compromise, we use case (1) now.

Anyway, I will not always stick to my point if we have a better solution. Actually, when I was implementing https://github.com/opencontainers/runc/issues/433, I found the string/integer conversion back-and-forth is really chatty due to JSON's limitation on non-decimal integer support.

wking: On Tue, Dec 22, 2015 at 09:22:58PM -0800, Xiaochen Shen wrote:
> Case (2), '4080' is very confusing as a bitmask value for end user.

The counter-argument here is that the target audience for the
config-file isn't end users [1,2,3], which can always insert some
tooling in between themselves and the JSON config.  For example, it
wouldn't be too bad to write a parser that converts a JSON-extension
form like:

  "intel_rdt": {
    "l3Cbm": 0xff0
  }

to official JSON.  And it's even easier (from an implementation
perspective, if not from a usability perspective) to do:

  $ python -c 'print(hex(4080))'
  0xff0

So my preference is “use integer types for integer values, and deal
with the lack of JSON hex support using other tools”.  But I don't
feel strongly enough about that put up a fight if the consensus is
that we want to use JSON strings for integers that are more intuitive
as hex.

[1]: https://www.opencontainers.org/news/faqs/how-broad-mission-oci
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/duzW7SIGDQAJ
     Subject: Re: removal of cgroups from the OCI Linux spec
     Date: Thu, 29 Oct 2015 20:15:13 -0700
     Message-ID: <CACfVidd8sV14rG6S2DX1mJ35ij1-Y7sAxir4p-U9EscCeqCMwA@mail.gmail.com>
[3]: https://github.com/opencontainers/specs/pull/275#discussion_r48074805

xiaochenshen: >wking added a note 18 hours ago

>The counter-argument here is that the target audience for the
>config-file isn't end users [1,2,3], which can always insert some
>tooling in between themselves and the JSON config.  For example, it
>wouldn't be too bad to write a parser that converts a JSON-extension
>form like:

@wking It makes sense.If JSON or similar config-file is not always UI-like thing,
integer is better than string even for 'bitmask' in the case.

Actually, I tried to use type 'uint64' instead of 'string' in the v1 patch of 
https://github.com/opencontainers/runc/pull/447, the code looks better, and the JSON config file looks like:

```json
"intel_rdt": {
    "l3Cbm": 4080
}
```
@others, please tell me if you have different opinions.
Otherwise, I will change the type of *IntelRdt.L3Cbm* from *string to *uint64 later.

xiaochenshen: >xiaochenshen added
>@others, please tell me if you have different opinions.
>Otherwise, I will change the type of IntelRdt.L3Cbm from *string to *uint64 later.

Ping @mrunalp @vishh @hqhq @philips

vishh: +1 for `*uint64`.

xiaochenshen: I have changed the type of 'IntelRdt.L3Cbm' from *string to *uint64 in commit e91f77f8c4ad6c964deaec449c7e3371206b43fa

xiaochenshen:
I have submitted latest commit e91f77f8c4ad6c964deaec449c7e3371206b43fa to address the comments of the data type discussion. But I found 2/3 of Travis CI build is failed (go: 1.4.3, 1.3.3), but 1 is passed (go: 1.5.1).

Who knows why?

vbatts:
hmmm. perhaps something in the upstream `vet` changed that is broken on go1.3 and go1.4?
https://travis-ci.org/opencontainers/specs/jobs/100780130#L117

vbatts:
@xiaochenshen i retriggered the tests, and they ran fine.

xiaochenshen:
@vbatts Thanks.

xiaochenshen:
The latest commit 38df0663594e98f78dcbf5da6f11bde2e072deb8 rebased e91f77f8c4ad6c964deaec449c7e3371206b43fa for unified config file according to #284.

hqhq:
@xiaochenshen Thanks for your PR, since it's blocked by Linux upstream, you don't need to rebase now as we won't merge until it's landed to Linux upstream.

But you can rebase your PR in runC, we can merge that without specs change.

xiaochenshen:
Fixed typo in config-linux.md. The latest commit: c3b1cf7f38a70314fd942617949633c1f44fbb93

xiaochenshen:
@hqhq I plan to update both specs and runc and make them synced.

vbatts:
Is this still waiting on upstream linux 4.6? (given that 4.5 was only released this week...)

xiaochenshen:
@vbatts Yes. Kernel patch is still pending. The interface/API of the kernel patch may change per community's comments. I will track the change and modify this PR accordingly.

tianon:
Can we get a status update here? :pray:

xiaochenshen:
@tianon This PR depends on Intel RDT kernel patch. But the kernel interface has changed from cgroup to "resource control" interface (cgroup-like). Currently, the kernel patch has finished code reviewed and will be likely to merge in 4.10 window.

Accordingly, this PR will be obsoleted by new PR which based on the new kernel interface soon. The implementation is working in progress now. Please find more information from https://github.com/opencontainers/runc/issues/433

xiaochenshen:
This PR could be closed for the new PR https://github.com/opencontainers/runtime-spec/pull/630 is open.

PR #270, opened 2015-12-16
Update cgroupsPath to cgroupName and clarify the semantics around that.

vishh:
@mrunalp

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -144,19 +144,29 @@ Also known as cgroups, they are used to restrict resource usage for a container 
 cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
-The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+The name of the cgroups can be specified in the Spec via `cgroupsName`.
+Cgroup names mimic filesystem paths closely since they express a hierarchy of cgroups.
+`cgroupsName` is expected to be absolute.
+An absolute name is one that is defined from the root cgroup `/`.
+For example, `/foo/bar` is acceptable. `foo/bar` is not acceptable.
+Allowable characters for cgroup names are,
+Alpha numeric ([a-zA-Z0-9]+)
+Underscores (_)
+Dashes (-)
+Periods (.)
+In general and unless otherwise specified, regular filesystem path rules apply.
+
+If `cgroupsName` is not specified, implementations can choose to use (or not use) any cgroups.
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
 The Spec does not support [split hierarchy](https://www.kernel.org/doc/Documentation/cgroups/unified-hierarchy.txt).
 The cgroups will be created if they don't exist.
 
 ```json
-   "cgroupsPath": "/myRuntime/myContainer"
+   "cgroupsName": "/foo-runtime/bar.container"
 ```
 
-`cgroupsPath` can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+`cgroupsName` can be used to either control the cgroups for new containers or to run a new process in an existing container.
 
 You can configure a container's cgroups via the `resources` field of the Linux configuration.
 Do not specify `resources` unless limits have to be updated.
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -25,10 +25,10 @@ type LinuxRuntime struct {
 	// Resources contain cgroup information for handling resource constraints
 	// for the container
 	Resources *Resources `json:"resources,omitempty"`
-	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
-	// The path is expected to be relative to the cgroups mountpoint.
-	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
-	CgroupsPath *string `json:"cgroupsPath,omitempty"`
+	// CgroupsName specifies the name of to cgroups that are created and/or joined by the container.
+	// The name is expected to be absolute.
+	// If resources are specified, the CgroupsName cgroups will be updated based on resources.
+	CgroupsName string `json:"cgroupsName,omitempty"`
 	// Namespaces contains the namespaces that are created and/or joined by the container
 	Namespaces []Namespace `json:"namespaces"`
 	// Devices are a list of device nodes that are created and enabled for the container


runtime-config-linux.md
```diff
@@ -144,19 +144,29 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
+Cgroup names mimic filesystem paths closely since they express a hierarchy of cgroups.
+`cgroupsName` is expected to be absolute.
+An absolute name is one that is defined from the root cgroup `/`.
+For example, `/foo/bar` is acceptible. `foo/bar` is not acceptible.
+Allowable characters for cgroup names are,
+Alpha numeric ([a-zA-Z0-9]+)
+Underscores (_)
+Dashes (-)
+Periods (.)
+In general and unless otherwise specified, regular filesystem path rules apply.
```
wking: Do we need the explicit allowed-character restrictions if we're punting to “regular filesystem rules” (which are a bit more relaxed than the character restrictions you list here)?

runtime-config-linux.md
```diff
@@ -144,19 +144,29 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
+
+If `cgroupsName` is not specified, implementations can choose to use (or not use) any cgroups.
```
wking: I'm not a fan of this.  What if you want to disable the runtime's cgroup handling so you can setup your own cgroups via hooks?  See #237 and the [associated sub-thread][1].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/c9mv6qXtDAAJ

wking:
I left a few nits inline, but except for [1] this PR is a wash for me.

[1]: https://github.com/opencontainers/specs/pull/270/files#r47825022

runtime-config-linux.md
wking: Clarification: my previous suggestions about disabling runtime cgroup handling involved the `resources` setting.  If the config includes a `resources` setting but doesn't include `cgroupsName`, I'm not sure how this proposed line would work.  Are some runtimes going to pick a path automatically and apply the configured limitations while others error out saying that `cgroupsName`  wasn't set?

runtime-config-linux.md
```diff
@@ -144,19 +144,29 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
 For more information, see the [kernel cgroups documentation](https://www.kernel.org/doc/Documentation/cgroups/cgroups.txt).
 
-The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+The name of the cgroups can be specified in the Spec via `cgroupsName`.
+Cgroup names mimic filesystem paths closely since they express a hierarchy of cgroups.
+`cgroupsName` is expected to be absolute.
+An absolute name is one that is defined from the root cgroup `/`.
+For example, `/foo/bar` is acceptible. `foo/bar` is not acceptible.
```
mrunalp: s/acceptible/acceptable/

runtime-config-linux.md
mrunalp: @wking IMO, it should mean pick their own default. We need a way to say don't do anything and for that resources == nil and cgroupsName == "" should work.

wking: On Wed, Dec 16, 2015 at 12:57:42PM -0800, Mrunal Patel wrote:
> > +If `cgroupsName` is not specified, implementations can choose to use (or not use) any cgroups.
> 
> @wking IMO, it should mean pick their own default. We need a way to
> say don't do anything and for that resources == nil and cgroupsName
> == "" should work.

Or cgroupsName = nil to avoid needing to qualify the slash-start
requirement (see also *string discussion in #233), but yeah.

For the requirements != nil && cgroupsName = nil case, are you also
recommending we keep the current:

  If `cgroupsName` is not specified, implementations can define the
  default cgroup path.

instead of the proposed “optional cgroups [and error if the runtime
choses not to use cgroups]” (keeping the current logic makes more
sense to me [1])?

[1]: https://github.com/opencontainers/specs/pull/270#discussion_r47825968

mrunalp: @wking Yes, if the resources != nil and no cgrouppath is specified then the runtime should pick a cgroup path on its own to apply those settings. It should be clear that this happens and runtime shouldn't opt out of applying the settings.

runtime-config-linux.md
hqhq: Yeah, a link would be better.

hqhq:
Any thought about how systemd implementation should perform on this? (not a question for this PR, but for the general cgroupsPath/Name definition)

mrunalp:
@hqhq systemd api work with a slice and a scope name, so would have to split the name into those parts and make the call to create a scope.

hqhq:
@mrunalp So we want users on systemd cgroup specify the exactly *right* `cgroupName` such as `/system.slice/my.slice/app.scope` and we need to validate it before apply?
Because for now, we are handling `name` and `parent` to assign slice and unitName. I'm facing the problem when working on https://github.com/opencontainers/runc/pull/397 as trying to remove `name` and `parent` in runc.

runtime-config-linux.md
vishh: In terms of characters, only the ones mentioned above are supported. But path construction follow filesystem path rules.
@hqhq: What link do you need?

runtime-config-linux.md
vishh: If resources are mentioned, runtimes will have to somehow satisfy them. How they setup and manage cgroups to achieve that can remain outside of the scope of the Spec.

vishh:
@hqhq @mrunalp: Comment addressed. PTAL.

runtime-config-linux.md
vishh: Fixed.

runtime_config_linux.go
```diff
@@ -25,10 +25,10 @@ type LinuxRuntime struct {
...
 	// for the container
 	Resources *Resources `json:"resources,omitempty"`
-	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
-	// The path is expected to be relative to the cgroups mountpoint.
-	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
-	CgroupsPath *string `json:"cgroupsPath,omitempty"`
+	// CgroupsName specifies the name of to cgroups that are created and/or joined by the container.
+	// The name is expected to be absolute.
+	// If resources are specified, the CgroupsName cgroups will be updated based on resources.
+	CgroupsName string `json:"cgroupsName,omitempty"`
```
wking: This should keep the `*string` from #233.

vishh: Why? There are only two options here, an empty string "" or a valid name.
There isn't a third system default.

runtime-config-linux.md
wking: On Mon, Dec 28, 2015 at 11:56:49AM -0800, Vish Kannan wrote:
> > +In general and unless otherwise specified, regular filesystem path rules apply.
> 
> In terms of characters, only the ones mentioned above are
> supported. But path construction follow filesystem path rules.
> @hqhq: What link do you need?

I can't find a reference to this in the Linux kernel docs, but POSIX
says pretty clearly that everything except the null and slash bytes
are fair game as long as the full byte name isn't ‘.’ or ‘..’ [1].

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_170

runtime_config_linux.go
wking: On Mon, Dec 28, 2015 at 12:14:05PM -0800, Vish Kannan wrote:
> > +	// CgroupsName specifies the name of to cgroups that are created and/or joined by the container.
> > +	// The name is expected to be absolute.
> > +	// If resources are specified, the CgroupsName cgroups will be updated based on resources.
> > +	CgroupsName string `json:"cgroupsName,omitempty"`
> 
> Why? There are only two options here, an empty string "" or a valid name.
> There isn't a third system default.

This:

  "cgroupsName": "",

seems illegal to me (it's not absolute).  If we use missing/null
cgroupsName, we don't have to put in an absolute-path loophole for the
empty string.  For more background on nil over empty strings for
“unset”, see [1] and surrounding comments, as well as [2].

[1]: https://github.com/opencontainers/specs/pull/233#discussion_r47829711
[2]: https://github.com/opencontainers/specs/pull/270/files#r47838511

runtime-config-linux.md
hqhq: @vishh I was thinking of a link about the definition of "what is the allowable characters of a regular filesystem path", if that does exist. Or I'm also OK if we can't find a proper link because I think that's common enough to be accepted.

mlaventure: Note that this is false for systemd, `/` within the parent path are rejected by the dbus service.

vishh:
@mrunalp @crosbymichael: Can you review this patch if you get a chance?

hqhq:
The major problem is it's not working on systemd cgroup. Should that be concerned in spec?

mlaventure:
Couldn't we just state that the format of the path depends on the backend and redirect the user to those for more information?

vishh:
@hqhq: What part of this PR is not working with systemd?

vishh:
In the case of systemd, would it be possible to translate the path into a
systemd compatible name?
The Spec currently aligns closely with Linux APIs. I'd like to stick to
that here as well.

mlaventure:
That's certainly doable, although in the unlikely event that systemd changes its syntax or that a new manager chooses something different we would have to update our implementation.

wking:
On Wed, Jan 27, 2016 at 01:57:31PM -0800, Kenfe-Mickaël Laventure wrote:
> That's certainly doable, although in the unlikely event that systemd
> changes its syntax…

I don't know how much stock I'd put in systemd's commitment to
backwards compatibility ;).  Decoupling cgroup managers from the
runtime, so you can use a systemd-driving manager or a
filesystem-driving-manager as you see fit still sounds good to me [1].
As far as I know, progress in that direction is still blocking on a
way to opt-out from runtime cgroup manipulation [2].

[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/qWHoKs8Fsrk/c9mv6qXtDAAJ
     Subject: Make runtime cgroups optional (was: removal of cgroups from the OCI Linux spec)
     Date: Thu, 29 Oct 2015 12:44:27 -0700
     Message-ID: <20151029194427.GA30073@odin.tremily.us>

hqhq:
The intention of `cgroupName` is to specify which cgroup will the container be in, for this purpose, cgroupfs API and systemd API need different information, a path can be used by cgroupfs API, but all systemd API needs is a slice name and a unit name (a scope according to the rule of systemd).

I'm not sure what to be defined in specs is the best to cover these. Maybe add a new entry like `slice`?

mrunalp:
We should probably leave this out of the spec and handle this in the runtime.

vishh:
Leave what out of the Spec? We have to include cgroup names right?

mlaventure:
@vishh: I think @mrunalp meant to leave the actual format taken by `cgroupsName` out of the spec, not the field itself. I tend to agree with this.

vishh:
@mrunalp @mlaventure @crosbymichael: How do we make progress here? The spec needs to define a standard way to specify cgroups. Systemd is an implementation detail. We need to standardize on the naming format. Otherwise the Spec and the corresponding bundle will not be portable. 
I'd really like for us to not tie the Spec to an specific implementation. We can always construct implementation specific names from a standardized naming format.

mlaventure:
The issue is that atm both implementations are hugely different.

systemd requires a name and a scope (which can't not include any `/` character) whereas the default linux usage is based on filepath.

So far the only way I can see us reconciliating both of them is by requiring a cgroup name of the form `x-y-z` then systemd would take the last element as the scope (i.e. `z` in that example) and everything else as the slice name  (i.e. `x-y`) whereas for fs we would just replace the `-` with `/`.

But I do not find this particularly elegant.

vishh:
What about converting the Linux format to systemd compatible one when
required within the runtime?

hqhq:
Close in favor of #493

PR #271, opened 2015-12-16
implementations: Reference the CNI as a Linux network hook

wking:
Based on @philips' [post][1].  Container-ID injection would take a
wrapping shell instance or some such (e.g., see [here][2]), but I
imagine config authors can figure that out (or just write an
alternative wrapper script that extracts the container ID from stdin's
state JSON on its own).

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/9VMBA2dAeuU/OGl8tsNBCwAJ
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/9VMBA2dAeuU/LzELsZO-CgAJ

### DIFF
--- a/implementations.md
+++ b/implementations.md
@@ -19,3 +19,21 @@ If you know of any associated projects that are not listed here, please file a p
 ## Testing
 
 * [huawei-openlab/oct](https://github.com/huawei-openlab/oct) - Open Container Testing framework for OCI configuration and runtime
+
+## Hooks
+
+For addressing container issues that are outside the scope of this specification, the following hooks may be useful.
+
+### Linux
+
+#### Networking
+
+* [CNI][], the Container Network Interface.
+  The [`exec-plugins.sh`][cni-exec-plugins.sh] script can be used in both [prestart][] (with `add`) and [poststop][] (with `del`) hooks insert and remove a container from networks [configured][cni-conf] in a host-side directory like `/etc/cni/net.d`.
+
+[prestart]: runtime-config.md#prestart
+[poststop]: runtime-config.md#poststop
+
+[CNI]: https://github.com/appc/cni
+[cni-exec-plugins.sh]: https://github.com/appc/cni/blob/v0.1.0/scripts/exec-plugins.sh
+[cni-conf]: https://github.com/appc/cni/blob/v0.1.0/SPEC.md


mrunalp:
LGTM

implementations.md
```diff
@@ -19,3 +19,25 @@ If you know of any associated projects that are not listed here, please file a p
 ## Testing
 
 * [huawei-openlab/oct](https://github.com/huawei-openlab/oct) - Open Container Testing framework for OCI configuration and runtime
+
+## Hooks
+
+For addressing container issues that are outside the scope of this
```
LK4D4: I wonder what became with `To keep consistency throughout the Markdown files in the Open Container spec all files should be formatted one sentence per line.` I personally hate that rule, but I know that you love it :)

crosbymichael:
Not against the change but there is literally nothing in this pr or the script that would tell me how to use it with an OCI container.  Its like a having a link to some random script saying that it could be used as a hook.

implementations.md
wking: On Thu, Dec 17, 2015 at 10:44:35AM -0800, Alexander Morozov wrote:
> > +For addressing container issues that are outside the scope of this
> 
> I wonder what became with … one sentence per line.

Fixed in 1c301f1 → 3f89557, thanks :).

> I personally hate that rule, but I know that you love it :)

I'm not a fan of that *choice* in particular, but I am a fan of having
a rule of some sort (e.g. I'd also be ok with “Markdown must be
wrapped at 72 characters”, but that's probably harder for folks to
implement consistently).  Bikeshedding about or churning the
repository for re-wrapping Markdown is just a waste of time.

wking:
On Thu, Dec 17, 2015 at 10:53:04AM -0800, Michael Crosby wrote:
> Not against the change but there is literally nothing in this pr or
> the script that would tell me how to use it with an OCI container.
> Its like a having a link to some random script saying that it could
> be used as a hook.

In the absence of a runtime that is compatible with v0.1.0 (currently
our only release) of this spec, providing detailed examples seemed
premature ;).  As it stands this PR is just to help discovery,
pointing folks upstream to CNI docs and over to our hook docs for the
details.

If we added a detailed example, would you prefer it to be in this
file?  Upstream in the CNI repository?  Somewhere else?

ChengTiesheng:
I think maybe a simple example or link can be given at the first, then we can enrich it step by step with the feedback from OCI community.

vbatts:
lets put examples on the wiki, until we formalize patterns like this 
https://github.com/opencontainers/specs/wiki

implementations.md
```diff
@@ -19,3 +19,21 @@ If you know of any associated projects that are not listed here, please file a p
...
+## Hooks
+
+For addressing container issues that are outside the scope of this specification, the following hooks may be useful.
+
+### Linux
+
+#### Networking
+
+* [CNI][], the Container Network Interface.
+  The [`exec-plugins.sh`][cni-exec-plugins.sh] script can be used in both [prestart][] (with `add`) and [poststop][] (with `del`) hooks insert and remove a container from networks [configured][cni-conf] in a host-side directory like `/etc/cni/net.d`.
```
vishh: minor nit: Can we avoid describing the usage of plugins here and instead just mention `CNI` as an option?
I'm thinking if this PR will become a reason for other net plugin vendors to include their specification here.

wking: On Tue, Jan 26, 2016 at 10:47:43AM -0800, Vish Kannan wrote:
> minor nit: Can we avoid describing the usage of plugins here and
> instead just mention `CNI` as an option?  I'm thinking if this PR
> will become a reason for other net plugin vendors to include their
> specification here.

I think this PR is on hold until someone has time to translate it into
a wiki page [1].  But personally, I'm +1 to net-plugin vendors pushing
docs (ideally, links to external docs) that describe how to setup OCI
hooks for their plugins.  Having a collection of references to the
various net-plugin options would help avoid confusion like [2].

[1]: https://github.com/opencontainers/specs/pull/271#issuecomment-165669973
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/9VMBA2dAeuU/_tlRFaNTCgAJ
     Subject: no vNIC in the container, what is a good default .json template?
     Date: Thu, 10 Dec 2015 12:18:07 -0800 (PST)
     Message-Id: <acbb4587-8d80-452a-8908-446f14a5107c@opencontainers.org>

wking: On Tue, Jan 26, 2016 at 10:54:42AM -0800, W. Trevor King wrote:
> I think this PR is on hold until someone has time to translate it into
> a wiki page [1]…

More food for an eventual wiki page: on IRC today [@crosbymichael
pointed out][1] [netns][2], which includes docs on specifically
integrating via a pre-start hook.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-02-05.log.html#t2016-02-05T21:34:49
[2]: https://github.com/jfrazelle/netns

vbatts:
Honestly, lets put tangential examples like this in the wiki for now.

PR #276, opened 2015-12-18
source and schema: differentiate with examples

vbatts:
The standard is on the JSON schema (not yet IETF spec JSON-schema), such
that it is not implemenations specific. Thus far, the reference has been
in how golang source renders the JSON documents.

Having the JSON source and the markdown documents in sync has been an
ongoing step to keep in sync.

Separating these two allows the golang source to continue being _a_
reference, but the JSON schema in the documentation to be _the_
reference.

As validation tooling is refined, then it will facilitate ensuring
the available golang source conforms to the reference JSON.


config.md
```diff
@@ -127,4 +127,77 @@ For Linux-based systems the user structure has the following fields:
 Interpretation of the platform section of the JSON file is used to find which platform-specific sections may be available in the document.
 For example, if `os` is set to `linux`, then a JSON object conforming to the [Linux-specific schema](config-linux.md) SHOULD be found at the key `linux` in the `config.json`.
 
+## Configuration Schema Example
+
+Here is a full exmample `config.json` contents for reference (generated with `runc spec`).
```
wking: nit: “exmample” → “example”, and I'd drop “contents” and “(generated with `runc spec`)”.  The generation note may be useful for a commit message, but doesn't seem useful in the spec itself.

config.md
```diff
@@ -127,4 +127,77 @@ For Linux-based systems the user structure has the following fields:
...
+	"platform": {
+		"os": "linux",
+		"arch": "amd64"
+	},
+	"process": {
+		"terminal": true,
+		"user": {
+			"uid": 0,
+			"gid": 0,
+			"additionalGids": null
```
wking: If we're looking for a full config (i.e. excercising all available options), we should probably include a real `additionalGids` value (and similar for other optional fields).  If the intention is just to show an example that explores most of the available options for anchoring the individual snippets (like [this one][1]), then I'd probably just drop the no-op `additionalGids` (because the surrounding entries provide enough context to anchor the existing user snippet).

[1]: https://github.com/opencontainers/specs/blob/e79365a749f86254dec96609caff2a8acd15f689/config.md#process-configuration

config.md
```diff
@@ -127,4 +127,77 @@ For Linux-based systems the user structure has the following fields:
...
+		},
+		"args": [
+			"sh"
+		],
+		"env": [
+			"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+			"TERM=xterm"
+		],
+		"cwd": ""
```
wking: Is an empty-string **`process.cwd`** valid?  What does it mean?  If it's “don't change the working directory”, then I'd just leave this out.  If it means setting a particular wording directory, then I'd explicitly set it to that directory (e.g. `"cwd": "/"`).  For what it's worth `cd ''` seems to be a no-op.

config.md
```diff
@@ -127,4 +127,77 @@ For Linux-based systems the user structure has the following fields:
...
+		},
+		{
+			"name": "sysfs",
+			"path": "/sys"
+		},
+		{
+			"name": "cgroup",
+			"path": "/sys/fs/cgroup"
+		}
+	],
```
wking: Excepting mqueue and cgroup, these are all covered by #164, so I'd leave the covered entries out (see mrunalp/ocitools#15).

runtime-config.md
```diff
@@ -120,3 +120,228 @@ Only [mounts from the portable config](config.md#mount-points) will be mounted.
...
+				"nosuid",
+				"noexec",
+				"nodev"
+			]
+		}
+	},
+	"hooks": {
+		"prestart": null,
+		"poststart": null,
+		"poststop": null
```
wking: [Another][1] location where I think we should fill in something useful.  If we remove the whole section, there wouldn't be any context here to anchor the [hook specs][2].

[1]: https://github.com/opencontainers/specs/pull/276/files#r48075833
[2]: https://github.com/opencontainers/specs/blob/e79365a749f86254dec96609caff2a8acd15f689/runtime-config.md#poststop

wking:
On Fri, Dec 18, 2015 at 01:50:31PM -0800, Vincent Batts wrote:
> Separating these two allows the golang source to continue being _a_
> reference, but the JSON schema in the documentation to be _the_
> reference.

I'm not sure these examples have much normative weight, but they
certainly are useful for showing how the normative sections (e.g. [1])
fit together.  And it would let runC and similar drop the examples
they currently embed [2], by providing a centrally-managed example.
So +1 on this going in (although I've left a few inline comments about
the example content).

[1]: https://github.com/opencontainers/specs/blob/e79365a749f86254dec96609caff2a8acd15f689/config.md#manifest-version
[2]: https://github.com/opencontainers/runc/pull/242#discussion_r38484100

vishh:
+1 for the overall intent of separating schema from the golang structures.

crosbymichael:
The only downside is that it makes the Go import path ugly.

wking:
On Fri, Dec 18, 2015 at 02:52:59PM -0800, Michael Crosby wrote:
> The only downside is that it makes the Go import path ugly.

We expect Go import paths to get more interesting soon anyway, no?
@mrunalp's platform-directory revival [1] seemed to go over well in
the 2015-12-16 meeting.  I don't remember hearing any pushback other
than @duglin's concern that it was *adding* platform-dependent
compilation instead of removing it [2], and once we cleared that up I
think everyone was in favor of it [3].

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2015/opencontainers.2015-12-16-18.00.log.html#l-39
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2015/opencontainers.2015-12-16-18.00.log.html#l-51
[3]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2015/opencontainers.2015-12-16-18.00.log.html#l-53

vishh:
@vbatts: How do we differentiate between optional and required fields? And how will this address extensible fields?

wking:
On Fri, Dec 18, 2015 at 05:46:21PM -0800, Vish Kannan wrote:
> How do we differentiate between optional and required fields? And
> how will this address extensible fields?

The normative specification will still be things like [1].  This is
just a global example to help tie all of that together.

[1]: https://github.com/opencontainers/specs/blob/e79365a749f86254dec96609caff2a8acd15f689/config.md#manifest-version

vbatts:
@vishh that has to be done per field and value. Like the comments that describe them.
```markdown

* **`path`** (string, required) ...
* **`readonly`** (bool, optional) ...

```

It's not very parsable, but is legible.

vbatts:
@crosbymichael you thinking those paths would be ugly was the first thing that came to mind. What do you recommend besides `github.com/opencontainers/specs/src/specs` ?

vbatts:
otherwise, I've rebased and updated the structs a little based on comments.
PTAL

wking:
On Mon, Dec 21, 2015 at 12:24:27PM -0800, Vincent Batts wrote:
> otherwise, I've rebased and updated the structs a little based on comments.

All the ecccada → 99d1df7 changes look good to me, but this is still
duplicating fields required by #164 [1].  And there are still a number
of nulls in the resources that can be left out now that #233 has
landed a pointer-based approach to “don't touch this”.

[1]: https://github.com/opencontainers/specs/pull/276/files#r48076330

vbatts:
rebased. PTAL

wking:
On Wed, Jan 13, 2016 at 01:22:20PM -0800, Vincent Batts wrote:
> rebased. PTAL

  $ diff -u <(git show -M50 99d1df7) <(git show -M50 998344)
  --- /dev/fd/63  2016-01-13 13:33:17.048080367 -0800
  +++ /dev/fd/62  2016-01-13 13:33:17.048080367 -0800
  @@ -1,4 +1,4 @@
  -commit 99d1df777601345d84e80fffd09f075c1b9d0436
  +commit 99834455b165462575821c0b3bb6e586fa8e8759
   Author: Vincent Batts <vbatts@hashbangbash.com>
   Date:   Fri Dec 18 15:47:21 2015 -0500

So that is just “rebased onto the new master to pickup changes in the
renamed files”.  I still see the nulls I suggest we remove [1].

[1]: https://github.com/opencontainers/specs/pull/276#issuecomment-166411095

mrunalp:
We can do this using ocitools to generate and a Makefile target to update this.

wking:
On Wed, Jan 13, 2016 at 02:43:21PM -0800, Mrunal Patel wrote:
> We can do this using ocitools to generate and a Makefile target to
> update this.

I'm concerned that:

1. Write normative Markdown (this repo)
2. Update informative Go types (this repo)
3. Update ocitools to pull in new Go types
4. Generate new example JSON with ocitools
5. Update these informative examples (this repo)

is too long a loop, involving a number of non-normative moving parts.
It also means that without some ocitools hoop-jumping, you'd have to
land a single semantic update in two PRs (one PR for (1) and (2), and
another PR for (5)).

But if the idea is that PR submitters will use whatever tooling they
want (including a Make-spawned ocitools or hand-editing) to update
this section, then +1 to anything folks think makes that easier.

vbatts:
rebased

vbatts:
brainstorming: would the sample json be better served as an `config.example.json` file sitting in the directory, rather than appended to the `*.md`?

wking:
On Wed, Jan 20, 2016 at 10:31:44AM -0800, Vincent Batts wrote:
> brainstorming: would the sample json be better served as an
> `config.example.json` file sitting in the directory, rather than
> appended to the `*.md`?

I'm ok with that (which file this example lives in doesn't seem
particularly important to me, but keeping it in Markdown may make it
easier to embed in the single-page spec from #263).

My main concern for normative docs is that we have some way to anchor
a given setting in the config, but I can address that with a separate
PR adding better linking between parent ↔ child settings (e.g. [1,2]).

[1]: https://github.com/opencontainers/specs/blob/v0.2.0/runtime-config-linux.md#control-groups
[2]: https://github.com/opencontainers/specs/blob/v0.2.0/runtime-config-linux.md#disable-out-of-memory-killer

vishh:
+1

vbatts:
i've rebased this, but presently the example shown in `config.md` is wrong. I'll have to update this example json.

vbatts:
rebased and updated the example JSON. PTAL.

config.md
```diff
@@ -212,4 +212,160 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
...
+		"path": "rootfs",
+		"readonly": true
+	},
+	"hostname": "slartibartfast",
+	"mounts": [
+		{
+			"destination": "/proc",
+			"type": "proc",
+			"source": "proc"
+		},
```
wking: This is [still][1] [duplicating][2] mounts required by #164.  It sounds like [runC is coming around][3] on this point as well.

[1]: https://github.com/opencontainers/specs/pull/276/files#r48076330
[2]: https://github.com/opencontainers/specs/pull/276#issuecomment-166411095
[3]: https://github.com/opencontainers/ocitools/pull/2#issuecomment-181303007

config.md
```diff
@@ -212,4 +212,160 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
...
+		"namespaces": [
+			{ "type": "pid" },
+			{ "type": "network" },
+			{ "type": "ipc" },
+			{ "type": "uts" },
+			{ "type": "mount" }
+		],
+		"devices": null,
+		"apparmorProfile": "",
+		"selinuxProcessLabel": "",
```
wking: If these are all the Go-default values, why specify them?  I'd guess we're just [missing a few `omitempty`s and/or pointers][1].

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-02-05.log.html#t2016-02-05T21:52:27

mrunalp:
@crosbymichael @LK4D4 @vishh @hqhq @dqminh PTAL

config.md
mrunalp: yeah, we could probably skip these.

vbatts:
rebased

wking:
On Thu, Mar 03, 2016 at 12:26:39PM -0800, Vincent Batts wrote:
> rebased

It looks like 2b39bd30e (your current commit) is missing updates for
the landed-yesterday #329.

vbatts:
fixed

config.md
```diff
@@ -231,4 +231,160 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
...
+			"TERM=xterm"
+		],
+		"cwd": "/",
+		"capabilities": [
+			"CAP_AUDIT_WRITE",
+			"CAP_KILL",
+			"CAP_NET_BIND_SERVICE"
+		],
+		"apparmorProfile": "",
+		"selinuxProcessLabel": ""
```
wking: Before these moved [@mrunalp and I were suggesting we drop the empty values][1].  Since #329 [they're omitempty][2], so I see no reason to keep them in this example.

[1]: https://github.com/opencontainers/specs/pull/276#discussion_r52622034
[2]: https://github.com/opencontainers/specs/pull/329/files#diff-b4bda758a2aef091432646c354b4dc59R39

crosbymichael: Its fine to have these in an example because how else would ppl know that they are there without them being in an EXAMPLE

wking: On Tue, Mar 08, 2016 at 10:43:21AM -0800, Michael Crosby wrote:
> > +		"apparmorProfile": "",
> > +		"selinuxProcessLabel": ""
> 
> Its fine to have these in an example because how else would ppl know
> that they are there without them being in an EXAMPLE

I'm fine listing them here with meaningful content.  And regardless of
whether they're listed here, I think people should be able to discover
them via the process documentation [1] (which seems to use
‘selinuxLabel’ and not ‘selinuxProcessLabel’ anyway).

[1]: https://github.com/opencontainers/specs/blob/3b7c15d90f4becef0767b66c0d175a3ad653ff7e/config.md#process-configuration

vbatts:
rebased

mrunalp:
Should we format the json using jq to get a consistent output?

crosbymichael:
```go

import "github.com/opencontainers/specs/spec-go"
```

vbatts:
@crosbymichael done

vbatts:
rebased again

crosbymichael:
LGTM

config.md
```diff
@@ -244,4 +244,175 @@ Annotations are key-value maps.
...
+                "nodev",
+                "relatime",
+                "ro"
+            ]
+        }
+    ],
+    "hooks": {
+        "prestart": [
+            {
+                "path": "/",
```
mrunalp: This should be /usr/bin/uptime
args[0] could match that or be something else or just empty.

vbatts: fixed

mrunalp:
LGTM

PR #279, opened 2015-12-22
Fix value of swappiness

hqhq:
It's officially pointer of uint64 now, no point it can be
-1, change it to 0 as other fields in example.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -214,7 +214,7 @@ The following parameters can be specified to setup the controller:
         "swap": 0,
         "kernel": 0,
         "kernelTCP": 0,
-        "swappiness": -1
+        "swappiness": 0
     }
 ```
 


wking:
On Mon, Dec 21, 2015 at 10:59:50PM -0800, Qiang Huang wrote:
> It's officially pointer of uint64 now, no point it can be
> -1, change it to 0 as other fields in example.

e4b4ba2c4 looks good to me.  But just to be clear, the old -1 (don't
touch this) maps to not not setting this key (or setting the value to
null) in the post-#233 pointer world.  0 swappiness has always meant
“prevents from any swapping even if there is a swap storage available”
[1].

[1]: https://www.kernel.org/doc/Documentation/cgroups/memory.txt

philips:
Yea, I think for illustrative purposes of showing the available fields this is fine. LGTM

vbatts:
LGTM

PR #280, opened 2015-12-23
README: add some user stories

philips:
I am trying to document who might want to read this spec. I constantly get asked who should be looking at OCI and why. Lets see how successful I am.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,10 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
 
 To provide context for users the following section gives example use cases for each part of the spec.
 
-## Filesystem Bundle & Configuration
+#### Filesystem Bundle & Configuration
 
 - A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+
 
 # Releases
 
--- a/README.md
+++ b/README.md
@@ -22,10 +22,20 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
 
 To provide context for users the following section gives example use cases for each part of the spec.
 
-#### Filesystem Bundle & Configuration
+#### Application Bundle Builders
 
-- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files required for launching an application as a container.
+The bundle contains OCI [configuration files](config.md) where the builder can specify host-independent details such as [which executable to launch](config.md#process-configuration) and host-specific settings such as [mount](runtime-config.md#mount-configuration) locations, [hook](runtime-config.md#hooks) paths, Linux [namespaces](runtime-config-linux.md#namespaces) and [cgroups](runtime-config-linux.md#control-groups).
+Because the configuration includes host-specific settings, application bundle directories copied between two hosts may require configuration adjustments.
 
+#### Hook Developers
+
+[Hook](runtime-config.md#hooks) developers can extend the functionality of an OCI-compliant runtime by hooking into a container's lifecycle with an external application.
+Example use cases include sophisticated network configuration, volume garbage collection, etc.
+
+#### Runtime Developers
+
+Runtime developers can build runtime implementations that run OCI-compliant bundles and container configuration, containing low-level OS and host specific details, on a particular platform.
 
 # Releases
 


README.md
```diff
@@ -22,9 +22,17 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
 
 To provide context for users the following section gives example use cases for each part of the spec.
 
-## Filesystem Bundle & Configuration
+**Container Configuration**
```
wking: I'd rather stick to headers (they go [all the way down to h6][1] ;), so we get link anchors.

[1]: http://daringfireball.net/projects/markdown/syntax#header

README.md
```diff
@@ -22,9 +22,17 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
...
 
-- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+User: Application writer. Container configuration defines the environment that an application can rely on being available when it is launched inside of a container, this includes things such as environment variables, a working directory, mount points, and operating system permissions.
```
wking: I'd use “bundle author” or maybe “packager” here instead of “application writer”, since the later might also sounds like an author of the bundled software (e.g. [Nginx authors][1]), when we mean the one who put the bundle together and wrote the OCI config(s).  And there's also some [dislike of the word “application”][2] (although I'm personally fine with it).

[1]: https://github.com/nginx/nginx/graphs/contributors
[2]: https://github.com/opencontainers/specs/pull/107#issuecomment-145615479

wking:
On Tue, Dec 22, 2015 at 10:10:47PM -0800, Brandon Philips wrote:
> I am trying to document who might want to read this spec.

+1 to this :).  The content sections look like they're currently
focusing on details about in the spec, and not on the relationships
between user-types that the spec enables.  How about something like:

  # User stories

  ## Bundle author

  Bundle authors can write and publish [bundles](bundle.md) to
  distribute software along with any required libraries or auxiliary
  tools.  They can provide OCI [configuration](config.md)
  [files](runtime-config.md) in the bundle to configure default
  resources limits so runtime callers have reasonable defaults
  (although runtime callers are free to adjust these limits).

  ## Runtime author

  Runtime authors can write runtime implementations to allow
  OCI-compliant bundles to run on a particular platform (or with a
  particular containerization-technology).

  ## Hook author

  [Hook](runtime-config.md#hooks) authors can write commands to handle
  functionality that is outside the scope of basic container
  technology (for example, sophisticated network configuration).

  ## Runtime callers

  Runtime callers can use a [consistent runtime
  API](https://github.com/wking/oci-command-line-api) to launch
  containers from bundles provided by bundle authors.  They can also
  adjust the bundle author's provided OCI config to tailor the
  container to their environment (for example, by injecting hooks or
  adjusting resource limits).

Although until the command-line API lands somewhere more official, we
should probably leave the runtime-callers section out.

philips:
@wking I like your approach by user better, I will fix if up.

mrunalp:
+1 to user approach

philips:
PTAL @wking @mrunalp

README.md
```diff
@@ -22,9 +22,21 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
 
 To provide context for users the following section gives example use cases for each part of the spec.
 
-## Filesystem Bundle & Configuration
+#### Application Bundle Builders
 
-- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files and some low-level details for launching an application as a container.
+The application can in-turn rely on certain OS configuration being available when it is launched inside of a container by a runtime.
```
jonboulle: in-turn -> in turn

jonboulle: It's pretty unclear what low level configuration the user (builder) must specify and what will be provided by something else (presumably the runtime). Maybe this is just conflating the two "users"? (builder and application).

README.md
```diff
@@ -22,9 +22,21 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
...
+
+#### Hook Writers
+
+[Hook](runtime-config.md#hooks) authors can extend the functionality of an OCI-compliant runtime by hooking into a containers lifecycle with an external application.
+A hook can manipulate the container using state information passed to the hook at the time of the lifecycle event.
+Example use cases include sophisticated network configuration, volume garbage collection, etc.
+
+#### Runtime Writer
+
+Runtime authors can build runtime implementations that run OCI-compliant filesystem bundles and container configuration, containing low-level OS and host specific details, on a particular platform.
```
wking: I'd drop “filesystem …” through the end of the line and just use “[bundles](bundle.md).”.  We shouldn't be explaining what a bundle is here.

My [initial suggestion][1] tried to sketch out why you would start a new runtime implementation when we already have a handful of projects, but I'm fine leaving that off if it seems too tangential.

[1]: https://github.com/opencontainers/specs/pull/280#issuecomment-166948973

README.md
```diff
@@ -22,9 +22,21 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
...
+#### Application Bundle Builders
 
-- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files and some low-level details for launching an application as a container.
+The application can in-turn rely on certain OS configuration being available when it is launched inside of a container by a runtime.
+The OS configuration will include things such as environment variables, a working directory, mount points, and operating system permissions.
+
+#### Hook Writers
+
+[Hook](runtime-config.md#hooks) authors can extend the functionality of an OCI-compliant runtime by hooking into a containers lifecycle with an external application.
```
wking: ”containers” → “container's”.

README.md
```diff
@@ -22,9 +22,21 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
...
+A hook can manipulate the container using state information passed to the hook at the time of the lifecycle event.
```
wking: I don't see a need for this level of detail since we already link to the relevant spec section.  The earlier “extend” sentence and the following “use cases include” sentence are all we need to cover here.

README.md
wking: On Thu, Dec 24, 2015 at 06:24:56AM -0800, Jonathan Boulle wrote:
> >  
> > -- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
> > +Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files and some low-level details for launching an application as a container.
> > +The application can in-turn rely on certain OS configuration being available when it is launched inside of a container by a runtime.
> 
> It's pretty unclear what low level configuration the user (builder)
> must specify and what will be provided by something else (presumably
> the runtime).

I'm pretty sure this is calling attention to the requirements added in
in #164.  And again, I don't think we need to get down to this level
of detail here.  Folks are packaging some software in a container
because they want it to be easy to run in a reliable, safe manner.
They pick the OCI format because it's a popular (hopefully ;) standard
with broad (hopefully ;) support.  Making a low-level sales pitch for
the OCI format (vs. other formats?) seems to be off-target for a
doc-section where we're trying to paint pictures of who these
spec-user classes are.

philips: OCI is low-level. It points and and in many cases requires things like cgroups paths. I think someone familiar with docker image format or appc would be surprised by these requirements.

wking: On Fri, Dec 25, 2015 at 07:46:34AM -0800, Brandon Philips wrote:
> > +Application bundle builders can create a [bundle](bundle.md)
> >  directory that includes all of the files and some low-level
> >  details for launching an application as a container.
> > +The application can in-turn rely on certain OS configuration
> >  being available when it is launched inside of a container by a
> >  runtime.
> 
> OCI is low-level. It points and and in many cases requires things
> like cgroups paths. I think someone familiar with docker image
> format or appc would be surprised by these requirements.

Clearing up that surprise makes more sense to me than the “can in-turn
rely on” phrasing (which doesn't link to docs about what OS
configuration it's referring to).  How about:

  Application bundle builders can create a [bundle](bundle.md)
  directory that includes all of the files required for launching an
  application as a container.  The bundle also contains OCI
  [configuration](config.md) [files](runtime-config.md) where you can
  declare both high-level settings ([which executable to
  launch](config.md#process-configuration)) and low-level settings
  (which [namespaces](runtime-config-linux.md#namespaces) and
  [cgroups](runtime-config-linux.md#control-groups) to join).

Although I still prefer my initial focus on resource limits [1], as it
avoids mentioning the Linux-specific cgroups/namespaces.

[1]: https://github.com/opencontainers/specs/pull/280#issuecomment-166948973

README.md
```diff
@@ -22,9 +22,20 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
...
+Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files and some low-level details for launching an application as a container.
+The application can in-turn rely on certain OS configuration being available when it is launched inside of a container by a runtime.
+The OS configuration will include things such as environment variables, a working directory, mount points, and operating system permissions.
+
+#### Hook Writers
+
+[Hook](runtime-config.md#hooks) authors can extend the functionality of an OCI-compliant runtime by hooking into a container's lifecycle with an external application.
+Example use cases include sophisticated network configuration, volume garbage collection, etc.
+
+#### Runtime Writer
```
vishh: nit: s/Writer/Developer

vbatts: +1

README.md
```diff
@@ -22,9 +22,20 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
 
 To provide context for users the following section gives example use cases for each part of the spec.
 
-## Filesystem Bundle & Configuration
+#### Application Bundle Builders
 
-- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files and some low-level details for launching an application as a container.
```
mrunalp: low-level details in a configuration file..

README.md
philips: I think we need to stop trying to try and be OS agnostic until we get details on a second target. We are just inventing language without a target abstraction.

philips:
Please take a look at this latest attempt.

README.md
```diff
@@ -22,9 +22,20 @@ In the specifications in the above table of contents, the keywords "MUST", "MUST
 
 To provide context for users the following section gives example use cases for each part of the spec.
 
-## Filesystem Bundle & Configuration
+#### Application Bundle Builders
 
-- A user can create a root filesystem and configuration, with low-level OS and host specific details, and launch it as a container under an Open Container runtime.
+Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files required for launching an application as a container.
+The bundle contains OCI [configuration files](config.md) where the builder can specify host-independent details such as [which executable to launch](config.md#process-configuration) and host-specific settings such as [mount](runtime-config.md#mount-configuration) locations, [hook](runtime-config.md#hooks) paths, Linux [namespaces](runtime-config-linux.md#namespaces) and [cgroups](runtime-config-linux.md#control-groups).
+Because the configuration includes host-specific settings application bundle directories copied between two hosts may not work and will require configuration adjustments.
```
wking: “settings application” → “settings, application”.  And I'd replace “may not work and will require configuration adjustments” with “may require configuration adjustments” (to soften the “will”), because there's a chance that it does work as copied.

wking:
On Wed, Jan 06, 2016 at 02:13:16AM -0800, Brandon Philips wrote:
> Please take a look at this latest attempt.

Left some trivial copy-edit suggestions [1], but 1ca1981 → b5b252d
looks good to me.

[1]: https://github.com/opencontainers/specs/pull/280#discussion_r48969158

README.md
philips: I will change it to may but in almost every non-trivial example it will require changes.

philips:
Can I get an lgtm and merge?

mrunalp:
LGTM

vishh:
LGTM

PR #282, opened 2015-12-23
Appropriately mark optional fields as omitempty

arcnmx:
`runc spec` creates files with plenty of empty arrays, `null` values, and empty strings. While I'm mostly concerned about the latter case (some parsers will consider `"" != undefined`), the rest is ugliness that could be done without anyway.

### DIFF
--- a/config.go
+++ b/config.go
@@ -13,7 +13,7 @@ type Spec struct {
 	// Root is the root information for the container's filesystem.
 	Root Root `json:"root"`
 	// Hostname is the container's host name.
-	Hostname string `json:"hostname"`
+	Hostname string `json:"hostname,omitempty"`
 	// Mounts profile configuration for adding mounts to the container's filesystem.
 	Mounts []MountPoint `json:"mounts"`
 }
@@ -27,10 +27,10 @@ type Process struct {
 	// Args specifies the binary and arguments for the application to execute.
 	Args []string `json:"args"`
 	// Env populates the process environment for the process.
-	Env []string `json:"env"`
+	Env []string `json:"env,omitempty"`
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
-	Cwd string `json:"cwd"`
+	Cwd string `json:"cwd,omitempty"`
 }
 
 // Root contains information about the container's root filesystem on the host.
--- a/config_linux.go
+++ b/config_linux.go
@@ -21,5 +21,5 @@ type User struct {
 	// GID is the group id.
 	GID uint32 `json:"gid"`
 	// AdditionalGids are additional group ids set for the container's process.
-	AdditionalGids []uint32 `json:"additionalGids"`
+	AdditionalGids []uint32 `json:"additionalGids,omitempty"`
 }
--- a/runtime_config.go
+++ b/runtime_config.go
@@ -15,19 +15,19 @@ type RuntimeSpec struct {
 // Hook specifies a command that is run at a particular event in the lifecycle of a container
 type Hook struct {
 	Path string   `json:"path"`
-	Args []string `json:"args"`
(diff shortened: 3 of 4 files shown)


wking:
Missing a Signed-off-by [1], but other than that 9b772149 looks good
to me.

[1]: https://github.com/opencontainers/specs#sign-your-work

arcnmx:
> sorry, no pseudonyms or anonymous contributions

yeah, um... that doesn't work for me. Can I assert (a) and (b) and I guess have someone else sign / commit it..?

hqhq:
> yeah, um... that doesn't work for me. Can I assert (a) and (b) and I guess have someone else sign / commit it..?

@arcnmx Sorry I don't quite get it, what do you mean by this?
We need every patch be signed before merge, and usually it should be signed by author him/herself.

arcnmx:
@hqhq I can certainly sign it, but it will say `Signed-off-by: arcnmx <arcnmx@users.noreply.github.com>` which seems to not be allowed by the guidelines as written in the readme.

hqhq:
@arcnmx You can change to your real name and email by hand when committing the patch.

jonboulle:
I think the issue is that s/he would like to remain anonymous, not that s/he doesn't know how to do it.

@arcnmx I'm happy to put up this patch on your behalf if you want.

wking:
On Thu, Dec 24, 2015 at 05:52:42AM -0800, Jonathan Boulle wrote:
> @arcnmx I'm happy to put up this patch on your behalf if you want.

I'm not a lawyer, and I don't understand what motivated the “no
pseudonyms” language.  Git landed similar restrictions [1,2] and the
phrasing there sounds like “we want to be able to convince a court
that the person sitting here is the DCO-signer”.  I'm not sure what
the OCI's motivation is, but allowing sign-offs by proxy seems to
defeat any purpose behind a “no pseudonym” restriction.  I'd rather
see OCI lift the restriction or explain why they need it.

[1]: https://git.kernel.org/cgit/git/git.git/commit/?id=672752470c85c8dc3dd9b49f72063a99f8475b4a
[2]: http://article.gmane.org/gmane.comp.version-control.git/103614

arcnmx:
@jonboulle thanks, that would be appreciated.

@wking 
> allowing sign-offs by proxy seems to defeat any purpose behind a “no pseudonym” restriction

Well, let's be pedantic developers for a moment here! It says the sign-off must be under a legal name, so the paper trail leads to someone, but the origin of a change does not necessarily have to be tied to a name. The signer just has to assert `(a)` or `(b)`, and I guess takes the heat if they're wrong. It simply fulfills the requirements!

But eh, I think it's a silly restriction. It's taken from [the linux kernel](http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/SubmittingPatches#n409)

jonboulle:
This is a change I'd also like to see happen so before waiting for https://github.com/opencontainers/specs/pull/282#issuecomment-167131398 I've put up #283

wking:
On Thu, Dec 24, 2015 at 08:37:41AM -0800, arcnmx wrote:
> The signer just has to assert `(a)` or `(b)`, and I guess takes the
> heat if they're wrong.

It's this last bit that I think it tricky.  But if @jonboulle is
willing to step up and do that, it works for me ;).

arcnmx:
> It's this last bit that I think it tricky. But if @jonboulle is willing to step up and do that, it works for me ;).

Eh, you're just asserting that, to the best of your knowledge, the code signed off is licensed under a license compatible with the project. It's no different than incorporating any other open source code or dependency really. You're not even on the hook for anything if it somehow goes wrong :P

In any case, I don't like the requirement :<

hqhq:
Picked up by #283

PR #286, opened 2015-12-29
define the default value for the `cwd`

laijs:

### DIFF
--- a/config.md
+++ b/config.md
@@ -69,7 +69,7 @@ The runtime MUST mount entries in the listed order.
 ## Process configuration
 
 * **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.
-* **`cwd`** (string, optional) is the working directory that will be set for the executable.
+* **`cwd`** (string, optional) is the working directory that will be set for the executable. Defaults to "/" on linux.
 * **`env`** (array of strings, optional) contains a list of variables that will be set in the process's environment prior to execution. Elements in the array are specified as Strings in the form "KEY=value". The left hand side must consist solely of letters, digits, and underscores `_` as outlined in [IEEE Std 1003.1-2001](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html).
 * **`args`** (string, required) executable to launch and any flags as an array. The executable is the first element and must be available at the given path inside of the rootfs. If the executable path is not an absolute path then the search $PATH is interpreted to find the executable.
 


wking:
On Mon, Dec 28, 2015 at 07:50:21PM -0800, Lai Jiangshan wrote:
>   * define the default value for the `cwd`

I think the root directory is a good default on Linux when the
container creates/joins a different mount namespace (and that seems to
be the default used by nsenter(1)).  However, when you *don't*
create/join a different mount namespace, I think it makes more sense
to preserve the current working directory and not call chdir(2) at
all.  That will allow your host-mount-namespace config.json process to
access data and such from your bundle directory by looking in its
working directory.

laijs:
> I think the root directory is a good default on Linux when the
the user may not be the root. "home directory" seems to be good one,
but it is a little hard for the runtime resolves the "home directory".
it requires parsing the /etc/passwd like #191. If #191 is accepted,
it is really good to define the default value of the `cwd` as "home directory".

laijs:
currently, both runc&runv use "/" as the default value.

philips:
On Tue, Dec 29, 2015 at 10:03 AM W. Trevor King <notifications@github.com>
wrote:

> On Mon, Dec 28, 2015 at 07:50:21PM -0800, Lai Jiangshan wrote:
> > * define the default value for the `cwd`
>
> I think the root directory is a good default on Linux when the
> container creates/joins a different mount namespace (and that seems to
> be the default used by nsenter(1)). However, when you *don't*
> create/join a different mount namespace,


I really really want to avoid codifying special behavior. We either do one,
the other, or just require someone to be explicit. My preference would be
require the CWD be set, then "/".


> I think it makes more sense
> to preserve the current working directory and not call chdir(2) at
> all. That will allow your host-mount-namespace config.json process to
> access data and such from your bundle directory by looking in its
> working directory.


I can see the use case but this is a little to magical IMHO.

wking:
On Tue, Dec 29, 2015 at 03:36:24AM -0800, Brandon Philips wrote:
> I really really want to avoid codifying special behavior.

I think this falls under my preferred “don't do anything if this isn't
set” [1], which is how I've handled this in ccon with a pivot-root
caveat [2].  I just checked with ccon, and using setns(2) to join a
mount namespace lands you in / (regardless of whether you pivoted root
in that namespace).  So I don't think “we only attempt to apply this
setting if it's set” is codifying special behavior.

[2]: https://github.com/wking/ccon/tree/v0.2.0#current-working-directory

mrunalp:
I think we don't need to add this specifically for Linux.

crosbymichael:
After reviewing in your OCI f2f we think this should be a required field so that runtimes don't default to something different if it is optional or different platforms have different defaults.  By allowing the user to specify this then there is no question what the cwd will be for their container.

PR #287, opened 2015-12-30
style: Collect established conventions in a discoverable location

### LINKED ISSUE #273: Define Spec API best practices
To make contributions easier, as we come up with best practices for the Spec, we should document them.
For example, we decide to use pointers for all fields that need to provide a "do not touch" option.

wking:
So we have something to cite to avoid rehashing established decisions.
Provide some motivation and links to the backing discussion so folks
can re-open these if they have new information that wasn't covered in
the original decision.

Like the glossary (18734986, glossary: Provide a quick overview of
important terms, 2015-08-11, #107), I've used subsection titles for
each entry to get link anchors.

I'm fine dropping any of the decisions from this PR if they turn out
to be contentious, and will happily add any that I missed if folks
bring them to my attention here.

Mailing-list suggestion [here][1], and I'm interpreting the absence of
feedback there, plus a maintainer proposing the same thing in #273 as
enough of a consensus to transition this idea into a PR.  If there is
pushback on the idea of listing policy/style decisions at all, we
should probably close this PR and post the pushback reason in the
mailing-list thread for further discussion.

Fixes #273.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/CbiDx1WpleI

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 Table of Contents
 
 - [Container Principles](principles.md)
+- [Specification Style](style.md)
 - [Filesystem Bundle](bundle.md)
 - Configuration
   - [Container Configuration](config.md)
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ DOC_FILES := \
 	README.md \
 	code-of-conduct.md \
 	principles.md \
+	style.md \
 	ROADMAP.md \
 	implementations.md \
 	bundle.md \
--- a/style.md
+++ b/style.md
@@ -1,0 +1,21 @@
+# Style and conventions
+
+## Traditionally hex settings should use JSON integers, not JSON strings
+
+For example, [`"classID": 1048577`][class-id] instead of `"classID": "0x100001"`.
+The config JSON isn't enough of a UI to be worth jumping through string ↔ integer hoops to support an 0x… form ([source][integer-over-hex]).
+
+## Constant names should keep redundant prefixes
+
+For example, `CAP_KILL` instead of `KILL` in [**`linux.capabilities`**][capabilities]).
+The redundancy reduction from removing the namespacing prefix is not useful enough to be worth trimming the upstream identifier ([source][keep-prefix]).
+
+## Optional settings should have pointer Go types
+
+So we have a consistent way to identify unset values ([source][optional-pointer]).
+
+[capabilities]: config-linux.md#capabilities
+[class-id]: runtime-config-linux.md#network
+[integer-over-hex]: https://github.com/opencontainers/specs/pull/267#discussion_r48360013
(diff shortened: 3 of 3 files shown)


wking:
Do we want to put scoping decisions in here (e.g. network setup beyond namespace creation is out-of-scope for this spec)?  Or are those going to be recorded somewhere else?

wking:
While [tagging][1] [open issues][2] from the mailing list, I re-discovered my [post asking about the “OCI Scope Table” referenced in the adopted charter][3].  Until we get more clarity on what that table contained, landing any scoping language via this PR is probably premature.  For now, I think we should focus on the style issues like those discussed in c41523b6 (this PR's current commit).

[1]: https://github.com/wking/nmbug-oci
[2]: https://wking.github.io/nmbug-oci/#Charter-changes
[3]: https://groups.google.com/a/opencontainers.org/d/msg/dev/26EyeqOibC4/bOFPOCi2CQAJ

policy.md
```diff
@@ -0,0 +1,19 @@
+# Policies and conventions
+
+## Traditionally hex settings should use JSON integers, not JSON strings
+
+The config JSON isn't enough of a UI to be worth jumping through string ↔ integer hoops to support an 0x… form ([source][integer-over-hex]).
```
vbatts: perhaps this could be clearer as to what the expectation is.

README.md
```diff
@@ -6,6 +6,7 @@
 Table of Contents
 
 - [Container Principles](principles.md)
+- [Specification Policies](policy.md)
```
vbatts: it's not so much policy, as a style, right?

policy.md
wking: On Tue, Jan 05, 2016 at 07:56:12AM -0800, Vincent Batts wrote:
> > +The config JSON isn't enough of a UI to be worth jumping through string ↔ integer hoops to support an 0x… form ([source][integer-over-hex]).
> 
> perhaps this could be clearer as to what the expectation is.

Sure.  In c41523b → 408412b, I've called out network.classID as an
example (although it's currently using a string.  If we accept the
integer style that seems popular in [1], I'll file a PR updating
classID).

[1]: https://github.com/opencontainers/specs/pull/267#discussion_r48287582

README.md
wking: On Tue, Jan 05, 2016 at 07:56:41AM -0800, Vincent Batts wrote:
> > @@ -6,6 +6,7 @@
> >  Table of Contents
> >  
> >  - [Container Principles](principles.md)
> > +- [Specification Policies](policy.md)
> 
> it's not so much policy, as a style, right?

Agreed.  Updated in c41523b → 408412b.

philips:
sure, lgtm

mrunalp:
LGTM

vishh:
Thanks for this PR @wking !

PR #290, opened 2016-01-04
Add NoNewPrivileges setting for linux

mrunalp:
This is a security setting that could be used to prevent processes in the
container from gaining additional privileges.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -503,3 +503,14 @@ Its value is either slave, private, or shared.
 ```json
     "rootfsPropagation": "slave",
 ```
+
+## No new privileges
+
+Setting `noNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
+[The kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call.
+
+###### Example
+
+```json
+    "noNewPrivileges": true,
+```
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -39,8 +39,10 @@ type LinuxRuntime struct {
 	SelinuxProcessLabel string `json:"selinuxProcessLabel"`
 	// Seccomp specifies the seccomp security settings for the container.
 	Seccomp Seccomp `json:"seccomp"`
-	// RootfsPropagation is the rootfs mount propagation mode for the container
+	// RootfsPropagation is the rootfs mount propagation mode for the container.
 	RootfsPropagation string `json:"rootfsPropagation,omitempty"`
+	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container.
+	NoNewPrivileges bool `json:"noNewPrivileges,omitempty"`
 }
 
 // Namespace is the configuration for a linux namespace


runtime_config_linux.go
```diff
@@ -39,8 +39,10 @@ type LinuxRuntime struct {
 	SelinuxProcessLabel string `json:"selinuxProcessLabel"`
 	// Seccomp specifies the seccomp security settings for the container.
 	Seccomp Seccomp `json:"seccomp"`
-	// RootfsPropagation is the rootfs mount propagation mode for the container
+	// RootfsPropagation is the rootfs mount propagation mode for the container.
 	RootfsPropagation string `json:"rootfsPropagation,omitempty"`
+	// disableNewPrivileges controls whether additional privileges could be gained by processes in the container.
+	DisableNewPrivileges `json:"disableNewPrivileges,omitempty"`
```
wking: Missing `*bool`?  At some point I think we should land general specs for the:

> [Null and missing entries are valid JSON and mean “don't touch this”.][1]

semantics .  If we're not comfortable landing that generally, I think this PR needs docs about what runtimes should do if `DisableNewPrivileges` is null or unset in the JSON.

[1]: https://github.com/opencontainers/specs/pull/233#issuecomment-152606572

wking: My suggested generic wording for null/missing is [here][1].

wking:
+1 to including this in the runtime, because setting it in the
container process before invoking the user-configured code is both
easier and safer (for runtime callers) than expecting bundle authors
to handle this themselves.

mrunalp:
@crosbymichael @LK4D4  @vbatts @philips PTAL

crosbymichael:
Should this just be a default in a runtime?

runtime-config-linux.md
```diff
@@ -503,3 +503,14 @@ Its value is either slave, private, or shared.
 ```json
     "rootfsPropagation": "slave",
 ```
+
+## Disable new privileges
+
+Setting `disableNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
+[The kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call.
+
```
vbatts: give the line item information on it here (that it's a bool and whether required)

vbatts:
almost seems like it could be a default, but still likely that folks would want to opt-in to

mrunalp:
@crosbymichael @vbatts This tightens up the security but has some downsides if enabled by default. For e.g. ping won't work if your container doesn't already have the net_bind cap. So, I think opt-in might be better for this feature.

rhatdan:
Somewhat equivalent to

docker run --cap-drop=all -u UID -g GID.

I am trying to find if there is any other differences.

So as @mrunalp says if you just ran 
docker run -u UID fedora sh

Then tools like ping, chsh, traceroute etc could theoretically work.

mheon:
@rhatdan It does have an affect on how layered Seccomp filters are treated, so that would be one thing that your example doesn't cover.

It would be nice to set this to be enabled by default, or at least note that passing a Seccomp configuration implies that it is enabled, given that it has a beneficial relationship with applications in a container that also use Seccomp.

wking:
On Tue, Jan 05, 2016 at 11:12:38AM -0800, Daniel J Walsh wrote:
> Somewhat equivalent to
> 
> docker run --cap-drop=all -u UID -g GID.
> 
> I am trying to find if there is any other differences.

I don't think it's equivalent.  The point of no_new_privs (as outlined
in the kernel docs [1]) is to close the execve loopholes that are
generally allowed for priviledge escallation.  E.g. after dropping
caps, you can still execute a setuid binary and increase your
privileges.  If you want to experiment some, I implemented this option
in ccon last night [2].

Re, opt-in or opt-out, I think “least surprise” points to opt-in.
Otherwise I expect a lot of “why don't my setuid binaries work” bug
reports.  And I also prefer opt-in as a general principle [3].

[1]: https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt
[2]: https://github.com/wking/ccon#disable-new-privileges
[3]: https://github.com/opencontainers/specs/pull/285#issuecomment-167823651

runtime-config-linux.md
```diff
@@ -503,3 +503,17 @@ Its value is either slave, private, or shared.
 ```json
     "rootfsPropagation": "slave",
 ```
+
+## Disable new privileges
+
+Setting `disableNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
```
philips: is there a reason to not call it noNewPrivileges so it matches the kernel version?

philips:
Why not have a generic prctl field?

rhatdan:
Any way we can get it in, I am all for. 

I am torn on the OptIn versus OptOut,  OptIn will probably be required.  The problem with OptIn, is as I have often said, almost no one ever increases the security.  Most executers of containers using -u UID probably expect that the process inside of the container can not become a different UID or get any privileges.  But they would need to understand, to get their expected behaviour they would need to do

docker run --nonewpriv --cap-drop=all -u UID ...

Note:  --cap-=drop=all might not be needed above.  

Have --prctl=NO_NEW_PRIVS=1  might allow us to easily add additional prctl in the furture.

philips:
On Wed, Jan 6, 2016 at 7:37 PM Daniel J Walsh <notifications@github.com>
wrote:

> Have --prctl=NO_NEW_PRIVS=1 might allow us to easily add additional prctl
> in the furture
>
The more I think about it the more I really want to avoid adding new terms
to this spec for Linux. Calling a prctl with the name NO_NEW_PRIVS
disableNewPrivelges reduces discoverability and opens the door for
"expanding behavior".

mrunalp:
Downside of using prctl is that not all prctl options make sense for run times. For e.g. There is one for making the parent a reaper but it doesn't make sense for all runtime to support that.

mrunalp:
We could however start with a whitelist of what options are supported.

rhatdan:
We have the same problem with --sysctl  in that some sysctls are namespaced and some are not.  Our pull requests are blocked because a user could shoot himself in the head.  We have recently proposed a whitelist approach for runc to allow us to get --sysctl in.  But bottom line, is I believe we need sysctl in and I believe we need  NO_NEW_PRIV support, however we can get it in.

One potential option from a docker point of view would be 

docker run --security-opt privs:NO_NEW_PRIV:1

Which would allow us to add a new security feature without messy-ing up the CLI, too much.

dqminh:
@mrunalp @philips i think prctls + a required list of options works here rather than adding a new fields. Runtime can choose to support additional prctl options.

rhatdan:
Any other comments on this?  prctl probably makes sense for runc, then we can argue about docker CLI.

wking:
On Wed, Jan 13, 2016 at 06:55:15AM -0800, Daniel J Walsh wrote:
> Any other comments on this?  prctl probably makes sense for runc…

Yeah, I'm fine with a generic prctl setting.

crosbymichael:
Negative on the generic prctl setting.  That is super horrible api and is the first step going down a dark path of 

```json

"syscall": [
    [0,2,3,4,1,0]
]
```

wking:
On Wed, Jan 13, 2016 at 10:22:10AM -0800, Michael Crosby wrote:
> Negative on the generic prctl setting…

The problem with adding prctl wrappers (like disableNewPrivileges) one
at a time is that we end up in the same situation that we have now for
cgroups: a thin wrapper around the kernel that takes work to maintain
and adds no value [1].  The difference vs. cgroups is that most (all?)
prctl invocations act on the calling thread/process, so it doesn't
decouple as nicely as cgroups (where you can use hooks outside of the
namespaced-process setup dance).  Having a generic prctl setting is
the next closest thing we can do to “keep that out of the spec”, and
it allows higher-level tooling to add thin or thick wrappers as they
see fit [2].

And yeah, you *could* do all of that with a generic syscall wrapper,
but I think everyone agrees that that tips the balance too far (really
easy to write runtimes, but really hard to use them).  While for a
generic prctl setting doesn't seem to be much of a usability cost (and
may be more usable [3]) over per-prctl wrapper settings.  If somebody
has a brilliant idea for a general abstraction that covers
no_new_privs without being a thin wrapper, then *that* would be a
usability win.  But that's not what we have here.

[2]: https://github.com/opencontainers/specs/pull/290#issuecomment-171319391
[3]: https://github.com/opencontainers/specs/pull/290#issuecomment-169373879

mrunalp:
I agree with @crosbymichael on this. NO_NEW_PRIVS is specific and closes a lot of holes around priviliege escalation and I think it makes sense to have switch for it. Do we have an example for any other prctl settings that make sense to be exposed as a counter argument?

rhatdan:
Some of the other prctl are interesting but they seem to be handled via capability handling and seccomp.

I still think we could handle NO_NEW_PRIVS via SecurityOptions.

wking:
On Wed, Jan 13, 2016 at 11:35:54AM -0800, Mrunal Patel wrote:
> I agree with @crosbymichael on this. NO_NEW_PRIVS is specific and
> closes a lot of holes around priviliege escalation and I think it
> makes sense to have switch for it. Do we have an example for any
> other prctl settings that make sense to be exposed as a counter
> argument?

I'm currently hard-coding PR_SET_PDEATHSIG to SIGKILL in ccon, but
that could potentially be user configurable.  And grepping for PR_SET_
in util-linux turns up setpriv(1) with PR_SET_NO_NEW_PRIVS,
PR_SET_KEEPCAPS, and PR_SET_SECUREBITS.

runtime_config_linux.go
```diff
@@ -39,8 +39,10 @@ type LinuxRuntime struct {
 	SelinuxProcessLabel string `json:"selinuxProcessLabel"`
 	// Seccomp specifies the seccomp security settings for the container.
 	Seccomp Seccomp `json:"seccomp"`
-	// RootfsPropagation is the rootfs mount propagation mode for the container
+	// RootfsPropagation is the rootfs mount propagation mode for the container.
 	RootfsPropagation string `json:"rootfsPropagation,omitempty"`
+	// disableNewPrivileges controls whether additional privileges could be gained by processes in the container.
```
estesp: not sure if `golint` will ever be run, but I think it will want the comment to match the name (capital "**D**" for *DisableNewPrivileges*)

runtime_config_linux.go
```diff
@@ -39,8 +39,10 @@ type LinuxRuntime struct {
...
+	DisableNewPrivileges *bool `json:"disableNewPrivileges,omitempty"`
```
vishh: If the system default `false`? If yes, then can we drop the `*` here?

mrunalp: Yes. Removed the pointer.

mrunalp:
Renamed to NoNewPrivileges. PTAL.

vbatts:
LGTM

runtime-config-linux.md
```diff
@@ -503,3 +503,14 @@ Its value is either slave, private, or shared.
 ```json
     "rootfsPropagation": "slave",
 ```
+
+## No new privileges
+
+Setting `noNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
```
vishh: Should the default be `true` ?

vbatts: nope

vishh: Why not?

wking: On Wed, Jan 20, 2016 at 02:15:18PM -0800, Vish Kannan wrote:

> > +Setting `noNewPrivileges` to true prevents the processes in the
> > container from gaining additional privileges.
> 
> Why not?

Search for ‘opt-’ in the PR discussion to get previous opt-in
vs. opt-out thoughts ;).

vishh: Acknowledged.

vishh:
LGTM

philips:
Thank you, LGTM

PR #292, opened 2016-01-05
runtime-config-linux: Add info to clarify oom_score_adj (carrying #236)

vbatts:
This is carrying #236, but deferring the description to the linux docs.

### DIFF
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -180,7 +180,11 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 
 #### Set oom_score_adj
 
-More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+oom_score_adj works with disableOOMKiller. If out-of-memory killer enabled("disableOOMKiller": false), oom_score_adj can be set to adjust the oom-killer score, the value of oom_score_adj is added to the badness score before it is used to determine which task to kill. If out-of-memory killer disabled("disableOOMKiller": true), the value of oom_score_adj is useless and will be omitted. For more information, see [the proc filesystem man page](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+
+* **`oomScoreAdj`** *(int, optional)* - adjust the oom-killer score
+
+###### Example
 
 ```json
     "oomScoreAdj": 0
--- a/runtime-config-linux.md
+++ b/runtime-config-linux.md
@@ -180,7 +180,10 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 
 #### Set oom_score_adj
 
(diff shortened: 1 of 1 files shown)


runtime-config-linux.md
```diff
@@ -180,7 +180,13 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 
 #### Set oom_score_adj
 
-More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+oom_score_adj works with disableOOMKiller.
+If out-of-memory killer disabled("disableOOMKiller": true), the value of oom_score_adj is useless and will be omitted.
```
vishh: https://github.com/opencontainers/specs/pull/236#issuecomment-152684866

vbatts: fixed

runtime-config-linux.md
```diff
@@ -180,7 +180,14 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
 
 #### Set oom_score_adj
 
-More information on `oom_score_adj` available [here](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+oom_score_adj sets heuristic regarding how the process is evaluated by the kernel during memory pressure.
```
wking: “oom_score_adj” should probaby be “**`oom_score_adj`**” or “`oom_score_adj`”.  And maybe reword to:

> **`oom_score_adj`** adjusts the kernel's process evaluation during memory pressure.

runtime-config-linux.md
```diff
@@ -180,7 +180,14 @@ For more information, see [the memory cgroup man page](https://www.kernel.org/do
...
+For more information, see [the proc filesystem documentation section 3.1](https://www.kernel.org/doc/Documentation/filesystems/proc.txt).
+This is a kernel/system level setting, where as disableOOMKiller is scoped for a memory cgroup.
```
wking: For consistency, I'd backtick/** disableOOMKiller too.

runtime-config-linux.md
vbatts: but it is per process

vbatts: and i agree, re backticks

wking:
I left a few inline copy-edit suggestions for 1bfadf67, but the thrust
looks good to me.  While we're cleaning up the docs for this setting,
we may want to consider clarifying the runtime action if the setting
is unset or null [1].

runtime-config-linux.md
wking: On Tue, Jan 05, 2016 at 12:47:42PM -0800, Vincent Batts wrote:
> but it is per process

Right.  I just think “sets heuristic regarding how the process is
evaluated by the kernel” could be condensed without losing
information.  How about something closer to the kernel's §3.1 lead in
[1]:

  `oom_score_adj` adjusts the OOM-killer badness heuristic for this
  process.

?

[1]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt

hqhq:
LGTM

philips:
lgtm

vishh:
LGTM

PR #293, opened 2016-01-05
Create image.md

duglin:
This is far from complete but I wanted to get it out there in time for people to look it over prior to the f2f next week.  Its kind of a merging of docker's image format and appc's image format.

### DIFF
--- a/image.md
+++ b/image.md
@@ -1,0 +1,136 @@
+# Image
+
+This section defines a format for encoding a container's *filesytem bundle* as a portable *image*.
+An image is a set of files organized in a certain way, and containing all of the necessary data and metadata, for any compliant runtime to verify the validity of the image's contents and to construct a filesystem bundle represented by the image.
+Images are meant to be portable such that they can be used to reliably move filesystem bundles between instances of OCI compliant implementations.
+
+## Image Contents
+
+A compliant image contains the following pieces of information:
+
+1. An OCI [`config.json`](config.md) file.
+This file is REQUIRED to be present in the image, unless this image is a dependent image, in which case it is OPTIONAL.
+The `config.json` file contains the host independent configuration information necessary to execute the container.
+See [`config.json`](config.md) for more information.
+
+1. The root filesystem of the filesystem bundle.
+This directory MUST be present in the image, however it MAY be empty (not contain any files or folders).
+As specified in the `config.json` file, this directory contains the files that make up the root filesystem of the container.
+This directory MAY contain only a portion of the complete set of files that make up the filesystem bundle, in which case the remaining set of files MUST be found in one of the dependent images.
+
+1. References to dependent images.
+If present, this OPTIONAL information MUST be in a file called `required.json`.
+This file MUST be a JSON encoded array of references to images as specified in the [Dependent Images](#dependent-images) section.
+
+1. Dependent images.
+An image MAY choose to include dependent images as directories within the image.
+If present, each dependent image MUST be placed in a directory whose name is the same as its ID.
+Within each directory MUST be a complete OCI compliant image definition.
+Note that within each dependent image's directory there MAY be a whiteout.json file.
+
+1. Whiteout list.
+An image MAY choose to include a list of files that are to be considered to be deleted from the root filesystem even though they are included within the image's root filesystem directory.
+It is an implementation detail to decide how this whiteout list is processed but an implementation MUST ensure that any container that is created from this image will behave the same as if those files were not in the root filesystem at all.
+The whiteout list MUST be in a file called `whiteout.json`.
+See the [Whiteout List](#whiteout-list) section for more details.
+
+1. Extension files.
+An image MAY choose to include additional files and directories in the image.
+
+1. The ID of the image.
+This REQUIRED information MUST be in a file called `ID`.
+See the [Image ID](#image-id) section for more details.
+
+## Serialization of an Image
+
+An image MUST be a `tar` of the pieces of data that make up an image as specified in the [Image Contents](#image-contents) section.
+The order in which the data is serialized in the `tar` MUST be the same as the order specified in that section.
+The serialization of the "dependent images" MUST be in the same order in which those dependent images are listed in the `requires.json` file.
+The serialization of the  "extension files" MUST be in case-sensitive alphabetical order.
+
+Each file or directory specified in the [Image Contents](#image-contents) section MUST appear at the root of the `tar` without a leading `/` in their names.
+
+For example, a listing of the `tar` of an image might look like this:
+```
+config.json
+rootfs/
+    bin/myapp
+requires.json
+sha512-1d62d181e9c1322d56ccd3a29d05018399147a16188dbd861c0279ad0dd7e14c/
+    config.json
+    rootfs/
+        bin/runit
+    whiteout.json
+    ID
+sha512-291e2a171ef9bef8a838c59406d9b0aeb6f2f0ebe5173415205733d3d18b8e03/
+    config.json
+    rootfs/
+        bin/monitor
+    whiteout.json
+    ID
+whiteout.json
+ID
+```
+
+## Dependent Images
+
+If an image has dependent images then it MUST include a `requires.json` file that references those images.
+The `requires.json` file MUST be a JSON encoded array matching this format:
+```
+{
+  "requires": [
+    {
+      "ID": "..."
+    },
+    {
+      "ID": "..."
+    }, ...
+  ]
+}
+```
+
+The order of the images, `ID`s, in the array MUST be in the order in which the images are to be "layed down" into the filesystem bundle that is generated from this image.
+In other words, the first image in the array is written to the filesystem first, and subsequent images are overlayed on top - potentially overlaying an ealier image's files.
+
+## Image ID
+
+An image's `ID` MUST be a cryptographic hash of image's contents in the order specified in the [Image Contents](#image-contents) section.
+
+An image's `ID` MUST be a string of the form `<algorithm>-<hash value>`.
+The `algorithm` part of the string MUST be in lower-case.
+For example, `sha512` not `SHA512`.
+
+It is RECOMMENDED that implementations use the sha512 algorithm.
+
+## Whiteout List
+
+An image's `whiteout.json` file MUST be a JSON encoded file matching this format:
+```
+{
+  "files": [
+    "filename",
+    ...
+  ]
+}
+```
+where each `filename` MUST be the absolute path to the file.
+
+## Expanding an Image to Create a Filesystem Bundle
+
+When an image is expanded to create a filesystem bundle the image's `config.json` file MUST be written to the root of the filesystem bundle's directory.
+The expansion process MUST create a directory for the container's root filesystem with a name as specified in the `config.json` file - ie. the `root/path` property.
+The contents of this directory MUST appear to be materizlied in such a way as to contain all of the files from the image, without the whiteout (deleted) files, and its dependent images.
+The exact mechanism, process and order, by which those files are materialized on disk is an implementation detail, however in the end it MUST appear as though all of the files from the images were written to disk in the order they were specified in the `requires.json` file.
+This means that each dependent image's `whiteout.json` file MUST appear to be processed during the materialization of that image so that if a subsequent image creates a file by the same name as one mentioned in the `whiteout.json` file then it will not be deleted.
+
+The `config.json` file of any dependent images, if present, are ignored by the expansion process.
+
+This specification does not mandate any requirements on the processing of the extension files.
+In other words, implementations are not required to materialize them on disk in the filesystem bundle.
+
+This specification does not mandate any ordering to when each piece of information from the image is materialized.
+
+## Misc Notes
+
+There is no requirement that importing an image into a compliant OCI implementation and then exporting it will result in the same image since each implementation might choose to create the root filesystem differently.
+Thus, exporting that root filesystem might look different from the original image.
--- a/image.md
+++ b/image.md
@@ -1,150 +1,309 @@
 # Image
 
 This section defines a format for encoding a container's *filesytem bundle* as a portable *image*.
-An image is a set of files organized in a certain way, and containing all of the necessary data and metadata, for any compliant runtime to verify the validity of the image's contents and to construct a filesystem bundle represented by the image.
+An image is a set of files organized in a certain way, containing all of the necessary data and metadata, for any compliant runtime to verify the validity of the image's contents and to construct a filesystem bundle represented by the image.
 Images are meant to be portable such that they can be used to reliably move filesystem bundles between instances of OCI compliant implementations.
+
+In its simplest form, an image is a tar file of the filesystem bundle (root file system directory plus a configuration file) along with some additional metadata.
+While the runtime only sees a single directory as the root filesystem, it might actually be stored as a layering of filesytems under the covers.
+When representing this layering within an image, each "layer" will be independently referenced.
 
 ## Image Contents
 
 A compliant image contains the following pieces of information:
 
-1. An OCI [`config.json`](config.md) file.
+1. A `manifest.json` file describing the contents of the image.
 This file is REQUIRED to be present in the image.
-The `config.json` file contains the host independent configuration information necessary to execute the container.
-See [`config.json`](config.md) for more information.
+This file extends the [`config.json`](config.md) file with some additional properties:
+  * `name`: This REQUIRED property is a URL-like string to aide in discovery of the image.
+  For example, `example.com/myApp`.
+  * `whiteout`: This OPTIONAL property is a JSON array of absolute file paths that are to be considered to be deleted from dependent layers, even though they are included within those layers root filesystems.
+  It is an implementation detail to decide how this whiteout list is processed, but an implementation MUST ensure that any container that is created from this image will behave as if those files were not in the root filesystem at all.
+  * `layers`: This OPTIONAL property contains references to the layers that make up the root filesystem of the image.
+  This property MUST be a JSON encoded array where each entry consists of a REQUIRED `name` and OPTIONAL `ID` properties.
+  The array MUST be in the order in which the images are to be "layed down" into the filesystem bundle that is generated from this image.
+  In other words, the first image in the array is written to the filesystem first, and subsequent images are overlayed on top - potentially overlaying an earlier image's files.
+  If a referenced layer is embedded within this image then an `ID` property MUST be included.
+  If a referenced layer is not embedded within this image then an `ID` property is OPTIONAL.
+  * `root/path`: ThiS REQUIRED property, defined within [`config.json`](config.md), is the name of the directory into which all layers MUST appear.
+  While the `manifest.json` file extends the `config.json` file, a compliant OCI image is NOT REQUIRED to have all of the required fields from `config.json` within its `manifest.json` file if that image is only used as a layer (dependent image) within other images.
+  This is due to the fact that a layer's `manifest.json` properties are not used to produce a container's `config.json` file and therefore not all of them are used.
+  However, if an image is used to create a container then all of the "required" fields of the `config.json` MUST be present in the `manifest.json`.
 
-1. References to dependent images.
-If present, this OPTIONAL information MUST be in a file called `required.json`.
-This file MUST be a JSON encoded array of references to images as specified in the [Dependent Images](#dependent-images) section.
+1. Layers.
+Each embedded layer MUST have a corresponding image (tar) file, at the root of the image.
+Each layer's tar file MUST be named with its ID followed by the appropriate file-extension based on how the tar file is written (e.g. `.tar` for uncompressed).
 
-1. Dependent images.
-An image MAY include dependent images as directories within the image.
-Each dependent image MUST be placed in a directory whose name is the same as its ID (see the [IDs](#ids) section for more details about IDs).
-The contents of each directory MUST adhere to the following:
-  * there MAY be a `whiteout.json` file (see [Whiteout List](#whiteout-list)).
-  * there MUST be a directory called `rootfs` which contains the root filesystem of that image.
-  *  there MAY be additional "extension files".
+1. `rootfs` directory.
+An image MAY optionally contain a directory that's the same as the `root/path` property of the `manifest.json` file.
+And if so, then it MUST be treated as the top-most layer of the root filesystem.
 
-  The ID MUST be generated over the above contents in the order specified.
-
-1. Whiteout list.
-An image MAY choose to include a list of files that are to be considered to be deleted from dependent images even though they are included within the dependent image's root filesystem directory.
-It is an implementation detail to decide how this whiteout list is processed but an implementation MUST ensure that any container that is created from this image will behave the same as if those files were not in the root filesystem at all.
-The whiteout list MUST be in a file called `whiteout.json`.
-See the [Whiteout List](#whiteout-list) section for more details.
-
-1. The root filesystem of the filesystem bundle.
-This directory MUST be present in the image, however it MAY be empty (not contain any files or folders).
-As specified in the `config.json` file, this directory contains the files that make up the root filesystem of the container.
-This directory MAY contain only a portion of the complete set of files that make up the filesystem bundle, in which case the remaining set of files MUST be found in one of the dependent images.
-
-1. ID of the root filesystem.
-This data MUST be a cryptographic hash of the whiteout list file, if present, and the root filesystem directory (including the directory itself), in that order.
-If there are dependent images then this MUST NOT include the dependent image's data.
-This REQUIRED information MUST be in a file called `rootfs.ID`
-See the [IDs](#ids) section for more details.
-
-1. Extension files.
-An image MAY choose to include additional files and directories in the image.
-
-1. The ID of the image.
-This data MUST be a cryptographic hash of all the data listed above, in the order specified.
-For the purposes of calculating the ID, the order of processing the extension files MUST be in case-sensitive alphabetical order.
-This REQUIRED information MUST be in a file called `ID`.
-See the [IDs](#ids) section for more details.
+1. There MAY be additional files within the image.
+This specification places no requirements on these files.
+A compliant OCI implementation is NOT REQUIRED to materialize these files into a container's filesystem bundle directory.
 
 ## Serialization of an Image
 
-An image MUST be a `tar` of the pieces of data that make up an image as specified in the [Image Contents](#image-contents) section.
-The order in which the data is serialized in the `tar` MUST be the same as the order specified in that section.
-The serialization of the "dependent images" directories, if any, MUST be in the same order in which those dependent images are listed in the `requires.json` file.
-The serialization of the  "extension files" MUST be in case-sensitive alphabetical order.
-Note that "extension files", irrespective of their names, MUST appear after the `rootfs.ID` file for the image.
-
-Each file or directory specified in the [Image Contents](#image-contents) section MUST appear at the root of the `tar` without a leading `/` in their names.
-
-For example, a listing of the `tar` of an image might look like this:
-```
-config.json
-requires.json
-sha512-1d62d181e9c1322d56ccd3a29d05018399147a16188dbd861c0279ad0dd7e14c/
-    rootfs/
-        bin/runit
-sha512-291e2a171ef9bef8a838c59406d9b0aeb6f2f0ebe5173415205733d3d18b8e03/
-    whiteout.json
-    rootfs/
-        bin/monitor
-sha512-98364ca873540185d83645e93745a94847f2f0ebe51733987154840afebc9921/
-whiteout.json
-rootfs/
-    bin/myapp
-rootfs.ID
-extraFile
-ID
-```
-
-## Dependent Images
-
-If an image has dependent images then it MUST include a `requires.json` file that references those images.
-The `requires.json` file MUST be a JSON encoded array matching this format:
-```
-{
-  "requires": [
-    {
-      "ID": "..."
-    },
-    {
-      "ID": "..."
-    }, ...
-  ]
-}
-```
-
-The order of the images, `ID`s, in the array MUST be in the order in which the images are to be "layed down" into the filesystem bundle that is generated from this image.
-In other words, the first image in the array is written to the filesystem first, and subsequent images are overlayed on top - potentially overlaying an ealier image's files.
-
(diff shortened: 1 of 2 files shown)


image.md
```diff
@@ -0,0 +1,140 @@
+# Image
+
+This section defines a format for encoding a container's *filesytem bundle* as a portable *image*.
+An image is a set of files organized in a certain way, and containing all of the necessary data and metadata, for any compliant runtime to verify the validity of the image's contents and to construct a filesystem bundle represented by the image.
+Images are meant to be portable such that they can be used to reliably move filesystem bundles between instances of OCI compliant implementations.
+
+## Notational Conventions
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this and other documents of the specification are to be interpreted as described in [RFC2119](http://tools.ietf.org/html/rfc2119).
```
wking: I think the phrasing from #260 already covers this, so we can remove this section here.

image.md
```diff
@@ -0,0 +1,140 @@
...
+
+A compliant image contains the following pieces of information:
+
+1. An OCI [`config.json`](config.md) file.
+This file is REQUIRED to be present in the image.
+The `config.json` file contains the host independent configuration information necessary to execute the container.
+See [`config.json`](config.md) for more information.
+
+1. The root filesystem of the filesystem bundle.
+This directory MUST be present in the image, however it MAY be empty (not contain any files or folders).
```
wking: What's the benefit to requiring a directory that may be empty?  E.g. why is it better to have an empty directory than no directory at all?  This is maybe part of the existing [list thread][1].  To stay DRY and avoid that fluxy issue, I'd recommend we have a single entry here like:

    1. The [artifacts that are part of the filesystem bundle](bundle.md).

followed by entries that are unique to the image format.

image.md
```diff
@@ -0,0 +1,140 @@
...
+This directory MAY contain only a portion of the complete set of files that make up the filesystem bundle, in which case the remaining set of files MUST be found in one of the dependent images.
+
+1. References to dependent images.
+If present, this OPTIONAL information MUST be in a file called `required.json`.
+This file MUST be a json encoded array of references to images as specified in the [Dependent Images](#dependent-images) section.
+
+1. Dependent images.
+An image MAY choose to include dependent images as directories within the image.
+If present, each dependent image MUST be placed in a directory whose name is the same as its ID.
+Within each directory MUST be a complete OCI compliant image definition.
```
wking: “`OCI compliant image definition`” → “`[bundle](bundle.md)`”?

image.md
```diff
@@ -0,0 +1,140 @@
...
+Note that within each dependent image's directory there MAY be a whiteout.json file.
+
+1. Whiteout list.
+An image MAY choose to include a list of files that are to be considered to be deleted from the root filesystem even though they are included within the image's root filesystem directory.
+It is an implementation detail to decide how this whiteout list is processed but an implementation MUST ensure that any container that is created from this image will behave the same as if those files were not in the root filesystem at all.
+The whiteout list MUST be in a file called `whiteout.json`.
+See the [Whiteout List](#whiteout-list) section for more details.
+
+1. Extension files.
+An image MAY choose to include additional files in the image.
```
wking: +1, although I'd say “files and directories” to avoid any uncertainty.  And I'd update bundle.md to say that this sort of thing was permissable in the filesystem bundle (see [this thread][1]), after which my suggested [bundle.md reference][2] would cover this point.

[2]: https://github.com/opencontainers/specs/pull/293/files#r48905528

image.md
wking: Ah, the contents are *images* (this file), not bundles.  So I'm removing my earlier comment.

image.md
```diff
@@ -0,0 +1,140 @@
...
+Each file or directory specified in the [Image Contents](#image-contents) section MUST appear at the root of the `tar` without a leading `/` in their names.
+
+For example, a listing of the `tar` of an image might look like this:
+```
+config.json
+rootfs/
+    bin/myapp
+requires.json
+sha512-1d62d181e9c1322d56ccd3a29d05018399147a16188dbd861c0279ad0dd7e14c/
+    config.json
```
wking: This dependent image doesn't have a rootfs, and is likely an example of [why I don't think we want to require one][1].

[1]: https://github.com/opencontainers/specs/pull/293/files#r48905528

image.md
```diff
@@ -0,0 +1,140 @@
...
+	rootfs/
+	    bin/runit
```
wking: Duplicate of the rootfs containing `myapp`?  Or maybe you mean to have this in `
+sha512-1d62d181e9c1322d56ccd3a29d05018399147a16188dbd861c0279ad0dd7e14c`, in which case I think it needs more indentation.

image.md
```diff
@@ -0,0 +1,140 @@
...
+      "ID": "..."
+    },
+	{
+	  "ID": "..."
+	}, ...
+  ]
+}
+```
+
+The order of the images, `ID`s, in the array MUST be in the order in which the images are to be "layed down" into the filsystem bundle that is generated from this image.
```
wking: nit: “filsystem” → “filesystem”.

image.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Expanding an Image to Create a Filesystem Bundle
+
+When an image is expanded to create a filesystem bundle the image's `config.json` file MUST be written to the root of the filesystem bundle's directory.
+The expansion process MUST create a directory for the container's root filesystem with a name as specified in the `config.json` file - ie. the `root/path` property.
+The contents of this directory MUST appear to be materizlied in such a way as to contain all of the files from the image, without the whiteout (deleted) files, and its dependent images.
+The exact mechanism, process and order, by which those files are materialized on disk is an implementation detail, however in the end it MUST appear as though all of the files from the images were written to disk in the order they were specified in the `requires.json` file.
+This means that each dependent image's `whiteout.json` file MUST appear to be processed during the materialization of that image so that if a subsequent image creates a file by the same name as one mentioned in the `whiteout.json` file then it will not be deleted.
+
+The `config.json` file of any dependent images are ignored by the expansion process.
```
wking: Then why require them?  If the image format is *just* a mechanism for constructing a rootfs from layers, I don't see how it improves on our existing mounts.

image.md
```diff
@@ -0,0 +1,140 @@
...
+See [`config.json`](config.md) for more information.
+
+1. The root filesystem of the filesystem bundle.
+This directory MUST be present in the image, however it MAY be empty (not contain any files or folders).
+As specified in the `config.json` file, this directory contains the files that make up the root filesystem of the container.
+This directory MAY contain only a portion of the complete set of files that make up the filesystem bundle, in which case the remaining set of files MUST be found in one of the dependent images.
+
+1. References to dependent images.
+If present, this OPTIONAL information MUST be in a file called `required.json`.
+This file MUST be a json encoded array of references to images as specified in the [Dependent Images](#dependent-images) section.
```
mrunalp: s/json/JSON ?

wking:
Feedback on tactics:

I'm not sure why we have all the explicit ordering requirements
(e.g. “The order in which the data is serialized in the `tar` MUST be
the same as the order specified in that section” and then back off
with “There is no requirement that importing an image into a compliant
OCI implementation and then exporting it will result in the same image
since each implementation might choose to create the root filesystem
differently”.  Is the ordering just intended for more efficient
unpacking to the filesystem, and not for stable hashing?  I'm in favor
of efficiency gains, but think that requiring them in a spec like this
may be a bit premature.  If we want to encourage that ordering right
off the bat, I'd suggest we start off with SHOULD instead of MUST for
any tar-ordering requirements.

Without a reproducible way to go from a filesystem to a hash, I don't
see a need to name vendored dependencies by hash
(e.g. sha512-1d62d181e9c1322d56ccd3a29d05018399147a16188dbd861c0279ad0dd7e14c
in your example).  On the other hand, hashing an image tarball so it
can be content-addressed from another image that does *not* vendor it
makes sense.  Is there a reason to require hashed names for vendored
images?

Feedback on strategy:

This sounds like a fancy way to construct and distribute a root
filesystem, and I don't see a need to encode that in the runtime spec.
Implementing and testing tools that create and unpack images like this
seems to be a completely orthogonal problem to launching containers
from filesystem bundles.  But Docker proves there is a large community
of people who prefer to distribute their rootfs' using this
layered/blacklisted tarball approach.  I'd recommend defining it in a
separate repository so folks writing and using implementations of this
layering/distribution spec can collaborate on it without distracting
the set of folks writing and using implemenations of the
container-launching runtime.  There may be considerable overlap
between those sets of people, but they aren't the same set (e.g. I'm
personally interested in the container-launching runtime, but not in
layered-filesystem distribution) [1].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/1T0z1IJWxw8/XKdv69e1FgAJ
     Subject: Re: Hashing and verifying a bundle
     Date: Wed, 02 Sep 2015 11:48:54 -0700
     Message-ID: <20150902184854.GU32473@odin.tremily.us>

mrunalp:
The suggested changes imply that the runtime would have to construct a rootfs that could be used in the container (what we pivot the root to) either by using a union filesystem or just copying the files to a new root. Should we clarify that further?

wking:
On Tue, Jan 05, 2016 at 03:22:55PM -0800, Mrunal Patel wrote:
> The suggested changes imply that the runtime would have to construct
> a rootfs that could be used in the container…

In the current commit (e021a7f08), unpacking and image to a filesystem
bundle is a separate step [1].  So why does “the runtime” (which
launches containers from bundles [2]) need to be involved at all.  I
see no benefit, and an increased runtime-author cost, if we require a
single entity to be both the container-launching implementation and
the image-unpacking implementation (which is why I want this image
spec in a different repository [3]).

[1]: https://github.com/opencontainers/specs/commit/e021a7f0800ca34fad01411fde39c9609b1f6a8c#diff-b24a04a898225bc33eec822085b5e3ebR122
[2]: https://github.com/opencontainers/specs/blob/dd17705087b74338dbd2e60f133635cb28071ccc/glossary.md#runtime
[3]: https://github.com/opencontainers/specs/pull/293#issuecomment-169166550

philips:
I think large design things like this should be discussed on the mailing lists and with Google Docs. GitHub is a very difficult forum for this.

Also, here is a proposal I worked up that tries to break an image format into a number of different layers with hashes at each of those layers: https://docs.google.com/document/d/1JPQqH9m9fwZIVsvpB5FSyPJjH0RZLnFkI2J2gohfPu0/edit

My concern here is that the image format is too monolithic and includes the host specific config.json file by default.

mrunalp:
I agree that we should discuss this on the mailing list first.

image.md
```diff
@@ -0,0 +1,140 @@
...
+
+QUESTION: what about deleted files? Do we want to mandate a whiteout scheme?
+
+## Image ID
+
+An image's `ID` MUST be a sha512 hash of image's contents in the order specified in the [Image Contents](#image-contents) section.
+
+An image's `ID` MUST be a string of the form `sha512-<hash>`.
+
+QUESTION: Do we really want to mandate sha512?
```
vbatts: It is a solid default. For a majority of x86{,_64} chips, it is faster than sha256. Though for architectures like ARM, this could be a hit. 
Here is my laptop. Both are speedy, but sha512 is faster:
```bash
$ dd if=/dev/urandom of=rand.img bs=1M count=64
64+0 records in
64+0 records out
67108864 bytes (67 MB) copied, 9.47004 s, 7.1 MB/s
$ time sha256sum rand.img 
2d4e6875974352af656630e8ed76ee036847888d69c507f73809d672c58f439d  rand.img

real	0m1.068s
user	0m1.026s
sys	0m0.041s

$ time sha512sum rand.img 
ba914581b051a017d9f7fc3169bd6036bfa24666eb692d891107ce237d60a091919f0cbce0408221c452bd0217ee1388a72ca33a2b65e6bb791714217f85a4d3  rand.img

real	0m0.692s
user	0m0.654s
sys	0m0.036s
```

Here is a raspberry pi. The sha512 is an order of magnitude slower:
```bash
$ dd if=/dev/urandom of=rand.img bs=1M count=64
64+0 records in
64+0 records out
67108864 bytes (67 MB) copied, 54.8794 s, 1.2 MB/s
$ time sha256sum rand.img
d286218629fccb13c9c1568b6ddb9f2fc5452299a4ee307e3cff353836d8901a  rand.img

real    0m5.552s
user    0m4.850s
sys     0m0.440s
$ time sha512sum rand.img
8dcaddd1337cc9dcef273b4a53243563a619747214256a7c6285e6e7a94c671d06121ba08c87ae8c0a33bde4eb142d80413782913f79d9fd2686799174be1061  rand.img

real    1m38.730s
user    1m33.010s
sys     0m1.120s
```

But like the pattern of `<algorithm>-<digest>` could allow folks to choose alternates that suite their needs better, or once there is the next great algorithm to be switched to.

image.md
duglin: Two reasons:
1 - since the on-the-disk bundle requires it, I thought that for consistency we should require it here.
2 - it seems wrong to have a field point to something that doesn't exist.
If we really want to allow for "no dir at all" then we should let the "rootfs" field to be nil in the config files.

image.md
duglin: actually it does, there are tabs in there, switching to spaces

wking: On Thu, Jan 07, 2016 at 09:57:56AM -0800, Doug Davis wrote:
> > +sha512-1d62d181e9c1322d56ccd3a29d05018399147a16188dbd861c0279ad0dd7e14c/
> > +    config.json
> 
> actually it does, there are tabs in there, switching to spaces

Looks good with the spaces e021a7f → c7b6c60.  Maybe a GitHub
rendering bug?  Anyway, it makes sense now so I'll delete my comment
(including this one).

image.md
wking: On Thu, Jan 07, 2016 at 09:49:07AM -0800, Doug Davis wrote:
> > +1. The root filesystem of the filesystem bundle.
> > +This directory MUST be present in the image, however it MAY be empty (not contain any files or folders).
> 
> Two reasons:
> 1 - since the on-the-disk bundle requires it, I thought that for
>     consistency we should require it here.

I agree on the requirement (you can't create an on-disk bundle if your
image is missing some required piece), but I don't agree on the need
to echo bundle.md here when we could just link to it [1].

> 2 - it seems wrong to have a field point to something that doesn't
>      exist.  If we really want to allow for "no dir at all" then we
>      should let the "rootfs" field to be nil in the config files.

Agreed.  Optional rootfs is [2], and that hasn't landed yet.  But
thinking about this some more, I guess “we have a rootfs and it's
empty” is valid with the current spec (and containers launched from it
would pivot into an empty directory, modulo required, runtime-provided
content [3])?  Anyhow, with that interpretation, it seems like this
condition would also be covered by a single entry punting to
bundle.md.

[1]: https://github.com/opencontainers/specs/pull/293/files#r48905528
     Subject: Re: Dropping the rootfs requirement and restoring arbitrary bundle content
     Date: Tue, 5 Jan 2016 13:52:13 -0800
     Message-ID: <20160105215212.GY3680@odin.tremily.us>
[3]: https://github.com/opencontainers/specs/pull/164

duglin:
created a 2nd commit, so I could keep the first draft around in the first commit, trying to add some of the concepts from https://docs.google.com/document/d/1JPQqH9m9fwZIVsvpB5FSyPJjH0RZLnFkI2J2gohfPu0/edit#heading=h.f5v7pjhtsws8

duglin:
re: ordering - I put the ordering requirements in there so that systems that might be resource constrained didn't need to stash away (potentially large amount of data) prior to it needing to be processed.  You could like at it as it was my attempt to allow a streaming decompression algorithm to be implemented.  May be premature as you suggest, but that was the reasoning behind it.

re: hash naming for dependent images - just seemed like the best way to find what we were looking for w/o then needing to define more properties in json files telling us what names/dirs to look for.

@mrunalp I'm not sure what more we should clarify w.r.t. materializing the rootfs (or its layers).  I tried to say a little as possible knowing that how people may merge the layers (if at all) would be up to them as long as it "appears" right to the runtime.  Open to suggestions though.

vbatts:
are we going to give any hint as to what `./extraFile` is? that is vague, and per the F2F, SHOULD include a manifest of some kind (e.g. an [mtree manifest](http://www.freebsd.org/cgi/man.cgi?mtree(8)), a [continuity file](https://github.com/stevvooe/continuity), a [tar-split data file](https://github.com/vbatts/tar-split), etc)

duglin:
@vbatts do we need to say anything about `./extraFile`?  The spec talks about how there could be extension files in there and that's one of them. Do we need to say anything other than that its ok for other files to be there? When you mention a manifest file, do you mean a manifest for the extra files?  If so, why? Isn't its presence in the tar enough?

philips:
We really need to move this discussion somewhere more productive like a google doc. I really don't think we have a good starting point. For example I have no idea what "extraFile" is and I don't think this proposal is well layered enough to allow for caching/signing of image contents from image configuration as it packages everything together.

image.md
```diff
@@ -0,0 +1,308 @@
...
+  If a referenced layer is embedded within this image then an `ID` property MUST be included.
+  If a referenced layer is not embedded within this image then an `ID` property is OPTIONAL.
+  * `root/path`: ThiS REQUIRED property, defined within [`config.json`](config.md), is the name of the directory into which all layers MUST appear.
+  While the `manifest.json` file extends the `config.json` file, a compliant OCI image is NOT REQUIRED to have all of the required fields from `config.json` within its `manifest.json` file if that image is only used as a layer (dependent image) within other images.
+  This is due to the fact that a layer's `manifest.json` properties are not used to produce a container's `config.json` file and therefore not all of them are used.
+  However, if an image is used to create a container then all of the "required" fields of the `config.json` MUST be present in the `manifest.json`.
+
+1. Layers.
+Each embedded layer MUST have a corresponding image (tar) file, at the root of the image.
+Each layer's tar file MUST be named with its ID followed by the appropriate file-extension based on how the tar file is written (e.g. `.tar` for uncompressed).
```
philips: why?

philips: File extensions don't matter, we should just use magic numbers like ACI does.

image.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+
+### Example 2 - An image with 4 layers, 2 of which are embedded:
+```
+sha512-DEF.tar:
+|-- manifest.json
+|   |== {
+|   |==   ...config.json properties...
+|   |==   "name": "myImage",
```
philips: Name is insufficient to locate a resource. For example there are considerations like architecture, operating system, Kernel ABI version, etc.

image.md
duglin: I wasn't focusing on the extension for any purpose, other than to say it needs to have one **after** the ID part of the name.

Not sure what your "why?" was referring to, but if its the ID part of the filename then I did that because we needed a way to go from the "layers" to the embedded tar file in a consistent fashion.  We can't use the "name" of the image because it might have characters that are not permitted on all filesystems - like a "/" - so having a tar file with that name might not be possible.  Using, and requiring an ID, also forces us to ensure there's a crypto key for each embedded image.

image.md
duglin: Wouldn't this be up to the user to decide that?  If "myImage" is sufficient for their needs then I'm not sure why we should force them into a specific naming scheme.  If we want to include other info about the image (like OS, arch, ...) then we should include it in the manifest.json as a field.

That reminds me, we need to add support for multiple architectures in here....

image.md
```diff
@@ -0,0 +1,308 @@
...
+
+In the following examples, the contents of some of the files within the images are included (in abbreviated form) for clarity.
+
+### Example 1 -  An image with one layer:
+```
+SHA512-ABC.tar:
+|-- manifest.json
+|   |==  {
+|   |==    ...config.json properties...
+|   |==    "name": "myApp",
```
vbatts: are these names arbitrary, or intended to have any implications on the remote service to pull them?

image.md
```diff
@@ -0,0 +1,308 @@
...
+sha512-DEF.tar:
+|-- manifest.json
+|   |== {
+|   |==   ...config.json properties...
+|   |==   "name": "myImage",
+|   |==   "root": {
+|   |==     "path": "myfs"
+|   |==   },
+|   |==   "layers": [
+|   |==     { "name": "example.com/ubuntu" },
```
vbatts: like here for example, is this name pointing to a docker registry? If so, would it be clearer to be `docker://`?
Or is it intended to be a namespacing like java `com.example.ubuntu` style?

image.md
```diff
@@ -0,0 +1,308 @@
...
+|   |   |== }
+|   |
+|   |-- myRoot/
+|   |   |-- home/jeff/runit
+```
+
+### Example 3 - An image with 1 layer, but that one layer references other images:
+
+```
+sha512-GHI.tar:
```
vbatts: nit: isn't hexidecimal

image.md
```diff
@@ -0,0 +1,308 @@
...
+|   |   |==   ],
+|   |   |== }
+|   |
+|   |-- myRoot/
+|   |   |-- home/jeff/runit
+```
+
+### Example 4 - Same as previous but all are included in the image.
+```
+sha512-JKL.tar
```
vbatts: nit: isn't hex

image.md
```diff
@@ -0,0 +1,308 @@
...
+An image MAY optionally contain a directory that's the same as the `root/path` property of the `manifest.json` file.
+And if so, then it MUST be treated as the top-most layer of the root filesystem.
+
+1. There MAY be additional files within the image.
+This specification places no requirements on these files.
+A compliant OCI implementation is NOT REQUIRED to materialize these files into a container's filesystem bundle directory.
+
+## Serialization of an Image
+
+An image MUST be a `tar` of the files that make up an image as specified in the [Image Contents](#image-contents) section.
```
vbatts: To be clear, if the archive includes a subset of tar archives, these tar archive payloads are included in the checksum of the image? but if it just leaves a json reference to them, then the tar archive payloads are not included in the checksum?

image.md
```diff
@@ -0,0 +1,308 @@
...
+  In other words, the first image in the array is written to the filesystem first, and subsequent images are overlayed on top - potentially overlaying an earlier image's files.
+  If a referenced layer is embedded within this image then an `ID` property MUST be included.
+  If a referenced layer is not embedded within this image then an `ID` property is OPTIONAL.
+  * `root/path`: ThiS REQUIRED property, defined within [`config.json`](config.md), is the name of the directory into which all layers MUST appear.
+  While the `manifest.json` file extends the `config.json` file, a compliant OCI image is NOT REQUIRED to have all of the required fields from `config.json` within its `manifest.json` file if that image is only used as a layer (dependent image) within other images.
+  This is due to the fact that a layer's `manifest.json` properties are not used to produce a container's `config.json` file and therefore not all of them are used.
+  However, if an image is used to create a container then all of the "required" fields of the `config.json` MUST be present in the `manifest.json`.
```
vbatts: perhaps an example here. I'm most curious about seeing the mount paths and hooks not being conveyed in a manifest

image.md
duglin: @vbatts yes that is what I was thinking - so the hash of an image is different depending on whether it embedded the layers or not - even if the layers are the same in both cases.   I'm adding a "note" to that effect at the end of this paragraph see if you think that covers it.

image.md
duglin: Been thinking about this more and I'm not sure we should support mutli-arch within this image format. I think it makes more sense to do it at a higher-level - meaning develop the notion of an uber-image that knows how to associate multiple images to each other.  My concern with merging the concept into this one is that if someone wants to pull down an image, we don't want them to have to pull down ALL versions of that image (for all architectures) - they should be able to just pull down the one they want.  There may be cases where all are needed but I think we should call that something else - and define it in a different spec..

image.md
duglin: I believe the intent is more like java namespacing, but not 100% the same - but need @philips's to weigh in.

image.md
duglin: fixed

image.md
```diff
@@ -0,0 +1,309 @@
...
+While the runtime only sees a single directory as the root filesystem, it might actually be stored as a layering of filesytems under the covers.
+When representing this layering within an image, each "layer" will be independently referenced.
+
+## Image Contents
+
+A compliant image contains the following pieces of information:
+
+1. A `manifest.json` file describing the contents of the image.
+This file is REQUIRED to be present in the image.
+This file extends the [`config.json`](config.md) file with some additional properties:
```
vbatts: I'm struggling on this.
As the `config.json` has evolved to essentially be a instance configuration for a container, I don't see how the `manifest.json` is particularly an extension. This comes back to the F2F conversations about having templates, or "suggests" or "requests" for resources. But the config.json is really not portable and would come with a list of caveats and exceptions, right?

duglin: I tend to think of this as a separate discussion. Whether the json in an image becomes the exact values we see in config.json or a template, or ranges, or something else is something that we can work on in parallel.  To me the key question for us in this PR is whether we're ok with a single json file (manifest.json) that contains: 1) what will be used to construct the config.json, and  2) extra fields to help process the image.  Or do we want to split these into multiple files.  I don't yet see value in splitting them.  Then in a different PR we can twiddle the json to be what we want based on whatever usecases we're looking to support.

Perhaps people could enumerate the usecases for something more verbose than the existing config.json.

wking: On Mon, Feb 15, 2016 at 06:40:49PM -0800, Doug Davis wrote:
> Or do we want to split these into multiple files.  I don't yet see
> value in splitting them.

The value I see in splitting files is clear separation of concerns.  A
manifest file contains only the bits needed for fetching and
validation.  A config.json contains only the bits needed by the
‘start’ command.  A process.json contains only the bits needed by the
‘exec’ command [1].  The state JSON contains only the bits needed to
manipulate (join, signal, tear down) a container.  I'm ok allowing
folks to wedge all this into one file [2], but it seems like it would
be easier to write the individual tools and to understand the results
of a particular action if each action had a separate config file.

Unless there's some overlap between the actions, I see no benefit to
configuring multiple actions in the same file.  And “start a container
from this OCI-bundle directory” and “create and OCI-bundle directory
from this tarball” don't seem like they have *any* overlap.

[1]: https://github.com/opencontainers/specs/pull/225/files#diff-b84a8d65d8ed53f4794cd2db7e8ea731R97
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/p_u2xps6jvo/bcMI6hm1DwAJ
     Subject: Re: Labels and extension meta data in containers #108
     Date: Wed, 30 Dec 2015 20:54:37 -0800
     Message-ID: <20151231045437.GA6362@odin.tremily.us>

image.md
philips: On Mon, Feb 15, 2016 at 8:14 AM Doug Davis <notifications@github.com> wrote:

> In image.md
> <https://github.com/opencontainers/specs/pull/293#discussion_r52916754>:
>
> > +|   |   |-- home/jeff/runit
> > +```
> > +
> > +
> > +### Example 2 - An image with 4 layers, 2 of which are embedded:
> > +```
> > +sha512-DEF.tar:
> > +|-- manifest.json
> > +|   |== {
> > +|   |==   ...config.json properties...
> > +|   |==   "name": "myImage",
> > +|   |==   "root": {
> > +|   |==     "path": "myfs"
> > +|   |==   },
> > +|   |==   "layers": [
> > +|   |==     { "name": "example.com/ubuntu" },
>
> I believe the intent is more like java namespacing, but not 100% the same
> - but need @philips <https://github.com/philips>'s to weigh in.
>

Yes, I think if we add a name it must suggest some namespacing. And using
DNS is a reasonable default suggestion because that is well understood for
ownership on the web.

image.md
mrunalp: I agree that adding an example here would help clarify.

image.md
mrunalp: I agree with @philips that having magic numbers might be a better option than having an extension. @duglin, could you explain the advantages of having an extension?

image.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Serialization of an Image
+
+An image MUST be a `tar` of the files that make up an image as specified in the [Image Contents](#image-contents) section.
+The name of the tar file MUST be the ID of the image followed by the appropriate extension based on the type of tar file (e.g. `.tar` for uncompressed tar).
+The ID of an image MUST be a cryptographic hash of the uncompressed tar file.
+An `ID` MUST be a string of the form `<algorithm>-<hash value>`.
+The `algorithm` part of the string MUST be in lower-case.
+For example, `sha512` not `SHA512`.
+It is RECOMMENDED that implementations use the sha512 algorithm.
```
mrunalp: We should perhaps back this with some reasoning and references and also suggest alternatives.

image.md
```diff
@@ -0,0 +1,309 @@
...
+
+Each top-level file, or directory, specified in the tar file MUST appear without a leading `/` in its name.
+
+Each file or directory specified in the [Image Contents](#image-contents) section MUST appear at the root of the file file.
+
+## Expanding an Image to Create a Filesystem Bundle
+
+When an image is expanded to create a filesystem bundle the image's `config.json` file MUST be written to the root of the filesystem bundle's directory.
+The content of `config.json` are dervived from the `manifest.json` file in the image.
+The expansion process MUST create a directory for the container's root filesystem with a name as specified in the `manigest.json` file - ie. the `root/path` property.
```
mrunalp: typo: manigest -> manifest

image.md
mrunalp: +1 to using DNS here.

image.md
duglin: I think you guys are reading too much into the extension - all files (well most) have extensions, tar/tgz/etc... 
In that sentence I was defining the format of the base filename nothing more - I can remove the comment about extensions - I just added it to make it clear that its still a normal file and whatever extension would normally be there should still be there..  I still expect there to be the appopriate magic number in the file - but that's out of scope of this doc - that's defined by 'tar.

image.md
duglin: typo fixed

image.md
duglin: Someone want to propose some text?  We need to be a little careful because I don't necessarily want to get into defining "discovery" within this doc, but providing a statement saying that DNS is a good example of how an implementation can resolve it might be good.

duglin:
rebased

philips:
This work is moving into the newly formed OCI Image Format. https://github.com/opencontainers/image-spec

PR #216, opened 2015-10-06
.travis: Convert TRAVIS_COMMIT_RANGE base...head to base..head

wking:
Work around travis-ci/travis-ci#4596 until that is fixed upstream.
This avoids pulling in commits from the base tip that aren't reachable
from the head tip (e.g. if master has advanced since the PR branched
off, and the PR is against master).  We only want to check commits
that are in the head branch but not in the base branch (more details
on the range syntax [here](http://git-scm.com/docs/gitrevisions#_specifying_ranges)).

Once the Travis bug does get fixed, the shell replacement will be a
no-op.  So we don't have to worry about checks breaking once the bug
gets fixed, and can periodically poll the bug and remove the
workaround at out leisure after the fix.

Spun off from #182 now that #215 has landed.

PR #295, opened 2016-01-07
Makefile: add a target to run tests

vbatts:
For now, just vet and lint. But would like to include the commit
validator, once a good range is selectable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ DOC_FILES := \
 	runtime-config.md \
 	runtime-config-linux.md \
 	glossary.md
+EPOCH_TEST_COMMIT := 041eb73d2e0391463894c04c8ac938036143eba3
 
 docs: pdf html
 .PHONY: docs
@@ -43,7 +44,7 @@ html:
 
 .PHONY: test .govet .golint .gitvalidation
 
-test: .govet .golint
+test: .govet .golint .gitvalidation
 
 # `go get golang.org/x/tools/cmd/vet`
 .govet:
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Jan 07, 2016 at 12:33:19PM -0800, Vincent Batts wrote:
> But would like to include the commit validator, once a good range is
> selectable.

With reliable Git configs, I'd replace ${TRAVIS_COMMIT_RANGE} with:

  @{upstream}..HEAD

although for me, @{upstream} is usually my *downstream* GitHub
repository (because I use ‘hub pull-request’ [1]).  Git lets you
distinguish between push and pull repositories (branch.<name>.remote
and branch.<name>.pushremote), but does not seem to allow that
distinction for branches (as of 2.7.0).

Alternatively, we could assume that all branches are being developed
for opencontainers/specs' master, so:

  REMOTE = $(shell git remote -v | sed -n 's|^\([^[:space:]]*\)[[:space:]].*github.com/opencontainers/specs.*|\1|p' | head -n1)
  …
  .gitvalidation:
      git-validation -run DCO,short-subject -v -range $(REMOTE)/master..HEAD

I'm not sure which is more reliable:

a. @{upstream} actually pointing at the upstream branch, or
b. ‘master’ being the right upstream branch and
   i. branch.<name>.remote actually having the upstream remote name, or
   ii. sed'ing through ‘git remote -v’ to find the right remote name

Alternatively, you could just require folks to set UPSTREAM
explicitly, and guess it's ‘@{upstream}’ or origin/master, or
whatever:

  UPSTREAM = @{upstream}
  …
  .gitvalidation:
      git-validation -run DCO,short-subject -v -range $(UPSTREAM)..HEAD

See #216 for why I prefer two dots to three.

[1]: https://hub.github.com/

crosbymichael:
LGTM

vishh:
LGTM

philips:
LGTM

PR #298, opened 2016-01-12
runtime-config-linux: Separate mknod from cgroups

wking:
This is #99 rebased onto the current master, because [the idea seems
to have new life][0].  It leaves mknod entries in **`linux.devices`**
and moves cgroups entries into **`linux.resources.devices`**.  Quiet
thread on the issue [here][1].

This makes it easy to distinguish between configs that call for cgroup
adjustments (which have linux.resources entries) from those that
don't.  Without this split, folks interested in making that
distinction would have to parse the device section to determine if it
included cgroup changes.  This will also make it easy to drop either
portion ([mknod][2] or [cgroups][3]) independently of the other if the
project decides to do so.

Using seperate sections for mknod and cgroups also allows us to avoid
the complicated validation rules needed for the combined format
mknod/cgroup (#101).

Now that there is a section specific to supplying devices, I [shifted
the default device listing over from config-linux][5].  The `/dev/ptmx`
entry is a bit awkward, since it's not a device, but it seemed to fit
better over here.  But I would also be fine leaving it with the other
mounts in config-linux.

The reference links are sorted into two blocks, with kernel-doc links
sorted alphabetically followed by man pages sorted alphabetically by
section.

[0]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-01-06-18.03.log.html#l-36
[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/y_Fsa2_jJaM
[2]: https://github.com/opencontainers/specs/pull/98
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/qWHoKs8Fsrk
[5]: https://github.com/opencontainers/specs/pull/171#discussion_r41190655

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -16,27 +16,19 @@ Valid values are the strings for capabilities defined in [the man page](http://m
     ]
 ```
 
-## Default Devices and File Systems
+## Default File Systems
 
 The Linux ABI includes both syscalls and several special file paths.
 Applications expecting a Linux environment will very likely expect these files paths to be setup correctly.
 
-The following devices and filesystems MUST be made available in each application's filesystem
+The following filesystems MUST be made available in each application's filesystem
 
-|     Path     |  Type  |  Notes  |
-| ------------ | ------ | ------- |
-| /proc        | [procfs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)    | |
-| /sys         | [sysfs](https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt)    | |
-| /dev/null    | [device](http://man7.org/linux/man-pages/man4/null.4.html)                 | |
-| /dev/zero    | [device](http://man7.org/linux/man-pages/man4/zero.4.html)                 | |
-| /dev/full    | [device](http://man7.org/linux/man-pages/man4/full.4.html)                 | |
-| /dev/random  | [device](http://man7.org/linux/man-pages/man4/random.4.html)               | |
-| /dev/urandom | [device](http://man7.org/linux/man-pages/man4/random.4.html)               | |
-| /dev/tty     | [device](http://man7.org/linux/man-pages/man4/tty.4.html)                  | |
-| /dev/console | [device](http://man7.org/linux/man-pages/man4/console.4.html)              | |
-| /dev/pts     | [devpts](https://www.kernel.org/doc/Documentation/filesystems/devpts.txt)  | |
-| /dev/ptmx    | [device](https://www.kernel.org/doc/Documentation/filesystems/devpts.txt)  | Bind-mount or symlink of /dev/pts/ptmx |
-| /dev/shm     | [tmpfs](https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt)    | |
+|   Path   |  Type  |
+| -------- | ------ |
+| /proc    | [procfs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)   |
+| /sys     | [sysfs](https://www.kernel.org/doc/Documentation/filesystems/sysfs.txt)   |
+| /dev/pts | [devpts](https://www.kernel.org/doc/Documentation/filesystems/devpts.txt) |
+| /dev/shm | [tmpfs](https://www.kernel.org/doc/Documentation/filesystems/tmpfs.txt)   |
 
 ## Namespaces
 
@@ -115,92 +107,58 @@ There is a limit of 5 mappings which is the Linux kernel hard limit.
 
 ## Devices
 
-`devices` is an array specifying the list of devices to be created in the container.
+`devices` is an array specifying the list of devices that MUST be available in the container.
+The runtime may supply them however it likes (with [mknod][mknod.2], by bind mounting from the runtime mount namespace, etc.).
 
 The following parameters can be specified:
 
-* **`type`** *(char, required)* - type of device: `c`, `b`, `u` or `p`. More info in `man mknod`.
-
-* **`path`** *(string, optional)* - full path to device inside container
-
-* **`major, minor`** *(int64, required)* - major, minor numbers for device. More info in `man mknod`. There is a special value: `-1`, which means `*` for `device` cgroup setup.
-
-* **`permissions`** *(string, optional)* - cgroup permissions for device. A composition of `r` (*read*), `w` (*write*), and `m` (*mknod*).
-
-* **`fileMode`** *(uint32, optional)* - file mode for device file
-
-* **`uid`** *(uint32, optional)* - uid of device owner
-
-* **`gid`** *(uint32, optional)* - gid of device owner
-
-**`fileMode`**, **`uid`** and **`gid`** are required if **`path`** is given and are otherwise not allowed.
+* **`type`** *(char, required)* - type of device: `c`, `b`, `u` or `p`.
+  More info in [mknod(1)][mknod.1].
+* **`path`** *(string, required)* - full path to device inside container.
+* **`major, minor`** *(int64, required unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
+* **`fileMode`** *(uint32, optional)* - file mode for the device.
+  You can also control access to devices [with cgroups](#device-whitelist).
+* **`uid`** *(uint32, optional)* - id of device owner.
+* **`gid`** *(uint32, optional)* - id of device group.
 
 ###### Example
 
 ```json
    "devices": [
         {
-            "path": "/dev/random",
+            "path": "/dev/fuse",
             "type": "c",
-            "major": 1,
-            "minor": 8,
-            "permissions": "rwm",
+            "major": 10,
+            "minor": 229,
             "fileMode": 0666,
             "uid": 0,
             "gid": 0
         },
         {
-            "path": "/dev/urandom",
-            "type": "c",
-            "major": 1,
-            "minor": 9,
-            "permissions": "rwm",
-            "fileMode": 0666,
-            "uid": 0,
-            "gid": 0
-        },
-        {
-            "path": "/dev/null",
-            "type": "c",
-            "major": 1,
-            "minor": 3,
-            "permissions": "rwm",
-            "fileMode": 0666,
-            "uid": 0,
-            "gid": 0
-        },
-        {
-            "path": "/dev/zero",
-            "type": "c",
-            "major": 1,
-            "minor": 5,
-            "permissions": "rwm",
-            "fileMode": 0666,
-            "uid": 0,
-            "gid": 0
-        },
-        {
-            "path": "/dev/tty",
-            "type": "c",
-            "major": 5,
+            "path": "/dev/sda",
+            "type": "b",
+            "major": 8,
             "minor": 0,
-            "permissions": "rwm",
-            "fileMode": 0666,
-            "uid": 0,
-            "gid": 0
-        },
-        {
-            "path": "/dev/full",
-            "type": "c",
-            "major": 1,
-            "minor": 7,
-            "permissions": "rwm",
-            "fileMode": 0666,
+            "fileMode": 0660,
             "uid": 0,
             "gid": 0
         }
     ]
 ```
+
+###### Default Devices
+
+In addition to any devices configured with this setting, the runtime MUST also supply:
+
+* [`/dev/null`][null.4]
+* [`/dev/zero`][zero.4]
+* [`/dev/full`][full.4]
+* [`/dev/random`][random.4]
+* [`/dev/urandom`][random.4]
+* [`/dev/tty`][tty.4]
+* [`/dev/console`][console.4]
+* [`/dev/ptmx`][pts.4].
+  A [bind-mount or symlink of the container's `/dev/pts/ptmx`][devpts].
 
 ## Control groups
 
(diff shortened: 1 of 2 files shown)


wking:
@mrunalp, you were [going to write up some more motivation for this][1]?  It's also possible that you intended to post that motivation to [the thread][2], and that I'm jumping the gun with this reroll.  If so, feel free to table/close this PR until the thread has reached a consensus.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-01-06-18.03.log.html#l-50
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/y_Fsa2_jJaM

mrunalp:
We can continue using this PR. i have responded to the mailing list discussion.

mrunalp:
There are two use cases for this separation:
1. The user doesn't want to specify any cgroup configuration. This is easy for the other values but the device cgroups are tied up the device configuration.
2. Running containers under a non-host usernamespace is an issue, because one can't modify the device cgroups in such a scenario.

So, separation will be good for consistency as well as technical reasons.

runtime-config-linux.md
```diff
@@ -77,93 +77,59 @@ There is a limit of 5 mappings which is the Linux kernel hard limit.
 
 ## Devices
 
-`devices` is an array specifying the list of devices to be created in the container.
+`devices` is an array specifying the list of devices that the MUST be available in the container.
```
vishh: Can devices be setup using hooks?

runtime_config_linux.go
```diff
@@ -238,6 +238,20 @@ type Device struct {
 	GID uint32 `json:"gid"`
 }
 
+// DeviceCgroup represents a device rule for the whitelist controller
+type DeviceCgroup struct {
+	// Allow or deny
+	Allow bool `json:"allow"`
+	// Device type, block, char, etc.
+	Type *rune `json:"type"`
```
vishh: This should include `"omitempty"` tag.

runtime_config_linux.go
```diff
@@ -238,6 +238,20 @@ type Device struct {
...
+	// Major is the device's major number.
+	Major *int64 `json:"major"`
```
vishh: Ditto.

runtime_config_linux.go
```diff
@@ -238,6 +238,20 @@ type Device struct {
...
+	// Minor is the device's minor number.
+	Minor *int64 `json:"minor"`
```
vishh: Ditto

runtime-config-linux.md
wking: On Wed, Jan 20, 2016 at 10:15:41AM -0800, Vish Kannan wrote:
> > -`devices` is an array specifying the list of devices to be created in the container.
> > +`devices` is an array specifying the list of devices that the MUST be available in the container.
> 
> Can devices be setup using hooks?

Yeah.  Or with bind-mounts, or by putting the device nodes in the
rootfs (#98).  My attempt to summarize the pushback for dropping the
mknod section is in [1].

[1]: https://github.com/opencontainers/specs/pull/98#issuecomment-128804303

wking:
Rerolled dfb5dad → 3f8105e adding @vishh's ‘omitempty’s [1,2,3].

[1]: https://github.com/opencontainers/specs/pull/298#discussion_r50295751
[2]: https://github.com/opencontainers/specs/pull/298#discussion_r50295858
[3]: https://github.com/opencontainers/specs/pull/298#discussion_r50295869

wking:
In today's meeting there was talk about where default-device creation
happens in the container-setup timeline [1], and whether/how config
authors should be able to opt-out of default-device creation [2].
Addressing either of those seems orthogonal to the split in this PR
[3], but I'd thought I'd mention them here in case anyone else
disagrees about their orthogonality.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-01-20-18.02.log.html#l-40
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-01-20-18.02.log.html#l-43
[3]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-01-20-18.02.log.html#l-59

mrunalp:
Needs rebase.

wking:
On Mon, Jan 25, 2016 at 11:57:44AM -0800, Mrunal Patel wrote:
> Needs rebase.

I'll wait until #284 lands.

wking:
On Mon, Jan 25, 2016 at 12:27:43PM -0800, W. Trevor King wrote:
> Mon, Jan 25, 2016 at 11:57:44AM -0800, Mrunal Patel:
> > Needs rebase.
> 
> I'll wait until #284 lands.

Rebased around #284 with 3f8105e → 81f726e.

  $ diff -u <(git show 3f8105ed4) <(git show 81f726e3)

shows no changes outside of diff context.

config-linux.md
```diff
@@ -115,93 +107,59 @@ There is a limit of 5 mappings which is the Linux kernel hard limit.
 
 ## Devices
 
-`devices` is an array specifying the list of devices to be created in the container.
+`devices` is an array specifying the list of devices that the MUST be available in the container.
```
mrunalp: s/that the/that/

wking: On Wed, Jan 27, 2016 at 10:47:11AM -0800, Mrunal Patel wrote:
> > -`devices` is an array specifying the list of devices to be created in the container.
> > +`devices` is an array specifying the list of devices that the MUST be available in the container.
> 
> s/that the/that/

Fixed in 81f726e → a855883.

config-linux.md
```diff
@@ -228,6 +186,46 @@ You can configure a container's cgroups via the `resources` field of the Linux c
 Do not specify `resources` unless limits have to be updated.
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
+#### Device whitelist
```
mrunalp: Should we call it an access list instead of a whitelist?

wking: On Wed, Jan 27, 2016 at 12:13:03PM -0800, Mrunal Patel wrote:
> > +#### Device whitelist
> 
> Should we call it an access list instead of a whitelist?

I'm just echoing the kernel docs [1] (which moved! [2]).  I've pushed
a855883 → ddd56d8 updating the link target.

[1]: https://www.kernel.org/doc/Documentation/cgroup-v1/devices.txt
[2]: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34a9304a96d6351c2d35dcdc9293258378fc0bd8

config-linux.md
```diff
@@ -115,93 +107,59 @@ There is a limit of 5 mappings which is the Linux kernel hard limit.
...
-
-* **`gid`** *(uint32, optional)* - gid of device owner
-
-**`fileMode`**, **`uid`** and **`gid`** are required if **`path`** is given and are otherwise not allowed.
+* **`type`** *(char, required)* - type of device: `c`, `b`, `u` or `p`.
+  More info in [mknod(1)][mknod.1].
+* **`path`** *(string, required)* - full path to device inside container.
+* **`major, minor`** *(int64, required)* - [major, minor numbers][devices] for the device.
+* **`fileMode`** *(uint32, required)* - file mode for the device.
+  You can also control access to devices [with cgroups](#device-whitelist).
```
vishh: Are the following fields required?

config-linux.md
```diff
@@ -115,93 +107,59 @@ There is a limit of 5 mappings which is the Linux kernel hard limit.
...
-
-* **`uid`** *(uint32, optional)* - uid of device owner
-
-* **`gid`** *(uint32, optional)* - gid of device owner
-
-**`fileMode`**, **`uid`** and **`gid`** are required if **`path`** is given and are otherwise not allowed.
+* **`type`** *(char, required)* - type of device: `c`, `b`, `u` or `p`.
+  More info in [mknod(1)][mknod.1].
+* **`path`** *(string, required)* - full path to device inside container.
+* **`major, minor`** *(int64, required)* - [major, minor numbers][devices] for the device.
```
vishh: This field should not be specified if the type is `p`. Hence it must be optional

config-linux.md
```diff
@@ -115,93 +107,59 @@ There is a limit of 5 mappings which is the Linux kernel hard limit.
...
+* **`fileMode`** *(uint32, required)* - file mode for the device.
```
vishh: This must be optional as well.

config_linux.go
```diff
@@ -251,6 +251,20 @@ type Device struct {
...
+	// Allow or deny
+	Allow bool `json:"allow"`
+	// Device type, block, char, etc.
+	Type *rune `json:"type,omitempty"`
+	// Major is the device's major number.
+	Major *int64 `json:"major,omitempty"`
+	// Minor is the device's minor number.
+	Minor *int64 `json:"minor,omitempty"`
+	// Cgroup access permissions format, rwm.
+	Access string `json:"access"`
```
vishh: AFAIK, this field must be optional

config-linux.md
wking: On Wed, Jan 27, 2016 at 01:13:52PM -0800, Vish Kannan wrote:
> > +* **`major, minor`** *(int64, required)* - [major, minor numbers][devices] for the device.
>
> This field should not be specified if the type is `p`. Hence it must
> be optional

It's hard to create a block or character special file if either of
those is left off.  How about:

* **`major, minor`** *(int64, required unless **`type`** is `p`)* - …

Obviously, that means they need to be optional in the Go type
definition, and we'd need more complicated validation logic to check
that condition.  But I think:

  {
    "path": "/dev/fuse",
    "type": "c",
    "fileMode": 0666,
    "uid": 0,
    "gid": 0
  }

is pretty clearly an invalid entry.

config-linux.md
wking: On Wed, Jan 27, 2016 at 01:13:14PM -0800, Vish Kannan wrote:
> > +* **`fileMode`** *(uint32, required)* - file mode for the device.
> > +  You can also control access to devices [with cgroups](#device-whitelist).
> 
> Are the following fields required?

I'm fine making fileMode, uid, and gid optional.  Would they default
to the effective uid/gid/umask of the runtime process?  Of the
container process as it will be when it execs any user-configured
arguments?

config_linux.go
wking: On Wed, Jan 27, 2016 at 01:16:29PM -0800, Vish Kannan wrote:
> > +	// Cgroup access permissions format, rwm.
> > +	Access string `json:"access"`
> 
> AFAIK, this field must be optional

I'm not sure about *must* (when would you not be able to supply a
value?), but I'm fine making it optional, defaulting to not writing an
access entry to the cgroup device file.  The filesystem API allows
[1]:

  $ echo a > /sys/fs/cgroup/1/devices.deny

which acts like:

  $ echo 'a *:* rwm' > /sys/fs/cgroup/1/devices.deny

so they have an implicit access there (although I don't see
documentation for that beyond the example quoted above).  

[1]: https://kernel.org/doc/Documentation/cgroup-v1/devices.txt

config-linux.md
vishh: Defaults can also depend on it setgid bit is set on the parent dir I think.
But it shouldn't matter from the Spec level.

vishh: What I was saying was that, it is possible to use the struct without
specifying this field. Hence it is optional.

wking: Ok, I just pushed ddd56d8 → 7d5b027, which I think addresses all the
optional/required issues.  I put some comments on the changes with
references, etc., in the new commit message if reviewers want to look
those over alongside the diff.  The commit message cites mknod, but
obviously the bind-mount approach to supplying devices would also work
without fileMode, uid, or gid specified.

config-linux.md
```diff
@@ -228,6 +186,46 @@ You can configure a container's cgroups via the `resources` field of the Linux c
...
+  `null` or unset values mean "all", mapping to [`*` in the filesystem API][cgroups-devices].
+* **`access`** *(string, optional)* - cgroup permissions for device.
+  A composition of `r` (read), `w` (write), and `m` (mknod).
+
+###### Example
+
+```json
+   "devices": [
+        {
+            "allow": false,
```
crosbymichael: I don't really understand this `allow` and `access` and how they work together.  Can you explain better to me?

tonistiigi: I think allow is `devices.allow` vs `devices.deny` and 'rwm` is read, write, mknod.

wking: On Wed, Jan 27, 2016 at 03:30:42PM -0800, Tõnis Tiigi wrote:
> > +* **`allow`** *(boolean, required)* - whether the entry is allowed or denied.
>
> I think allow is `devices.allow` vs `devices.deny` and 'rwm` is
> read, write, mknod.

That's right.  What wording should I add to make it more clear?

crosbymichael: ahh, i guess i get what would happen

crosbymichael:
LGTM

mrunalp:
LGTM

config-linux.md
vishh: Whitelist typically carries a positive connotation. In this case, this whitelist can be used to deny access to devices as well.

PR #300, opened 2016-01-13
Update ROADMAP.md

duglin:

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -17,12 +17,12 @@ Although OCI doesn't define a transport method we should have a cryptographic di
 
 *Owner:* philips
 
-### Review the need for runtime.json
+### Review the need for runtime.json (Target release: v0.3.0)
 
 There are some discussions about having `runtime.json` being optional for containers and specifying defaults.
 Runtimes would use this standard set of defaults for containers and `runtime.json` would provide overrides for fine tuning of these extra host or platform specific settings.
 
-*Owner:*
+*Owner:* mrunalp
 
 ### Define Container Lifecycle
 
@@ -31,15 +31,19 @@ The lifecycle events of a container also help identify areas to implement hooks 
 
 *Owner:* mrunalp
 
-### Define Standard Container Actions
+### Define Standard Container Actions (Target release: v0.3.0)
 
 Define what type of actions a runtime can perform on a container without imposing hardships on authors of platforms that do not support advanced options.
 
-*Owner:*
+*Owner:* duglin
 
 ### Clarify rootfs requirement in base spec
 
 Is the rootfs needed or should it just be expected in the bundle without having a field in the spec?
+
+Resolution: keep it as is - user definable
+Submit as a different PR
+For reference: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/6ZKMNWujDhU
 
 *Owner:*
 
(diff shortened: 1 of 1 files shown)


vbatts:
Thanks for writing up this from our f2f discussions.
I'm taking a look.

duglin:
this is from the OCI F2F (Jan 11/12)

ROADMAP.md
```diff
@@ -31,38 +31,48 @@ The lifecycle events of a container also help identify areas to implement hooks
...
+Proposal: make it an optional feature
+
+*Owner:* hqhq (was vishh) robdolinms, bcorrie
 
 ### Protobuf Config
 
 We currently have only one language binding for the spec and that is Go.
 If we change the specs format in the respository to be something like protobuf then the generation for multiple language bindings become effortless.
 
+Not going to do it
```
vbatts: This is my wording verbatim.

vbatts:
Per discussion, perhaps I should open an PR that includes the discussion of a common distribution format spec in this ROADMAP.md for 1.0

LK4D4:
LGTM

vbatts:
LGTM

ROADMAP.md
wking: On Wed, Jan 13, 2016 at 11:24:14AM -0800, Vincent Batts wrote:
> >  
> >  ### Protobuf Config
> >  
> >  We currently have only one language binding for the spec and that is Go.
> >  If we change the specs format in the respository to be something like protobuf then the generation for multiple language bindings become effortless.
> >  
> > +Not going to do it
> 
> This is my wording verbatim. 

So close #185 with some notes on why?  I'm fine with closing that
door, but I expect we can explain the reasoning more clearly than
“this needs a step back” [1].  Maybe just “we like #276 better because
$REASONS”?

[1]: https://github.com/opencontainers/specs/pull/185#issuecomment-165172658

wking:
On Wed, Jan 13, 2016 at 11:26:05AM -0800, Vincent Batts wrote:
> Per discussion, perhaps I should open an PR that includes the
> discussion of a common distribution format spec in this ROADMAP.md
> for 1.0

Isn't there already a distribution-format PR (#293)?  And I think the
feeling there was that discussion should be happening on the list
[1,2].

[1]: https://github.com/opencontainers/specs/pull/293#issuecomment-169276204
[2]: https://github.com/opencontainers/specs/pull/293#issuecomment-169400365

vbatts:
@wking Two things, I'm referring to having something like #293 formally on the roadmap for 1.0. Secondly, as has been already said, there is active, open discussion of this happening in the face2face meeting. 
https://docs.google.com/document/d/1AtpEgQOc0lzuwRIJuPZgCHdYz4olpTwszsCoeRR1_r4/edit are the notes and agenda.

wking:
On Wed, Jan 13, 2016 at 11:38:15AM -0800, Vincent Batts wrote:
> @wking Two things, I'm referring to having something like #293
> formally on the roadmap for 1.0.

Ah, putting #293 on the roadmap makes more sense to me than opening a
new PR.

> Secondly, as has been already said, there is active, open discussion
> of this happening in the face2face meeting.
> https://docs.google.com/document/d/1AtpEgQOc0lzuwRIJuPZgCHdYz4olpTwszsCoeRR1_r4/edit
> are the notes and agenda.

Yeah, I'm occasionally checking in on the Google Doc, but pushing
comments in there seems like it would just be annoying, and I don't
have time to join in on the audio.  On the other hand, by the time
something from that discussion bubbles back over here, it seems like
GitHub comments are appropriate.

ROADMAP.md
vbatts: That's what we're working on right now. :-)

wking: On Wed, Jan 13, 2016 at 12:11:23PM -0800, Vincent Batts wrote:
> That's what we're working on right now. :-)

Excellent :).  I don't see the need to land #300 with WIP wording and
follow-up PRs, but as long as the motivation gets documented
somewhere, I'm happy.

PR #307, opened 2016-01-13
Clarify cwd

vbatts:
carrying forward #286 by making the field required

### DIFF
--- a/config.md
+++ b/config.md
@@ -69,7 +69,7 @@ The runtime MUST mount entries in the listed order.
 ## Process configuration
 
 * **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.
-* **`cwd`** (string, optional) is the working directory that will be set for the executable.
+* **`cwd`** (string, optional) is the working directory that will be set for the executable. Defaults to "/" on linux.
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

wking:
On Wed, Jan 13, 2016 at 01:41:00PM -0800, Vincent Batts wrote:
> carrying forward #286 by making the field required

Should this have dropped omitempty from Process.Cwd in config.go?

Re [1], I think “different platforms have different defaults” is a
feature (platforms can try and differentiate themselves by picking
more useful defaults, and spec wording like [2] would ensure “runtimes
don't default to something different”).  For an example where we
decided to expose platform defaults (as I think we should do
consistently), see [3].

[1]: https://github.com/opencontainers/specs/pull/286#issuecomment-171442542
[2]: https://github.com/opencontainers/specs/pull/285#issuecomment-167823651
[3]: https://github.com/opencontainers/specs/pull/285#issuecomment-171425759

wking:
On Wed, Jan 13, 2016 at 01:41:00PM -0800, Vincent Batts wrote:
>   * config: cwd is a required default

I'm not sure what the plan is for maintaining the ChangeLog is [1],
but if it's not “added right before cutting a release” we probably
don't want to wait too long before landing an entry for this
backwards-incompatible change.

[1]: https://github.com/opencontainers/specs/pull/294#issuecomment-171106678

Issue #110, opened 2015-08-14
"version" should not be used for the manifest format version


stuartpb:
I think `"version"` should be reserved for future expansion and/or metadata extensions by the environment, not used for the *manifest version* the JSON document complies with. That should be specified with a key like Chrome's [`manifest_version`](https://developer.chrome.com/extensions/manifestVersion).

crosbymichael:
SGTM

wking:
On Fri, Aug 14, 2015 at 03:51:45PM -0700, Stuart P. Bentley wrote:
> I think `"version"` should be reserved for future expansion and/or
> metadata extensions by the environment…

I'm not sure what else would go here, but if we want to avoid claiming
‘version’, I'd suggest ‘ocsVersion’, ‘OCSVersion’, or
‘openContainerSpecificationVersion’.  The other spec tags are all
mixedCase, so I'd avoid underscores or hyphens.  And if we're
namespacing to distinguish between multiple versioned entities, the
special thing about *this* version entry is that it references this
Open Container Specification [1].

[1]: https://github.com/opencontainers/specs/tree/9ad789fa16156841074275172530fe7db6bcb26b#open-container-specifications

jonboulle:
FWIW we went through this exact conversation a very long time ago in appc
land and settled on "acVersion". The nice part about this (vs. the more
generic manifest_version) is it also effectively doubles as a magic header
for identifying some generic JSON object as an ocSomething

wking:
On Fri, Aug 21, 2015 at 03:12:58PM -0700, Jonathan Boulle wrote:
> The nice part about this (vs. the more generic manifest_version) is
> it also effectively doubles as a magic header for identifying some
> generic JSON object as an ocSomething

config.json is also not a manifest ;).  And I prefer ‘ocs’ as a prefix
to just ‘oc’, since I think we want to leave room for additional OCI
specs in the future [1].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/BIxya5eSNLo

wking:
On Fri, Aug 14, 2015 at 03:51:45PM -0700, Stuart P. Bentley wrote:
> I think `"version"` should be reserved for future expansion…

Thinking this over some more, if we do want to namespace our versions,
it might be better to nest a map under ‘version’:

  "version": {
    "openContainerSpecification": "0.1.0",
    …
  }

(or: "ocs": "0.1.0").

That avoids polluting the root JSON namespace with a bunch of
[...]

PR #309, opened 2016-01-13
config: qualify the name of the version field

vbatts:
https://github.com/opencontainers/specs/issues/110

### DIFF
--- a/config.go
+++ b/config.go
@@ -5,7 +5,7 @@ package specs
 // bundle is packaged for distribution.
 type Spec struct {
 	// Version is the version of the specification that is supported.
-	Version string `json:"version"`
+	Version string `json:"ociVersion"`
 	// Platform is the host information for OS and Arch.
 	Platform Platform `json:"platform"`
 	// Process is the container's main process.
--- a/config.md
+++ b/config.md
@@ -8,14 +8,17 @@ This includes the process to run, environment variables to inject, sandboxing fe
 
 Below is a detailed description of each field defined in the configuration format.
 
-## Manifest version
+## Specification version
 
-* **`version`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OCF specification with which the container bundle complies. The Open Container spec follows semantic versioning and retains forward and backward compatibility within major versions. For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series. NOTE that there is no guarantee for forward or backward compatibility for version 0.x.
+* **`ociVersion`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OpenContainer specification with which the bundle complies.
+The OpenContainer spec follows semantic versioning and retains forward and backward compatibility within major versions.
+For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series.
+NOTE that there is no guarantee for forward or backward compatibility for version 0.x.
 
 *Example*
 
 ```json
-    "version": "0.1.0"
+    "ociVersion": "0.1.0"
 ```
 
 ## Root Configuration


vbatts:
Closes #110

tianon:
LGTM

vishh:
LGTM

config.go
```diff
@@ -5,7 +5,7 @@ package specs
 // bundle is packaged for distribution.
 type Spec struct {
 	// Version is the version of the specification that is supported.
-	Version string `json:"version"`
+	Version string `json:"spec_version"`
```
wking: I'd [rather have a more specific name][1].  Presumably the other versioned things were distinguishing from also have specs (and the current JSON pattern is camelCase, not underscores).

[1]: https://github.com/opencontainers/specs/issues/110#issuecomment-171470005

config.md
```diff
@@ -8,14 +8,17 @@ This includes the process to run, environment variables to inject, sandboxing fe
 
 Below is a detailed description of each field defined in the configuration format.
 
-## Manifest version
+## Specification version
 
-* **`version`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OCF specification with which the container bundle complies. The Open Container spec follows semantic versioning and retains forward and backward compatibility within major versions. For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series. NOTE that there is no guarantee for forward or backward compatibility for version 0.x.
+* **`spec_version`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OpenContainer specification with which the bundle complies.
```
mrunalp: should be specVersion

vbatts: i already pushed a change for this. PTAL

wking: On Thu, Jan 14, 2016 at 02:37:10PM -0800, Vincent Batts wrote:
> > +* **`spec_version`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OpenContainer specification with which the bundle complies.
> 
> i already pushed a change for this. PTAL

4df6c3bb (the current tip) has specVersion in config.go but
spec_version in config.md (twice).  Are we sure we don't want a more
specific name [1,2]?

[1]: https://github.com/opencontainers/specs/pull/309#discussion_r49667293
[2]: https://github.com/opencontainers/specs/issues/110#issuecomment-171470005

jonboulle: +1 to more specific name.

duglin: `ociVersion`

mrunalp: +1 to ociVersion

vbatts:
updated to `ociVersion`. PTAL

duglin:
+1

tianon:
re-LGTM

wking:
ociVersion is better, but we may end up with multiple OCI specs.
We're calling this spec “Open Container Specifications” [1], so why
not “ocsVersion”?

[1]: https://github.com/opencontainers/specs/tree/v0.2.0#open-container-specifications

duglin:
Not to take us down too much of a tangent, but do we have one spec or multiple in this project? If we have multiple then "ocs" doesn't help since having "s=spec" when there are multiple is kind of ambiguous. Perhaps "Open Container Runtime (ocr)" for the core spec we've been talking about.   But if we view this all as one big spec (which might be nice to avoid each one having a different version number), then just "oci" would work.

wking:
On Fri, Jan 15, 2016 at 10:35:34AM -0800, Doug Davis wrote:
> Not to take us down too much of a tangent, but do we have one spec
> or multiple in this project?

I would love to have this question answered, since I'm frequently
suggesting we handle orthogonal issues in separate OCI specs
(e.g. [1]).  But to stay on topic here, maybe we should move this
discussion to the list?

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/uo11avcWlQQ/Tn0nFymREAAJ
     Subject: Re: appc + oci harmonization progress
     Date: Thu, 13 Aug 2015 11:29:08 -0700
     Message-ID: <20150813182908.GS412@odin.tremily.us>

hqhq:
We should leave the discussion about the needs of multiple OCI spec to the list.
LGTM

PR #310, opened 2016-01-13
config: platform dependent reference source

vbatts:
This introduces verbiage of fields that may occur in json (technically
optional), but is required on certain platforms (e.g. Linux).

Not adding a "name" string, as that is not a requirement yet.

In the event a windows runtime shows up, I could imagine an `sid`, but
we'll get to that when it happens.

Closes https://github.com/opencontainers/specs/pull/135
Related to https://github.com/opencontainers/specs/issues/166


config.go
```diff
@@ -33,6 +33,17 @@ type Process struct {
 	Cwd string `json:"cwd"`
 }
 
+// User specifies linux specific user and group information for the container's
+// main process.
+type User struct {
+	// UID is the user ID the Process is executed as. (this field is platform dependent)
```
vishh: Should we use a json tag to indicate platform dependency? It will help with tooling as well.

vishh:
LGTM

wking:
On Wed, Jan 13, 2016 at 03:53:12PM -0800, Vincent Batts wrote:
> In the event a windows runtime shows up, I could imagine an `sid`,
> but we'll get to that when it happens.

+1 on this PR and its punt, but until we have no platform-specific
build decisions I think #135 should stay open.  #190 implies that
having *_linux.go is enough to trigger platform-specific build
decisions.

config.go
vbatts: Good idea

vbatts:
This PR now addresses all the platform dependent source, not just the `User` struct.
And added field tags that we can build tooling around.

vbatts:
PTAL

duglin:
looks nice

config.md
```diff
@@ -46,24 +46,24 @@ The runtime MUST mount entries in the listed order.
 *Example*
 
 ```json
-"mounts": [
+  "mounts": [
     {
-        "name": "proc",
-        "path": "/proc"
+      "name": "proc",
+      "path": "/proc"
```
wking: Most of our other JSON examples use four-space indents, so we probably just want a blanket `s/^/    /` on this block instead of dropping to two-space indents.

wking: And either way, this seems orthogonal to platform-dependent Go types ;).

config.md
```diff
@@ -117,6 +116,8 @@ For Linux-based systems the user structure has the following fields:
 * **`os`** (string, required) specifies the operating system family this image must run on. Values for os must be in the list specified by the Go Language document for [`$GOOS`](https://golang.org/doc/install/source#environment).
 * **`arch`** (string, required) specifies the instruction set for which the binaries in the image have been compiled. Values for arch must be in the list specified by the Go Language document for [`$GOARCH`](https://golang.org/doc/install/source#environment).
 
+*Example*
+
```
wking: Drive by?  I'd prefer migrating config.md to linkable headers (like runtime-config-linux.md has now), but that should happen in a separate PR.

wking: Although I see you're dropping those headers to bold later on in this PR :p.  Is the header styling really so bad that avoiding it is more important than linkable examples?

config.md
```diff
@@ -125,6 +126,46 @@ For Linux-based systems the user structure has the following fields:
 ```
 
 Interpretation of the platform section of the JSON file is used to find which platform-specific sections may be available in the document.
-For example, if `os` is set to `linux`, then a JSON object conforming to the [Linux-specific schema](config-linux.md) SHOULD be found at the key `linux` in the `config.json`.
+For example, if `os` is set to `linux`, then a JSON object conforming to the [Linux-specific schema](#linux-specific-container-configuration) SHOULD be found at the key `linux` in the `config.json`.
+
+## Linux-specific Container Configuration
+
+The Linux container specification.
+uses various kernel features like namespaces, cgroups, capabilities, LSM, and file system jails to fulfill the spec.
```
wking: This sentence was broken with a period inserted.  I'd leave it as a single sentence.

wking:
On Thu, Jan 14, 2016 at 01:56:28PM -0800, Vincent Batts wrote:
> This PR now addresses all the platform dependent source…

You squashed config-linux.md into config.md but left
runtime-config-linux.md as its own file.  I'm fine either going with
platform-specific files or per-config files, but I think we should
pick one way or the other.

config.md
```diff
@@ -76,9 +76,9 @@ The runtime MUST mount entries in the listed order.
 The user for the process is a platform-specific structure that allows specific control over which user the process runs as.
 For Linux-based systems the user structure has the following fields:
 
-* **`uid`** (int, required) specifies the user id.
-* **`gid`** (int, required) specifies the group id.
-* **`additionalGids`** (array of ints, optional) specifies additional group ids to be added to the process.
+* **`uid`** (int, required on Linux) specifies the user id.
+* **`gid`** (int, required on Linux) specifies the group id.
+* **`additionalGids`** (array of ints, optional on Linux) specifies additional group ids to be added to the process.
```
hqhq: It's already said "For Linux-based systems the user structure has the following fields", isn't this additional change kind of redundant?

hqhq:
Is the plan to remove `*-linux.md` and won't add any `*-windows.md` but all combined into `config.md`? It's gonna make it really fat, isn't there other way we can do to make specs platforms agnostic?

vbatts:
@hqhq that's a fair point

wking:
On Fri, Jan 15, 2016 at 01:29:00AM -0800, Qiang Huang wrote:
> Is the plan to remove `*-linux.md` and won't add any `*-windows.md`
> but all combined into `config.md`? It's gonna make it really fat,
> isn't there other way we can do to make specs platforms agnostic?

Combining the Go (or just changing *_linux.go to something else.  Does
Go treat linux/*.go as platform-specific?) seems orthogonal to
combining the Markdown docs.  We can leave *-linux.md and still have
cross-platform Go.

wking:
On the other hand, lots of folks want a single-page spec (#263), and a
single-page Markdown would make that easier (fixing links [1], etc.).

[1]: https://github.com/opencontainers/specs/pull/263#issuecomment-163341758

duglin:
I have a preference for one file with sections for platform specific stuff.  Less of a chance of having to duplicate things (and keep them in-sync). And it easier to then compare the differences between what each platform expects.  Besides, the size of these platform specific bits should be small.

vbatts:
I started to do runtime*.md, but it was much longer. It was really an arbitrary decision.

config.md
vbatts: I'll drop that part of the sentence

config.go
```diff
@@ -33,6 +36,17 @@ type Process struct {
 	Cwd string `json:"cwd"`
 }
 
+// User specifies linux specific user and group information for the container's
+// main process.
+type User struct {
```
vishh: Should this type be named `LinuxUser` for the sake of readability?

runtime_config.go
```diff
@@ -10,6 +12,9 @@ type RuntimeSpec struct {
 	Mounts map[string]Mount `json:"mounts"`
 	// Hooks are the commands run at various lifecycle events of the container.
 	Hooks Hooks `json:"hooks"`
+
+	// LinuxRuntime is platform specific configuration for linux based containers. (this field is platform dependent)
+	Linux LinuxRuntime `json:"linux,omitempty"`
```
vishh: nit: `*LinuxRuntime`

config.go
vbatts: No, because the intention is that as other platforms have their own fields, they still fall in here.

runtime_config.go
```diff
@@ -40,3 +45,297 @@ type Mount struct {
...
+	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
+	// The path is expected to be relative to the cgroups mountpoint.
+	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
+	CgroupsPath *string `json:"cgroupsPath,omitempty"`
+	// Namespaces contains the namespaces that are created and/or joined by the container
+	Namespaces []Namespace `json:"namespaces"`
+	// Devices are a list of device nodes that are created and enabled for the container
+	Devices []Device `json:"devices"`
+	// ApparmorProfile specified the apparmor profile for the container.
+	ApparmorProfile string `json:"apparmorProfile"`
```
vishh: Why is there no `omitempty`?

runtime_config.go
```diff
@@ -40,3 +45,297 @@ type Mount struct {
...
+	// SelinuxProcessLabel specifies the selinux context that the container process is run as.
+	SelinuxProcessLabel string `json:"selinuxProcessLabel"`
```
vishh: ditto here

wking:
On Fri, Jan 15, 2016 at 10:19:34AM -0800, Doug Davis wrote:
> Besides, the size of these platform specific bits should be small.

Heh.

  $ git describe
  v0.2.0
  $ wc -l *config*.md
     39 config-linux.md
    130 config.md
    535 runtime-config-linux.md
    122 runtime-config.md
    826 total

so 70% of our config docs are Linux specific, and that's not even
counting Linux-specific examples like the process entry in config.md
and the mount entry in runtime-config.md.  Of course, [1] would bring
that number down a lot, but since we still haven't landed an opt-out
like #237, I'd be surprised if cgroups are moved outside of this spec
before summer (if they get moved out at all).

[1]: https://groups.google.com/a/opencontainers.org/forum/#!original/dev/qWHoKs8Fsrk/eXxZsxmWDAAJ

config.go
duglin: +1 and while in there perhaps renaming "Linux" above too - when I first saw `Linux Linux ...` it made me laugh.

runtime_config.go
```diff
@@ -40,3 +45,297 @@ type Mount struct {
...
+	// Seccomp specifies the seccomp security settings for the container.
+	Seccomp Seccomp `json:"seccomp"`
```
vishh: nit: Make this optional.

runtime_config.go
vbatts: alrighty

runtime_config.go
vbatts: and a pointer. got it.

config.go
vishh: Should these tags be added to all Linux specific fields? It is perfectly fine to tackle that in a separate PR, if you agree.
These tags can help with validation.

wking:
On Fri, Jan 15, 2016 at 10:34:21AM -0800, Vincent Batts wrote:
> I started to do runtime*.md, but it was much longer.

If we're not comfortable rolling that in, that's fine with me [1], and
if we don't want to develop in a single page, but still provide a
single page view that's also fine with me [2].  But if we don't want
squash *-linux.md for the runtime.json spec, I'd suggest rerolling
this series to avoid squashing *-linux.md for the config.json spec.
Otherwise we'll get odd half-squashing if/when we unify runtime.json
and config.json (#284).

[1]: https://github.com/opencontainers/specs/pull/310#issuecomment-172023477
[2]: https://github.com/opencontainers/specs/pull/310#issuecomment-172024305

vbatts:
will need more rebasing, with https://github.com/opencontainers/specs/pull/329

vbatts:
Rebased

vbatts:
Rebased

vbatts:
@crosbymichael if you would give this a once-over. It ought not be a code/structure change at all.

crosbymichael:
LGTM

PR #316, opened 2016-01-26
Remove pointers for slices preferring omitempty tag instead 

mrunalp:
Converting to pointers to be consistent with other cgroups settings.

### DIFF
--- a/runtime_config_linux.go
+++ b/runtime_config_linux.go
@@ -139,15 +139,15 @@ type BlockIO struct {
 	// Specifies tasks' weight in the given cgroup while competing with the cgroup's child cgroups, range is from 10 to 1000, CFQ scheduler only
 	LeafWeight *uint16 `json:"blkioLeafWeight,omitempty"`
 	// Weight per cgroup per device, can override BlkioWeight
-	WeightDevice []*WeightDevice `json:"blkioWeightDevice,omitempty"`
+	WeightDevice []WeightDevice `json:"blkioWeightDevice,omitempty"`
 	// IO read rate limit per cgroup per device, bytes per second
-	ThrottleReadBpsDevice []*ThrottleDevice `json:"blkioThrottleReadBpsDevice,omitempty"`
+	ThrottleReadBpsDevice []ThrottleDevice `json:"blkioThrottleReadBpsDevice,omitempty"`
 	// IO write rate limit per cgroup per device, bytes per second
-	ThrottleWriteBpsDevice []*ThrottleDevice `json:"blkioThrottleWriteBpsDevice,omitempty"`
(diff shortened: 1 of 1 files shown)


mrunalp:
@crosbymichael @vishh PTAL

mrunalp:
We can wait for #284 to merge before this.

runtime_config_linux.go
```diff
@@ -195,7 +195,7 @@ type Network struct {
 	// Set class identifier for container's network packets
 	ClassID *uint32 `json:"classID"`
 	// Set priority of network traffic for container
-	Priorities []InterfacePriority `json:"priorities"`
+	Priorities []*InterfacePriority `json:"priorities"`
```
vishh: Why? If we add an `omitempty` tag, even an empty list signifies non-existence of this field.

runtime_config_linux.go
```diff
@@ -213,7 +213,7 @@ type Resources struct {
 	// BlockIO restriction configuration
 	BlockIO *BlockIO `json:"blockIO,omitempty"`
 	// Hugetlb limit (in bytes)
-	HugepageLimits []HugepageLimit `json:"hugepageLimits,omitempty"`
+	HugepageLimits []*HugepageLimit `json:"hugepageLimits,omitempty"`
```
vishh: ditto

runtime_config_linux.go
mrunalp: @vishh I am fine one way or the other. But we have them as pointers in more places compared to where we don't.

vishh: In general, we don't need pointers for slices and maps. I can go cleanup
the Spec again once I get some spare cycles.

runtime_config_linux.go
```diff
@@ -195,7 +195,7 @@ type Network struct {
 	// Set class identifier for container's network packets
 	ClassID *uint32 `json:"classID"`
 	// Set priority of network traffic for container
-	Priorities []InterfacePriority `json:"priorities"`
+	Priorities []*InterfacePriority `json:"priorities,omitempty"`
```
vishh: For this PR just adding `omitempty` here should do. Kindly revert the pointer changes.

runtime_config_linux.go
mrunalp: @vishh No issues going one way or the other. I can change the PR.

vishh: Thanks :)

mrunalp:
@vishh Updated to remove pointers from slices.

vishh:
LGTM. Thanks for the cleanup @mrunalp :+1:

hqhq:
LGTM

PR #317, opened 2016-01-26
style: Document recent Go-pointer exceptions

wking:
The general rule seems to be:

> If Go's default value has the same semantics we'd use for an unset
> value, don't bother with a pointer.

I'm not sure how well that squares with [the pointer motivation][1],
which sounded like:

> We want a consistent way to identify unset settings.

But if the falsy values count as “unset”, maybe the “null is a
consistent identifier for unset” approach was never really viable.

I'm also not sure if the new style extends to integers where zero has
the same semantics as unset values.  It sounds like Michael [was ok
with no pointers for those values][2], but `OOMScoreAdj` (where zero
clearly means “do nothing”) [got a pointer in #233][3].  More clarity
on the threshold would be nice.

[1]: https://github.com/opencontainers/specs/pull/233#discussion_r47829711
[2]: https://github.com/opencontainers/specs/pull/233#issuecomment-155250592
[3]: https://github.com/opencontainers/specs/pull/233/files#diff-34c30be66233f08b447fb608ea0e66bbR206

### DIFF
--- a/style.md
+++ b/style.md
@@ -13,9 +13,14 @@ The redundancy reduction from removing the namespacing prefix is not useful enou
 ## Optional settings should have pointer Go types
 
 So we have a consistent way to identify unset values ([source][optional-pointer]).
+The exceptions are entries where the Go default for the type is a no-op in the spec, in which case `omitempty` is sufficient and no pointer is needed (sources [here][no-pointer-for-slices], [here][no-pointer-for-boolean], and [here][pointer-when-updates-require-changes]).
+
 
 [capabilities]: config-linux.md#capabilities
 [class-id]: runtime-config-linux.md#network
 [integer-over-hex]: https://github.com/opencontainers/specs/pull/267#discussion_r48360013
 [keep-prefix]: https://github.com/opencontainers/specs/pull/159#issuecomment-138728337
+[no-pointer-for-boolean]: https://github.com/opencontainers/specs/pull/290#discussion_r50296396
+[no-pointer-for-slices]: https://github.com/opencontainers/specs/pull/316/files#r50782982
 [optional-pointer]: https://github.com/opencontainers/specs/pull/233#discussion_r47829711
+[pointer-when-updates-require-changes]: https://github.com/opencontainers/specs/pull/317/files#r50932706


style.md
```diff
@@ -13,9 +13,12 @@ The redundancy reduction from removing the namespacing prefix is not useful enou
 ## Optional settings should have pointer Go types
 
 So we have a consistent way to identify unset values ([source][optional-pointer]).
+The exceptions are slices, maps, and booleans where the default should be false, in which case `omitempty` is sufficient and no pointer is needed ([source][no-pointer-for-slices] and [source][no-pointer-for-boolean]).
```
vishh: I disagree with `bool` values. How can I express "don't change a value of a flag" if the flag was boolean? "false" does not always mean "don't change". 
I think it still makes sense to have pointers to booleans.

wking: On Tue, Jan 26, 2016 at 10:45:05AM -0800, Vish Kannan wrote:
> > +The exceptions are slices, maps, and booleans where the default
> >  should be false, in which case `omitempty` is sufficient and no
> >  pointer is needed ([source][no-pointer-for-slices] and
> >  [source][no-pointer-for-boolean]).
>
> I disagree with `bool` values. How can I express "don't change a
> value of a flag" if the flag was boolean? "false" does not always
> mean "don't change".

But sometimes false does mean “don't change” (and those are the
booleans where the default should be false).  For example, the source
I link is you asking for *bool → bool change [1] ;).

If my wording here comes across as “slices, maps, and booleans should
never use pointers” instead of the intended “slices, maps, and
booleans should not use pointers when Go's default for that type means
‘don't do anything’”, then I'm happy to push alternate wording.

[1]: https://github.com/opencontainers/specs/pull/290#discussion_r50296396

hqhq: > But sometimes false does mean “don't change” (and those are the
booleans where the default should be false)

That'll be a problem if we want some `bool` values to be updated (live update), we surely want to change some booleans from `true` to `false`.

wking: On Tue, Jan 26, 2016 at 06:03:44PM -0800, Qiang Huang wrote:
> > But sometimes false does mean “don't change” (and those are the
> > booleans where the default should be false)
> 
> That'll be a problem if we want some `bool` values to be updated
> (live update), we surely want to change some booleans from `true` to
> `false`.

Yeah, booleans where ‘false’ requires action should use pointers.  I
guess the wording in 60b83df is based on my assumption that “default”
equals “don't do anything” [1].  I've pushed 60b83df → d715acf with
wording to explicitly pick out “Go's default for the type is a no-op
in the spec” cases.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #319, opened 2016-01-27
state: remove the file API path

vbatts:
After much discussion, we are not going to accommodate a compatible file
API, but rather have a response expected from a runtime to query for a
container's state.

This should like include the addition of paths to where the namespace references are bind mounted, at least for the linux platform

### DIFF
--- a/config_linux.go
+++ b/config_linux.go
@@ -2,9 +2,6 @@ package specs
 
 import "os"
 
-// LinuxStateDirectory holds the container's state information
-const LinuxStateDirectory = "/run/opencontainer/containers"
-
 // LinuxSpec is the full specification for linux containers.
 type LinuxSpec struct {
 	Spec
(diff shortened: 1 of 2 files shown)


wking:
On Wed, Jan 27, 2016 at 01:41:52PM -0800, Vincent Batts wrote:
> After much discussion, we are not going to accommodate a compatible file
> API, but rather have a response expected from a runtime to query for a
> container's state.

As of 2614bcefe2, this is already handled in #225 (slated to land in
the next few days before #225).  I think #225 is biting off too much
for one PR [1], but it's probably too late to start splitting it out
now.  If we bump #225 to v0.4.0, then splitting it up would be more
feasible.

[1]: https://github.com/opencontainers/specs/pull/225/files#r50440584

vbatts:
ah yeah. I misremembered, and thought I saw a PR for this. After reviewing the unified config.json PR, I wondered why this path was still around. I'll close this PR

PR #322, opened 2016-02-12
README: BlueJeans -> Zoom for the weekly call

wking:
And use generic link text and a generic label so we can just bump the
URL the next time we switch providers.  Catches up with [this
post][1].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/vPfm0jYCmjY/tqPwWT_bDQAJ

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ When in doubt, start on the [mailing-list](#mailing-list).
 ## Weekly Call
 
 The contributors and maintainers of the project have a weekly meeting Wednesdays at 10:00 AM PST.
-Everyone is welcome to participate in the [BlueJeans call][BlueJeans].
+Everyone is welcome to participate in the [call][].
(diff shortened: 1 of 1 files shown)


vbatts:
Simmer down. We haven't even used it yet.

wking:
On Thu, Feb 11, 2016 at 08:47:06PM -0800, Vincent Batts wrote:
> Simmer down. We haven't even used it yet.

Next week, folks are going to be wondering where the meeting is at
(we've been using BlueJeans for a while, and folks still pop up on IRC
asking for the meeting URL).  I don't mind leaving this PR open until
we have more Zoom-time under our belts, but I don't see any harm in
landing it now either.  It didn't take me long to write up, and it
won't take long to swap in the next URL if/when we dump Zoom.

crosbymichael:
We will let people know on the mailing list where its at, lets give it a try first before changing all the docs.

PR #323, opened 2016-02-17
Fix type of devices type

hqhq:
Fixes: https://github.com/opencontainers/runc/issues/566

For type rune, we can assign char as 'c' in struct, but after
marshal, it'll be presented as int32. So it should be a
number in our example of json file. Otherwise we'll got error:
`json: cannot unmarshal string into Go value of type int32`
If we trying to unmarshal the json file.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -112,7 +112,7 @@ The runtime may supply them however it likes (with [mknod][mknod.2], by bind mou
 
 The following parameters can be specified:
 
-* **`type`** *(char, required)* - type of device: `c`, `b`, `u` or `p`.
+* **`type`** *(string, required)* - type of device: `c`, `b`, `u` or `p`.
   More info in [mknod(1)][mknod.1].
 * **`path`** *(string, required)* - full path to device inside container.
 * **`major, minor`** *(int64, required unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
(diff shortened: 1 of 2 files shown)


hqhq:
I found the example of `fileMode` is also wrong, pushing a patch to fix it.

crosbymichael:
I think we should change it from rune to string so that you can actually write "b", "c" in the json spec and you can easily know what type of device it is.  What do you think?

vishh:
+1 for switching to `string`.

wking:
On Mon, Feb 22, 2016 at 11:32:06AM -0800, Vish Kannan wrote:
> +1 for switching to `string`.

ditto

hqhq:
Updated, thanks.

wking:
On Mon, Feb 22, 2016 at 05:24:03PM -0800, Qiang Huang wrote:
> Updated, thanks.

I prefer backticks to quotes for Markdown literals, but other than
that minor quibble, both 9cd1ae59 and 96c5ff07 look good to me.

hqhq:
@wking Using quotes so people know that's a `string`, after a second thought, I think backticks won't cause confusion, and it's more consistent with other usage in specs, updated, thanks.

wking:
On Mon, Feb 22, 2016 at 10:36:46PM -0800, Qiang Huang wrote:
> … I think backticks won't cause confusion… updated…

9bab93004 and ccf3a246c look good to me.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #329, opened 2016-03-02
Move process specific settings to process

crosbymichael:
This moves process specific settings like caps, apparmor, and selinux
process label onto the process structure to allow the same settings to
be changed at exec time.

### DIFF
--- a/config.go
+++ b/config.go
@@ -33,6 +33,14 @@ type Process struct {
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
 	Cwd string `json:"cwd"`
+	// Capabilities are linux capabilities that are kept for the container.
+	Capabilities []string `json:"capabilities,omitempty"`
+	// ApparmorProfile specified the apparmor profile for the container.
+	ApparmorProfile string `json:"apparmorProfile,omitempty"`
+	// SelinuxProcessLabel specifies the selinux context that the container process is run as.
+	SelinuxLabel string `json:"selinuxLabel,omitempty"`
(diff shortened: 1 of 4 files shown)


mrunalp:
LGTM

vbatts:
LGTM

config_linux.go
```diff
@@ -35,16 +33,10 @@ type Linux struct {
 	Namespaces []Namespace `json:"namespaces"`
 	// Devices are a list of device nodes that are created for the container
 	Devices []Device `json:"devices"`
-	// ApparmorProfile specified the apparmor profile for the container.
-	ApparmorProfile string `json:"apparmorProfile"`
-	// SelinuxProcessLabel specifies the selinux context that the container process is run as.
-	SelinuxProcessLabel string `json:"selinuxProcessLabel"`
 	// Seccomp specifies the seccomp security settings for the container.
 	Seccomp Seccomp `json:"seccomp"`
```
vishh: Shouldn't we move seccomp as well?

config.go
```diff
@@ -33,6 +33,14 @@ type Process struct {
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
 	Cwd string `json:"cwd"`
```
vishh: I'd like to move `Resources` as well. Use case is that of running exec processes in a sub-container inside the main container. Imagine not letting your editor exec session resulting in a OOM kill of the container.

crosbymichael: cgroups is more of a container level feature instead of just one process

vishh: How can one then place resource restrictions on an exec'ed process? Run a
new container?

PR #331, opened 2016-03-04
Add annotations and labels to the Spec.

### LINKED ISSUE #108: Labels and extension meta data in containers

vishh:
Fixes #108

### DIFF
--- a/config.go
+++ b/config.go
@@ -18,6 +18,8 @@ type Spec struct {
 	Mounts []Mount `json:"mounts"`
 	// Hooks are the commands run at various lifecycle events of the container.
 	Hooks Hooks `json:"hooks"`
+	// Annotations is an unstructured key value map that may be set by external tools to store and retrieve arbitrary metadata.
+	Annotations map[string]string `json:"annotations,omitempty"`
 }
 
 // Process contains information to start a specific application inside the container.
--- a/config.md
+++ b/config.md
@@ -231,4 +231,17 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 `args` and `env` are optional.
 The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://golang.org/pkg/os/exec/#Cmd).
 
+## Annotations
+
+Annotations are optional arbitrary non-identifying metadata that can be attached to containers.
+This information may be large, may be structured or unstructured.
+Annotations are key-value maps.
+
+```json
+"annotations": {
+	"key1" : "value1",
+	"key2" : "value2"
+}
+```
+
 [uts-namespace]: http://man7.org/linux/man-pages/man7/namespaces.7.html


mrunalp:
Do we need Labels and Annotations in both spec and state? I thought Labels were more of a runtime grouping mechanism so belong to the state and Annotations static and hence should be in the spec config.

vbatts:
@mrunalp the labels are useful in both places. And as far as annotations, there may be runtimes that need to stash runtime specific metadata, and it would provide a clean place to do so.

wking:
It looks like the runtime will be ignoring ‘annotations’ [1].  Why not
follow the approach #188 proposes and use a single, opaque field
(‘interface{}’ in Go and ‘JSON’ or ‘object’ in the Markdown [2])?
Then the higher-level tools that will be consuming this information
can specify their own schema, Go types, etc.  I don't see a benefit to
requiring a particular schema for data that the runtime will ignore,
and requring a map[string]string seems like it conflicts with “This
information may be … structured” [1].

I'm -1 on labels (which do impact the runtime interface) until we have
clarity on that interface.  For example, the ‘list’ command used as an
example [3] is not one of our current operations [4].  If we want to
add label-based identifiers for stop and exec, I'd rather have links
from those actions to the labels documentation.  And “This operation
MUST generate an error if it is not provided the container ID” [5]
would probably need to be softened.

[1]: https://github.com/vishh/specs/commit/ce688a750f5de242a2d0572e860876b0e42612d7#diff-c9c91c29b41257aea3a3403cc606ad99R298
[2]: https://github.com/opencontainers/specs/pull/188/files#r45020143
[3]: https://github.com/vishh/specs/commit/ce688a750f5de242a2d0572e860876b0e42612d7#diff-c9c91c29b41257aea3a3403cc606ad99R293
[4]: https://github.com/opencontainers/specs/blob/a1e32a8ead2ba57adce3e36e956b4dc32c1b85c4/runtime.md#operations
[5]: https://github.com/opencontainers/specs/blob/a1e32a8ead2ba57adce3e36e956b4dc32c1b85c4/runtime.md#stop

vishh:
How much do we care about the `state` from a Spec perspective?

wking:
On Fri, Mar 04, 2016 at 01:54:50PM -0800, Vish Kannan wrote:
> How much do we care about the `state` from a Spec perspective?

I care because it is the key that allows container access without
going through an established runtime operation [1].  If you need to
adjust something that doesn't have an OCI knob, attach some monitoring
that doesn't have an OCI integration, etc., the state gives you the
platform primitives you need to interact with the container directly.

[1]: https://github.com/opencontainers/specs/blob/a1e32a8ead2ba57adce3e36e956b4dc32c1b85c4/runtime.md#operations

vishh:
@philips @crosbymichael @mrunalp @vbatts: How do we want to move forward on this PR?

mrunalp:
@vishh Could you give use cases for each as we had discussed in the call? I think that will help move this.

crosbymichael:
I'm still unclear why we need to.  its like 2x the complexity and ppl could be confused when and why to use each one.

wking:
On Tue, Mar 08, 2016 at 10:42:38AM -0800, Michael Crosby wrote:
> I'm still unclear why we need to.  its like 2x the complexity and
> ppl could be confused when and why to use each one.

One is for opaque-to-the-runtime data, and the other is for grouping
(as defined in this PR).  I think that's a clear enough distinction.
I'd prefer folks used other files for opaque-to-the-runtime data, but
I'm ok with a blessed field in config.json for opaque-to-the-runtime
data [1].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/p_u2xps6jvo/bcMI6hm1DwAJ
     Subject: Re: Labels and extension meta data in containers #108
     Date: Wed, 30 Dec 2015 20:54:37 -0800
     Message-ID: <20151231045437.GA6362@odin.tremily.us>

vishh:
Updates PR to not include Labels. Removed Annotations from `state.go`

crosbymichael:
LGTM

mrunalp:
LGTM

config.go
```diff
@@ -18,6 +18,8 @@ type Spec struct {
 	Mounts []Mount `json:"mounts"`
 	// Hooks are the commands run at various lifecycle events of the container.
 	Hooks Hooks `json:"hooks"`
+	// Annotations is an unstructured key value map that may be set by external tools to store and retrieve arbitrary metadata.
+	Annotations map[string]string `json:"annotations,omitempty"`
```
wking: Are we sure we don't [want an opaque type here][1]?  Or will we just revisit that later when somebody needs something that doesn't fit into a `map[string]string`?

[1]: https://github.com/opencontainers/specs/pull/331#issuecomment-192477931

config.md
```diff
@@ -231,4 +231,17 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 `args` and `env` are optional.
 The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://golang.org/pkg/os/exec/#Cmd).
 
+## Annotations
+
+Annotations are optional arbitrary non-identifying metadata that can be attached to containers.
```
wking: “arbitrary” and ”non-identifying” seem to conflict.  I'd expect we want to drop “non-identifying” and just missed it in the reroll that dropped labels.

PR #340, opened 2016-03-10
add omitempty to 'Devices

liangchenye:
1. Add 'omitempty' to 'Device'.
2. remove the 'seccomp' and 'device' from the example since they are not necessary
3. change the device type description `c` to "c" since it became a string.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -128,9 +128,9 @@ type Linux struct {
 	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
 	CgroupsPath *string `json:"cgroupsPath,omitempty"`
 	// Namespaces contains the namespaces that are created and/or joined by the container
-	Namespaces []Namespace `json:"namespaces"`
+	Namespaces []Namespace `json:"namespaces,omitempty"`
 	// Devices are a list of device nodes that are created for the container
-	Devices []Device `json:"devices"`
+	Devices []Device `json:"devices,omitempty"`
 	// Seccomp specifies the seccomp security settings for the container.
 	Seccomp *Seccomp `json:"seccomp,omitempty"`
 	// RootfsPropagation is the rootfs mount propagation mode for the container.
--- a/config.md
+++ b/config.md
@@ -404,12 +404,7 @@ Here is a full example `config.json` for reference.
             {
                 "type": "mount"
             }
-        ],
-        "devices": null,
-        "seccomp": {
-            "defaultAction": "",
-            "architectures": null
-        }
+        ]
     }
 }
 ```


mrunalp:
LGTM

specs-go/config.go
```diff
@@ -130,7 +130,7 @@ type Linux struct {
 	// Namespaces contains the namespaces that are created and/or joined by the container
 	Namespaces []Namespace `json:"namespaces"`
```
vishh: Aren't namespaces optional as well?

wking:
On Thu, Mar 10, 2016 at 10:25:56AM -0800, Michael Crosby wrote:
> Why do we have to remove fields from the EXAMPLES?  They are
> examples, if you don't have fields in examples where are people
> going to learn about them?

I think we should either remove optional fields or fill them in with
meaningful values [1].  Discovery should be through the parent field's
documentation [2,3].

[1]: https://github.com/opencontainers/specs/pull/276#discussion_r55407392
[2]: https://github.com/opencontainers/specs/pull/325#issuecomment-187834414

liangchenye:
Same with Wking, I think we don't need the meanless unnecessary value.

config-linux.md
```diff
@@ -99,7 +99,7 @@ The runtime may supply them however it likes (with [mknod][mknod.2], by bind mou
 
 The following parameters can be specified:
 
-* **`type`** *(string, required)* - type of device: `c`, `b`, `u` or `p`.
+* **`type`** *(string, required)* - type of device: "c", "b", "u" or "p".
   More info in [mknod(1)][mknod.1].
 * **`path`** *(string, required)* - full path to device inside container.
 * **`major, minor`** *(int64, required unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
```
hqhq: You might want to change `p` here and below in `Device whitelist` section as well. Anyway I don't see this necessary, because I don't think there is inkling for a specific type with backticks since it's just consistent usage for markdown. See discussion in https://github.com/opencontainers/specs/pull/323#issuecomment-187568649

liangchenye:
@hqhq yes, remove the unnecessary change.
@vishh  new update also adds omitempty to 'namespace'

config-linux.md
wking: On Thu, Mar 10, 2016 at 06:36:58PM -0800, Qiang Huang wrote:
> > @@ -99,7 +99,7 @@ The runtime may supply them however it likes (with [mknod][mknod.2], by bind mou
> >  
> >  The following parameters can be specified:
> >  
> > -* **`type`** *(string, required)* - type of device: `c`, `b`, `u` or `p`.
> > +* **`type`** *(string, required)* - type of device: "c", "b", "u" or "p".
> >    More info in [mknod(1)][mknod.1].
> >  * **`path`** *(string, required)* - full path to device inside container.
> >  * **`major, minor`** *(int64, required unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
>
> You might want to change `p` here and below in `Device whitelist`
> section as well. Anyway I don't see this necessary, because I don't
> think there is inkling for a specific type with backticks since it's
> just consistent usage for markdown. See discussion in
> https://github.com/opencontainers/specs/pull/323#issuecomment-187568649

I think the style we choose is less important thant that we pick
something and document it in [1] for consistency.  Having to think
about this whenever someone inserts a string literal in new docs is
wasted time ;).

[1]: https://github.com/opencontainers/specs/blob/master/style.md

vbatts:
LGTM

and while this removes the empty values, i'll fix via https://github.com/opencontainers/specs/issues/344

PR #341, opened 2016-03-10
Move rlimits to process

julz:
As discussed at the 9/3/16 weekly call and on IRC, this moves the rlimits struct on to process to match capabilities, apparmor, selinux etc which were moved in 5a8a779.

Also fixed a typo (`systemd` -> `systems`)

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -455,24 +455,6 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
    }
 ```
 
-## Rlimits
-
-rlimits allow setting resource limits.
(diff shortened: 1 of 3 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

crosbymichael:
We can update the formatting post merge

wking:
Through 9d9ed06d looks good to me, with a minor backtick nit [1].

I think summarizing the discussion for the PR or commit message would
be useful for folks who wonder about this later on, and am happy to
take a stab at writing that summary if folks want.

[1]: https://github.com/opencontainers/specs/pull/341/files#r55725249

PR #313, opened 2016-01-19
json-schema: initial pass at schema and validator

vbatts:
Conforming to https://tools.ietf.org/html/draft-zyp-json-schema-03
and http://json-schema.org/latest/json-schema-core.html

This work would ideally pair with separating the current reference source effort #276 

* [x] provide validation tool to validate a configuration file
* [x] full structures
* [x] define basic primitives
* [x] account for all the pointer usage
* [x] discuss versioned schema-$ver.json

The gist of the `config.json` being tested with (may change periodically): https://gist.github.com/vbatts/5ac2723c9598875ffb0e

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

PR #342, opened 2016-03-10
config: Make Markdown canonical

wking:
It's pretty clear that the Go bindings cannot be canonical on their
own, because they do not define limits (e.g. the 0 through 512 range
for `FileMode`).  The JSON Schema is closer, but still does not cover
everything (e.g. “a directory must exist at **`root.path`**”).  Both
the Go bindings and the JSON Schema could grow to cover the full spec
by adding that sort of thing to comments and descriptions, but that's
not how things seem to be working now.

Also:

* Update links to Go bindings after 7bf06d5 (source and schema:
  differentiate with examples, 2015-12-18, #276).
* Add a reference to the JSON Schema after cdcabde (schema: JSON
  Schema and validator for `config.json`, 2016-01-19, #313).

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,7 +1,7 @@
 # Container Configuration file
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
-For now the canonical schema is defined in [config.go](config.go) and [config_linux.go](config_linux.go), but this will be moved to a formal JSON schema over time.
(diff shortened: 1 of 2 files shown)


vbatts:
LGTM

mrunalp:
LGTM

PR #343, opened 2016-03-16
Improve json encoding performance for specs-go

crosbymichael:
By using ffjson for json encoding we can reduce the time for encoding
and decoding.

Without:

BenchmarkMarsalSpec-4     100000             18276 ns/op
BenchmarkUnmarshal-4       30000             55115 ns/op

With:

BenchmarkMarsalSpec-4     100000             13649 ns/op
BenchmarkUnmarshal-4       50000             24747 ns/op

This is a reduces time about 25% on marshal and 50% on unmarshal.

I think its worth it to use ffjson as there is really no changes other than needing to generate the code after a change and we get good benefits in return.


vbatts:
not that i'm disagreeing with it, but I we don't much do any marshaling in the source of this repo presently. Though perhaps it makes sense to be present in the same package?

crosbymichael:
@vbatts ya, its for consumers of the package that we provide this.  It really helps out alot.

jfrazelle:
imma use it :innocent: i just updated all my tools

vbatts:
Then ought we put a target in the `Makefile` or go generate in the other source?

crosbymichael:
@vbatts updated makefile

specs-go/serialize_test.go
```diff
@@ -0,0 +1,149 @@
...
+				Type: "uts",
+			},
+			{
+				Type: "mount",
+			},
+		},
+	},
+}
+
+func BenchmarkMarsalSpec(b *testing.B) {
```
runcom: typo s/Marsal/Marshal/

mrunalp:
LGTM

Makefile
```diff
@@ -56,6 +56,11 @@ test: .govet .golint .gitvalidation
 .gitvalidation:
 	git-validation -q -run DCO,short-subject -v -range $(EPOCH_TEST_COMMIT)..HEAD
 
+# `go get https://github.com/pquerna/ffjson`
```
vbatts: s!https://!!

vbatts:
LGTM beside silly nit

marcosnils:
@crosbymichael @vbatts time to move to easyjson?. https://github.com/mailru/easyjson

PR #346, opened 2016-03-16
Add timeout field to hooks

crosbymichael:
So that runtimes don't lock up forever waiting on a hook to return I
added a timeout field.  This allows the author to specify the timeout in
seconds that they want the hook to run before aborting the hook and
container execution.

I think the hook is the proper place to specify a timeouot like this
because any type of overall timeout in the runtime will probably not
work for all hooks and will not be flexable enough.

Let me know that you all think.  

### DIFF
--- a/config.md
+++ b/config.md
@@ -226,6 +226,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
         "poststart": [
             {
                 "path": "/usr/bin/notify-start"
+                "timeout": 5
             }
         ],
         "poststop": [
@@ -239,6 +240,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 
 `path` is required for a hook.
 `args` and `env` are optional.
+`timeout` is the number of seconds before aborting the hook.
 The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://golang.org/pkg/os/exec/#Cmd).
 
 ## Annotations
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -96,9 +96,10 @@ type Mount struct {
 
 // Hook specifies a command that is run at a particular event in the lifecycle of a container
 type Hook struct {
-	Path string   `json:"path"`
-	Args []string `json:"args,omitempty"`
-	Env  []string `json:"env,omitempty"`
+	Path    string   `json:"path"`
+	Args    []string `json:"args,omitempty"`
+	Env     []string `json:"env,omitempty"`
+	Timeout *int     `json:"timeout,omitempty"`
 }
 
 // Hooks for container setup and teardown


vishh:
This change makes sense. But adding a create step feels like a better solution :)

wking:
On Wed, Mar 16, 2016 at 04:54:22PM -0700, Michael Crosby wrote:
> I think the hook is the proper place to specify a timeouot like this
> because any type of overall timeout in the runtime will probably not
> work for all hooks and will not be flexable enough.

I'd rather folks used ‘timeout’ [1] or similar in the hook itself.

[1]: http://man7.org/linux/man-pages/man1/timeout.1.html

crosbymichael:
@wking i don't think that is a good general solution for everyone.  having this enforced by the runtime seems better and we don't have to rely on an external binary.

marcosnils:
Agree with @crosbymichael. It helps when developing a spec implementation for multiple platforms.

config.md
```diff
@@ -226,6 +226,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
         "poststart": [
             {
                 "path": "/usr/bin/notify-start"
```
vbatts: adding that line will require a comma on the end of this line

vbatts:
While we have hooks, this LGTM,
though like @vishh said, if things were broken into create sandbox and start then it could be left up to the runtime

wking:
On Wed, Mar 16, 2016 at 05:02:38PM -0700, Michael Crosby wrote:
> @wking i don't think that is a good general solution for everyone.
> having this enforced by the runtime seems better and we don't have
> to rely on an external binary.

I don't think this argument scales particularly well.  The GNU Core
Utilities provide timeout(1) and a few other wrappers [1].  I don't
have any systems without coreutils installed, and a timeout wrapper
seems like a lightweight dependency on systems that don't have
coreutils.  What happens when somebody wants to set nice on a hook?
Do we add a nice setting, or point them at [2,3]?

And hooks are setup / audited by the runtime-caller right [4]?  They
should know if they have a local timeout(1) or equivalent, and know
how to get it installed if they don't have it.

[1]: http://www.gnu.org/software/coreutils/manual/html_node/Modified-command-invocation.html#Modified-command-invocation
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/nice.html
[3]: http://www.gnu.org/software/coreutils/manual/html_node/nice-invocation.html#nice-invocation
[4]: https://github.com/opencontainers/specs/pull/284#issuecomment-169137733

wking:
On Wed, Mar 16, 2016 at 06:42:33PM -0700, Vincent Batts wrote:
> though like @vishh said, if things were broken into create sandbox
> and start then it could be left up to the runtime

Breaking create / (exec|start) doesn't get you out of hooks entirely,
since anything that depends on the container PID will still need to
happen in an exec|start hook [1,2].

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-01-20.log.html#t2016-01-20T18:24:40
[2]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-02-25.log.html#t2016-02-25T00:34:15

vishh:
This PR is not the place to talk about hooks. @wking it will be helpful if
you can enumerate the use cases for hooks that cannot be satisfied by the
split.

hqhq:
It makes sense for me, LGTM.

mrunalp:
LGTM

PR #347, opened 2016-03-17
config: clarify the uname mapping

### LINKED ISSUE #38: linux: username: how to figure out uid reliably

vbatts:
Closes #38

### DIFF
--- a/config.md
+++ b/config.md
@@ -104,12 +104,19 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
 * **`noNewPrivileges`** (bool, optional) setting `noNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
 [The kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call.
 
+### User
+
 The user for the process is a platform-specific structure that allows specific control over which user the process runs as.
+
+#### Linux User
+
 For Linux-based systems the user structure has the following fields:
 
 * **`uid`** (int, required) specifies the user id.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -104,13 +104,20 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
...
+
+#### Linux User
+
 For Linux-based systems the user structure has the following fields:
 
 * **`uid`** (int, required) specifies the user id.
 * **`gid`** (int, required) specifies the group id.
 * **`additionalGids`** (array of ints, optional) specifies additional group ids to be added to the process.
 
+_Note: symbolic name for uid and gid, such as uname and gname respectively, are left to upper levels to derive (i.e. `/etc/passwd` parsing, NSS, etc)_
```
wking: Maybe:

```markdown
_Note: mapping from symbolic names (user and group names) to numeric IDs, for example by parsing `/etc/passwd` or using [getent][getent.1], is left to higher-level tools._

[getent.1]: http://www.man7.org/linux/man-pages/man1/getent.1.html
```

mrunalp:
LGTM

vishh:
LGTM

PR #349, opened 2016-03-17
Makefile: add target for tools

vbatts:
As we have several tools used in the Makefile, might as well make them
easier to install.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,30 +17,30 @@ DOC_FILES := \
 	glossary.md
 EPOCH_TEST_COMMIT := 041eb73d2e0391463894c04c8ac938036143eba3
 
-docs: pdf html
+default: docs
+
 .PHONY: docs
+docs: output/docs.pdf output/docs.html
 
-pdf:
-	@mkdir -p output/ && \
+output/docs.pdf:
+	mkdir -p output/ && \
 	$(DOCKER) run \
 	-it \
 	--rm \
 	-v $(shell pwd)/:/input/:ro \
 	-v $(shell pwd)/output/:/output/ \
 	-u $(shell id -u) \
-	vbatts/pandoc -f markdown_github -t latex -o /output/docs.pdf $(patsubst %,/input/%,$(DOC_FILES)) && \
-	ls -sh $(shell readlink -f output/docs.pdf)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -15,8 +15,10 @@ DOC_FILES := \
 	glossary.md
 EPOCH_TEST_COMMIT := 041eb73d2e0391463894c04c8ac938036143eba3
 
-docs: pdf html
+default: docs
+
 .PHONY: docs
+docs: pdf html
```
wking: Since you're touching these targets, maybe adjust the `pdf` and `html` targets to actual filenames (`output/docs.pdf`?) or mark them phony as well.  If you go the filename route, you'll want to specify prerequisites for both targets.

vbatts: fair

vbatts:
Updated, fixed and rebased. PTAL.

Makefile
```diff
@@ -51,19 +51,41 @@ test: .govet .golint .gitvalidation
...
 .gitvalidation:
+	@which git-validation > /dev/null 2>/dev/null || (echo "ERROR: git-validation not found. Consider 'make install.tools' target" && false)
 	git-validation -q -run DCO,short-subject -v -range $(EPOCH_TEST_COMMIT)..HEAD
 
+.PHONY: install.tools
+install.tools: .install.golint .install.govet .install.gitvalidation
+
+# golint does not even build for <go1.5
+.install.golint:
+ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
```
wking: We can probably just remove the `ALLOWED_GO_VERSION` stuff, since c506ce6f (Fix the build by getting rid of go get for vet, 2016-04-07, #372) completely dropped Go < 1.5.  We can always drag it out of the PR history if we need that sort of logic to partially support Go 1.5 (or whatever) in the future.

vbatts: it's being used on the install of go vet, and someone's local go version
may vary.

wking: On Tue, Apr 12, 2016 at 06:49:07AM -0700, Vincent Batts wrote:
> it's being used on the install of go vet, and someone's local go
> version may vary.

Ah, good point.  I guess folks might be running these checks outside
of Travis ;).

crosbymichael:
LGTM

PR #352, opened 2016-03-19
travis: add go1.6 and work around golint

### LINKED ISSUE #350: golint now requires go1.5
we use golint for basic regression checks. As of https://github.com/golang/lint/commit/a113636da96a2efee02051f523c74104dbe4d202 it now fails for earlier go versions.
Example: https://travis-ci.org/opencontainers/specs/jobs/116826264

vbatts:
closes #350

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: go
 go:
+  - 1.6
   - 1.5.3
   - 1.4.3
   - 1.3.3
@@ -8,13 +9,13 @@ sudo: false
 
 before_install:
   - go get golang.org/x/tools/cmd/vet
-  - go get github.com/golang/lint/golint
+  - go version | (grep -q 'go1.[56]' || exit 0 && go get -u github.com/golang/lint/golint )
   - go get github.com/vbatts/git-validation
 
 install: true
 
 script:
   - go vet -x ./...
-  - $HOME/gopath/bin/golint ./...
+  - make .golint
   - $HOME/gopath/bin/git-validation -run DCO,short-subject -v -range ${TRAVIS_COMMIT_RANGE}
   
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,11 @@ html:
 	vbatts/pandoc -f markdown_github -t html5 -o /output/docs.html $(patsubst %,/input/%,$(DOC_FILES)) && \
 	ls -sh $(shell readlink -f output/docs.html)
 
+
+HOST_GOLANG_VERSION	= $(shell go version | cut -d ' ' -f3 | cut -c 3-)
+# this variable is used like a function. First arg is the minimum version, Second arg is the version to be checked.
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -8,13 +9,13 @@ sudo: false
 
 before_install:
   - go get golang.org/x/tools/cmd/vet
-  - go get github.com/golang/lint/golint
+  - go version | grep -qE '(go1.5|go1.6)' && go get github.com/golang/lint/golint
```
wking: `grep -q 'go1.[56]'` should be sufficient.

vbatts: yeah, but now i'm trying to figure out the proper subterminal return codes,
because this is now failing as grep is not finding the version

marcosnils: @vbatts what about doing something like what I did for containerd to check for a minimum version?: https://github.com/docker/containerd/pull/132/commits/c6680da78b7777fc1c814c5f35ee166a83e5408a

vbatts:
my goodness. Travis is having API threshold issues, but this finally passes.

`golint` is only done for go1.5 and go1.6.

PTAL

wking:
cb02ebd looks good to me.

marcosnils:
@wking @vbatts this won't work with any version greater that 1.6. Have you checked my comment? 

https://github.com/opencontainers/specs/pull/352#discussion_r56752921 ?

wking:
On Mon, Mar 21, 2016 at 10:42:32AM -0700, Marcos Nils wrote:
> @wking @vbatts this won't work with any version greater that
> 1.6. Have you checked my comment?
> 
> https://github.com/opencontainers/specs/pull/352#discussion_r56752921 ?

This approach looks fine to me to, although bumping @vbatts' approach
as we add/remove Go implementations from the Travis config doesn't
seem like it would be too hard.

vbatts:
@marcosnils I looked, but it's not like the specs code doesn't support prior golang versions, but that we can't use the `golint` tool.

marcosnils:
@vbatts exactly. By the approach I'm proposing you wouldn't have to change the regex each time a new golang version is added to the travis file. You just specify the minimum version which `golint` needs and that's it. 

It might be a good idea also to add the `tip` version to the travis file so we can know beforehand if any changed made into Go might break something in the specs.

vbatts:
k. I've updated it, but the travis `before_install` step would end up overlapping a bit with this PR https://github.com/opencontainers/specs/pull/349/files#diff-b67911656ef5d18c4ae36cb6741b7965R70

I can consolidate one or the other, for it be be cleaner and all done in the `Makefile`

marcosnils:
@vbatts +1 to consolidate.

mrunalp:
@vbatts Is the build expected to pass here?

vbatts:
@mrunalp all green

crosbymichael:
LGTM

mrunalp:
LGTM

PR #355, opened 2016-03-22
fix go get path

laijs:
go get https://github.com/pquerna/ffjson
package https:/github.com/pquerna/ffjson: "https://" not allowed in
import path

so we remove the "https://"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ test: .govet .golint .gitvalidation
 .gitvalidation:
 	git-validation -q -run DCO,short-subject -v -range $(EPOCH_TEST_COMMIT)..HEAD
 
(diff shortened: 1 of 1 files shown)


wking:
On Mon, Mar 21, 2016 at 07:25:25PM -0700, Lai Jiangshan wrote:
> go get https://github.com/pquerna/ffjson
> package https:/github.com/pquerna/ffjson: "https://" not allowed in
> import path

This is also covered by a79fdefce (Makefile: add target for tools,
2016-03-17, #349), where it's a drive-by.  I expect it doesn't matter
particularly much which PR lands the fix, but the sooner we land one
of them the less likely we'll be to get another duplicate ;).

laijs:
thanks, close it

PR #361, opened 2016-03-30
Fix blkio major/minor json marshaling

crosbymichael:
Currently ffjson does not support embedded structs so we can just remove
that.  Its only saving us ~2 lines of code to have this embeded struct
so we can just remove it.

Signed-off-by: Michael Crosby <crosbymichael@gmail.com>

PR #364, opened 2016-04-01
Add masked and readonly paths

### LINKED ISSUE #320: Add support for MaskPaths and ReadonlyPaths
It is currently not possible for a user to specify `libcontainer` [MaskPaths](https://github.com/opencontainers/runc/blob/d66c9632bfc945547f7df4390bff35681023a24b/libcontainer/configs/config.go#L157) and [ReadonlyPaths](https://github.com/opencontainers/runc/blob/d66c9632bfc945547f7df4390bff35681023a24b/libcontainer/configs/config.go#L161) fields content through the bundle `json` configuration.

They should be added to the spec.

crosbymichael:
Fixes #320

This adds the maskedPaths and readonlyPaths fields to the spec so that
proper masking and setting of files in /proc can be configured.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -443,7 +443,7 @@ The following parameters can be specified to setup the controller:
 
 ## Sysctl
 
-sysctl allows kernel parameters to be modified at runtime for the container.
+`sysctl` allows kernel parameters to be modified at runtime for the container.
 For more information, see [the man page](http://man7.org/linux/man-pages/man8/sysctl.8.html)
 
 ###### Example
@@ -511,7 +511,7 @@ Operator Constants:
 
 ## Rootfs Mount Propagation
 
-rootfsPropagation sets the rootfs's mount propagation.
+`rootfsPropagation` sets the rootfs's mount propagation.
 Its value is either slave, private, or shared.
 [The kernel doc](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt) has more information about mount propagation.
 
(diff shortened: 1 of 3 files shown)


specs-go/config.go
```diff
@@ -136,6 +136,10 @@ type Linux struct {
 	Seccomp *Seccomp `json:"seccomp,omitempty"`
 	// RootfsPropagation is the rootfs mount propagation mode for the container.
 	RootfsPropagation string `json:"rootfsPropagation,omitempty"`
+	// MaskedPaths masks over the provided paths inside the container.
+	MaskedPaths []string `json:"maskedPaths"`
```
vishh: nit: Add `omitempty` here and below

vishh:
LGTM

config-linux.md
```diff
@@ -521,6 +521,30 @@ Its value is either slave, private, or shared.
...
+
+```json
+    "maskedPaths": [
+        "/proc/kcore"
+    ]
+```
+
+## Readonly Paths
+
+readonlyPaths will set the provieded paths as readonly inside the container.
```
wking: “provieded” → “provided”.  And probably **`readonlyPaths`** (and similar for your earlier **`maskedPaths`** docs) to match the rest of the spec.

wking:
On Thu, Mar 31, 2016 at 05:15:51PM -0700, Michael Crosby wrote:
> This adds the maskedPaths and readonlyPaths fields to the spec so
> that proper masking and setting of files in /proc can be configured.

I'm still not sure we need these [1], but if they're going in, should
they also come with JSON Schema entries?  Or is the policy there
“@vbatts does his best to keep them up to date on his own” [2] ;)?

[1]: https://github.com/opencontainers/runtime-spec/issues/320#issuecomment-204160097
[2]: https://github.com/opencontainers/runtime-spec/pull/361#issuecomment-203674306

crosbymichael:
@vbatts i tried to update the scheme stuff, let me know if I did it wrong

mrunalp:
The schema changes look good. (I also tested them using the validate tool).
LGTM

PR #368, opened 2016-04-06
[Weekly call] Call is for all OCI projects and TZ

RobDolinMS:
Per discussion on today's call, updates to indicate the call is for all OCI projects (rather than having multiple calls.)

Also updates time zone description from "PST" to "USA Pacific" since this time zone switches between PST and PDT.  

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ When in doubt, start on the [mailing-list](#mailing-list).
 
 ## Weekly Call
 
-The contributors and maintainers of the project have a weekly meeting Wednesdays at 10:00 AM PST.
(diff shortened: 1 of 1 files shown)


wking:
2cd54fb5 looks good to me, although it's missing a sign-off [1].

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/121217027#L181

RobDolinMS:
Thanks @wking.  'Just made the edit via web and forgot to add sign-off.  Have now added to my initial comment.

vbatts:
carried to #378

PR #369, opened 2016-04-06
style: Move one-sentence-per-line rule over from the README

wking:
This is style information, even though it's about the spec and not
about the behavior covered by the spec.  The Go-pointer rule is also
about more peripheral stuff though, and the README has a lot of stuff
in it, so it seems like a better fit after the move.


vbatts:
LGTM

mrunalp:
LGTM

PR #348, opened 2016-03-17
Anchor examples

vbatts:
This cleans up the anchoring of examples and adds to the style and conventions of such.

PR #370, opened 2016-04-06
Json schema and examples

### LINKED ISSUE #344: Build out example JSON
There json examples, ought to have more full example values.

vbatts:
Fixes #344

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -241,7 +241,7 @@ For more information on how these two settings work together, see [the memory cg
 ###### Example
 
 ```json
-    "oomScoreAdj": 0
+    "oomScoreAdj": 100
 ```
 
 #### Memory
@@ -251,9 +251,9 @@ For more information, see [the memory cgroup man page][cgroup-v1-memory].
 
 The following parameters can be specified to setup the controller:
 
-* **`limit`** *(uint64, optional)* - sets limit of memory usage
+* **`limit`** *(uint64, optional)* - sets limit of memory usage in bytes
 
-* **`reservation`** *(uint64, optional)* - sets soft limit of memory usage
+* **`reservation`** *(uint64, optional)* - sets soft limit of memory usage in bytes
 
 * **`swap`** *(uint64, optional)* - sets limit of memory+Swap usage
 
@@ -267,9 +267,9 @@ The following parameters can be specified to setup the controller:
 
 ```json
     "memory": {
-        "limit": 0,
-        "reservation": 0,
-        "swap": 0,
+        "limit": 536870912,
+        "reservation": 536870912,
+        "swap": 536870912,
         "kernel": 0,
         "kernelTCP": 0,
         "swappiness": 0
@@ -301,13 +301,13 @@ The following parameters can be specified to setup the controller:
 
 ```json
     "cpu": {
-        "shares": 0,
-        "quota": 0,
-        "period": 0,
-        "realtimeRuntime": 0,
-        "realtimePeriod": 0,
-        "cpus": "",
-        "mems": ""
+        "shares": 1024,
+        "quota": 1000000,
+        "period": 500000,
+        "realtimeRuntime": 950000,
+        "realtimePeriod": 1000000,
+        "cpus": "2-3",
+        "mems": "0-7"
     }
 ```
 
@@ -337,8 +337,8 @@ The following parameters can be specified to setup the controller:
 
 ```json
     "blockIO": {
-        "blkioWeight": 0,
-        "blkioLeafWeight": 0,
+        "blkioWeight": 10,
+        "blkioLeafWeight": 10,
         "blkioWeightDevice": [
             {
                 "major": 8,
--- a/config.md
+++ b/config.md
@@ -232,7 +232,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
         ],
         "poststart": [
             {
-                "path": "/usr/bin/notify-start"
+                "path": "/usr/bin/notify-start",
                 "timeout": 5
             }
         ],
@@ -269,7 +269,7 @@ Here is a full example `config.json` for reference.
 
 ```json
 {
-    "ociVersion": "0.3.0",
+    "ociVersion": "0.5.0-dev",
     "platform": {
         "os": "linux",
         "arch": "amd64"
@@ -284,6 +284,20 @@ Here is a full example `config.json` for reference.
                 6
             ]
         },
+        "uidMappings": [
+            {
+                "hostID": 1000,
+                "containerID": 0,
+                "size": 32000
+            }
+        ],
+        "gidMappings": [
+            {
+                "hostID": 1000,
+                "containerID": 0,
+                "size": 32000
+            }
+        ],
         "args": [
             "sh"
         ],
@@ -299,13 +313,19 @@ Here is a full example `config.json` for reference.
         ],
         "rlimits": [
             {
+                "type": "RLIMIT_CORE",
+                "hard": 1024,
+                "soft": 1024
+            },
+            {
                 "type": "RLIMIT_NOFILE",
                 "hard": 1024,
                 "soft": 1024
             }
         ],
-        "apparmorProfile": "",
-        "selinuxLabel": ""
+        "apparmorProfile": "acme_secure_profile",
+        "selinuxLabel": "system_u:system_r:svirt_lxc_net_t:s0:c124,c675",
+        "noNewPrivileges": true
     },
     "root": {
         "path": "rootfs",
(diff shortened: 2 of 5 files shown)


config.md
```diff
@@ -284,6 +284,20 @@ Here is a full example `config.json` for reference.
                 6
             ]
         },
+        "uidMappings": [
+            {
+                "hostID": 1000,
+                "containerID": 0,
+                "size": 10
```
mrunalp: 32k will be a better example for the mappings.

config.md
```diff
@@ -390,20 +405,166 @@ Here is a full example `config.json` for reference.
...
+            "pids": {
+                "limit": 32771
+            },
+            "hugepageLimits": [
+                {
+                    "pageSize": "2MB",
+                    "limit": 9223372036854772000
+                }
+            ],
+            "oomScoreAdj": 0,
```
mrunalp: 100 here

config.md
```diff
@@ -390,20 +405,166 @@ Here is a full example `config.json` for reference.
...
+            "memory": {
+                "limit": 0,
```
mrunalp: I think non-zero limit would be a better example

config.md
```diff
@@ -269,7 +269,7 @@ Here is a full example `config.json` for reference.
 
 ```json
 {
-    "ociVersion": "0.3.0",
+    "ociVersion": "0.5.0-dev",
```
wking: Are we going to bump this for every release and post-release?

schema/defs-linux.json
```diff
@@ -93,7 +93,9 @@
             "type": "string"
         },
         "FileType": {
-            "type": "integer"
+            "description": "Type of a block or special character device",
+            "type": "string",
+            "pattern": "^[c|b|u|p]$"
```
wking: Is this JSON Schema following [the RFC draft 4][1]?  I don't see anything about `pattern` in there, but would expect a regexp to look like `^[cbup]$`.

[1]: https://tools.ietf.org/html/draft-zyp-json-schema-04

schema/schema.json
```diff
@@ -139,6 +139,31 @@
                 "noNewPrivileges": {
                     "id": "https://opencontainers.org/schema/bundle/process/linux/noNewPrivileges",
                     "type": "boolean"
+                },
+                "rlimits": {
+                    "id": "https://opencontainers.org/schema/bundle/linux/rlimits",
+                    "type": "array",
+                    "items": [
+                        {
```
wking: JSON Schema's `items` takes an object, not an array of objects.  The docs are pretty sparse here, but see [this example][1].

[1]: https://tools.ietf.org/html/draft-zyp-json-schema-04#section-7.2.3

wking: Better (but non-normative?) docs [here][1].

[1]: http://spacetelescope.github.io/understanding-json-schema/reference/array.html#items

wking: And those non-normative docs do have a [list of objects][1], but only for validating fixed-length arrays.  And we want to allow more than one entry in **`rlimits`**.

[1]: http://spacetelescope.github.io/understanding-json-schema/reference/array.html#tuple-validation

schema/defs-linux.json
vbatts: http://json-schema.org/latest/json-schema-validation.html#anchor33

schema/schema.json
vbatts: fair. though it is not strict. I can change it, but it works as is.

schema/defs-linux.json
wking: On Wed, Apr 06, 2016 at 07:38:15PM -0700, Vincent Batts wrote:
> > @@ -93,7 +93,9 @@
> >              "type": "string"
> >          },
> >          "FileType": {
> > -            "type": "integer"
> > +            "description": "Type of a block or special character device",
> > +            "type": "string",
> > +            "pattern": "^[c|b|u|p]$"
> 
> http://json-schema.org/latest/json-schema-validation.html#anchor33

Thanks :).  That references ECMA 262, but [1] is too dense for me.
[2] has:

  [xyz] [a-c] A character set. Matches any one of the enclosed
              characters. You can specify a range of characters by
              using a hyphen, but if the hyphen appears as the first
              or last character enclosed in the square brackets it is
              taken as a literal hyphen to be included in the
              character set as a normal character.

              For example, [abcd] is the same as [a-d]. They match the
              "b" in "brisket" and the "c" in "chop".

And in my browser console:

  >> /^[cbup]$/.test('c');
  ← true
  >> /^[cbup]$/.test('|');
  ← true
  >> /^[c|b|u|p]$/.test('|');
  ← true

So I'm even more sure that we don't want the pipes inside the
character set.

[1]: http://www.ecma-international.org/ecma-262/5.1/#sec-15.10.2.13
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#character-classes

vbatts: fair

vbatts:
updated. PTAL.

schema/schema.json
wking: On Wed, Apr 06, 2016 at 07:43:10PM -0700, Vincent Batts wrote:
> > @@ -139,6 +139,31 @@
> >                  "noNewPrivileges": {
> >                      "id": "https://opencontainers.org/schema/bundle/process/linux/noNewPrivileges",
> >                      "type": "boolean"
> > +                },
> > +                "rlimits": {
> > +                    "id": "https://opencontainers.org/schema/bundle/linux/rlimits",
> > +                    "type": "array",
> > +                    "items": [
> > +                        {
> 
> fair. though it is not strict. I can change it, but it works as is.

Does it work with zero or multiple entries as is?  I'd expect a schema
array with one object to only validate the first entry in the config
array.

wking:
On Thu, Apr 07, 2016 at 07:14:17AM -0700, Vincent Batts wrote:
> updated. PTAL.

It looks like fc6670d4 → 1e2a72d2 addresses all the points raised so
far, except for the version issue which we aren't sure how to handle
[1].

There are still some zero values that don't make sense (e.g. in
linux.resources.cpu.period, but [2] has “The minimum quota allowed for
the quota or period is 1ms.”).

[1]: https://github.com/opencontainers/runtime-spec/pull/370/files#r58783380
[2]: https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt

vbatts:
@wking fair. Though it also says that a -1 to quotas indicates no restriction, though we have it as a uint64. That does not line up. 

I've updated the example.

config.md
```diff
@@ -390,20 +405,166 @@ Here is a full example `config.json` for reference.
...
+                {
+                    "pageSize": "2MB",
+                    "limit": 9223372036854772000
+                }
+            ],
+            "oomScoreAdj": 100,
+            "memory": {
+                "limit": 536870912,
+                "reservation": 0,
+                "swap": 0,
```
mrunalp: Perhaps make the swap same as the limit?

vbatts: done

config.md
```diff
@@ -390,20 +405,166 @@ Here is a full example `config.json` for reference.
...
+            "memory": {
+                "limit": 536870912,
+                "reservation": 0,
+                "swap": 536870912,
+                "kernel": 0,
+                "kernelTCP": 0,
+                "swappiness": 0
+            },
+            "cpu": {
+                "shares": 0,
```
wking: To pick another suspicious example, it looks like 0 here is “never run this if anyone else a share value” ([docs][1] [here][2]).  Maybe follow the kernel docs and use 1024?

[1]: https://www.kernel.org/doc/Documentation/scheduler/sched-design-CFS.txt
[2]: https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-kernel-uids

config.md
```diff
@@ -390,20 +405,166 @@ Here is a full example `config.json` for reference.
...
+                "quota": 1000000,
+                "period": 500000,
+                "realtimeRuntime": 0,
+                "realtimePeriod": 0,
```
wking: And we probably want the same non-zero value we use for **`period`** here.

config.md
```diff
@@ -390,20 +405,166 @@ Here is a full example `config.json` for reference.
...
+                "swap": 536870912,
+                "kernel": 0,
+                "kernelTCP": 0,
+                "swappiness": 0
+            },
+            "cpu": {
+                "shares": 0,
+                "quota": 1000000,
+                "period": 500000,
+                "realtimeRuntime": 0,
```
wking: This looks like another meaningless value.  I'm not sure how it relates to the other settings, but zero probably means either "never run this" or "there is no limit on this cgroup's continuous CPU access”, neither of which sound like the goal of a user who explicitly specifies this setting.

mrunalp:
@vbatts Could you address the latest comments from @wking ?

config.md
vbatts: again, these can be `-1` in the kernel https://www.kernel.org/doc/Documentation/scheduler/sched-rt-group.txt
but in the specs-go/config.go they are `*uint64`. Not sure there is a direct mapping, but I'll set a value.

vbatts:
@mrunalp done.
PTAL.

config.md
wking: On Mon, Apr 11, 2016 at 03:20:10PM -0700, Vincent Batts wrote:
> > +            ],
> > +            "oomScoreAdj": 100,
> > +            "memory": {
> > +                "limit": 536870912,
> > +                "reservation": 0,
> > +                "swap": 536870912,
> > +                "kernel": 0,
> > +                "kernelTCP": 0,
> > +                "swappiness": 0
> > +            },
> > +            "cpu": {
> > +                "shares": 0,
> > +                "quota": 1000000,
> > +                "period": 500000,
> > +                "realtimeRuntime": 0,
> > +                "realtimePeriod": 0,
> 
> again, these can be `-1` in the kernel
> https://www.kernel.org/doc/Documentation/scheduler/sched-rt-group.txt
> but in the specs-go/config.go they are `*uint64`. Not sure there is
> a direct mapping, but I'll set a value.

Ah, that also has “A run time of -1 specifies runtime == period,
ie. no limit.”, so probably best to use half of the period or some
such.

And I agree that if -1 is valid in the kernel's filesystem API, we
should probably be using signed integers in Go and reflecting the -1
minimum in the JSON Schema.  But that doesn't mean 0 is a good example
value ;).

config-linux.md
```diff
@@ -267,9 +267,9 @@ The following parameters can be specified to setup the controller:
 
 ```json
     "memory": {
-        "limit": 0,
+        "limit": 536870912,
         "reservation": 0,
```
wking: This is probably not a good example of a soft-limit ;).

vbatts: http://vignette3.wikia.nocookie.net/starwars/images/2/21/Stahp.jpg/revision/latest?cb=20140401024542

wking:
a341686 → c1612af look good to me, although it's still pretty easy to
find apparently-meaningless example values [1].  I thought part of the
purpose of this PR was fixing these meaningless examples (see also
#348), but we can obviously push some of that to follow-up PRs /
issues if we don't want to check them now.

[1]: https://github.com/opencontainers/runtime-spec/pull/370/files#r59293784

mrunalp:
I think we can do cleanups in follow on PRs. This LGTM overall.

mrunalp:
@crosbymichael @vishh PTAL

config-linux.md
vbatts: now I have a question. @mrunalp or @crosbymichael, does this `reservation` correspond to `memory.soft_limit_in_bytes`? if so, that can take a string like `1G` or `256M`. So is this `uint64` value just the number of bytes for a soft limit?

vishh:
LGTM

config-linux.md
vishh: AFAIK, it is supposed to be in `bytes`.

vishh: If we were to be pedantic, I'd rename all the fields to include the units. For example, `reservation` becomes `reservationBytes`

mrunalp: It is in bytes.

vbatts: fixed. and add `in bytes` to the field docs

wking: On Mon, Apr 11, 2016 at 03:52:52PM -0700, Vish Kannan wrote:
> > @@ -267,9 +267,9 @@ The following parameters can be specified to setup the controller:
> >  
> >  ```json
> >      "memory": {
> > -        "limit": 0,
> > +        "limit": 536870912,
> >          "reservation": 0,
>
> If we were to be pedantic, I'd rename all the fields to include the
> units. For example, `reservation` becomes `reservationBytes`

Ugh :p.  I think that's what docs are for ;).

vishh: I disagree. API should be user-friendly and readable. Having to keep
referring to docs while programming against an API is not a good user
experience. Simple changes like including units in the field names will
save a lot of time in the long run.

wking: On Mon, Apr 11, 2016 at 04:08:14PM -0700, Vish Kannan wrote:
> > @@ -267,9 +267,9 @@ The following parameters can be specified to setup the controller:
> >  
> >  ```json
> >      "memory": {
> > -        "limit": 0,
> > +        "limit": 536870912,
> >          "reservation": 0,
>
> I disagree. API should be user-friendly and readable. Having to keep
> referring to docs while programming against an API is not a good
> user experience.

Some ideas don't pack down nicely into a JSON key, so I think *not*
referring to docs while programming against an API is a bad idea.  If
the only interesting thing to say is “this is in bytes”, then reading
the docs won't take very long ;).  That said, the kernel devs are with
you on this one (since they chose limit_in_bytes),

PR #371, opened 2016-04-06
config.md: Rename ociVersion -> ociRuntimeVersion

wking:
Now that we have [multiple][1] [OCI][2] specs.  [Reported][3] by
@jonboulle.

Generated with:

    $ sed -i 's/ociVersion/ociRuntimeVersion/g' $(git grep -l ociVersion)

[1]: https://github.com/opencontainers/image-spec
[2]: https://github.com/opencontainers/runtime-spec
[3]: https://github.com/opencontainers/image-spec/pull/6/files/bf5c3f6b30cf326ee3e64849c8eb4c953500dd25#r58699357

### DIFF
--- a/config.md
+++ b/config.md
@@ -10,7 +10,7 @@ Below is a detailed description of each field defined in the configuration forma
 
 ## Specification version
 
-* **`ociVersion`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OpenContainer specification with which the bundle complies.
+* **`ociRuntimeVersion`** (string, required) must be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the OpenContainer specification with which the bundle complies.
 The OpenContainer spec follows semantic versioning and retains forward and backward compatibility within major versions.
 For example, if an implementation is compliant with version 1.0.1 of the spec, it is compatible with the complete 1.x series.
 NOTE that there is no guarantee for forward or backward compatibility for version 0.x.
@@ -18,7 +18,7 @@ NOTE that there is no guarantee for forward or backward compatibility for versio
 *Example*
 
 ```json
-    "ociVersion": "0.1.0"
+    "ociRuntimeVersion": "0.1.0"
 ```
 
 ## Root Configuration
@@ -269,7 +269,7 @@ Here is a full example `config.json` for reference.
 
 ```json
 {
-    "ociVersion": "0.3.0",
+    "ociRuntimeVersion": "0.3.0",
     "platform": {
         "os": "linux",
         "arch": "amd64"
--- a/runtime.md
+++ b/runtime.md
@@ -9,7 +9,7 @@ Whether other entities using the same, or other, instance of the runtime can see
 
 The state of a container MUST include, at least, the following propeties:
 
-* **`ociVersion`**: (string) is the OCI specification version used when creating the container.
+* **`ociRuntimeVersion`**: (string) is the OCI specification version used when creating the container.
 * **`id`**: (string) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
@@ -22,7 +22,7 @@ This is provided so that consumers can find the container's configuration and ro
 When serialized in JSON, the format MUST adhere to the following pattern:
 ```json
 {
-    "ociVersion": "0.2.0",
+    "ociRuntimeVersion": "0.2.0",
     "id": "oci-container1",
     "pid": 4422,
     "bundlePath": "/containers/redis"
--- a/schema/schema.json
+++ b/schema/schema.json
@@ -4,9 +4,9 @@
     "id": "https://opencontainers.org/schema/bundle",
     "type": "object",
     "properties": {
-        "ociVersion": {
+        "ociRuntimeVersion": {
             "description": "The version of OpenContainer specification configuration complies with",
-            "id": "https://opencontainers.org/schema/bundle/ociVersion",
+            "id": "https://opencontainers.org/schema/bundle/ociRuntimeVersion",
             "type": "string"
         },
         "hooks": {
@@ -147,7 +147,7 @@
         }
     },
     "required": [
-        "ociVersion",
+        "ociRuntimeVersion",
         "platform",
         "process",
         "root",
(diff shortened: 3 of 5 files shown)


vbatts:
Hold on this.

wking:
There was some discussion on this proposal in today's meeting [1].
The main alternative proposal seemed to be image-spec-style media
types [2].  Looking over [3,4], the media-type approach looks similar
to JSON-LD's @context [5,6].  Using JSON-LD would certainly give us an
approach that has had some thought put into it, although I'm not
particularly familiar with it myself.  And it also makes life easier
for consumers, since they can notice and complain about an
unrecognized @context value, but would have a harder time tracking
version → ociVersion → ociRuntimeVersion → … without a human reading
the spec ChangeLog.  @context also gives us the multiple-schema
support I was stumbling towards with [7,8], and that gives bundle
authors a convenient way for to clarify the interpretation of whatever
they put in ‘annotations’ [9] (or anywhere else in their JSON).

I haven't been able to dig up anyone talking about putting JSON-LD and
SemVer together.  The Schema.org folks store frozen versions of their
schemas [10,11], but say “Note that schema.org release numbers are not
generally included when you use schema.org” [10,11].  There's also a
schemaVersion property, but it lists only ~10 domains using it [12].
And see [13] where Dan Brickley says:

  Semver.org is interesting but the terminology doesn't directly map
  from APIs to schemas.

I'm not convinced (the distinction between a schema and an API is not
clear to me).  But we could either write our own ‘semanticVersion’
property, or leave the SemVer-ness implicit in the @context.

All of these approaches (ociRuntimeVersion, mediaType, and @context)
avoid the current ociVersion ambiguity.  Is there a reason to prefer
mediaType over @context?

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-04-13-17.01.html
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-04-13-17.01.log.html#l-63
[3]: https://github.com/opencontainers/image-spec/blob/6a4fcccd11cc4ed7117854c671a3123e6506e0cd/manifest.md#media-types
[4]: https://github.com/opencontainers/image-spec/blob/6a4fcccd11cc4ed7117854c671a3123e6506e0cd/manifest.md#manifest-list-field-descriptions
[5]: https://www.w3.org/TR/json-ld/#the-context
[6]: https://www.w3.org/TR/json-ld/#advanced-context-usage
[7]: https://github.com/opencontainers/runtime-spec/issues/15
[8]: https://github.com/opencontainers/runtime-spec/issues/110#issuecomment-133587646
[9]: https://github.com/opencontainers/runtime-spec/pull/331
[10]: http://schema.org/version/2.1/
[11]: http://schema.org/version/2.2/
[12]: http://schema.org/schemaVersion
[13]: https://github.com/schemaorg/schemaorg/pull/117#issuecomment-55224693

wking:
“How are we going to version config.json?” came up again today [1].
The issue there revolved around hypothetically dropping support for a:

  "mac-address": "default"

on Solaris even though the ‘mac-address’ property remains in the spec
and the ‘default’ value was never mentioned (instead, the in-flight
Solaris docs refer to external docs [2]).

To me, versioning is about communicating the intended semantics for
the config settings (like JSON-LD's @context and @id properties).  So
if a value (like ‘default’) is meaningful in an earlier iteration of a
property, and not meaningful (or meaning a different thing) in later
iteration of that property, it's a SemVer major bump for that property
[3].  Runc has currently (permanently?) dropped versioning [4] to
avoid having to bump ociVersion to get an old-version bundle running
on a new-version runtime even though the changes between versions
didn't impact the subset of properties used in bundle's config.  And
there has been similar discussion around versioning image manifests
[5,6].

While getting an IRI for each property [7] may seem tedious, it's easy
to automate, and we already have steps in that direction in the JSON
Schema [8,9].  The JSON Schema ids are RFC 3986 URIs [9], while
JSON-LD @ids are RFC 3987 IRIs [7], but that distinction is probably
not a big deal for us.

Anyhow, I'm happy to work up a JSON-LD PR if folks feel like that has
a chance of being accepted ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/411/files#r61637534
[2]: https://github.com/opencontainers/runtime-spec/pull/411/files#r61636723
[3]: http://semver.org/
[4]: https://github.com/opencontainers/runc/pull/521
[5]: https://github.com/opencontainers/image-spec/pull/6/files/bf5c3f6b30cf326ee3e64849c8eb4c953500dd25#r58780265
[6]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-04-06.log.html#t2016-04-06T20:58:29
[7]: https://www.w3.org/TR/json-ld/#dfn-iri
[8]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/schema/schema.json#L4
[9]: http://tools.ietf.org/html/draft-zyp-json-schema-04#section-7.2.2

crosbymichael:
Yes, lets hold on doing this because it would be much better to keep the two in sync instead of having two different versions that interop with each other.

wking:
On Fri, Apr 29, 2016 at 03:03:09PM -0700, Michael Crosby wrote:
> Yes, lets hold on doing this because it would be much better to keep
> the two in sync instead of having two different versions that
> interop with each other.

The two what in sync?  Is that “we want similar versioning for the
image-spec and runtime-spec repositories”?  If so (and that makes
sense to me), I'm happy to spin off the JSON-LD ideas to dev@.
Although I think I have a clear enough idea of how they'll work to
submit a PR to anchor discussion.  In any case, clarity on what you
think the next steps are and when/where they will be taken would be
good ;).

wking:
On Fri, Apr 29, 2016 at 03:45:46PM -0700, W. Trevor King wrote:
> On Fri, Apr 29, 2016 at 03:03:09PM -0700, Michael Crosby wrote:
> > Yes, lets hold on doing this because it would be much better to
> > keep the two in sync instead of having two different versions that
> > interop with each other.
> 
> The two what in sync?  Is that “we want similar versioning for the
> image-spec and runtime-spec repositories”?  If so (and that makes
> sense to me)…

Ah, maybe you meant “we want both image-spec and runtime-spec to use
the same ociVersion identifier, and will lock their release cycles
together”.  That moves in a *less* granular direction, and doesn't
sound like a good plan to me (why would the runtime care if the image
format got bumped).  And it was sounding like locked release cycles
were not part of the plan [1] (which does make sense to me).

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-04-13-17.01.log.html#l-62

PR #377, opened 2016-04-11
config-linux.md: Reword kernelTCP docs and mention bytes

wking:
Avoid the dangling “using” from e9a6d948 (cgroup: Add support for
`memory.kmem.tcp.limit_in_bytes`, 2015-10-26, #235).  I've tried to
echo the [kernel docs][1] by mentioning buffer memory.  I'd personally
prefer linking to the kernel docs and mentioning
`memory.kmem.tcp.limit_in_bytes`, but that seemed like too big of a
break from the existing style for this commit.

[1]: https://kernel.org/doc/Documentation/cgroup-v1/memory.txt

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -259,7 +259,7 @@ The following parameters can be specified to setup the controller:
 
 * **`kernel`** *(uint64, optional)* - sets hard limit for kernel memory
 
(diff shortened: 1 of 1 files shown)


vishh:
LGTM

mrunalp:
LGTM

PR #378, opened 2016-04-12
README: clarify weekly call

vbatts:
Weekly call is for all OCI projects and per discussion on today's call,
updates to indicate the call is for all OCI projects (rather than having
multiple calls.)

And updates time zone description from PST/PDT to Pacific time

This is carried PR from #368

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ When in doubt, start on the [mailing-list](#mailing-list).
 
 ## Weekly Call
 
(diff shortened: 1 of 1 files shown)


wking:
Only change since #368's 2cd54fb is to the commit message, which now
has @RobDolinMS's Signed-off-by, so 524d101d looks great to me.

mrunalp:
LGTM

hqhq:
LGTM

PR #382, opened 2016-04-12
schema: reduce indentation to 2

vbatts:
With a few of these nested structures, the 4 space indent made the page
extend further than really needed. This is just a simple tidy for
viewing ease.

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -8,7 +8,7 @@ help:
 	@echo " * 'validate' - build the validation tool"
 
 fmt:
-	for i in *.json ; do jq --indent 4 -M . "$${i}" > xx && cat xx > "$${i}" && rm xx ; done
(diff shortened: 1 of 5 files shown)


hqhq:
Hmm, do we have to do this? I kind of prefer 4 spaces, just 2 cents.

vbatts:
need, no. It just seemed much. On even a wide screen, was difficult to have
a vertical-split that was usable.

wking:
On Thu, Apr 14, 2016 at 06:53:35PM -0700, Vincent Batts wrote:
> need, no. It just seemed much. On even a wide screen, was difficult
> to have a vertical-split that was usable.

Sounds like an argument for tabs to me.  Then everyone can configure
their editor to use their favorite depth ;).

PR #383, opened 2016-04-12
ReadMe: Add hyperlink to releases view

RobDolinMS:
Hopefully makes it easier for folks to find HTML and PDF versions of spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ Runtime developers can build runtime implementations that run OCI-compliant bund
 
 There is a loose [Road Map](./ROADMAP.md).
 During the `0.x` series of OCI releases we make no backwards compatibility guarantees and intend to break the schema during this series.
+Current and previous releases of the spec in HTML and PDF formats can be found in [Releases](./releases).
+Release notes summarizing key changes are also included here.
 
 # Contributing
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Runtime developers can build runtime implementations that run OCI-compliant bund
 There is a loose [Road Map](./ROADMAP.md).
 During the `0.x` series of OCI releases we make no backwards compatibility guarantees and intend to break the schema during this series.
 Current and previous releases of the spec in HTML and PDF formats can be found in [Releases](./releases).
-Release notes summarizing key changes are also included here.
+Release notes summarizing key changes are also included there.
 
 # Contributing
 


README.md
```diff
@@ -47,6 +47,8 @@ Runtime developers can build runtime implementations that run OCI-compliant bund
 
 There is a loose [Road Map](./ROADMAP.md).
 During the `0.x` series of OCI releases we make no backwards compatibility guarantees and intend to break the schema during this series.
+Current and previous releases of the spec in HTML and PDF formats can be found in [Releases](./releases).
+Release notes summarizing key changes are also included here.
```
wking: “here” → “`[here](ChangeLog)`”?  We don't have a change summary in the README anyway, and a repo-local link seems nicer than linking to GitHub's release page.

RobDolinMS: @wking Thanks.  I was thinking "also included here" referred to the previous location hyperlinked in the prior sentence.

wking: On Tue, Apr 26, 2016 at 05:27:56PM -0700, Rob Dolin (MSFT) wrote:
> > +Current and previous releases of the spec in HTML and PDF formats can be found in [Releases](./releases).
> > +Release notes summarizing key changes are also included here.
>
> @wking Thanks.  I was thinking "also included here" referred to the
> previous location hyperlinked in the prior sentence.

In that case I'd use “there”, but as I mentioned briefly in [1], I
like the repo-local link for the ChangeLog more than a releases link.
It's distributed with the repo, so it's under obvious version control
and the link works even when you or GitHub are offline.  Do you have
reasons for preferring the releases link to a ChangeLog link?

[1]: https://github.com/opencontainers/runtime-spec/pull/383/files/1fa3741d673086cf35d399ecc40492b79c39d57c#r59437515

RobDolinMS: I've updated "here" --> "there"
IMHO, I think the ChangeLog may be too much detail for folks just looking for a printable version of the spec.

wking: On Wed, Apr 27, 2016 at 04:14:58PM -0700, Rob Dolin (MSFT) wrote:
> > +Current and previous releases of the spec in HTML and PDF formats can be found in [Releases](./releases).
> > +Release notes summarizing key changes are also included here.
> 
> I've updated "here" --> "there"
> IMHO, I think the ChangeLog may be too much detail for folks just looking for a printable version of the spec.  

Right, which is why I like the “Current and previous releases…” line
just the way it is.  Folks interested in HTML and PDF formats *should*
be going to the GitHub release page.

I just think folks who are interested in release notes should be going
to the local ChangeLog (which seems to contain exactly the same
information that is included on the release page, compare [1,2]).

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/ChangeLog#L3-L32
[2]: https://github.com/opencontainers/runtime-spec/releases/tag/v0.5.0

crosbymichael:
Please reopen once you have squash these commits and fixed the DCO signoff.

Thanks!

PR #385, opened 2016-04-14
Fix example in schema README

hqhq:
We should only download dependence without installing,
otherwise we'll probably get error:
go install: no install location for directory /home/qhuang/specs/schema outside GOPATH
        For more details see: go help gopath


hqhq:
Append a patch to modify README to keep it consistent with Makefile.

vbatts:
LGTM

wking:
Through aaf05c4 looks good to me.

mrunalp:
LGTM

Issue #345, opened 2016-03-16
Exec semantics


vishh:
The current Spec attempts to express `exec` functionality by splitting the container into a sandbox and a process. This split fails to address some of the use cases like as follows - 
1. Imposing additional resource restrictions on the exec process.
2. Imposing alternate/additional security policy on the exec process.

It is proving to be difficult to represent all the use cases and also preserve the semantics around the sandbox and the process.
Have we considered representing `exec` as a separate container with a process? That separate container can share some parts of the existing container's sandbox.
IIRC, the Spec tried to represent `exec` as a separate container in the beginning, but I cannot recall when it diverged from the original plan.

cc @crosbymichael @mrunalp

wking:
On Wed, Mar 16, 2016 at 12:37:22PM -0700, Vish Kannan wrote:
> IIRC, the Spec tried to represent `exec` as a separate container in
> the beginning, but I cannot recall when it diverged from the
> original plan.

‘exec’ only landed three weeks ago with #225, so there's not a lot of
history in the spec itself.  There has been wording that looks sort of
like separate create / exec since 77d44b10 (Update runtime.md,
2015-06-16) [1], but that's from the ancient past before changes were
publically motivated and the wording went away with the initial
lifecycle docs in #231.

Public out-of-spec discussion around separating sandbox creation from
process execution goes back a lot further (e.g. see [2,3,4,5]).  I
think the main point in favor of distinguishing between sandboxes
(create) and processes (exec) is that it makes it easy to launch new
processes in an existing container [6].

> Have we considered representing `exec` as a separate container with
> a process?

I think the namespace creation and cgroup joining in the current spec
should allow you to:

1. Join an existing container (or not).
2. Create a new container (or not).
3. Execute a new process inside (1) and/or (2).

As far as cgroups are concerned, join-and-create-a-subgroup is easy
with cgroupsPath (just use ‘{existing-parent-path}/{new-child-path}’,
see #68).  Namespaces are a bit trickier, because of the very-narrow
visibility scoping in the state registry [7,8].  It's possible that:

  $ nsenter <parent-namespaces> funC start <container-id> .

would create a new container whose state was only accessible via:

  $ nsenter <parent-namespaces> funC state <container-id>

and not directly via:

  $ funC state <container-id>

An extra nsenter call doesn't seem like that big a deal though.

And this sort of nested-namespace approach will also make clarifying
PID namespaces more important [9].

[...]

PR #388, opened 2016-04-19
Remove exec from required runtime functionalities.

vishh:
Exec can be implemented as running additional containers.

For #345

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -91,42 +91,6 @@ Deleting a container MUST delete the associated namespaces and resources associa
 Once a container is deleted, its `id` MAY be used by subsequent containers.
 Attempting to stop a container that is not running MUST have no effect on the container and MUST generate an error.
 
-### Exec
-
-`exec <container-id> <path-to-json>`
-
-This operation MUST generate an error if it is not provided the container ID and a path to the JSON describing the process to start.
-The JSON describing the new process MUST adhere to the [Process configuration](config.md#process-configuration) definition.
-This operation MUST create a new process within the scope of the container.
-If the container is not running then this operation MUST have no effect on the container and MUST generate an error.
-Executing this operation multiple times MUST result in a new process each time.
-
-#### Example
-
-```json
-{
-    "terminal": true,
-    "user": {
-        "uid": 0,
-        "gid": 0,
-        "additionalGids": null
-    },
-    "args": [
-        "/bin/sleep",
-        "60"
-    ],
-    "env": [
-        "version=1.0"
-    ],
-    "cwd": "...",
-}
-```
-
-This specification does not mandate the name of this JSON file.
-See the specification of the `config.json` file for the definition of these fields.
-The stopping, or exiting, of these secondary process MUST have no effect on the state of the container.
-In other words, a container (and its PID 1 process) MUST NOT be stopped due to the exiting of a secondary process.
-
 ## Hooks
 
 Many of the operations specified in this specification have "hooks" that allow for additional actions to be taken before or after each operation.


vishh:
cc @mrunalp @crosbymichael

wking:
On Tue, Apr 19, 2016 at 10:53:58AM -0700, Vish Kannan wrote:
>   * Remove exec from required runtime functionalities. Exec can be

Probably want two newlines instead of the period to avoid the
long-subject error [1].

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/124259128#L181

wking:
Do we want to put suggestions about implementing exec in an
informative sections somewhere?  I'd guess the template is:

  {
    "ociVersion": … usual stuff…,
    "platform": … usual stuff…,
    "process": … what you want to exec…,
    "root": what to put here?
    "linux": {
      "cgroupsPath": … extracted from state JSON…,
      "namespaces": [
        {
          "type": "pid",
          "path": … extracted from state JSON…
        },
        …
      ]
    }
  }

The namespace paths and cgroupsPath would come from a procfs entry,
which you'd find by:

1. Get state JSON using the ‘state <container-id>’ operation.
2. Get the container PID from the state (and maybe the process
   namespace that PID is from [1]).
3. Find a procfs mount for the process namespace listed in the state
   JSON [1]).

The namespace paths would be from /proc/{pid}/ns/…, and the
cgroupsPath would be from /proc/self/cgroup (any particular cgroup in
that list?).

Any ideas about ‘root’?

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/ujtABQoCmgk
     Subject: Linux: Adding the PID namespace inode to state JSON
     Date: Wed, 30 Dec 2015 21:34:57 -0800
     Message-ID: <20151231053457.GG3680@odin.tremily.us>

crosbymichael:
LGTM after travis is happy

vishh:
@wking We need blessed cookbook for the Spec.

wking:
On Tue, Apr 19, 2016 at 03:20:51PM -0700, Vish Kannan wrote:
> @wking We need blessed cookbook for the Spec. 

Sounds good to me.  And that can certainly happen in a separate
thread/issue/PR if folks want.  But I think it makes sense to back up
the “can be implemented…” claim with explicit “… and here's how …”
directions in this PR.  There aren't many missing pieces in [1], so it
shouldn't take that much work to clear them up.

[1]: https://github.com/opencontainers/runtime-spec/pull/388#issuecomment-212049907

crosbymichael:
We can work on that stuff later, we just need to get the spec finalized and in good shape before we start talking about how ppl should implement something

wking:
On Tue, Apr 19, 2016 at 04:09:13PM -0700, Michael Crosby wrote:
> We can work on that stuff later, we just need to get the spec
> finalized and in good shape before we start talking about how ppl
> should implement something

Agreed, I just think “in good shape” includes exec-like behavior.  To
avoid reducing functionality with this PR, I think we need to also
fill in the gaps to make ‘exec’ emulation possible.

For example, making ‘root’ optional [1] would be one way to address a
question from [2].  If folks have alternative approaches for handling
that question (supported by the current spec?), that seems like the
sort of thing we want hashed out before we remove the current ‘exec’
feature.

     Subject: Dropping the rootfs requirement and restoring arbitrary bundle content
     Date: Wed, 26 Aug 2015 12:54:47 -0700
     Message-ID: <20150826195447.GX21585@odin.tremily.us>
[2]: https://github.com/opencontainers/runtime-spec/pull/388#issuecomment-212049907

mrunalp:
@wking I think we can handle examples/cookbook etc. after this PR. To answer your question about `root`, it isn't used in exec if there is a mount namespace.

mrunalp:
LGTM

PR #390, opened 2016-04-20
config: Move (u|g)idMappings from 'process' to 'linux'

wking:
To match [where they're defined in the JSON Schema][1].  The old
location is from d4e7326d (config: JSON examples, 2016-04-06, #370),
and seems to have been accidental.

[1]: https://github.com/opencontainers/runtime-spec/blob/0982071b288ddddc1ae84d21c4bd682c96942f5c/schema/schema-linux.json#L21-L48

### DIFF
--- a/config.md
+++ b/config.md
@@ -286,20 +286,6 @@ Here is a full example `config.json` for reference.
                 6
             ]
         },
-        "uidMappings": [
-            {
-                "hostID": 1000,
-                "containerID": 0,
-                "size": 32000
-            }
-        ],
-        "gidMappings": [
-            {
-                "hostID": 1000,
-                "containerID": 0,
-                "size": 32000
-            }
-        ],
         "args": [
             "sh"
         ],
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -463,6 +449,20 @@ Here is a full example `config.json` for reference.
                 "gid": 0
             }
         ],
+        "uidMappings": [
+            {
+                "hostID": 1000,
+                "containerID": 0,
+                "size": 32000
```
hqhq: 32000 looks like a non sense magic number, 32768 might be better, what do you think?

wking: On Wed, Apr 20, 2016 at 07:23:00PM -0700, Qiang Huang wrote:
> > @@ -463,6 +449,20 @@ Here is a full example `config.json` for reference.
> >                  "gid": 0
> >              }
> >          ],
> > +        "uidMappings": [
> > +            {
> > +                "hostID": 1000,
> > +                "containerID": 0,
> > +                "size": 32000
> 
> 32000 looks like a non sense magic number, 32768 might be better, what do you think?

This PR is just about moving the existing example to the right place,
not about adjusting the example content.

If you want to file a separate issue/PR adjusting the number, that
size is from #370 based on @mrunalp's “32k” suggestion [1].  I'm happy
with anything > 0 and still well short of overflowuid.

[1]: https://github.com/opencontainers/runtime-spec/pull/370#discussion_r58783031

hqhq: I'm OK if we do this in a separate PR, I think 32k is supposed to be 32768.

hqhq:
LGTM

crosbymichael:
LGTM

PR #393, opened 2016-04-21
Add support for Selinux mount context labels

mrunalp:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -545,6 +545,16 @@ Its value is either slave, private, or shared.
     ]
 ```
 
+## Mount Label
+
+`mountLabel` will set the Selinux context for the mounts in the container.
+
+###### Example
+
+```json
+    "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
+```
+
 [cgroup-v1]: https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
 [cgroup-v1-blkio]: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt
 [cgroup-v1-cpusets]: https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt
--- a/schema/schema-linux.json
+++ b/schema/schema-linux.json
@@ -330,6 +330,10 @@
             "readonlyPaths": {
                 "id": "https://opencontainers.org/schema/bundle/linux/readonlyPaths",
                 "$ref": "defs.json#/definitions/ArrayOfStrings"
+            },
+            "mountLabel": {
+                "id": "https://opencontainers.org/schema/bundle/linux/mountLabel",
+                "type": "string"
             }
         }
     }
--- a/config.md
+++ b/config.md
@@ -605,7 +605,8 @@ Here is a full example `config.json` for reference.
             "/proc/irq",
             "/proc/sys",
             "/proc/sysrq-trigger"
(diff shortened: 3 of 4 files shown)


mrunalp:
@crosbymichael @LK4D4 ping
cc: @rhatdan

crosbymichael:
Why does this not get placed in the Mount.Options array?

mrunalp:
@crosbymichael This is for convenience. We could pass it to individual mounts as well. (We had a MountLabel in libcontainer configuration).

mrunalp:
An argument in favor is that the value of the label is the same for all the mount points (that we need to label).

crosbymichael:
Could it every need to be different or will it **always** be the same?

mrunalp:
AFAIK it wouldn't be but would defer to @rhatdan on that.

config-linux.md
```diff
@@ -545,6 +545,18 @@ Its value is either slave, private, or shared.
...
 ```
 
+## Mount Label
+
+`mountLabel` will set the Selinux context for the mounts in the container.
+
+###### Example
+
+```json
+    "mountLabel": [
```
crosbymichael: This should not be an array `[]` right?

crosbymichael:
ok, looking at the existing code this looks good.  Only thing is that your json example has this as an array and not just a string

mrunalp:
@crosbymichael Fixed.

crosbymichael:
LGTM

rhatdan:
Right now we are still always planning on labeling content inside of the container with the same label.  
LGTM

vbatts:
@mrunalp would you mind adding it to the JSON schema too? otherwise I can do a follow-up PR.

vbatts:
LGTM based on your response ^^

mrunalp:
I'll add it.

mrunalp:
@vbatts Added to schema.

config-linux.md
```diff
@@ -545,6 +545,16 @@ Its value is either slave, private, or shared.
...
 ```
 
+## Mount Label
+
+`mountLabel` will set the Selinux context for the mounts in the container.
+
+###### Example
+
+```json
+    "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
```
vbatts: sorry. One more nit. Please add this line to the "complete json" example in the config.md.

schema/schema-linux.json
```diff
@@ -330,6 +330,10 @@
             "readonlyPaths": {
                 "id": "https://opencontainers.org/schema/bundle/linux/readonlyPaths",
                 "$ref": "defs.json#/definitions/ArrayOfStrings"
+            },
+            "mountLabel": {
+                "id": "https://opencontainers.org/schema/bundle/linux/mountLabel",
+                "type": "string"
```
vbatts: cool

config-linux.md
mrunalp: Good catch. Fixed.

config-linux.md
```diff
@@ -545,6 +545,16 @@ Its value is either slave, private, or shared.
     ]
 ```
 
+## Mount Label
+
+`mountLabel` will set the Selinux context for the mounts in the container.
```
wking: Do we want cross-links between this and config.md's **`process.selinuxLabel`**?  I expect folks interested in one would also be interested in the other.

vbatts:
LGTM

PR #394, opened 2016-04-22
Remove requirement for rootfs path to be relative

### LINKED ISSUE #389: Why does rootfs path have to be relative?
Ideally it is good to have the rootfs inside the bundle but I don't think it should be required.  When dealing with overlay drivers or things like btrfs/zfs it would be better to not required extra bind mounts just to put the rootfs inside the bundle.  Can we look at changing this requirement?

mlaventure:
Closes #389

Signed-off-by: Kenfe-Mickael Laventure <mickael.laventure@gmail.com>

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -8,10 +8,10 @@ See also [OS X application bundles](http://en.wikipedia.org/wiki/Bundle_%28OS_X%
 The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local file system so that it can be consumed by a compliant runtime.
 
 A Standard Container bundle contains all the information needed to load and run a container.
-This includes the following artifacts which MUST all reside in the same directory on the local filesystem:
+This MUST include the following artifacts:
 
 1. `config.json` : contains configuration data.
-This REQUIRED file, which MUST be named `config.json`.
+This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
 When the bundle is packaged up for distribution, this file MUST be included.
 See [`config.json`](config.md) for more details.
 
--- a/config.md
+++ b/config.md
@@ -25,7 +25,7 @@ NOTE that there is no guarantee for forward or backward compatibility for versio
 
 Each container has exactly one *root filesystem*, specified in the *root* object:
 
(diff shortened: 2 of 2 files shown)


crosbymichael:
LGTM

These lines where added to the spec from the initial import but there is no real technical reason why this needs to be included in the directory when we have the information about where its located.  

The initial spec also said that all "bind" mounts must be included in the bundle and be relative but we also learned that this was too much of a burden for no real technical gain.

bundle.md
```diff
@@ -8,7 +8,7 @@ See also [OS X application bundles](http://en.wikipedia.org/wiki/Bundle_%28OS_X%
 The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local file system so that it can be consumed by a compliant runtime.
 
 A Standard Container bundle contains all the information needed to load and run a container.
-This includes the following artifacts which MUST all reside in the same directory on the local filesystem:
+This MUST include the following artifacts:
```
duglin: I agree with remove the bit about "same dir" from above, but then I think you need to add it below for each artifact that still need to be in that dir - like config.json.

duglin: To be clear, what I'm worried about is someone thinking its ok to put config.json in some nested dir instead of the root of the bundle dir.  And w/o some kind of -f flag I don't think we support that yet.

crosbymichael: What would be a better sentence to add that clarifies that they should be located at the root?

duglin: I was thinking that on line 14 we say:
```
This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
```

mlaventure: That sounds good to me. Will update shortly

mlaventure:
Updated wording to ensure that it is made clear that `config.json` must be at the root of the bundle. Thanks @duglin for the wording.

duglin:
thanks!

duglin:
ping @mrunalp or @vbatts for the merge!

mrunalp:
LGTM

GGOemea:
Ok

PR #405, opened 2016-04-27
Add VM-based runtime configuration details.

jamesodhunt:
First pass, based on discussions on the mailing list:

```
https://groups.google.com/a/opencontainers.org/d/topic/dev/mM_DWZmXst0/discussion
```

Signed-off-by: James Hunt james.o.hunt@intel.com

### DIFF
--- a/config-vm.md
+++ b/config-vm.md
@@ -1,0 +1,37 @@
+# Virtual-machine-specific Configuration
+
+Virtual-machine-based runtimes require additional configuration to that specified in the [default spec configuration](config.md).
+
+This optional configuration is specified in a "VM" object:
+
+* **`imagePath`** (string, required) path to file that represents the root filesystem for the virtual machine.
+* **`kernel`** (object, required) specifies details of the kernel to boot the virtual machine with.
+
+Note that `imagePath` refers to a path on the host (outside of the virtual machine).
+This field is distinct from the **`path`** field in the [Root Configuration](config.md#Root-Configuration) section since in the context of a virtual-machine-based runtime:
+
+* **`imagePath`** will represent the root filesystem for the virtual machine.
+* The container root filesystem specified by **`path`** from the [Root Configuration](config.md#Root-Configuration) section will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
+
+The virtual-machine-based runtime will use these two path fields to arrange for the **`path`** from the [Root Configuration](config.md#Root-Configuration) section to be presented to the process to run as the root filesystem.
+
+## Kernel
+
+Used by virtual-machine-based runtimes only.
+
+* **`path`** (string, required) specifies the path to the kernel used to boot the virtual machine.
+* **`parameters`** (string, optional) specifies a space-separated list of parameters to pass to the kernel.
+* **`initrd`** (string, optional) specifies the path to an initial ramdisk to be used by the virtual machine.
+
+## Example of a fully-populated `VM` object
+
+```json
+"vm": {
+    "imagePath": "path/to/rootfs.img",
+    "kernel": {
+        "path": "path/to/vmlinuz",
+        "parameters": "foo=bar hello world",
+        "initrd": "path/to/initrd.img"
+    },
+}
+```
--- a/config-vm.md
+++ b/config-vm.md
@@ -4,13 +4,13 @@ Virtual-machine-based runtimes require additional configuration to that specifie
 
 This optional configuration is specified in a "VM" object:
 
-* **`imagePath`** (string, required) path to file that represents the root filesystem for the virtual machine.
+* **`imagePath`** (string, required) host path to file that represents the root filesystem for the virtual machine.
 * **`kernel`** (object, required) specifies details of the kernel to boot the virtual machine with.
 
 Note that `imagePath` refers to a path on the host (outside of the virtual machine).
 This field is distinct from the **`path`** field in the [Root Configuration](config.md#Root-Configuration) section since in the context of a virtual-machine-based runtime:
 
-* **`imagePath`** will represent the root filesystem for the virtual machine.
+* **`imagePath`** will represent the host path to the root filesystem for the virtual machine.
 * The container root filesystem specified by **`path`** from the [Root Configuration](config.md#Root-Configuration) section will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
 
 The virtual-machine-based runtime will use these two path fields to arrange for the **`path`** from the [Root Configuration](config.md#Root-Configuration) section to be presented to the process to run as the root filesystem.
@@ -19,9 +19,9 @@ The virtual-machine-based runtime will use these two path fields to arrange for 
 
 Used by virtual-machine-based runtimes only.
 
-* **`path`** (string, required) specifies the path to the kernel used to boot the virtual machine.
+* **`path`** (string, required) specifies the host path to the kernel used to boot the virtual machine.
 * **`parameters`** (string, optional) specifies a space-separated list of parameters to pass to the kernel.
-* **`initrd`** (string, optional) specifies the path to an initial ramdisk to be used by the virtual machine.
+* **`initrd`** (string, optional) specifies the host path to an initial ramdisk to be used by the virtual machine.
 
 ## Example of a fully-populated `VM` object
 
--- a/config.md
+++ b/config.md
@@ -462,6 +462,12 @@ Instead they MUST ignore unknown properties.
 Runtimes that are reading or processing this configuration file MUST generate an error when invalid or unsupported values are encountered.
 Unless support for a valid value is explicitly required, runtimes MAY choose which subset of the valid values it will support.
 
+## VM
+
+VM is an optional object used by virtual-machine-based runtimes.
+
+See [Virtual-machine-specific schema](config-vm.md) for details.
+
 ## Configuration Schema Example
 
 Here is a full example `config.json` for reference.
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -18,6 +18,8 @@ type Spec struct {
 	Hooks *Hooks `json:"hooks,omitempty" platform:"linux,solaris"`
 	// Annotations contains arbitrary metadata for the container.
 	Annotations map[string]string `json:"annotations,omitempty"`
+	// VM specifies configuration for virtual-machine-based runtimes.
+	VM VM `json:"vm,omitempty"`
 
 	// Linux is platform-specific configuration for Linux based containers.
 	Linux *Linux `json:"linux,omitempty" platform:"linux"`
@@ -25,6 +27,24 @@ type Spec struct {
 	Solaris *Solaris `json:"solaris,omitempty" platform:"solaris"`
 	// Windows is platform-specific configuration for Windows based containers.
 	Windows *Windows `json:"windows,omitempty" platform:"windows"`
+}
+
+// VM contains information for virtual-machine-based runtimes.
+type VM struct {
+	// Kernel specifies kernel-related configuration for virtual-machine-based runtimes.
+	Kernel Kernel `json:"kernel"`
+	// ImagePath is the path to the root filesystem image on the host which can be used by a virtual-machine-based runtime.
+	ImagePath string `json:"imagePath"`
+}
+
+// Kernel contains information about the kernel to use for a virtual machine.
+type Kernel struct {
+	// Path is the path to the kernel used to boot the virtual machine.
+	Path string `json:"path"`
+	// Parameters specifies parameters to pass to the kernel.
+	Parameters string `json:"parameters,omitempty"`
+	// InitRd is the path to an initial ramdisk to be used by the kernel.
+	InitRd string `json:"initrd,omitempty"`
 }
 
 // Process contains information to start a specific application inside the container.
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -43,8 +43,8 @@ type VMKernel struct {
 	Path string `json:"path"`
 	// Parameters specifies parameters to pass to the kernel.
 	Parameters string `json:"parameters,omitempty"`
-	// InitRd is the host path to an initial ramdisk to be used by the kernel.
-	InitRd string `json:"initrd,omitempty"`
+	// InitRD is the host path to an initial ramdisk to be used by the kernel.
+	InitRD string `json:"initrd,omitempty"`
 }
 
 // Process contains information to start a specific application inside the container.
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -33,17 +33,17 @@ type Spec struct {
 type VM struct {
 	// Kernel specifies kernel-related configuration for virtual-machine-based runtimes.
 	Kernel Kernel `json:"kernel"`
-	// ImagePath is the path to the root filesystem image on the host which can be used by a virtual-machine-based runtime.
+	// ImagePath is the host path to the root filesystem image on the host which can be used by a virtual-machine-based runtime.
 	ImagePath string `json:"imagePath"`
 }
 
 // Kernel contains information about the kernel to use for a virtual machine.
 type Kernel struct {
-	// Path is the path to the kernel used to boot the virtual machine.
+	// Path is the host path to the kernel used to boot the virtual machine.
 	Path string `json:"path"`
 	// Parameters specifies parameters to pass to the kernel.
 	Parameters string `json:"parameters,omitempty"`
-	// InitRd is the path to an initial ramdisk to be used by the kernel.
+	// InitRd is the host path to an initial ramdisk to be used by the kernel.
 	InitRd string `json:"initrd,omitempty"`
 }
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -32,13 +32,13 @@ type Spec struct {
 // VM contains information for virtual-machine-based runtimes.
 type VM struct {
 	// Kernel specifies kernel-related configuration for virtual-machine-based runtimes.
-	Kernel Kernel `json:"kernel"`
+	Kernel VMKernel `json:"kernel"`
 	// ImagePath is the host path to the root filesystem image on the host which can be used by a virtual-machine-based runtime.
 	ImagePath string `json:"imagePath"`
 }
 
-// Kernel contains information about the kernel to use for a virtual machine.
-type Kernel struct {
+// VMKernel contains information about the kernel to use for a virtual machine.
+type VMKernel struct {
 	// Path is the host path to the kernel used to boot the virtual machine.
 	Path string `json:"path"`
 	// Parameters specifies parameters to pass to the kernel.
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -191,6 +191,9 @@
                 }
             }
         },
+        "vm": {
+            "$ref": "schema-vm.json#/vm"
+        },
         "linux": {
             "$ref": "config-linux.json#/linux"
         },
--- a/schema/schema-vm.json
+++ b/schema/schema-vm.json
@@ -9,7 +9,7 @@
         ],
         "properties": {
             "imagePath": {
-                "description": "path to rootfs image on host system which is used for VM-based runtimes",
+                "description": "host path to rootfs image on host system which is used for VM-based runtimes",
                 "id": "https://opencontainers.org/schema/bundle/vm/imagePath",
                 "type": "string"
             },
@@ -23,7 +23,7 @@
                 "properties": {
                     "path": {
                         "id": "https://opencontainers.org/schema/bundle/vm/kernel/path",
-                        "description": "path to kernel image",
+                        "description": "host path to kernel image",
                         "type": "string"
                     },
                     "parameters": {
@@ -32,7 +32,7 @@
                         "type": "string"
                     },
                     "initrd": {
-                        "description": "path to initial ramdisk image",
+                        "description": "host path to initial ramdisk image",
                         "id": "https://opencontainers.org/schema/bundle/vm/kernel/initrd",
                         "type": "string"
                     }
--- a/schema/schema-vm.json
+++ b/schema/schema-vm.json
@@ -1,0 +1,43 @@
+{
+    "vm": {
+        "description": "configuration for virtual-machine-based runtimes",
+        "id": "https://opencontainers.org/schema/bundle/vm",
+        "type": "object",
+        "required" : [
+            "imagePath",
+            "kernel"
+        ],
+        "properties": {
+            "imagePath": {
+                "description": "path to rootfs image on host system which is used for VM-based runtimes",
+                "id": "https://opencontainers.org/schema/bundle/vm/imagePath",
+                "type": "string"
+            },
+            "kernel": {
+                "description": "kernel config used by VM-based runtimes",
+                "id": "https://opencontainers.org/schema/bundle/vm/kernel",
+                "type": "object",
+                "required": [
+                    "path"
+                ],
+                "properties": {
+                    "path": {
+                        "id": "https://opencontainers.org/schema/bundle/vm/kernel/path",
+                        "description": "path to kernel image",
+                        "type": "string"
+                    },
+                    "parameters": {
+                        "description": "space-separated list of kernel parameters",
+                        "id": "https://opencontainers.org/schema/bundle/vm/kernel/parameters",
+                        "type": "string"
+                    },
+                    "initrd": {
+                        "description": "path to initial ramdisk image",
+                        "id": "https://opencontainers.org/schema/bundle/vm/kernel/initrd",
+                        "type": "string"
+                    }
+                }
+            }
+        }
+    }
+}


config.md
```diff
@@ -265,6 +265,27 @@ Annotations are key-value maps.
...
+
+HypervisorPath is an optional absolute path to the hypervisor binary used by VM-based runtimes.
+
+## ImagePath
+
+ImagePath is an optional absolute path to the root filesystem image used by VM-based runtimes.
+
+## RuntimeDriver
+
+RuntimeDriver is an optional driver name to be passed to VM-based runtimes.
```
avsm: Is the intention of `RuntimeDriver` to select between Xen, KVM, Qemu and other hypervisors?

config.md
```diff
@@ -265,6 +265,27 @@ Annotations are key-value maps.
 }
 ```
 
+## HypervisorPath
+
+HypervisorPath is an optional absolute path to the hypervisor binary used by VM-based runtimes.
+
+## ImagePath
+
+ImagePath is an optional absolute path to the root filesystem image used by VM-based runtimes.
```
wking: I haven't checked the thread for earlier discussion, but this sounds like it should go in **`root.path`**.  What else goes into an image file besides the root filesystem?

config.md
```diff
@@ -265,6 +265,27 @@ Annotations are key-value maps.
 }
 ```
 
+## HypervisorPath
+
+HypervisorPath is an optional absolute path to the hypervisor binary used by VM-based runtimes.
```
wking: This sounds like a runtime concern.  What is the benefit to putting it in config.json?

config.md
laijs: When `ImagePath` and `root.path` are both specific, the rootfs  will the the
`filepath.Join(ImageMountPoint, root.path)`  (root.path can be absolute or not)
the `ImageMountPoint` is mounted by the guest inside the VM. the `ImagePath` is the block device passed to VM by the hypervisor.

Take docker's devicemapper dirver as an example, `ImagePath` will  be the path of the block device of the dm snapshot, and `root.path` will be "rootfs" or "/rootfs" or "./rootfs". (docker dons't put the real rootfs in the root directory of the block device. it put the real rootfs in /rootfs inside the block device).

laijs:
@jamesodhunt  could you add descriptions for the behavior when `ImagePath` and `root.path` are both specific.

config.md
wking: On Wed, Apr 27, 2016 at 07:46:13AM -0700, Lai Jiangshan wrote:
> > +## ImagePath
> > +
> > +ImagePath is an optional absolute path to the root filesystem image used by VM-based runtimes.
> 
> When `ImagePath` and `root.path` are both specific, the rootfs  will the the
> `filepath.Join(ImageMountPoint, root.path)`  (root.path can be absolute or not)
> the `ImageMountPoint` is mounted by the guest inside the VM. the `ImagePath` is the block device passed to VM by the hypervisor.

I think you mean ‘ImagePath’ where you have ‘ImageMountPoint’ in your
comment.

If the runtime needs to pass a block device to the VM, it seems like
it could expand root.path into an absolute path, and then keep running
dirname or similar until it found a block-device mount-point.  Is
there really a need to split it into two fields?  See also #133, which
sounds similar.

laijs: `ImagePath` is the path of the image outside the vm. ImageMountPoint is the path inside the vm (the path is chosen by the runtime-implmentation) for mounting the image.
The image can't not be mounted in the host (if the fs inside the image is not disbuted filesystem) when the container is running, it is mount only in the vm. so there is not mount-point in the host and there is no way to "find a block-device mount-point".
And the root directory inside the image may not be the rootfs of the container, so root.path is needed for pointing out where is the rootfs for the container. See the devicemapper driver of docker.

wking: On Wed, Apr 27, 2016 at 05:19:49PM -0700, Lai Jiangshan wrote:
“`ImagePath` is the path of the image outside the vm”

Ah.  So it's something like:

  "ImagePath": "path/to/some-image"

where some-image is a tarball-like thing that includes the root
filesystem.  That sounds more widely applicable than the VM case.  For
example, it would be nice to be able to run Linux namespace/cgroup
containers with a tarred rootfs without having to unpack the tarball.
Why can't you have:

  "root": {
    "path": "path/to/some-image/my/root"
  }

and teach the VM runtime logic like:

* Nibble path components from root.path until you find a path that
  exists on the host filesystem.
* The existing path (‘path/to/some-image’ in the example above) is the
  image file (your ImagePath).
* The nibbled components (‘my/root’ in the example above) are the path
  to root inside the image (your root.path).

The fact that your filesystem switches from Btrfs (or some such) to
OVA (or some such) at path/to/some-image doesn't seem like information
that requires a VM-specific setting to convey.

config.md
jamesodhunt: That was the intention, yes. Since HypervisorPath is already specified, maybe this field is redundant as named. Maybe rather than RuntimeDriver, something like HypervisorPersonality would be more useful.  I added RuntimeDriver specifically for the benefit of runv so would be good to get input from them as to whether HypervisorPath is sufficient.

config.md
jamesodhunt: Possibly, but I think it depends on whether the runtime should be making decisions, or whether it should just do what it's told by acting on the contents of config.json.

Since that runtime is running in a potentially restricted environment and having its actions directed by entities such as docker/containerd, I've opted for the latter imperative approach. I think that is also potentially safer since those higher-level entities can potentially perform the appropriate validity checks before handing the config to the runtime (see docker/docker#21324).

Ultimately, I'd prefer that the runtime avoid having to use heuristics, PATH searching and custom-config files to allow it to handle the workload. By encoding all configuration the runtime needs in the OCI configuration, it should be fully portable to any OCI-compliant environment.

If we imagine that the VM configuration does not form part of the OCI spec, the runtime would then have to find its custom config file (to know which hypervisor to use, etc).  This might not be as trivial as it sounds given that various distros treat config file locations differently. That's not even considering the Windows aspect.

wking: On Thu, Apr 28, 2016 at 01:53:53AM -0700, James Hunt wrote:
“This might not be as trivial as it sounds given that various distros
treat config file locations differently.”

I expect they also have different paths for their hypervisor binaries
;).  I'm not sure I have a clear picture of where the line is between
“belongs in config.json”, “belongs in /etc/{runtime}/… or similar”,
and “belongs in a runtime CLI option”.

Do we expect a single runtime on a single host to use different
hypervisor paths on a per-container basis?  That would seem to rule
out the /etc approach as a complete solution.

Do we expect bundle authors (vs. runtime callers) to care about which
hypervisor launches the container?  I'm not terribly familar with the
hypervisor ecosystem, but it sounds to me like HypervisorPersonality
[1] may be a better fit for config.json, and an /etc/{runtime}/…
setting would translate those personalities to paths that were
appropriate for the local system.

[1]: https://github.com/opencontainers/runtime-spec/pull/405/files/6194f0b24749bedaf3f384787b73cf1c9f899f4b#r61391724

config.md
```diff
@@ -265,6 +265,27 @@ Annotations are key-value maps.
 }
 ```
 
+## HypervisorPath
```
crosbymichael: I don't understand why you would need this if you already have your own runtime binary, `runv` `lkvm` that is starting the thing.  Why is this needed?

jamesodhunt: I think it depends where the line of demarcation lies- who/what is the ultimate arbiter in the OCI world?

In the current docker world, my understanding is that docker itself will "own" the OCI config file (but it would be great to get clarification on docker/docker#21324). Currently, docker is hard-coded to launch containerd (well, "docker-containerd"). But in that world, containerd gets to decide which OCI runtime it will launch ("containerd --runtime runc").  And if that runtime were, say, runv rather than runc, then runv gets to decide which actual hypervisor is ultimately launched to host the workload.

Clearly, this is but one example, but is this the way we want the system to work generally? Or should the OCI config specify:

1) which runtime to use ("HypervisorPath" here, but should probably be renamed "RuntimePath").

2) which hypervisor that runtime should use (optional).

If the spec shouldn't specify this, then yes the field can be dropped.

But if these values should be encoded in the config, docker could continue to call containerd, but containerd would then simply consume the config.json and launch the appropriate runtime.  Assuming that docker owns config.json and that file is generated by docker, docker can perform path checks, etc when passed the appropriate configuration data (again, see docker/docker#21324) which it will use to generate config.json. And docker would know which runtimes were available on the system since the admin would have provided the details to docker.

Currently, the OCI bundle includes config.json, but does that imply the file is "read-only"? Or will "OCI managers" (is there an official term?) like docker be able to supplement (in other words rewrite) this configuration based on locally available resources? If so, what are they allowed to change in the file?  Can they only supplement, or are they permitted to change anything and everything?

Looking at [application bundle builders](/opencontainers/specs/blob/master/README.md#application-bundle-builders), we see:

    > Because the configuration includes host-specific settings, application
    > bundle directories copied between two hosts may require configuration
    > adjustments.

This implies that config.json can be modified to atleast add supplementary data.  Further, some of the config specified by config-linux.md almost guarantees the file will need to be modified before passing to the runtime (memory, CPU shares, cgroups, etc).

To me, this implies it might be better to have a "static" (effectively read-only) config file that contains the minimum possible config for the bundle (config.json). That file would be owned not by the "OCI manager", but by the bundle author.  Hence, it would be illegal for the OCI manager to, say, change process.args.

Additionally, the bundle could also provide an optional separate runtime.json that could specify reasonable defaults for certain settings. However, that runtime.json could be modifiable by the OCI manager in any way it deems appropriate (the file is owned by the OCI manager). Regarding runtime.json, yes, I know we've already been there :-)

wking: On Fri, Apr 29, 2016 at 02:01:45AM -0700, James Hunt wrote:
“1) which runtime to use ("HypervisorPath" here, but should probably
be renamed "RuntimePath").”

This approach assumes a shim like containerd between the config author
and the runtime.  I'm fine with settings like this being part of the
contract between Docker and containerd, but if they are, containerd
should remove the setting before passing the config down to the
runtime.  This repo is about the runtime ↔ caller relationship, and in
that context “which runtime should I use?” has already been decided.

“Currently, the OCI bundle includes config.json, but does that imply
the file is "read-only"? Or will "OCI managers" (is there an official
term?) like docker be able to supplement (in other words rewrite) this
configuration based on locally available resources?  If so, what are
they allowed to change in the file?  Can they only supplement, or are
they permitted to change anything and everything?”

This spec is just about the final bundle passed to the runtime, so
this seems out of scope here.  I recommend “OCI container as a
service” providers clearly document their policies (e.g. “we clobber
‘hooks’ for safety”), but not in this repo ;).

“To me, this implies it might be better to have a "static"
(effectively read-only) config file…”

We tried this for a time.  See #284 for why we went back to a single
config.

crosbymichael: Ok, I don't think this should be part of the config.  The high levels will provide a way for the user to specify what runtime it should be launched with.  

This should be safe to remove.

specs-go/config.go
```diff
@@ -22,11 +22,36 @@ type Spec struct {
 	Hooks Hooks `json:"hooks"`
 	// Annotations is an unstructured key value map that may be set by external tools to store and retrieve arbitrary metadata.
 	Annotations map[string]string `json:"annotations,omitempty"`
-
+	// HypervisorPath is the absolute path to the hypervisor binary
+	// for VM-based runtimes.
+	HypervisorPath string `json:"hypervisorPath"`
```
mrunalp: These settings should be optional and probably don't belong in the top level struct.

wking: On Mon, May 02, 2016 at 11:33:45AM -0700, Mrunal Patel wrote:
> > +	// HypervisorPath is the absolute path to the hypervisor binary
> > +	// for VM-based runtimes.
> > +	HypervisorPath string `json:"hypervisorPath"`
> 
> These settings should be optional and probably don't belong in the top level struct. 

Re: optional, see [1,2].  Re: top level, see [3] and later in that
thread.

[1]: https://github.com/opencontainers/runtime-spec/pull/405/files/6194f0b24749bedaf3f384787b73cf1c9f899f4b#r61262329
[2]: https://github.com/opencontainers/runtime-spec/pull/405/files/6194f0b24749bedaf3f384787b73cf1c9f899f4b#r61473810
[3]: https://groups.google.com/a/opencontainers.org/d/msg/dev/mM_DWZmXst0/uDPFv8-MBQAJ
     Subject: Re: Make the container spec unified for both native and virtual machine
     Date: Mon, 22 Feb 2016 15:12:46 -0800
     Message-ID: <20160222231246.GV4265@odin.tremily.us>

specs-go/config.go
```diff
@@ -22,11 +22,30 @@ type Spec struct {
 	Hooks Hooks `json:"hooks"`
 	// Annotations is an unstructured key value map that may be set by external tools to store and retrieve arbitrary metadata.
 	Annotations map[string]string `json:"annotations,omitempty"`
-
+	// KernelConfig specifies kernel-related configuration for
+	// virtual-machine-based runtimes.
+	KernelConfig KernelConfig `json:"kernelConfig"`
```
gao-feng: Don't need omitempty?

specs-go/config.go
```diff
@@ -22,11 +22,30 @@ type Spec struct {
...
+	// ImagePath is the absolute path to the root filesystem image
+	// used by a virtual-machine-based runtime.
+	ImagePath string `json:"imagePath"`
 	// Linux is platform specific configuration for Linux based containers.
 	Linux Linux `json:"linux" platform:"linux"`
 }
 
+// Kernel contains information about the kernel to use for a virtual
+// machine.
```
laijs: s/Kernel contains/KernelConfig contains/

and comments need to be single line.

jamesodhunt: Thanks for the feedback - fixed.

laijs:
LGTM

config.md
```diff
@@ -265,6 +265,19 @@ Annotations are key-value maps.
 }
 ```
 
+## ImagePath
+
+ImagePath is an optional absolute path to the root filesystem image used by VM-based runtimes.
```
wking: Maybe:

* **`imagePath`** (string, optional) path to the root filesystem image used by VM-based runtimes.

Why do you currently require an absolute path?

config.md
```diff
@@ -265,6 +265,19 @@ Annotations are key-value maps.
...
+
+## Kernel-configuration
+
+Used by VM-based runtimes only.
+
+* **`path`** (string, required) specifies an absolute path to the kernel
+  used to boot the virtual machine.
+* **`parameters`** (string, optional) specifies a space-separated list of parameters to pass to the kernel.
+* **`initrd`** (string, optional) specifies an absolute path to an initial ramdisk to be used by the kernel.
```
wking: More absolute path stuff here (both `path` and `initrd`).  Why require absolute paths?

config.md
wking: And docs around the need for this setting vs. `root.path` would be nice (cherry-pick from [here][1]?).  I'm backing off of my earlier [suggestion to put all of this information in `root.path`][2], because it would be awkward to combine a POSIX path-to-image with a Windows path-to-root-inside-the-image (if you were launching a Windows VM on from a POSIX system).

[1]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61358293
[2]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61373485

jamesodhunt: imagePath needs to be absolute since it will not form part of the bundle, so the runtime needs to know exactly where on the host to find it.

I've updated the PR with further details on imagePath as suggested.

config.md
jamesodhunt: I think we might want to consider adding ``RuntimePath``. Currently, ``dockerd`` will launch ``containerd`` and ``containerd`` will launch the runtime configured when it was started.

But this implies that if a single host is to support multiple runtimes, *you would need two instances of* ``containerd``:

* ``containerd --runtime foo --state-dir /run/containerd-foo --listen unix:///run/containerd/containerd-foo.sock --runtime /path/to/foo/runtime``

* ``containerd --runtime bar --state-dir /run/containerd-bar --listen unix:///run/containerd/containerd-bar.sock --runtime /path/to/bar/runtime``

But, *you would also need two instances of* ``dockerd``, one to talk to each of the above:

* ``dockerd --containerd /path/to/containerd1.sock``
* ``dockerd --containerd /path/to/containerd2.sock``

This doesn't sound ideal.

But, if we were to introduce ``RuntimePath``, only a single ``dockerd`` and a single ``containerd`` instance would be required. We could then remove ``--runtime`` and ``--runtime-args`` from ``containerd`` and have it launch the runtime specified by ``RuntimePath`` (we'd also need a ``RuntimeArgs`` of course).

Maybe another way to handle this is to allow ``RuntimePath`` to be specified. If it is, ``containerd`` (for example) would honour that (if the path existed). If not specified ``containerd `` would default to the value of ``--runtime`` (or "``runc``" if ``--runtime`` is not specified).

config.md
patux: What's the problem with absolute paths?

config.md
wking: On Wed, May 11, 2016 at 02:12:16AM -0700, James Hunt wrote:
> > +ImagePath is an optional absolute path to the root filesystem image used by VM-based runtimes.
>
> imagePath needs to be absolute since it will not form part of the
> bundle, so the runtime needs to know exactly where on the host to
> find it.

With Linux-namespace containers, the runtime knows exactly where to
find root.path without it having to be absolute (relative paths are
interpreted relative to the directory containining config.json).  So I
don't see a reason to *require* absolute paths.  Then folks could put
the image in the bundle if they wanted, or point at an absolute path
if they wanted.

config.md
wking: On Wed, May 11, 2016 at 10:34:58AM -0700, Geronimo Orozco wrote:
> > +* **`path`** (string, required) specifies an absolute path to the kernel
> > +  used to boot the virtual machine.
> > +* **`parameters`** (string, optional) specifies a space-separated list of parameters to pass to the kernel.
> > +* **`initrd`** (string, optional) specifies an absolute path to an initial ramdisk to be used by the kernel.
> 
> What's the problem with absolute paths? 

I'm fine allowing absolute paths, but I don't see why we shouldn't
allow relative paths too [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/405/files/8d359458aef03c76837ed17eb3fcfb935f0708b2#r62714446

config.md
wking: On Wed, May 11, 2016 at 03:12:24AM -0700, James Hunt wrote:
> I think we might want to consider adding ``RuntimePath``. Currently,
> ``dockerd`` will launch ``containerd`` and ``containerd`` will
> launch the runtime configured when it was started.
> 
> But this implies that if a single host is to support multiple
> runtimes, *you would need two instances of* ``containerd``:

I acknowledge the problem, but think that this is a dockerd ↔
containerd issue [1].  There's no need to put this in runtime-spec if
the containerd API changes to support:

  launch(pathToBundle, runtimePath, runtimeArgs)

[1]: https://github.com/opencontainers/runtime-spec/pull/405/files/6194f0b24749bedaf3f384787b73cf1c9f899f4b#r61617026

WeiZhang555: @jamesodhunt 

I agree with you that we should create container with different runtime in one single containerd, but I think you are using the wrong way.

If I understand correctly, config.json should only be interpreted by runtime but not containerd, if we put the runtime path or name in config.json, we need to handle it first in containerd then let container make decision of calling which runtime.

I think the correct way is making the `--runtime` and `--runtime-args` per-container flags but not containerd daemon flags, we don't need to remove all these two from daemon but leave them also as default configure. 

I think this is also what @crosbymichael 's saying " I don't think this should be part of the config" mean.  Actually I want to make per-container `--runtime` and `--runtime-args` long time ago, do you think it's acceptable or not? If OK, I can start working on this @crosbymichael . Sorry this is a little bit beyond the topic.

jamesodhunt: @WeiZhang555 - yes, making ``--runtime`` a container-specific flag would indeed solve the issue. +1.

The only complication would be that the user would need to remember to specify it to "``docker run``". Maybe it would be useful to be able to configure ``dockerd`` to specify a default runtime to use? That way, if the user runs, "``docker run $image``", they get the admin-configured default runtime **unless** they explicitly specify "``docker run --runtime=$foo $image``"?

WeiZhang555: @jamesodhunt Agree. Configurable default runtime will be good.

I think we can keep `--runtime` daemon flag of containerd as default, so we can explicitly start containerd with some custom runtime e.g. `dockerd --runtime foo`,  then docker can connect to it via `docker --containerd xxxxx`.  If user want to use another runtime, then `docker run --runtime bar`

config.md
jamesodhunt: Looking through the existing docs, the only paths I can see that represents entities that must exist but which are allowed to be relative are:

- `process.args[0]`: this is reasonable given `$PATH` can be searched plus it permits the command to be absolute.

- `root.path`: this is potentially reasonable given that that path must exist below `$bundle_dir`.
  However, [the runtime start command](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#start) does not specify that the `start` command be passed an absolute path as it's second argument (`start <container-id> <path-to-bundle>`). This seems surprising - if it's not absolute, how is the runtime to find it? Is this an oversight or is there a contract somewhere stating that the runtime will always be launched from, or set its cwd to "`$bundle_dir`"?

There are only two points of reference that a runtime has from what I can see:

- it's cwd.
- the bundle directory provided to it.

The elements referenced by this PR (`kernelConfig.path`, `kernelConfig.initrd` and `imagePath`):

- are not commands so cannot be found using `$PATH`.
- will not be available below `$bundle_dir`.

Hence, I believe they must be absolute paths. Further, it's safer to specify absolute paths. On that topic, I note that all hook paths must be absolute. Also, absolute paths are more in line with the imperative nature of the configuration file - don't give the runtime choices (unless there's are good reason to ;-)

Since paths can be troublesome, I wonder if all paths mentioned by the spec should be reviewed to ensure that for each path that is permitted to be relative, a rationale be provided as to why it can be relative?

wking: On Mon, May 16, 2016 at 06:35:07AM -0700, James Hunt wrote:
> - it's cwd.

Ccon uses it's cwd (e.g. see wording in [1]), and that approach hasn't
given me any trouble.

> - will not be available below `$bundle_dir`.

Why force this restriction at the runtime level?  I'm fine if you want
to enforce it at a higher level for your implementation, but I see no
technical reason to keep the kernel binary outside the bundle
directory (and of course, the absolute path could lead you back into
the bundle directory, just not in a very portable manner).

> On that topic, I note that all hook paths must be absolute.

Yeah, I'm also not a fan of that restriction ;).

> Also, absolute paths are more in line with the imperative nature of
> the configuration file - don't give the runtime choices (unless
> there's are good reason to ;-)

If they're relative to the runtime's cwd (or use runtime PATH lookups
for hooks), then there are still no *runtime* choices.  It just makes
some workflows more convenient for the bundle author and runtime
caller (e.g. you can bundle your kernel instead of transmitting it
out-of-band).

> Since paths can be troublesome, I wonder if all paths mentioned by
> the spec should be reviewed to ensure that for each path that is
> permitted to be relative, a rationale be provided as to why it can
> be relative?

I think this is a policy decision that is probably useful to do at a
higher level, but I'd rather keep policy out of the runtime spec.

specs-go/config.go
```diff
@@ -22,11 +22,24 @@ type Spec struct {
 	Hooks Hooks `json:"hooks"`
 	// Annotations is an unstructured key value map that may be set by external tools to store and retrieve arbitrary metadata.
 	Annotations map[string]string `json:"annotations,omitempty"`
-
+	// KernelConfig specifies kernel-related configuration for virtual-machine-based runtimes.
+	KernelConfig KernelConfig `json:"kernelConfig"`
```
crosbymichael: Can we move these two to a Vm *Vm field?

vishh: Optional fields must have `optional` json tag.

specs-go/config.go
```diff
@@ -22,11 +22,24 @@ type Spec struct {
...
+	// ImagePath is the absolute path to the root filesystem image on the host which can be used by a virtual machine-based runtime.
+	ImagePath string `json:"imagePath"`
 	// Linux is platform specific configuration for Linux based containers.
 	Linux Linux `json:"linux" platform:"linux"`
```
vishh: May be `Linux` should be renamed to `LinuxContainer` & then VM based options can be added as `LinuxVirtualMachine`..

cc @crosbymichael @wking

crosbymichael: idk, vm can be mutli OS/platform right?

crosbymichael: not sure we need that type of granularity in the fields when we have the platform struct in the spec

wking: On Wed, May 18, 2016 at 10:40:15AM -0700, Michael Crosby wrote:
> >  	// Linux is platform specific configuration for Linux based containers.
> >  	Linux Linux `json:"linux" platform:"linux"`
>
> not sure we need that type of granularity in the fields when we have
> the platform struct in the spec

I'm not sure about the granularity.  If the new VM fields are intended
to support VMs running different guest platforms (e.g. guest Linux,
guest Windows, …) then generic ‘virtual-machine’ namespacing sounds
good.  If the new fields are only for Linux guests, then specific
‘linux-virtual-machine’ sounds good.

In the latter case, changing the existing ‘linux’ to ‘linux-container’
makes sense (no need to make the container implementation special),
but maybe isn't worth breaking backwards compatibility.  It does mean
that *new* entries should be as specific as possible to avoid future
ambiguity sticking around for backwards compatibility.

And with VM's in the picture, we may have to overhaul the ‘platform’
entry to allow for a host/guest distinction.

crosbymichael: @jamesodhunt lets just add a Vm subsection for these fields then I think we are good to go on this.

tianon: "LinuxContainer" would be redundant anyhow IMO given that this whole thing
is "the container runtime spec" :smile:

vishh: @tianon What does a container really mean 😃 ?

crosbymichael: noooo

mrunalp: 😃

wking: On Wed, May 18, 2016 at 04:32:53PM -0700, Tianon Gravi wrote:
> "LinuxContainer" would be redundant anyhow IMO given that this whole thing
> is "the container runtime spec" :smile:

Yeah, “linux-namespace-cgroup” or something would probably be better,
since those are the main components the current “linux” section is
based on.  I'm still not sure how much common ground there is to be
found with a big-tent approach that also covers VMs, although I guess
we'll see ;).

jamesodhunt:
@crosbymichael, @wking - how does this look?

config.md
```diff
@@ -265,6 +265,41 @@ Annotations are key-value maps.
...
 ```
 
+## VM
+
+VM is an optional object used by virtual machine-based runtimes.
+
+* **`imagePath`** (string, required) path to file that represents the root filesystem for the virtual machine.
+* **`kernel`** (object, required) specifies details of the kernel to boot the virtual machine with.
+
+Note that `imagePath` refers to a path on the host (outside of the virtual machine). This field is distinct from the **`path`** field in the (#Root-Configuration) section since in the context of a virtual machine-based runtime:
```
wking: I may be hyphen-happy, but I'm pretty sure a “machine-based” hyphen should also come with a “virtual-machine” hyphen.  Whether “virtual-machine-based” or “virtual-machine based” is better is beyond me ;).

wking: Also, [one sentence per line][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/style.md#one-sentence-per-line

config.md
```diff
@@ -265,6 +265,41 @@ Annotations are key-value maps.
...
+
+* **`ImagePath`** will represent the root filesystem for the virtual machine.
+* The container root filesystem specified by **`path`** will be mounted inside the virtual machine at a location chosen by the virtual machine-based runtime.
+
+The virtual machine-based runtime will use these two path fields to arrange for **`path`** to be presented to the process to run as the root filesystem.
+
+An example of a fully-populated `VM` object:
```
wking: Can this example get [a header][1] for convenient linking?

[1]: https://github.com/opencontainers/runtime-spec/blob/d8d0bc6f601f98122f8908eb21158e32432ab563/style.md#examples

wking:
4c8a03c looks pretty good to me (I just have copy-edit suggestions
[1,2]).  It would also be nice to get a ‘vm’ entry in here [3].  And
with Solaris and Linux docs in separate config-{platform}.json files,
it may make sense to put the VM stuff in config-vm.json too.

[1]: https://github.com/opencontainers/runtime-spec/pull/405/files#r63914921
[2]: https://github.com/opencontainers/runtime-spec/pull/405/files#r63914921
[3]: https://github.com/opencontainers/runtime-spec/blob/d8d0bc6f601f98122f8908eb21158e32432ab563/config.md#platform-specific-configuration

jamesodhunt:
Hi @wking - I've pushed updates to all your comments except [3] - I know we discussed this on the mailing list, but I'm still slightly wary of having "`platform.os`" set as "`vm`" as that precludes the vm-based runtimes that use Linux from using any details from the "`linux`" object (I think?)

Also, using a real-world example, I believe docker/containerd would be required to know that a containers runtime was set to a vm-based one and therefore not generate a "`linux`" object. Or alternatively, something like `containerd` would need to in-flight modify `config.json` to:

* remove the "`linux`" object.
* change "`platform.os`" to `vm`.

This doesn't sound ideal, but maybe @crosbymichael could comment on this.

How about instead we add something like "`platform.type`" or "`platform.variant`"?

```json
        "platform": {
                "os": "linux",
                "arch": "amd64",
                "variant": "vm"
        },
```
Where `variant` would be optional but if set a runtime could consume both the "`os`" and "`variant`" objects.

The only further thought I had was on the kernel parameters: it might be more elegant to encode them as an array, but more practical to specify them in string form.

[3] - https://github.com/opencontainers/runtime-spec/blob/d8d0bc6f601f98122f8908eb21158e32432ab563/config.md#platform-specific-configuration

wking:
On Fri, May 20, 2016 at 07:15:46AM -0700, James Hunt wrote:
> Hi @wking - I've pushed updates to all your comments except [3]…

Thanks.  And just a reminder that I'm not a maintainer, so please take
my comments with a grain of salt ;).

>         "platform": {
>                 "os": "linux",
>                 "arch": "amd64",
>                 "variant": "vm"
>         },

I'm not clear on how VM-based containers are intended to manage the
host/guest split [1].  Are platform.os and platform.arch for the
guest?  If so, how does a runtime know if the config is targetting the
right host OS?  For example, hooks, imagePath, etc. all need to be in
the host-OS language.  Instead of adding a variant field, I think VM
support calls for an explicit declaration like:

  "platform": {
     "host": {
       "os": "linux",
       "arch": "amd64",
     },
     "guest: {
       "os": "linux",
       "arch": "amd64",
     }
  }

Where runtimes that support VMs require ‘guest’ and look for a ‘vm’
section, runtimes that don't support VMs error if ‘guest’ is present,
and the validator requires a ‘vm’ section if and only if ‘guest’ is
present.

[1]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r63751591

wking:
On Fri, May 20, 2016 at 11:25:28AM -0700, W. Trevor King wrote:
> Are platform.os and platform.arch for the guest?  If so, how does a
> runtime know if the config is targetting the right host OS?

The other reasonable approach here is to punt platform information to
higher levels (it's already in the image spec:
opencontainers/image-spec#60).  Then have the runtime just assume that
the config targets platforms it can handle.  The platform fields seem
to be designed to let the runtime guess ahead of time whether it can
launch the given container, and since #418 we don't have to check
ahead of time (we can just try open the kernel path, for example, and
error out if that path doesn't exist because the config uses
forward-slash path separators but your host uses backslash path
separators).  With this approach:

* We drop ‘platform’ entirely.
* VM-based runtimes read the ‘vm’ section, and error out if it's
  missing.
* Non-VM runtimes ignore ‘vm’ even if it is present.
* Linux(-guest)-based runtimes read a ‘linux’ section, ignoring
  ‘solaris’, etc. if they're present.
* …

It seems unlikely that the ‘platform’ section ever covers the full
scope of what a container needs (e.g. was your kernel compiled with
the right settings?  Where any of the bundled binaries compiled with
-march=native (or other extremely-specific arch setting)?).  On the
other hand, ‘platform’ is a fairly reliable solution even if it isn't
perfect, and folks for whom it is not specific enough likely know what
they're up to.

So I'm fine keeping it (and splitting host/guest) or dropping it (and
punting to image-spec), but I don't think keeping it as it stands is
viable.

crosbymichael:
@jamesodhunt i don't think we need to change anything around populating the platform to VM.

LGTM

wking:
On Fri, May 20, 2016 at 02:16:25PM -0700, Michael Crosby wrote:
> i don't think we need to change anything around populating the
> platform to VM.

How do you see the existing structure being populated for the various
permutations of Linux/Windows host/guest?  Or are you only planning to
support VM configs where the host and guest are the same platform?

config-vm.md
```diff
@@ -0,0 +1,37 @@
...
+
+This optional configuration is specified in a "VM" object:
+
+* **`imagePath`** (string, required) path to file that represents the root filesystem for the virtual machine.
+* **`kernel`** (object, required) specifies details of the kernel to boot the virtual machine with.
+
+Note that `imagePath` refers to a path on the host (outside of the virtual machine).
+This field is distinct from the **`path`** field in the (#Root-Configuration) section since in the context of a virtual-machine-based runtime:
+
+* **`ImagePath`** will represent the root filesystem for the virtual machine.
```
mrunalp: s/ImagePath/imagePath ?

config-vm.md
```diff
@@ -0,0 +1,37 @@
...
+* The container root filesystem specified by **`path`** will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
```
mrunalp: This line isn't clear to me. Looking at example below, it implies that `vmlinuz` carries the rootfs of the container which doesn't sound right to me.

config-vm.md
```diff
@@ -0,0 +1,37 @@
...
+
+The virtual-machine-based runtime will use these two path fields to arrange for **`path`** to be presented to the process to run as the root filesystem.
```
mrunalp: I think this also needs to be clarified.

wking: On Fri, May 20, 2016 at 03:28:37PM -0700, Mrunal Patel wrote:
> > +* **`ImagePath`** will represent the root filesystem for the virtual machine.
> > +* The container root filesystem specified by **`path`** will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
> > +
> > +The virtual-machine-based runtime will use these two path fields to arrange for **`path`** to be presented to the process to run as the root filesystem.
> 
> I think this also needs to be clarified. 

It's good to clarify the docs, but see [1,2] for previous discussion
in this PR.

[1]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61261776
[2]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r62715247

crosbymichael:
LGTM

We can squash on merge

philips:
NACK

philips:
I am sorry, I don't understand why this configuration data must live in the runtime specification and isn't just part of the flags or configuration given to the runtime. These three parameters don't seem fundamental to me. In `rkt` when someone wants to run their container in a VM they just add a flag that changes the execution to a VM. They don't put in a path to a kernel, initrd, etc.

philips:
Rejected

wking:
On Thu, Jun 02, 2016 at 09:28:31PM -0700, Brandon Philips wrote:
> In `rkt` when someone wants to run their container in a VM they just
> add a flag that changes the execution to a VM. They don't put in a
> path to a kernel, initrd, etc.

How does the bundle author tell you which kernel they want?  E.g. if I
send you an image that needs v4.6, do I just expect you to know that?
It seems like bundling the required kernel into the image would make
that easier for VM hosts.

crosbymichael:
@philips @jamesodhunt this was my first thought when I need this PR and it does make alot of sense.  Why can these things not be default in your VM runtime?  The config is a constant and the runtime is the variable to how it is run.

philips:
@wking I don't think bundle authors should care about the Kernel that is executing their stuff. Non-VM bundle users don't specify Kernel versions. Also, I would argue that the use case here is not for the bundle author to provide a Kernel ever. Instead it is for someone to execute a container in a pre-vetted system Kernel.

wking:
On Fri, Jun 03, 2016 at 12:41:34PM -0700, Brandon Philips wrote:
> I don't think bundle authors should care about the Kernel that is
> executing their stuff.

That's crazy; my containers will certainly not run on 2.4 ;).  And
lots of other things depend on which particular kernel-config options
you enabled.

> Also, I would argue that the use case here is not for the bundle
> author to provide a Kernel ever.  Instead it is for someone to
> execute a container in a pre-vetted system Kernel.

I don't see a need to *force* that approach.  I agree that I'd be very
cautious about executing a kernel supplied by an untrusted user (even
in a VM), but that caution applies to many other config settings
already (e.g. hooks).  If folks find it convenient to push this
information around in the config, and their VM-based runtime consumes
it, then it's certainly no worse a fit for config.json than some other
things we already have ;).

wking:
Especially after all the time put into this change, dating back to the
2015-09-24 thread [1], I find it surprising that there are a run of
“this is out of scope” calls *now*.  If that ends up sinking this PR,
it would have been much more considerate to have mentioned those
concerns last fall :p.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/mM_DWZmXst0/discussion
     Subject: Make the container spec unified for both native and virtual machine
     Date: Thu, 24 Sep 2015 08:08:59 -0700 (PDT)
     Message-Id: <343f21e9-460d-4e35-9ff3-37d77f609741@opencontainers.org>

jamesodhunt:
@philips - I agree that the bundle author shouldn't necessarily care about a runtimes precise configuration (the ideal is for a bundle to be "runnable" by any runtime). However, the runtime that needs to handle that bundle clearly does care about its configuration.

Stepping back, for the purposes of this discussion we have the following parties in a generic OCI environment:

1) The bundle creator.
2) The container engine (such as docker).
3) The runtime.

The runtime *could* deal with its configuration using a standard application-specific configuration file. However, that might not be so trivial given:

- Different distros use different config file locations.
- Different platforms (`c:\some\where\ ?`)
- Read-only `/etc/`.

That aside, `runc` is a runtime, and it is being passed its configuration in `config.json`. Therefore, why should other runtimes not also have their configuration passed in the same manner? The only difference is that VM-based runtimes require a few extra configuration items. By ensuring that all runtime configuration comes from the container manager, potentially extra checks could be performed by the container manager before launching the runtime.

Further, these extra VM configuration details correlate to the extra files required by the VM-based runtimes. And those files will potentially need to be updated (new features, security fixes, etc). Hence, if we imagine that a particular bundle is unchanging for a period of time, the VM-configuration *might* change over that time frame. Since the pristine bundle should not be modified, we are left with either configuration the VM-based runtime via a runtime-specific configuration file, or at the container engine level. As explained, I'd vote for the container engine level since that (I believe) owns the config.json file at container startup and is responsible for modifying it as necessary before passing to the runtime.

If the top-level "`vm`" object is not palatable to the OCI, it would seem that the only alternative would be to make use of [Annotations](opencontainers/runtime-spec/blob/master/config.md#annotations). However, if with that approach, we will need some sort of namespacing (see https://github.com/docker/docker/issues/21324).

It would help immensely in my opinion if we could get some clarity on who actually "owns" the configuration file. As I've outlined on https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61550431, the config file ownership appears to be passed from the bundle owner to the container engine at container creation time. If my understanding is correct and given the above, I think there needs to be a way to pass runtime-specific configuration to the runtime via the container manager.

wking:
On Mon, Jun 06, 2016 at 06:50:52AM -0700, James Hunt wrote:
> If the top-level "`vm`" object is not palatable to the OCI, it would
> seem that the only alternative would be to make use of Annotations.

Another alternative is to have VM-based runtimes extend their API to
accept both a bundle/config path and these additional options.  This
is what happened with the previously-proposed hypervisorPath [1].
This approach requires callers to know about the API extentions, but
if they're the ones setting up the VM-oriented configuration
(e.g. kernel path), they probably will.

The distinction between “a runtime-specific option” and “an OCI-spec
config entry” is probably “does this meaninfully impact the
environment seen within the container?”.  And I think the kernel
you're using does [2], while the hypervisor you're using should not.

> It would help immensely in my opinion if we could get some clarity
> on who actually "owns" the configuration file. As I've outlined on
> https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61550431,
> the config file ownership appears to be passed from the bundle owner
> to the container engine at container creation time. If my
> understanding is correct and given the above, I think there needs to
> be a way to pass runtime-specific configuration to the runtime via
> the container manager.

The container engine is free to override whatever it wants in the
configuration file and also free to pass additional options to runtime
commands (just like you're free to patch your local copy of Nginx, or
adjust the stock Nginx config however you like, or pass Nginx
additional options).  However, providers offering containers as a
service should probably be documenting what they do [3], otherwise the
service will be difficult to use (client: “Why isn't my pre-start hook
getting executed?”.  Service: “Oh, we remove all user-provided
hooks”).

[1]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r62985654
[2]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-223678348
[3]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61617026

philips:
@jamesodhunt I apologize for the delay in replying. 

In my mind the Kernel, initrd, etc are not the first class citizens of the open container runtime spec because the spec concerns itself with application processes. The way that process is containerized is left to the runtime to sort out on its own.

To put it another way: The spec drops details that aren't directly related to the container processes configuration or lifecycle are dropped where necessary. For example details about the network configuration and isolation aren't configured by the OCI runtime but instead relies on an existing network setup. Or another example: the container process might be an init process or child of an init process.

In the case of a VM executing a container process I see it as part of the environment details and not necessarily the container process configuration or lifecycle.

Another way to think about it: perhaps the Kernel, initrd, etc are just shared libraries for this particular runtimes isolation method.

I understand this line is fuzzy here but that is my point of view on the subject.

jamesodhunt:
HI @philips 

> The way that process is containerized is left to the runtime to sort out on its own.

I'm struggling to understand this comment since `runc` for example is told very directly how to containerize the process via the namespace and cgroup configuration details. Admittedly, those are Linux-specific and hence are defined in `config-linux.md`. By the same token, I created `config-vm.md` to make it clear the VM configuration only applies in certain circumstances: the proposed VM configuration data is optional so will not add any overhead for non-VM-based runtimes.

Ultimately, this PR is trying to:

- avoid runtimes having to use multiple configuration files when a single one -- `config.json` -- should be sufficient.
- ensure that all runtimes are treated as first-class citizens.

sameo:
@philips Out of curiosity, when you write: 
> In rkt when someone wants to run their container in a VM
> they just add a flag that changes the execution to a VM. 

the flag you're talking about is the `--stage1` one for passing e.g. an lkvm based stage1 ACI ?

cyphar:
> I'm struggling to understand this comment since runc for example is told very directly how to containerize the process via the namespace and cgroup configuration details.

The cgroup details are resource control knobs. Admittedly, they're written to be just as granular as cgroups, but that's so that users have maximum control of the resource restrictions. There's no option called "cgroup", just as there's no option called "root filesystem disk image".

Namespaces are a bit more of a mixed bag (though without that option you couldn't effectively use rootless containers). However, given the fact that some namespaces have certain operational and security concerns it is understandable that they are optional.

HuKeping:
> the flag you're talking about is the --stage1 one for passing e.g. an lkvm based stage1 ACI ?

I think so @sameo

vbatts:
From F2F: Regardless of the overlap with the stage1 concept, this type of state information seems feasible to included. @stevvooe recommends checking the MS hyperv way of describing this information so we can re-use code/concepts.

tianon:
Has this seen any progress since @vbatts summary of the F2F discussion?  Should we close this and open an issue tracking the status of this addition instead?

config-vm.md
```diff
@@ -0,0 +1,37 @@
...
+This field is distinct from the **`path`** field in the [Root Configuration](config.md#Root-Configuration) section since in the context of a virtual-machine-based runtime:
+
+* **`imagePath`** will represent the root filesystem for the virtual machine.
+* The container root filesystem specified by **`path`** from the [Root Configuration](config.md#Root-Configuration) section will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
+
+The virtual-machine-based runtime will use these two path fields to arrange for the **`path`** from the [Root Configuration](config.md#Root-Configuration) section to be presented to the process to run as the root filesystem.
+
+## Kernel
+
+Used by virtual-machine-based runtimes only.
```
cyphar: Maybe you should specify that these are all host paths (or if they could be other paths, you could make it if they're absolute paths then they're host paths).

config-vm.md
```diff
@@ -0,0 +1,37 @@
...
+## Example of a fully-populated `VM` object
+
+```json
+"vm": {
+    "imagePath": "path/to/rootfs.img",
+    "kernel": {
+        "path": "path/to/vmlinuz",
+        "parameters": "foo=bar hello world",
+        "initrd": "path/to/initrd.img"
+    },
```
cyphar: Surely these examples should use absolute paths?

specs-go/config.go
```diff
@@ -20,13 +20,32 @@ type Spec struct {
...
+}
+
+// Kernel contains information about the kernel to use for a virtual machine.
+type Kernel struct {
+	// Path is the path to the kernel used to boot the virtual machine.
+	Path string `json:"path"`
+	// Parameters specifies parameters to pass to the kernel.
+	Parameters string `json:"parameters,omitempty"`
+	// InitRd is the path to an initial ramdisk to be used by the kernel.
+	InitRd string `json:"initrd,omitempty"`
```
cyphar: This should either be `Initrd` or `InitRD`.

specs-go/config.go
```diff
@@ -20,13 +20,32 @@ type Spec struct {
...
+// VM contains information for virtual-machine-based runtimes.
+type VM struct {
+	// Kernel specifies kernel-related configuration for virtual-machine-based runtimes.
+	Kernel Kernel `json:"kernel"`
+	// ImagePath is the path to the root filesystem image on the host which can be used by a virtual-machine-based runtime.
+	ImagePath string `json:"imagePath"`
+}
+
+// Kernel contains information about the kernel to use for a virtual machine.
+type Kernel struct {
```
cyphar: Given the recent aggressive namespacing PR this should be `VMKernel`.

specs-go/config.go
cyphar: Also, I forgot to mention that I'm still iffy about making the `initrd` a first-class citizen (but at least it's optional).

config-vm.md
wking: What, no bring-your-own-kernel? ;)  I think the semantics of relative paths can be made clear (relative to the bundle, to match the [pattern set by `root.path`][1] since #558).  Whether allowing that sort of thing is a good idea is a policy decision best left to higher levels.

[1]: https://github.com/opencontainers/runtime-spec/blob/41ef05e714f2e629796664aae640c777b2d34e9a/config.md#root-configuration

stevvooe:
Let's make sure we use the `platform` fields correctly. The `os` field is for the target kernel, usually. In this case, it would be a `vm` or a particular virtual machine target. The `variant` field, short for cpu variant, operates on `arch`, so using `vm` here is completely incorrect.

It should look something like this:

```
"platform": {
                "os": "vm",
                "arch": "amd64",
        }
```

wking:
On Thu, Nov 17, 2016 at 01:47:10PM -0800, Stephen Day wrote:
> "platform": {
>                 "os": "vm",
>                 "arch": "amd64",
>         }

For this to work, you'd need to be able to guess the guest OS so you
can slot in the appropriate handling for platform-specific fields
(e.g. process.user.uid [1] vs. plrocess.user.username [2]).  Once you
have the ability to do that, why don't we just drop ‘platform’
entirely and punt its handling to higher levels [3]?

[1]: https://github.com/opencontainers/runtime-spec/blob/7839cbb5dc30600f294fe8c31fdf635bddadff62/config.md#linux-and-solaris-user
[2]: https://github.com/opencontainers/runtime-spec/blob/7839cbb5dc30600f294fe8c31fdf635bddadff62/config.md#windows-user
[3]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-220697296

stevvooe:
@wking Please let the contributor respond. You've already made this point above.

specs-go/config.go
jodh-intel: Fixed.

config-vm.md
jodh-intel: Fixed.

specs-go/config.go
jodh-intel: Fixed.

jodh-intel:
Hi @stevvooe - again, this depends to some degree on docker/docker#21324 I think.

Taking the docker container manager as an example, it generates `config.json` (including `platform.*`) based on data from the docker image. That works perfectly right now for our runtime [1] since it means the runtime can run with any docker hub image where `platform.os` matches the O/S used by the VM and `platform.arch` matches the architecture used in the VM. If however `config.json` contained `platform.os = vm`, that would tie the container (and any images based off it) to some VM-based runtime (only) [2]. This is not desirable behaviour. Further, imagine a system with a docker daemon configured with multiple different VM-based runtimes - `vm` is not precise enough to know which VM-based runtime the container was created with. Also, the VM-based runtime would now need to guess the *actual* O/S and architecture rather than being told directly.

Also, I don't think we can expect all VM-based runtimes to be compatible. Since `parameters` and `initrd` are optional attempting to run a container using an image created with a different VM-based runtime is not guaranteed to succeed with the `platform.os = vm` strategy unless the container manager has full knowledge of how the configuration differs for each.

---
[1] - https://github.com/01org/clr-oci-runtime
[2] - I'm pre-supposing that docker in this example would retain the `platform.*` settings from `config.json` as a result of a `docker commit` followed by a `docker run` using the previously-committed image, but I don't know categorically if that would happen).

cyphar:
> vm is not precise enough to know which VM-based runtime the container was created with. Also, the VM-based runtime would now need to guess the actual O/S and architecture rather than being told directly.

If your VM configuration is not sufficient to be interoperable with more than one runtime then it's not good enough to be in the spec IMO. The whole point of this spec is to avoid problems like this. If you need to add more fields then please add them -- don't make it so that users can't be sure that anything will work with VM-based `config.json`s because there aren't enough JSON fields to describe how they want their images to run.

jodh-intel:
@cyphar - no further config is needed if `platform.os` retains its current meaning. `platform.os = vm` should not be necessary - we already have the `vm` object to describe that. Further, a VM-based runtime (like any other) needs to be able to interrogate `platform.*`.

stevvooe:
@jodh-intel I see. Re-wording your response, we want `platform.os` to always refer to the "guest", or the inside of the container, while some `vm` config configures the host or runtime. Retaining this meaning makes a lot of sense, if you plan for the container to run the same on a vm or vanilla container runtime.

Do we need to introduce an abstraction to configure the runtime, be it vm or otherwise?

config-vm.md
```diff
@@ -0,0 +1,37 @@
+# Virtual-machine-specific Configuration
+
+Virtual-machine-based runtimes require additional configuration to that specified in the [default spec configuration](config.md).
+
+This optional configuration is specified in a "VM" object:
+
+* **`imagePath`** (string, required) host path to file that represents the root filesystem for the virtual machine.
```
stevvooe: This might make sense as a structure, rather than just a path, to allow different kinds of images in the future.

RobDolinMS:
@jodh-intel Can you respond to @stevvooe's comments on this?

jodh-intel:
Hi @stevvooe - regarding a runtime configuration abstraction, that might make sense. For [cc-oci-runtime](https://github.com/01org/clr-oci-runtime), we are currently having to generate the `vm` object snippet but also require a file that contains the arguments for the hypervisor (https://github.com/01org/cc-oci-runtime#hypervisor-args). However, the configuration in that latter file also has to contain variables expanded by the runtime.

I think we may need input from other parties to decide upon a good way to represent all common use-cases, unless we simply introduce an object/array variable that can contain arbitrary configuration values for the runtime, something like `Annotations`, but which are private to the runtime and which are not passed to the container environment itself.

config-vm.md
jodh-intel: Hi @stevvooe - what members were you thinking would be in this structure initially aside from `path`? We could introduce an `imageType` possibly, but then the runtime could detect the type itself potentially from the file header. There isn't a `kernelType` member currently although arguably there could be, but the hypervisor we use detects the kernel type at runtime.

Also, if type members get added, presumably the behaviour would need to be documented should the type specified not match that detected by the runtime. But that implies the runtime would need to validate the type itself which makes me wonder if we anything beyond `imagePath`. Maybe such type members could be optional and default to `auto-detect` meaning the runtime is responsible for determining if it is a supported type?

stevvooe: @jodh-intel I am not quite sure here. This isn't exactly the right layer to be relying on autodetect. That seems like something that would belong at a higher-level (ie whatever creates this file). Containers are mostly different in that the rootfs is either pre-expanded or a few host mounts away.

With an image file, its quite different. Does this image file get mounted? How are the parameters for that mount resolved? Is `kernel.path` relative to the image root or to the host root? What happens to the regular container `rootfs`? Does it get replaced by this image?

hqhq:
Are we still considering this to be compliant with windows hyperv container? Otherwise current version generally looks good to me.

vbatts:
per the weekly discussion, we would like to see some commonality for VMs, but that would require some feedback from folks working on vm-containers. We need to have VM folks on the call (intel, vmware?, hypersh).

sameo:
@vbatts I'll attend. Is this the wednesday 2PM weekly call ?

vbatts:
2pm POST, yes

hqhq:
@opencontainers/runtime-spec-maintainers Can we starting reviewing this again? What's the major block here?

warmchang:
"This branch has conflicts that must be resolved", should rebase and commit again.

hqhq:
After reading through this thread, the only blocker I see is to see some commonality for VMs, and we had a weekly meeting to discuss on March, according to the minutes: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-03-15-21.02.log.html , @RobDolinMS have pinged some Windows folks to take a look, how's that going? @RobDolinMS 

Also ping @opencontainers/runtime-spec-maintainers how's other maintainers think on this? I'd like to push this as it makes sense and is very useful for container ecosystem, it's been too long without any progress.

sameo:
@hqhq with CC 3.0 use a libcontainer based agent on the VM where we're basically able to apply most of the OCI runtime resource constraints to the container inside the VM.
We will rebase this PR very shortly and may add it one of the next OCI meetings agenda to re-discuss it.

hqhq:
@sameo Thanks for your response, after more thoughts, I have a question about this PR.

If we add an entry to OCI config, what should happen if we try to validate or run a non-vm-based runtime (such like runc) with a config file which has a vm entry in it? Should we say the config file is invalid? As we have definition in [`runtime.md`](https://github.com/opencontainers/runtime-spec/blob/v1.0.1/runtime.md#create)

> All of the properties configured in config.json except for process MUST be applied. process.args MUST NOT be applied until triggered by the start operation. The remaining process properties MAY be applied by this operation. If the runtime cannot apply a property as specified in the configuration, it MUST generate an error and a new container MUST NOT be created.

Or vise-versa, what if we validate or run a vm-based runtime with a config without a vm entry? Since this entry is optional, runtime should run a container without an optional property, right?

I kind of understand @philips 's concern on this, spec should contain properties those are directly related to the container processes configuration. But I also think the container word in `Open Container Initiative` should not be tied to the concept of Linux container (namespace + cgroup), all isolated environment can be called `container`, so container can be more than processes.

Also ping @liangchenye for runtime-tools 's perspective on this PR.

liangchenye:
The runtime-tools will have four tests: 

| case | runtime | config with vm | expected result |
|-----| ------|----|------|
| 1 | runc | yes | yes |
| 2 | runc | no | yes |
| 3 | runv | yes | yes |
| 4 | runv | no | no |

Just as @hqhq refers, it is inconsistent with current spec (case 1).
`runc` could/should not apply `vm` config, so maybe we should change the spec to:
```
If the runtime cannot apply a required property as specified in the configuration,
it MUST generate an error and a new container MUST NOT be created.
```

Another problem is it is not possible to determine if a runtime is vm-based (case 4).
So runtime-tools should category runtimes and provide different test cases for vm-based runtime and non-vm-based runtime.

liangchenye:
Also, we might need to change image spec to add a 'vm' config.
An image fetcher needs to pull the right 'vm' as well as pull with the right arch and os.

sameo:
@hqhq 
> Or vise-versa, what if we validate or run a vm-based runtime with a config without a vm entry? Since this entry is optional, runtime should run a container without an optional property, right?

The VM section is optional so non VM based runtimes should simply ignore it. The same way I assume runc today is capable of creating and running a container based on a config.json even if that file contains both a `Windows` and `Linux` sections. It simply ignores the `Windows` one.

And VM based runtimes must support configs without a VM section today and will have to do so in the future, as it's an optional section.

>  I kind of understand @philips 's concern on this, spec should contain properties those are directly related to the container processes configuration. But I also think the container word in Open Container Initiative should not be tied to the concept of Linux container (namespace + cgroup), all isolated environment can be called container, so container can be more than processes.

Yes, I fully agree. A VM is one addtional, optional isolation layer for containers. And it defines how the process could run.
There's already another example for this in the current spec, in the `Windows` section:

```Go
	// HyperV contains information for running a container with Hyper-V isolation.
	HyperV *WindowsHyperV `json:"hyperv,omitempty"`
```
So it seems hardware virtualization is already an accepted isolation mechanism for containers.
BTW, Windows/MSFT folks should be involved in that discussion as the VM section we're trying to add would basically supersede the current `WindowsHyperV` section for Windows containers.

sameo:
@liangchenye 
> Also, we might need to change image spec to add a 'vm' config.
An image fetcher needs to pull the right 'vm' as well as pull with the right arch and os.

I would disagree with that. The container image is orthogonal to the VM it's going to run into, there should not be any link between a container image and a VM definition.

sameo:
@liangchenye 
> Another problem is it is not possible to determine if a runtime is vm-based (case 4).

Pretty much, yes. So I think case 4 should be positive, i.e. runv/clearcontainers should continue supporting config files without a VM section. The same way they support it today.

liangchenye:
> i.e. runv/clearcontainers should continue supporting config files without a VM section. The same way they support it today.

It means runc/cc has and will use its default VM section, right?

> I would disagree with that. The container image is orthogonal to the VM it's going to run into, there should not be any link between a container image and a VM definition.

Images follows this flow: `ocihub -- fetcher--- scheduler ---- runtime`.
So the scenario will be like this?
1. deploy a scheduler
2. add a runtime to the scheduler
3. tell the scheduler the runtime's VM status(if it is vm based)
4. the scheduler uses a fetcher to get an image from ocihub
5. scheduler add the VM section
6. runtime runs the bundle with VM secion

sameo:
@liangchenye 
> It means runc/cc has and will use its default VM section, right?

We don't append a VM section in the OCI config file, but we use configuration files for setting default kernel and guest image paths.

> * deploy a scheduler
> * add a runtime to the scheduler
> * tell the scheduler the runtime's VM status(if it is vm based)
Not sure you need to do this. From e.g. kubernetes, you could add specific annotations to your pod to specify a kernel and image path you want to use for this specific pod. The scheduler/orchestrator does not really need to know that the runtime is VM based, it will at some point add a VM section to the OCI config file based on the pod/workload metadata it received.

> * the scheduler uses a fetcher to get an image from ocihub
> * scheduler add the VM section
Yes, although I would replace scheduler with "one of the scheduler's component".

> * runtime runs the bundle with VM secion
Yes.

cyphar:
Has the plan for this changed since the announcement of https://katacontainers.io/ ? Personally I think Kata should inform the spec work.

vbatts:
https://github.com/opencontainers/runtime-spec/pull/949 was merged.

PR #379, opened 2016-04-12
Bump version for v0.5.0 release

vbatts:

PR #406, opened 2016-04-27
version: include version in produced docs

### LINKED ISSUE #398: Version not included on first page of runtime spec
When viewing a printed PDF or HTML version of the runtime spec, it is not clear what version (or date) you're looking at.  A possible fix would be to include the version (and date) in the title.  

For example:
* vX.Y.Z-dev during development
* vX.Y.Z (and date) when a release is created
* vX.Y+1.Z-dev after release

vbatts:
Fixes #398

This produces a markdown page as a header for the PDF and HTML pages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 output
 schema/validate
 code-of-conduct.md
+version.md
--- a/.tool/version-doc.go
+++ b/.tool/version-doc.go
@@ -1,0 +1,25 @@
+// +build ignore
+
+package main
+
+import (
+	"fmt"
+	"html/template"
+	"os"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+var markdownTemplateString = `
+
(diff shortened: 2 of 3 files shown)


wking:
On Wed, Apr 27, 2016 at 07:54:30AM -0700, Vincent Batts wrote:
“     A .tool/version-doc.go (25)”

This is depressingly large when an inline:

  $ sed -n 's/.*"ociVersion": *"\([^"]*\)",/**Specification Version:** *\1*/p' config.md

would do the same thing (assuming we update the example [1]).  But
maybe relying on Go is more portable than relying on sed?  On the
other hand, systems that have ‘make’ are likely to have ‘sed’.

Aside from that minor grump, 73ef76db looks good to me.

[1]: https://github.com/opencontainers/runtime-spec/pull/379#issuecomment-209062341

crosbymichael:
LGTM

plz rebase

PR #408, opened 2016-04-28
specs-go/config: Make Spec.Mounts omitempty

wking:
And use “MAY” (RFC 2119) instead of “can” in the Markdown docs.

Otherwise:

    $ ocitools generate --mount-cgroups=no --template <(echo {})
    $ grep mounts config.json
            "mounts": null,


vbatts:
LGTM. These commits can be squashed.

PR #410, opened 2016-04-28
[ Lifecycle ] Increase importance of logging errors

RobDolinMS:
"errors, if any, MAY be logged" --> "errors, if any, SHOULD be logged"


wking:
On Thu, Apr 28, 2016 at 03:08:11PM -0700, Rob Dolin (MSFT) wrote:
> "errors, if any, MAY be logged" --> "errors, if any, SHOULD be logged"

I don't think this is particularly important until we get more clarity
on how logged errors are accessed.  But ignoring that, SHOULD seems
like a better fit than MAY, so e07ce303 looks good to me.

crosbymichael:
LGTM

PR #411, opened 2016-04-29
Introducing Solaris in OCI

anuthan:
This PR introduces Solaris application container configuration parameters to OCI.

Signed-off-by: Abhijeeth Nuthan <abhijeeth.nuthan@oracle.com>

### DIFF
--- a/config-solaris.md
+++ b/config-solaris.md
@@ -1,0 +1,116 @@
+# Solaris Application Container Configuration
+
+Solaris application containers can be configured using the following properties, all of the below properties have mappings to properties specified under zonecfg(1M) man page, except milestone.
+The Solaris specification is entirely optional.
+
+## milestone
+The SMF(Service Management Facility) FMRI which should go to "online" state before we start the desired process within the container.
+
+**`milestone`** *(string, optional)*
+
+### Example
+```json
+"milestone": "svc:/milestone/container:default"
+```
+
+## limitpriv
+The maximum set of privileges any process in this container can obtain.
+The property should consist of a comma-separated privilege set specification as described in priv_str_to_set(3C) man page for the respective release of Solaris.
+
+**`limitpriv`** *(string, optional)*
+
+### Example
+```json
+"limitpriv": "default"
+```
+
+## maxShmMemory
+The maximum amount of shared memory allowed for this application container.
+A scale (K, M, G, T) can be applied to the value for each of these numbers (for example, 1M is one megabyte).
+Mapped to max-shm-memory in zonecfg(1M) man page.
+
+**`maxShmMemory`** *(string, optional)*
+
+### Example
+```json
+"maxShmMemory": "512m"
+```
+
+## cappedCPU
+Sets a limit on the amount of CPU time that can be used by a container.
+The unit used translates to the percentage of a single CPU that can be used by all user threads in a container, expressed as a fraction (for example, .75) or a mixed number (whole number and fraction, for example, 1.25).
+An ncpu value of 1 means 100% of a CPU, a value of 1.25 means 125%, .75 mean 75%, and so forth.
+When projects within a capped container have their own caps, the minimum value takes precedence.
+cappedCPU is mapped to capped-cpu in zonecfg(1M) man page.
+
+* **`ncpus`** *(string, optional)*
+
+### Example
+```json
+"cappedCPU": {
+        "ncpus": "8"
+}
+```
+
+## cappedMemory
+The physical and swap caps on the memory that can be used by this application container.
+A scale (K, M, G, T) can be applied to the value for each of these numbers (for example, 1M is one megabyte).
+cappedMemory is mapped to capped-memory in zonecfg(1M) man page.
+
+* **`physical`** *(string, optional)*
+* **`swap`** *(string, optional)*
+
+### Example
+```json
+"cappedMemory": {
+        "physical": "512m",
+        "swap": "512m"
+}
+```
+
+## Network
+
+### Automatic Network (anet)
+anet is specified as an array that is used to setup networking for Solaris application containers.
+The anet resource represents the automatic creation of a network resource for an application container.
+The zones administration daemon, zoneadmd, is the primary process for managing the container's virtual platform.
+One of the daemons is responsibilities is creation and teardown of the networks for the container.
+For more information on the daemon check the zoneadmd(1M) man page.
+When such a container is started, a temporary VNIC(Virtual NIC) is automatically created for the container.
+The VNIC is deleted when the container is torn down.
+The following properties can be used to setup automatic networks.
+For additional information on properties check zonecfg(1M) man page for the respective release of Solaris.
+
+* **`linkname`** *(string, optional)* Specify a name for the automatically created VNIC datalink.
+* **`lowerLink`** *(string, optional)* Specify the link over which the VNIC will be created.
+Mapped to lower-link in the zonecfg(1M) man page.
+* **`allowedAddress`** *(string, optional)* The set of IP addresses that the container can use might be constrained by specifying the allowedAddress property.
+If allowedAddress has not been specified, then they can use any IP address on the associated physical interface for the network resource.
+Otherwise, when allowedAddress is specified, the container cannot use IP addresses that are not in the allowedAddress list for the physical address.
+Mapped to allowed-address in the zonecfg(1M) man page.
+* **`configureAllowedAddress`** *(string, optional)* If configureAllowedAddress is set to true, the addresses specified by allowedAddress are automatically configured on the interface each time the container starts.
+When it is set to false, the allowedAddress will not be configured on container start.
+Mapped to configure-allowed-address in the zonecfg(1M) man page.
+* **`defrouter`** *(string, optional)* The value for the optional default router.
+* **`macAddress`** *(string, optional)* Set the VNIC's MAC addresses based on the specified value or keyword.
+If not a keyword, it is interpreted as a unicast MAC address.
+For a list of the supported keywords please refer to the zonecfg(1M) man page of the respective Solaris release.
+Mapped to mac-address in the zonecfg(1M) man page.
+* **`linkProtection`** *(string, optional)* Enables one or more types of link protection using comma-separated values.
+See the protection property in dladm(8) for supported values in respective release of Solaris.
+Mapped to link-protection in the zonecfg(1M) man page.
+
+#### Example
+```json
+"anet": [
+    {
+        "allowedAddress": "172.17.0.2/16",
+        "configureAllowedAddress": "true",
+        "defrouter": "172.17.0.1/16",
+        "linkProtection": "mac-nospoof, ip-nospoof",
+        "linkname": "net0",
+        "lowerLink": "net2",
+        "macAddress": "02:42:f8:52:c7:16"
+    }
+]
+```
--- a/config.md
+++ b/config.md
@@ -108,15 +108,17 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
 
 The user for the process is a platform-specific structure that allows specific control over which user the process runs as.
 
-#### Linux User
+#### Linux and Solaris User
 
-For Linux-based systems the user structure has the following fields:
+For Linux and Solaris based systems the user structure has the following fields:
 
 * **`uid`** (int, required) specifies the user id.
 * **`gid`** (int, required) specifies the group id.
 * **`additionalGids`** (array of ints, optional) specifies additional group ids to be added to the process.
 
 _Note: symbolic name for uid and gid, such as uname and gname respectively, are left to upper levels to derive (i.e. `/etc/passwd` parsing, NSS, etc)_
+
+_Note: For Solaris, uid and gid specify the uid and gid of the process inside the container and need not be same as in the host._
 
 ### Example (Linux)
 
@@ -151,6 +153,26 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
             "soft": 1024
         }
     ]
+}
+```
+### Example (Solaris)
+
+```json
+"process": {
+    "terminal": true,
+    "user": {
+        "uid": 1,
+        "gid": 1,
+        "additionalGids": [2, 8]
+    },
+    "env": [
+        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+        "TERM=xterm"
+    ],
+    "cwd": "/root",
+    "args": [
+        "/usr/bin/bash"
+    ],
 }
 ```
 
@@ -607,6 +629,29 @@ Here is a full example `config.json` for reference.
             "/proc/sysrq-trigger"
         ],
         "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
+    },
+    "solaris": {
+        "anet": [
+            {
+                "allowedAddress": "172.17.0.2/16",
+                "configureAllowedAddress": "true",
+                "defrouter": "172.17.0.1/16",
+                "linkProtection": "mac-nospoof, ip-nospoof",
+                "linkname": "net0",
+                "lowerLink": "net2",
+                "macAddress": "02:42:f8:52:c7:16"
+            }
+        ],
+        "cappedCPU": {
+            "ncpus": "0.8"
+        },
+        "cappedMemory": {
+            "physical": "1G",
+            "swap": "512m"
+        },
+        "maxShmMemory": "256m",
+        "limitpriv": "default",
+        "milestone": "svc:/milestone/container:default"
     },
     "annotations": {
         "key1": "value1",
--- a/schema/schema-solaris.json
+++ b/schema/schema-solaris.json
@@ -1,0 +1,36 @@
+{
+    "solaris": {
+        "description": "Solaris platform-specific configurations",
+        "id": "https://opencontainers.org/schema/bundle/solaris",
+        "type": "object",
+        "properties": {
+            "milestone": {
+                "id": "https://opencontainers.org/schema/bundle/solaris/milestone",
+                "type": "string"
+            },
+            "limitpriv": {
+                "id": "https://opencontainers.org/schema/bundle/solaris/limitpriv",
+                "type": "string"
+            },
+            "maxShmMemory": {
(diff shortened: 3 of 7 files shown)


vbatts:
nicely. On the whole, LGTM

config-solaris.md
```diff
@@ -0,0 +1,88 @@
+# Solaris Application Container Configuration
+
+Solaris application containers can be configured using the following properties, all of the below properties are the same as given in zonecfg(1M) man page, except milestone. The Solaris specification is entirely optional.
```
wking: I'm not clear on “entirely optional”.  If you're aiming that at runtime authors (e.g. “authors of Linux runtimes can ignore this”), I don't think we need to call it out explicitly (all platform-specific specification can be ignored by runtime targeting other platforms).  If it's targeting bundle authors (e.g. “all settings in this section are optional, so even if you're writing a bundle targeting Solaris, you can leave this section out”), then I'd rather that get spelled out in more detail.  For example:

**`solaris`** *(object, optional)* Solaris-specific configuration.  Even configurations targeting Solaris MAY leave this field unset.

config-solaris.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Example
+```json
+"max-shm-memory": "512m"
+```
+
+## capped-cpu
+Sets a limit on the amount of CPU time that can be used by a container. The unit used translates to the percentage of a single CPU that can be used by all user threads in a container, expressed as a fraction (for example, .75) or a mixed number (whole number and fraction, for example, 1.25). An ncpu value of 1 means 100% of a CPU, a value of 1.25 means 125%, .75 mean 75%, and so forth. When projects within a capped container have their own caps, the minimum value takes precedence.
+
+* **`ncpus`** *(string, optional)*
```
wking: Docs for **`capped-cpu`**?  For an example of this sort of nesting in the Linux docs, see [here][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config-linux.md#memory

config.md
```diff
@@ -108,16 +108,18 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
...
-For Linux-based systems the user structure has the following fields:
+For Linux and Solaris based systems the user structure has the following fields:
 
 * **`uid`** (int, required) specifies the user id.
 * **`gid`** (int, required) specifies the group id.
 * **`additionalGids`** (array of ints, optional) specifies additional group ids to be added to the process.
 
 _Note: symbolic name for uid and gid, such as uname and gname respectively, are left to upper levels to derive (i.e. `/etc/passwd` parsing, NSS, etc)_
 
+_Note: For Solaris, uid and gid specify the uid and gid of the process inside the container and need not be same as in the host._
```
wking: This is also true for Linux.  Instead of this note, I've filed #412 to make this explicit for both platforms.

config.md
```diff
@@ -608,6 +630,29 @@ Here is a full example `config.json` for reference.
         ],
         "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
     },
+    "solaris": {
```
wking: We need a better way to handle this, because putting a `solaris` section in a config that has `linux` in `platform.os` may be confusing.

config-solaris.md
anuthan: "entirely optional" means as you said, "solaris (object, optional) Solaris-specific configuration. Even configurations targeting Solaris MAY leave this field unset." 
Having said that, would it be more appropriate to the the above text in config.md as opposed to config-solaris.md and have a entry for each platform to say if the platform-specific section is required by the said platform.

config-solaris.md
anuthan: For every property , the documentation is zonecfg(1M) man page for that release of solaris, except "milestone" (which has no other docs apart from the explanation) . Same is mentioned at the top of the file. I did not see a need to call it out for each specific property.

schema/schema-solaris.json
```diff
@@ -0,0 +1,36 @@
...
+                "id": "https://opencontainers.org/schema/bundle/solaris/limitpriv",
+                "type": "string"
+            },
+            "max-shm-memory": {
+                "id": "https://opencontainers.org/schema/bundle/solaris/max-shm-memory",
+                "type": "string"
+            },
+            "capped-cpu": {
+                "id": "https://opencontainers.org/schema/bundle/solaris/capped-cpu",
+                "$ref": "defs.json#/definitions/mapStringString"
```
wking: This is accurate, but it could be more specific if you explicitly lists allowed keys.  That can happen with a [separate definition][1] or [inline][2].

[1]: https://github.com/opencontainers/runtime-spec/blob/e5e603f26fe8119b564e506d66e2b3426a68f25f/schema/schema-linux.json#L83-L86
[2]: https://github.com/opencontainers/runtime-spec/blob/e5e603f26fe8119b564e506d66e2b3426a68f25f/schema/schema-linux.json#L218-L243

specs-go/config.go
```diff
@@ -25,6 +25,8 @@ type Spec struct {
 
 	// Linux is platform specific configuration for Linux based containers.
 	Linux Linux `json:"linux" platform:"linux"`
+	// Solaris is platform specific configuration for Solaris containers.
+	Solaris Solaris `json:"solaris" platform:"solaris"`
```
wking: I'm not very familiar with Go structures, but I'd expect both `Linux` and `Solaris` should be pointers.

schema/schema-solaris.json
anuthan: This is just the initial PR, we will expand(more properties to come) and make it more structured in the future. For now, I think this would be fine.

config-solaris.md
wking: On Fri, Apr 29, 2016 at 11:06:00AM -0700, Abhijeeth Nuthan wrote:
> Having said that, would it be more appropriate to the the above text
> in config.md as opposed to config-solaris.md and have a entry for
> each platform to say if the platform-specific section is required by
> the said platform.

Yeah, putting something more structured in [1] would be good.  It
looks like you're softening the existing SHOULD to a MAY.  I think
that's also the case for Linux, but I'm not sure, and having each
field listed with our usual:

**`{key}`** *({type}, {when-needed})* {description}

would make it more obvious.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#platform-specific-configuration

config-solaris.md
wking: On Fri, Apr 29, 2016 at 11:09:04AM -0700, Abhijeeth Nuthan wrote:
> > +* **`ncpus`** *(string, optional)*
>
> For every property, the documentation is zonecfg(1M) man page…

Right.  I just meant “why list ncpus, but not capped-cpu?”.

schema/schema-solaris.json
wking: On Fri, Apr 29, 2016 at 11:15:24AM -0700, Abhijeeth Nuthan wrote:
> > +            "capped-cpu": {
> > +                "id": "https://opencontainers.org/schema/bundle/solaris/capped-cpu",
> > +                "$ref": "defs.json#/definitions/mapStringString"
>
> This is just the initial PR, we will expand(more properties to come)
> and make it more structured in the future. For now, I think this
> would be fine.

Sounds good to me.

config.md
anuthan: I was under the impression that we could have all platform specified as such in config.json . The runtime would only parse the required fields based on image and/or the platform section of the config.json.
We can refactor this however we see fit though. Need not necessarily be the way I suggested as above.

config-solaris.md
anuthan: We want to keep the configuration same for all our flavors of virtualization. It's been this way for eons. So just continuing down that path made logical sense.

config-solaris.md
anuthan: Sounds like we need a separate PR to address the restructuring, if required. :)

config.md
wking: On Fri, Apr 29, 2016 at 11:19:26AM -0700, Abhijeeth Nuthan wrote:
> > +    "solaris": {
> 
> I was under the impression that we could have all platform specified
> as such in config.json . The runtime would only parse the required
> fields based on image and/or the platform section of the
> config.json.  We can refactor this however we see fit though. Need
> not necessarily be the way I suggested as above.

For previous multi-platform config discussion, see #73, #74, and #76.
The consensus was “this should be handled in a higher level” [1,2].

[1]: https://github.com/opencontainers/runtime-spec/issues/73#issuecomment-171449101
[2]: https://github.com/opencontainers/image-spec

config-solaris.md
wking: On Fri, Apr 29, 2016 at 11:22:23AM -0700, Abhijeeth Nuthan wrote:
> We want to keep the configuration same for all our flavors of
> virtualization…

I don't think I'm suggesting anything major, just:

* **`capped-cpu`** *(object, optional)* The following properties can be used to cap CPU usage:
  * **`ncpus`** *(string, optional)*

config-solaris.md
wking: On Fri, Apr 29, 2016 at 11:28:12AM -0700, Abhijeeth Nuthan wrote:
> Sounds like we need a separate PR to address the restructuring, if required. :) 

I just filed #414.

config-solaris.md
vishh: If renaming is an option, why not just `cpus` :) ?

config-solaris.md
```diff
@@ -0,0 +1,88 @@
...
+
+**`max-shm-memory`** *(string, optional)*
+
+### Example
+```json
+"max-shm-memory": "512m"
+```
+
+## capped-cpu
+Sets a limit on the amount of CPU time that can be used by a container. The unit used translates to the percentage of a single CPU that can be used by all user threads in a container, expressed as a fraction (for example, .75) or a mixed number (whole number and fraction, for example, 1.25). An ncpu value of 1 means 100% of a CPU, a value of 1.25 means 125%, .75 mean 75%, and so forth. When projects within a capped container have their own caps, the minimum value takes precedence.
```
vishh: The difference between `ncpus` and `capped-cpu` is not clear by reading these comments. Can you explain based on an example with both values set, and/or provide links to additional documentation?

vishh:
@anuthan Overall LGTM. Is there an implementation of this Spec that users can play with today?

config-solaris.md
anuthan: Both are actually one and the same , its just that structurally we expect them to be this way for legacy reasons and we can't really change the names or the way they are represented.  The value of ncpus is the value of capped-cpu. 

Example: Set container to use 80% of cpu.
"cappped-cpu" {
    "ncpus": "0.8"
}

Link to zonecfg(1M) manpage for Solaris 11.2: https://docs.oracle.com/cd/E36784_01/html/E36871/zonecfg-1m.html
You can find all the properties explained. It changes for each release of Solaris, hence we ask user to refer to zonecfg(1M) for additional information.

vishh: Ok. Can you add the link to the PR itself?

config-solaris.md
anuthan: @wking : I'm assuming you are suggesting adding an (object, optional) for capped-cpu and capped-memory, etc. That would require rework for all the properties even in config-linux.md . I prefer not to do that in this PR.  

@vishh  : Sorry can't change names for legacy reasons.

config-solaris.md
wking: On Fri, Apr 29, 2016 at 01:02:28PM -0700, Abhijeeth Nuthan wrote:
“It changes for each release of Solaris, hence we ask user to refer to
zonecfg(1M) for additional information.”

My intial thought on reading this was that you'll want some
target-Solaris-version information in the config.  But previous
discussion of that sort of thing (e.g. #39) ended up getting punted
over to image-spec.  Still, I'd suggest explicitly pointing out the
instability in your first config-solaris.md section.

config-solaris.md
wking: On Fri, Apr 29, 2016 at 01:08:38PM -0700, Abhijeeth Nuthan wrote:
> > +* **`ncpus`** *(string, optional)*
>
> @wking : I'm assuming you are suggesting adding an (object,
> optional) for capped-cpu and capped-memory, etc. That would require
> rework for all the properties even in config-linux.md . I prefer not
> to do that in this PR.

I feel like we can work towards that goal in baby steps instead of
shifting everything at once.  But yeah, if it doesn't feel right to
you, we can certainly defer until someone is bothered enough to PR the
change.

anuthan:
@vishh : The implementation of this spec is actively being worked on. I cannot comment on the timeline when this would be available. All I can say is coming soon to a Solaris store near you. :)

anuthan:
@vbatts and @vishh : Thanks for the LGTMs

config-solaris.md
anuthan: @vishh : We try to keep the meaning of the properties same for all releases but the values that the properties can take are release specific. I do not want to put release specific info here in OCI. 
We can have certain keywords supported in one release but not the next/previous release. Giving a hypothetical example, "mac-address": "default", the "default" may be supported in Solaris 11.2 but need not be supported in future releases, so adding the link to Solaris11.2 would send the wrong information across. We advise the users to look at the man page for that specific release they are running on for valid properties values.  
Hope that clears things up.

@wking 
We had thought to introduce versioning to solaris object, but seemed unnecessary. Mainly because if we EOF or add properties we can up rev the ociVersion itself.

wking: On Fri, Apr 29, 2016 at 01:42:57PM -0700, Abhijeeth Nuthan wrote:
“We can have certain keywords supported in one release but not the
next/previous release… We had thought to introduce versioning to
solaris object, but seemed unnecessary. Mainly because if we EOF or
add properties we can up rev the ociVersion itself.”

I am not looking forward to an runtime-spec major bump because Solaris
is dropping “default” support from “mac-address” in a future release
;).  In fact, I don't think we want to major bump the runtime-spec
because of any per-platform change.  I think re-opening per-feature
versioning (see #15 and the JSON-LD discussion in [1]) would be much
less invasive approaches to handling that sort of backwards
incompatibility.

[1]: https://github.com/opencontainers/runtime-spec/pull/371#issuecomment-209684002
     Subject: config.md: Rename ociVersion -> ociRuntimeVersion

anuthan: I understand your concerns.
Let me just clarify what I said, might have confused you a bit.
1. The reason for not putting the a release specific link of a man page in this PR is to avoid unnecessary up-rev of versions.  That is, if we hypothetically remove support for "default" of "mac-address" , it would affect the spec.  No need to up-rev any versioning. 
2. If we remove a property/add property from the spec , it need not be even platform specific we still need to up-rev the spec.  I'm not saying we need to up-rev major/minor/patch but we still need to uprev. Whether it is platform specific version we need to up-rev or the ociVersion.  
Having said that, We should move this discussion of per platform specific versioning elsewhere as that is orthogonal to this PR.  
Thanks for raising your concerns. We should definitely discuss this further.

specs-go/config.go
```diff
@@ -358,6 +360,51 @@ type Seccomp struct {
...
+	Physical string `json:"physical,omitempty"`
+	Swap     string `json:"swap,omitempty"`
+}
+
+// Anet provides the specification for automatic creation of network resources for this container.
+type Anet struct {
+	// Specify a name for the automatically created VNIC datalink.
+	Linkname string `json:"linkname,omitempty"`
+	// Specify the link over which the VNIC will be created.
+	Lowerlink string `json:"lower-link,omitempty"`
```
crosbymichael: These need to be camel case and not include `-` to be consistent with the rest of the spec.

specs-go/config.go
```diff
@@ -358,6 +360,51 @@ type Seccomp struct {
...
+	// SMF FMRI which should go "online" before we start the container process.
+	Milestone string `json:"milestone,omitempty"`
+	// Maximum set of privileges any process in this container can obtain.
+	LimitPriv string `json:"limitpriv,omitempty"`
+	// The maximum amount of shared memory allowed for this container.
+	MaxShmMemory string `json:"max-shm-memory,omitempty"`
+	// Specification for automatic creation of network resources for this container.
+	Anet []Anet `json:"anet,omitempty"`
+	// Set limit on the amount of CPU time that can be used by container.
+	CappedCPU CappedCPU `json:"capped-cpu,omitempty"`
```
crosbymichael: camel case

crosbymichael:
Looks good overall but all your json fields need to be camel case and not include `-` because it makes it inconsistent with the spec as a whole.  

`max-shm-memory` to `maxShmMemory`

config-solaris.md
```diff
@@ -0,0 +1,88 @@
...
+"capped-memory": {
+        "physical": "512m",
+        "swap": "512m"
+}
+```
+
+## Network
+
+### Automatic Network (anet)
+anet is specified as an array that is used to setup networking for Solaris application containers. The anet resource represents the automatic creation of a network resource for an application container. When such a container is started, a temporary VNIC(Virtual NIC) is automatically created for the container. The VNIC is deleted when the container is torn down.
```
mrunalp: One sentence per line.

specs-go/config.go
anuthan: @crosbymichael  : As stated previously, the names of the properties which appear in the config.json , such as capped-cpu, capped-memory, lower-link, etc., have been part of Solaris virtualization for a long time. 
We prefer to keep the same name and structure to how it is defined in https://docs.oracle.com/cd/E36784_01/html/E36871/zonecfg-1m.html . 
Sorry if that ruins consistency.

config-solaris.md
anuthan: If this is the only change you request, I will be happy to do it later today.

specs-go/config.go
anuthan: Same as previous comment.

philips:
LGTM, I would like @anuthan to add a link or some explanatory text about the lifecycle of networking in Solaris is tied to a particular process as this is quite different than Linux where a majority of the expertise is coming from.

anuthan:
@philips : I have added text for life-cycle of networking and a pointer to our documentation, which is a man page of our administrative daemon. 

@crosbymichael : removed camel case.

crosbymichael:
LGTM

vbatts:
LGTM

anuthan:
@vbatts Thanks.

Also thanks to all who reviewed.

PR #412, opened 2016-04-29
config: Explicit container namespace for uid, gid, and additionalGids

wking:
In the degenerate case where the container does not create a user
namespace, the "container namespace" distinction is unimportant, but
the phrasing is still accurate (the container and runtime namespaces
are the same).

### DIFF
--- a/config.md
+++ b/config.md
@@ -113,9 +113,9 @@ The user for the process is a platform-specific structure that allows specific c
 
 For Linux and Solaris based systems the user structure has the following fields:
 
-* **`uid`** (int, required) specifies the user id.
-* **`gid`** (int, required) specifies the group id.
-* **`additionalGids`** (array of ints, optional) specifies additional group ids to be added to the process.
+* **`uid`** (int, required) specifies the user ID in the [container namespace][container-namespace].
+* **`gid`** (int, required) specifies the group ID in the [container namespace][container-namespace].
+* **`additionalGids`** (array of ints, optional) specifies additional group IDs (in the [container namespace][container-namespace]) to be added to the process.
 
 _Note: symbolic name for uid and gid, such as uname and gname respectively, are left to upper levels to derive (i.e. `/etc/passwd` parsing, NSS, etc)_
 
@@ -668,6 +668,7 @@ Here is a full example `config.json` for reference.
 }
 ```
 
+[container-namespace]: glossary.md#container-namespace
+[go-environment]: https://golang.org/doc/install/source#environment
 [runtime-namespace]: glossary.md#runtime-namespace
 [uts-namespace]: http://man7.org/linux/man-pages/man7/namespaces.7.html
-[go-environment]: https://golang.org/doc/install/source#environment


philips:
We should make some concrete note about userns in Linux either in the glossary or as an example here. I understand the context from the PR but the doc here is a bit insufficient.

wking:
I updated docs (both in the glossary and with an explicit UID example) with 8a567d972 → c87e78a18.

wking:
Rebased around #415 with c87e78a → 5ba66f9.

glossary.md
```diff
@@ -16,6 +16,15 @@ For example, namespaces, resource limits, and mounts are all part of the contain
...
 On Linux, a leaf in the [namespace][namespaces.7] hierarchy in which the [configured process](config.md#process-configuration) executes.
+For example, with a PID namespace hierarchy like:
+
+| PID namespace | Parent namespace | Runtime process PID | Container process PID |
+| ------------- | -----------------| ------------------- | --------------------- |
+| *A*           | -                | 123                 | 678                   |
+| *B*           | *A*              | 45                  | 90                    |
+| *C*           | *B*              | 0                   | 1                     |
+
+where the container process was launched in a new [PID namespace][pid_namespaces.7], namespace *C* (in which the container process is executing) would be the container PID namespace.
```
dqminh: im not sure if we want to have nested user namespace clarification / education here. Seems much better to link to the kernel pages instead http://man7.org/linux/man-pages/man7/user_namespaces.7.html

glossary.md
```diff
@@ -16,6 +16,15 @@ For example, namespaces, resource limits, and mounts are all part of the contain
 ## Container namespace
 
 On Linux, a leaf in the [namespace][namespaces.7] hierarchy in which the [configured process](config.md#process-configuration) executes.
+For example, with a PID namespace hierarchy like:
```
vishh: This is confusing... Why don't we instead say "If the container init process pid is `100` in the runtime pid namespace, the container init process pid will be `0` inside the containers namespace"...

glossary.md
wking: On Wed, May 11, 2016 at 10:23:57AM -0700, Daniel, Dao Quang Minh wrote:
> im not sure if we want to have nested user namespace clarification /
> education here. Seems much better to link to the kernel pages
> instead http://man7.org/linux/man-pages/man7/user_namespaces.7.html

I don't currently talk about user namespaces at all in this entry, and
a link to namespaces(7) is already in this section.  The goal with the
changes here was to make “container namespace” a clearer idea.  If the
previous docs accomplished that, then I'm fine dropping this addition
;).

glossary.md
wking: On Wed, May 11, 2016 at 10:51:37AM -0700, Vish Kannan wrote:
> > +For example, with a PID namespace hierarchy like:
>
> This is confusing... Why don't we instead say "If the container init
> process pid is `100` in the runtime pid namespace, the container
> init process pid will be `0` inside the containers namespace"...

I can replace the table with words, but I think it's useful to have
the example distinguish between the root namespace and the runtime
namespace.  In the current table, there's the root PID namespace (A),
the runtime PID namespace (B), and the container PID namespace (C).

wking:
On Wed, May 04, 2016 at 10:23:31AM -0700, Brandon Philips wrote:
> We should make some concrete note about userns in Linux either in
> the glossary or as an example here. I understand the context from
> the PR but the doc here is a bit insufficient.

After a few unappealing attempts at improved docs, we backed off of
this in today's meeting [1].  I rerolled back to the initial commit
and rebased it onto master with 5ba66f9 -> 8a567d9.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-06-01-17.01.log.html#l-46

wking:
On Wed, Jun 01, 2016 at 08:56:25PM -0700, W. Trevor King wrote:
> I rerolled back to the initial commit and rebased it onto master
> with 5ba66f9 -> 8a567d9.

Oops, 8a567d9 was before the rebase onto the current master.  I just
pushed 8a567d9 → 08908d6 with the rebase.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #414, opened 2016-04-29
config: Split platform-specific configuration into its own section

wking:
To make it clear that the whole `linux` section is optional.

See discussion [here][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/411/files/d57479d6f48fc50f584d566c26c9dc06da140946#r61618585


crosbymichael:
LGTM

PR #417, opened 2016-05-01
State the default user a container process runs

amitsaha:
This patch adds information about the default user/group a process runs on Linux. 

Signed-off-by: Amit Saha <amitsaha.in@gmail.com>

### DIFF
--- a/config.md
+++ b/config.md
@@ -154,6 +154,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
 }
 ```
 
+If left empty, the user id and group id both defaults to 0 (root), and no additional group IDs are added.
 
 ## Hostname
 


vbatts:
SGTM, but the commit is not signed (try `git commit -s --amend .`)
Also, as a nit on the commit message, the subject line more appropriately be "config.md: args is an array of strings"

amitsaha:
Thanks @vbatts , done.

wking:
On Sun, May 01, 2016 at 02:17:13AM -0700, Amit Saha wrote:
> This patch adds information about the default user/group a process
> runs on Linux.

I'm not sure this default is a good idea.  For example, a config that
does not create a new user namespace may want the process to run as
the current user/group instead of as root.  I think it's better to
follow unshare(1) and require folks who want a process to run as root
to specify that explicitly in their config.

  $ unshare --user id
  uid=65534(nobody) gid=65534(nobody) groups=65534(nobody)
  $ unshare --user --map-root-user id
  uid=0(root) gid=0(root) groups=0(root),65534(nobody)

Wording for that approach would be something like (also pulling in
#412):

  * **`uid`** (int, required) specifies the user ID in the container
    namespace.  If unset, the runtime will not attempt to manipulate
    the user ID (e.g. not calling setuid(2) or similar).

amitsaha:
@wking But that *is* what is the default now, isn't it?

wking:
On Sun, May 01, 2016 at 03:21:16PM -0700, Amit Saha wrote:
> @wking But that *is* what is the default now, isn't it? 

It's not in the spec, so the current default is “undefined” ;).
Bundle authors who expect this behavior may get it with runC (I'm not
sure), but may see different behavior on other runtimes (e.g. ccon-oci
will not call setuid [1]).  If we want to make the “default to root”
behavior explicit, I'd rather see that also land with motivation
explaining why it's a better approach than “do nothing by default”
[2].

[1]: https://github.com/wking/ccon/tree/v0.3.0/examples/good/oci/0.1.0
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/DWdystx5X3A
     Subject: Exposing platform defaults
     Date: Thu, 14 Jan 2016 15:36:26 -0800
     Message-ID: <20160114233625.GN6362@odin.tremily.us>

philips:
I think the current text that makes it required is best. Implied behavior that goes straight to the highest privledge level isn't super helpful.

crosbymichael:
Because we have uid/gid required in the spec we do not need to talk about what the default is.  We made it required so we don't have to specify defaults for this.

amitsaha:
PR #425

PR #418, opened 2016-05-02
runtime: Explicitly allow 'start' to not validate config.json

wking:
This spec places RFC-2119 requirements on both bundles (bundle.md,
config.md, ...)  and runtimes (runtime.md, runtime-linux.md).  While
it's possible to envision a system where bundle validation is required
before container setup begins, it is also possible to decoupled
validation and allow the runtime to blindly stumble through as far as
it can.

We [already link to ocitools and OCT for testing both runtimes and
bundles][1], so users interested in pre-start validation can use those
tools.

This commit explicitly documents the non-requirement and links to
those tools, to make life less surprising for everybody.

This is the decoupled alternative to #404.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/implementations.md#testing--tools


mrunalp:
LGTM

crosbymichael:
LGTM

PR #423, opened 2016-05-04
*: Remove "bundle" and go straight to config.json

wking:
I think [the root filesystem should be optional][1], but even folks who disagree on that point have come to the conclusion that it doesn't need to be in the bundle (#389, #394).  #394 seems partially unfinished, with “[MUST be included][2]” and “[single directory][3]” wording still applying to the root filesystem, but I think the intention was clear (why allow absolute paths if the root filesystem must still be in the bundle directory?).  Once you relax the “bundle must contain the root filesystem” requirement, the only thing that the bundle must contain is config.json.  It doesn't seem to be worth the trouble to name a “bundle” construct if its only meaning is “the directory that holds config.json”, so this commit removes all remaining references to the term “bundle”.

We use the word “bundle” a lot, so this PR is split up into a few commits.  The earlier commits stand alone (at least, to my eye), and I expect they, or something like them, should land regardless of how the bundle removal goes down.  I'm happy to spin them off into separate PRs if that makes them easier to review.  Once the early commits are out of the way, the final commit clears out any remaining “bundle” wording.  More detailed motivation for each commit is in the commit message.

[2]: https://github.com/opencontainers/runtime-spec/blame/cbea66a27544683a819ef56ea1b89d4b11ed819b/bundle.md#L20
[3]: https://github.com/opencontainers/runtime-spec/blame/cbea66a27544683a819ef56ea1b89d4b11ed819b/bundle.md#L23


vbatts:
...

wking:
The bottom two commits from the original 42fbdb8 have landed (with
some other changes via #435.  With b08f2f4, I've rebased this PR onto
master to build on the version that landed in #435.

I've also spun off the new root commit into it's own PR (#451), since
“split off the bottom chunk” approach worked well in #435.  If this PR
is still to big to swallow, folks can review #451 instead.

wking:
And I've spun the config-linux commit off into #452.

crosbymichael:
We discussed this on the call and all said that there is benefits of keeping the bundle and we don't want to remove it.

wking:
On Tue, May 24, 2016 at 03:01:14PM -0700, Michael Crosby wrote:
> We discussed this on the call and all said that there is benefits of
> keeping the bundle and we don't want to remove it.

There were certainly folks who wanted to keep it (most people who
chimed in?).  But the end result (at least as I recorded it in the
minutes) was “needs more discussion in the PR” [1].

Let me pull out the the schema commit into it's own PR, and we can
take another look at this once it's down to just the final removal.
If we end up keeping the “bundle” word (which I don't think we want,
but that's currently not at consensus), I expect we'll still want at
least some parts of that last commit.

And this is rebased again with b08f2f4 → 39c3618 now that #451 and
#452 have landed.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-75

wking:
On Tue, May 24, 2016 at 03:32:00PM -0700, W. Trevor King wrote:
> Let me pull out the the schema commit into it's own PR…

Done with #453, and I tweaked the commit to also adjust the Solaris
schema (which landed via #411).  Rebased the final bundle-dropping
commit onto #453 with 39c3618 → 5a3bda6.

wking:
On Tue, May 24, 2016 at 03:41:18PM -0700, W. Trevor King wrote:
> Rebased the final bundle-dropping commit onto #453 with 39c3618 →
> 5a3bda6.

And I adjusted that tip commit to remove some new bundle references
with 5a3bda6 → bc306b5.

crosbymichael:
-1

dpc:
Hi. So I am a random dev that is doing something using `runc`, and this bundle thing doesn't click in my head. It doesn't seem to have any purpose **in this layer of abstraction**. I have a use-case that uses rootfs from outside of "bundle directory" and my initial prototypes seem to work.

So, is there any technical reason for requiring rootfs to be in the same directory? If not (and clearly it's very useful to be able to get it from the outside), and it's officially supported, then the spec file is all that matters, and I don't see why I need to put the spec file in a directory and have its name hardcoded.

I get that it might be useful during packaging, image managment and delivery, etc. to talk about a "bundle" of some kind, and have a standard format of it, but then again: that's another level of abstraction, above just "running a container".

PR #425, opened 2016-05-06
UID and GID for the main process are not optional

amitsaha:
The spec requires UID and GID to be specified, so we shouldn't ignore if they are not specified.

Signed-off-by: Amit Saha <amitsaha.in@gmail.com>


wking:
815ce4e2 looks good to me.

hqhq:
LGTM

crosbymichael:
LGTM

PR #427, opened 2016-05-06
config: Explicitly list 'hooks' as optional

wking:
And make it omitempty, otherwise:

```
$ cat config.json | jq -S .
{
  "hooks": {},
  ...
}
```

To provide space for the type information and `optional`, I've shuffled the hook docs to follow our usual:

```
* **`{property}`** ({type}, {when-needed}) {notes}
```

format.  I've kept the separate event-trigger sections (e.g. `### Prestart`) since they go into more detail on the timing, purpose, and exit handling for the different events (and that seemed like too much information to put into the nested lists).

This may be unnecessary if #384 kills hooks, although:
1. I'm not sure if the `run` command will keep hooks, and
2. The current #384 tip [still has these hook docs](https://github.com/opencontainers/runtime-spec/blob/0430aaf1dc22da65b66c28e8ba7ffb5ed912774a/config.md#hooks).

### DIFF
--- a/config.md
+++ b/config.md
@@ -125,9 +125,7 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
   * **`width`** (uint, REQUIRED)
 * **`cwd`** (string, REQUIRED) is the working directory that will be set for the executable.
   This value MUST be an absolute path.
-* **`env`** (array of strings, OPTIONAL) contains a list of variables that will be set in the process's environment prior to execution.
-  Elements in the array are specified as Strings in the form "KEY=value".
-  The left hand side MUST consist solely of letters, digits, and underscores `_` as outlined in [IEEE Std 1003.1-2001](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html).
+* **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) executable to launch and any flags as an array.
   The executable is the first element and MUST be available at the given path inside of the rootfs.
   If the executable path is not an absolute path then the search $PATH is interpreted to find the executable.
@@ -762,6 +760,7 @@ Here is a full example `config.json` for reference.
 
 [container-namespace]: glossary.md#container-namespace
 [go-environment]: https://golang.org/doc/install/source#environment
+[ieee-1003.1-2001-xbd-c8.1]: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html#tag_08_01
 [runtime-namespace]: glossary.md#runtime-namespace
 [uts-namespace]: http://man7.org/linux/man-pages/man7/namespaces.7.html
 [mount.8-filesystem-independent]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS
--- a/config.md
+++ b/config.md
@@ -310,19 +310,24 @@ For Windows based systems the user structure has the following fields:
 
 ## Hooks
 
-**`hooks`** (object, OPTIONAL) configures callbacks for container lifecycle events.
 Lifecycle hooks allow custom events for different points in a container's runtime.
-Presently there are `Prestart`, `Poststart` and `Poststop`.
 
-* [`Prestart`](#prestart) is a list of hooks to be run before the container process is executed
-* [`Poststart`](#poststart) is a list of hooks to be run immediately after the container process is started
-* [`Poststop`](#poststop) is a list of hooks to be run after the container process exits
+* **`hooks`** (object, OPTIONAL) MAY contain any of the following properties:
+  * **`prestart`** (array, OPTIONAL) is an array of [pre-start hooks](#prestart).
+    Entries in the array contain the following properties:
+    * **`path`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execv`'s *path*][ieee-1003.1-2001-xsh-exec].
+      This specification extends the IEEE standard in that **`path`** MUST be absolute.
+    * **`args`** (array of strings, REQUIRED) with the same semantics as [IEEE Std 1003.1-2001 `execv`'s *argv*][ieee-1003.1-2001-xsh-exec].
+    * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
+    * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
+  * **`poststart`** (array, OPTIONAL) is an array of [post-start hooks](#poststart).
+    Entries in the array have the same schema as pre-start entries.
+  * **`poststop`** (array, OPTIONAL) is an array of [post-stop hooks](#poststop).
+    Entries in the array have the same schema as pre-start entries.
 
 Hooks allow one to run programs before/after various lifecycle events of the container.
 Hooks MUST be called in the listed order.
-The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
-
-Hook paths are absolute and are executed from the host's filesystem in the [runtime namespace][runtime-namespace].
+The [state](runtime.md#state) of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
 
 ### Prestart
 
@@ -373,11 +378,6 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
         ]
     }
 ```
-
-`path` is REQUIRED for a hook.
-`args` and `env` are OPTIONAL.
-`timeout` is the number of seconds before aborting the hook.
-The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://golang.org/pkg/os/exec/#Cmd).
 
 ## Annotations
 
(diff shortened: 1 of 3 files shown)


config.md
```diff
@@ -229,17 +229,23 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
...
 
-* [`Prestart`](#prestart) is a list of hooks to be run before the container process is executed
-* [`Poststart`](#poststart) is a list of hooks to be run immediately after the container process is started
-* [`Poststop`](#poststop) is a list of hooks to be run after the container process exits
+* **`hooks`** (object, optional) MAY contain any of the following properties:
+  * **`prestart`** (array, optional) is an array of [pre-start hooks](#prestart).
+    Entries in the array contain the following properties:
+    * **`path`** (string, required) with the same semantics as Go's [`Cmd.Path`][go-cmd].
+      Hook paths are absolute and are executed from the host's filesystem in the [runtime namespace][runtime-namespace].
+    * **`args`** (array of strings, optional) with the same semantics as Go's [`Cmd.Args`][go-cmd].
```
crosbymichael: Don't make this go specific, its args and env like every args and env ever

wking: On Fri, May 06, 2016 at 10:36:29AM -0700, Michael Crosby wrote:
> > +    * **`path`** (string, required) with the same semantics as Go's [`Cmd.Path`][go-cmd].
> > +      Hook paths are absolute and are executed from the host's filesystem in the [runtime namespace][runtime-namespace].
> > +    * **`args`** (array of strings, optional) with the same semantics as Go's [`Cmd.Args`][go-cmd].
> 
> Don't make this go specific, its args and env like every args and env ever

The Go reference is inherited from 48049d2 (Clarify the semantics of
hook elements, 2015-11-25, #255).  I'm fine changing it to an
execve(2) reference or some such if you'd rather, but folks might
consider that platform-specific.  POSIX also has execve(3p) [1], but
it's in unistd.h, which may be POSIX-specific.  Is there a generic C
spec that covers this idea?

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/fexecve.html

wking:
This got some [airtime in Wendesday's meeting][1], with the consensus being:

* Don't reference Go (fine with me).
* Do reference POSIX (also fine with me), but phrase it so the POSIX reference only applies to POSIX systems (I [thought this would be ok][3], but no longer like it).

I've just pushed cb4cfd8 → a77a35a where I try to stick to the letter of that consensus (also updating the docs for **`process.args`** and **`process.env`** for consistency).  However, now that I see it in action, I'd much rather make remove the “platform-appropriate way” wiggle words and point to POSIX for all platforms.

Attempting to dig into Windows support (@RobDolinMS, please correct me if I'm getting this wrong), it seems like Visual Studio 2015 [does not support POSIX's `execv` and similar][4] but [does support C++ analogs in `_execv` and similar][5].  Those look like drop-in replacements to me (although the docs are very sparse).  And the [compatibility notes discuss the leading underscore and say][6]:

> Only the default POSIX names are deprecated, not the functions

More generally, the compatibility notes say:

> The UCRT also implements a large subset of the POSIX.1 (ISO/IEC 9945-1:1996, the POSIX System Application Program Interface) C library, but is not fully conformant to any specific POSIX standard.

which sounds like enough grounds for POSIX links here.  If that sounds plausible, I can try to dig up a linkable version of the 1996 standard (although [it's been withdrawn][7]).  Or we can keep linking to the 2001/2004 version and assume there haven't been serious changes at the basic level we're referencing (seems plausible based on the [history][8], although I haven't tracked down the relevant System V docs).

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-44
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-54
[3]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-55
[4]: https://msdn.microsoft.com/en-us/library/ms235416.aspx
[5]: https://msdn.microsoft.com/en-us/library/886kc0as.aspx
[6]: https://msdn.microsoft.com/en-us/library/sk54f3f5.aspx
[7]: http://www.iso.org/iso/catalogue_detail.htm?csnumber=24426
[8]: http://pubs.opengroup.org/onlinepubs/009695399/functions/exec.html#tag_03_130_11

wking:
Short form of the previous comment:

I'd like to drop the “platform-appropriate” wiggle and just require POSIX.  I think that's ok based on the age of the `execv` interface and apparent Windows (and cross-platform Go) support.  I'm also ok keeping the old Go links, since Go covers the obvious platforms.  Can I do either of those, or does everyone else really want the wordy “platform-appropriate” wiggle?

wking:
With a77a35a → b0f0096 I just went ahead and dropped the “platform-appropriate” wiggle.  Details in each commit message with links to Windows' Visual Studio 2015 supporting the POSIX semantics.

wking:
Now that we've closed #483 and are aiming for a certifiable/testable hook specification, I think we want to land this PR or something like it.

wking:
Rebased onto master (no serious conflicts) with ad33f9c → d40e9b1.

vbatts:
Sorry, why is this needed?

wking:
On Fri, Aug 26, 2016 at 07:48:39AM -0700, Vincent Batts wrote:
> Sorry, why is this needed?

I've tried to motivate each change in its commit message.  Is there
something in those commit messages that is unclear?  Or something in
the diff that's not motivated by the commit message?

The motivation for omitempty is also copied from the commit message
[1] into the topic post for this PR [2].  The POSIX adjustments
happened later, in response to [3], and most of the rest of my
comments in this thread talk about how those commits evolved.  It's
probably easier to read the commit messages though, since that lets
you skip some of the evolution and get motivation for the end result.

[1]: https://github.com/opencontainers/runtime-spec/pull/427/commits/d40e9b17011262e6868162e1a42c6481cb0e7f70
[2]: https://github.com/opencontainers/runtime-spec/pull/427#issue-153483614
[3]: https://github.com/opencontainers/runtime-spec/pull/427#discussion_r62362761

crosbymichael:
LGTM

vbatts:
overall SGTM. The `./schema/config-schema.json` still has `hooks` as a required field.

wking:
On Mon, Aug 29, 2016 at 07:49:58AM -0700, Vincent Batts wrote:
> overall SGTM. The `./schema/config-schema.json` still has `hooks` as
> a required field.

Oops, fxed with d40e9b1 → abc704a.

wking:
Anything I can do to help this along?  It had an LGTM and an SGTM two
weeks back before a tiny rebase ;).

vbatts:
On 15/09/16 20:37 -0700, W. Trevor King wrote:
>Anything I can do to help this along?  It had an LGTM and an SGTM two
>weeks back before a tiny rebase ;).

Re-reading the patches, one thing I don't like about it is that it
doesn't directly convey the form. Just simply linking out to IEEE is a
punt, and if an implementer only has this document then it's not quite
helpful. 

It would be more helpful to convey more of the form, even if quoting
from the IEEE link and including the link for more reference.

wking:
On Fri, Sep 16, 2016 at 05:22:52AM -0700, Vincent Batts wrote:
> Re-reading the patches, one thing I don't like about it is that it
> doesn't directly convey the form. Just simply linking out to IEEE is a
> punt, and if an implementer only has this document then it's not quite
> helpful.

Punting was the plan ;).  When would an implementer have this document
handy but not be able to click on the links to POSIX?

> It would be more helpful to convey more of the form, even if quoting
> from the IEEE link and including the link for more reference.

We don't quote when we punt MUST, MAY, etc. to RFC 2119, or when we
punt the code of conduct to the TOB, or when we punt UTF-8 to its
spec, or when we punt SemVer to its spec, or when we punt the GOOS
listing to Go's spec, or when we punt mount options to mount(8), ….
So I'm fine not quoting here.  It would make our spec wordier and have
the usual not-DRY exposure to deviations.

But those are small downsides, so if for whatever reason you really
want me to inline some POSIX here I can do that.  How much would you
like?  Do you want me to follow the environment variable definition
[1] down into the Portable Character Set [2]?  Do you want me to keep
going as far as “It is unwise to conflict…” [1]?  From the exec docs
[3], do you want me to quote the whole definition for argv, file, and
path?  Should these excerpts go inline in hooks, or be pushed down to
an appendix or some such?

[1]: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html#tag_08_01
[2]: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap06.html#tagtcjh_3
[3]: http://pubs.opengroup.org/onlinepubs/009695399/functions/exec.html

vbatts:
Quite literally, having the example of "NAME=VALUE" demonstration with
referring the particulars to IEEE.

wking:
On Fri, Sep 16, 2016 at 12:30:11PM -0700, Vincent Batts wrote:
> Quite literally, having the example of "NAME=VALUE" demonstration
> with referring the particulars to IEEE.

Don't we already have that name=value example down here [1]?

[1]: https://github.com/wking/opencontainer-runtime-spec/blob/abc704a36ec6f3044e3c5ba7ff610bbec4f3639e/config.md#example-4

wking:
Changed `Spec.Hooks` to a pointer in abc704a → 9bd270e to avoid the issue from #569.

wking:
Anything I can do to help this along?

wking:
Rebased onto master resolving a few conflicts (mostly with #574 but also shifting some wording that had landed in #525).

hqhq:
LGTM

wking:
I've just pushed b281671 → a78f255 rebasing onto master (no conflicts) to avoid the [golint compilation failure][1].  @hqhq, can you re-LGTM?

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/167825089#L272

hqhq:
LGTM, thanks.

mrunalp:
LGTM

PR #428, opened 2016-05-06
Makefile: Replace vbatts/pandoc with a PANDOC variable

wking:
And use `command -v` instead of `which`.  Details in the commit messages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 
-DOCKER ?= $(shell which docker)
+DOCKER ?= $(shell command -v docker)
 # These docs are in an order that determines how they show up in the PDF/HTML docs.
 DOC_FILES := \
 	version.md \
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 
 DOCKER ?= $(shell command -v docker)
+PANDOC ?= $(shell command -v pandoc)
 # These docs are in an order that determines how they show up in the PDF/HTML docs.
 DOC_FILES := \
 	version.md \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,6 @@
 
-DOCKER ?= $(shell which docker)
+DOCKER ?= $(shell command -v docker)
```
vbatts: I suppose this is fine.

vbatts:
LGTM

Makefile
hqhq: @wking It looks like it should work fine, but I got error `make: command: Command not found`, I have to change it back to `which`, any insight on this? Maybe related to bash version?

Makefile
```diff
@@ -32,7 +33,7 @@ output/docs.pdf: $(DOC_FILES)
 	-v $(shell pwd)/:/input/:ro \
 	-v $(shell pwd)/output/:/output/ \
 	-u $(shell id -u) \
-	vbatts/pandoc -f markdown_github -t latex -o /$@ $(patsubst %,/input/%,$(DOC_FILES))
+	$(PANDOC) -f markdown_github -t latex -o /$@ $(patsubst %,/input/%,$(DOC_FILES))
```
laijs: I'm just curious before any test or investigation, $(PANDOC) is a **path to the pandoc executive** when default, but we expect **pandoc image** here, how this replacement works?

wking: On Thu, May 19, 2016 at 08:40:21PM -0700, Lai Jiangshan wrote:
> > -	vbatts/pandoc -f markdown_github -t latex -o /$@ $(patsubst %,/input/%,$(DOC_FILES))
> > +	$(PANDOC) -f markdown_github -t latex -o /$@ $(patsubst %,/input/%,$(DOC_FILES))
> 
> I'm just curious before any test or investigation, $(PANDOC) is a
> **path to the pandoc executive** when default, but we expect
> **pandoc image** here, how this replacement works?

Oops, I had missed this.  See #440 for some paths forward.

PR #429, opened 2016-05-07
specs-go/config.go: Make Process.Terminal omitempty

wking:
It's an optional setting, and this change will fix the wordy:

    $ grep terminal config.json
        "terminal": false,

Instead, `config.json` will not contain a `terminal` entry at all.


crosbymichael:
LGTM

hqhq:
LGTM

PR #430, opened 2016-05-07
specs-go/config.go: Make Root.Readonly omitempty

wking:
It's an optional setting, and this change will fix the wordy:

    $ grep readonly config.json
        "readonly": false,

Instead, `config.json` will not contain a `readonly` entry at all.


crosbymichael:
LGTM

hqhq:
LGTM

PR #432, opened 2016-05-12
[Config] Mount options fstab format is for Linux

RobDolinMS:
The fstab format (https://wiki.archlinux.org/index.php/Fstab) may be appropriate for Linux, but may not be appropriate for other OS platforms.  

This edit clarifies that mount options for Linux MUST be in fstab format but leaves the format of options as an open question for other operating systems to clarify.  

### DIFF
--- a/config.md
+++ b/config.md
@@ -46,7 +46,8 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
 * **`destination`** (string, required) Destination of mount point: path inside container.
 * **`type`** (string, required) Linux, *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). Windows: ntfs
 * **`source`** (string, required) a device name, but can also be a directory name or a dummy. Windows, the volume name that is the target of the mount point. \\?\Volume\{GUID}\ (on Windows source is called target)
-* **`options`** (list of strings, optional) in the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
+* **`options`** (list of strings, optional) in a platform-appropriate format for each operating system.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -46,7 +46,8 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
 * **`destination`** (string, required) Destination of mount point: path inside container.
 * **`type`** (string, required) Linux, *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). Windows: ntfs
 * **`source`** (string, required) a device name, but can also be a directory name or a dummy. Windows, the volume name that is the target of the mount point. \\?\Volume\{GUID}\ (on Windows source is called target)
-* **`options`** (list of strings, optional) in the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
+* **`options`** (list of strings, optional) in a platform-appropriate format for each operating system.
+For the Linux OS, if present, options MUST be in the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
```
wking: This looks fine to me, although it would be nice to specify what other platforms expect so authors know what to put in here.  @jlb13, @anuthan, can you clarify what Solaris wants here?

Also, I'd like to replace the Arch link with a [link to the man page][1], but I'm happy to file that in a subsequent PR if you don't want to roll it in here.

[1]: http://www.man7.org/linux/man-pages/man5/fstab.5.html

wking: And as a minor copy-edit, maybe “When **`platform.os`** is `linux`, any entries MUST be in the [fstab][1] format.”, to avoid folks associating “if present” with Linux, instead of associating it with the options.

[1]: http://www.man7.org/linux/man-pages/man5/fstab.5.html

jlb13: Solaris is fine with a format similar to the Linux one. I guess that could be "Linux or Unix", for what it's worth. I would support platform-specific mounts, I'm not sure I like trying to enforce it in a global scope.

RobDolinMS:
Replaced by PR #439

PR #431, opened 2016-05-07
specs-go/config: Make Linux and Solaris omitempty

wking:
Also document `solaris` in the platform-specific configuration
section.  Details in the commit messages.

PR #436, opened 2016-05-16
config.md: Remove 'solaris' from full example

wking:
This should have been part of 759ee79c (config: Add platform-specific entry for 'solaris', 2016-05-06, #431), since the example has `platform.os` set to `linux`.

There was some (brief) [discussion][1] of this point before the `solaris` section landed, but the “should only be set if” wording landed in parallel via b373a15 (config: Split platform-specific configuration into its own section, 2016-05-02, #414), and I'd forgotten to go back and apply that logic to #411.

Having a full Solaris example would be useful, but I think it should be a separate, Solaris-only example.

[1]: https://github.com/opencontainers/runtime-spec/pull/411#discussion_r61621001

### DIFF
--- a/config.md
+++ b/config.md
@@ -654,29 +654,6 @@ Here is a full example `config.json` for reference.
         ],
         "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
     },
-    "solaris": {
-        "anet": [
-            {
-                "allowedAddress": "172.17.0.2/16",
-                "configureAllowedAddress": "true",
-                "defrouter": "172.17.0.1/16",
-                "linkProtection": "mac-nospoof, ip-nospoof",
-                "linkname": "net0",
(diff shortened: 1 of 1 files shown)


mrunalp:
Yeah, it makes sense to create a different solaris example.

wking:
On Wed, May 18, 2016 at 07:12:13AM -0700, Mrunal Patel wrote:
> Yeah, it makes sense to create a different solaris example.

Should that happen in this PR?  I'm not familiar enough with Solaris
to know what reasonable values are for the rest of the config, but it
would be nice if we were copying the Solaris-specific block somewhere
new instead of deleting it here and then restoring it later.

crosbymichael:
LGTM

mrunalp:
LGTM

anuthan:
This is fine, I will work on the Solaris only example in the coming days.

PR #433, opened 2016-05-12
[Config] Mount destinations MUST NOT be nested in Windows

RobDolinMS:
This was raised during reviews with folks working on Windows Containers.  I suspect the prohibition against nested mount destinations applies across all operating systems, but if not please feel free to edit.

Signed-off-by: Rob Dolin <RobDolin@microsoft.com>

PR #437, opened 2016-05-18
[Config] Mount destinations MUST NOT be nested in Windows

RobDolinMS:
This was raised during reviews with folks working on Windows Containers.  

This squashes commits from PR #433

### DIFF
--- a/config.md
+++ b/config.md
@@ -44,6 +44,7 @@ The runtime MUST mount entries in the listed order.
 The parameters are similar to the ones in [the Linux mount system call](http://man7.org/linux/man-pages/man2/mount.2.html).
 
 * **`destination`** (string, required) Destination of mount point: path inside container.
+For the Windows operating system, one mount destination MUST NOT be nested within another mount.  (Ex: c:\\foo and c:\\foo\\bar).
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

wking:
bef035ce looks good to me.

mrunalp:
LGTM

wking:
I was thinking about this a bit more, and have a few questions:

1. At what level are nested mounts not supported?
   a. Is it a Windows-kernel restriction?  If so:

      i. Is the restriction because of a policy (e.g. “allowing nested
         mounts is an unacceptable foot-gun and will not be
         supported”), or
      ii. Is it because the kernel devs just haven't gotten around to
          adding support for it yet?

   b. If it is not a kernel restriction, is it a runtime restriction?
      If so, the same policy questions apply, as well as:

      i. Is the platform sufficiently open to support third-party
         runtimes (in which case those runtimes might support nested
         paths).

2. What counts as a nested path?  The order is significant [1], and
   I'm curious about the following cases:

   i. Mounting into a subdirectory (e.g. c:\foo followed by
      c:\foo\bar).
   ii. Mounting exactly over (e.g. c:\foo followed by c:\foo).
   iii. Mointing over a parent directory (e.g. c:\foo\bar followed by
        c:\foo).

Under a few of the (1) cases, I'd suggest relaxing this PR's MUST NOT
to a MAY NOT.

And I'm guessing that at least (2i) is illegal, but I'm not sure about
the others.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config.md#mounts

PR #439, opened 2016-05-18
[Config] Make mount options format per-platform

RobDolinMS:
The fstab format (https://wiki.archlinux.org/index.php/Fstab) may be appropriate for Linux, but may not be appropriate for other OS platforms.

This edit clarifies that mount options for Linux (and Solaris per @jlb13) MUST be in fstab format but leaves the format of options as open for other operating systems to clarify. 

This replaces PR #432

### DIFF
--- a/config.md
+++ b/config.md
@@ -46,7 +46,9 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
 * **`destination`** (string, required) Destination of mount point: path inside container.
 * **`type`** (string, required) Linux, *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). Windows: ntfs
 * **`source`** (string, required) a device name, but can also be a directory name or a dummy. Windows, the volume name that is the target of the mount point. \\?\Volume\{GUID}\ (on Windows source is called target)
-* **`options`** (list of strings, optional) in the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
+* **`options`** (list of strings, optional) in a platform-appropriate format for each operating system.
+  * For the Linux OS, if the options object is present, it MUST be in the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
+  * For the Solaris OS, if the options object is present, it MUST be in the the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
 
 ### Example (Linux)
 


wking:
On Wed, May 18, 2016 at 11:51:26AM -0700, Rob Dolin (MSFT) wrote:
> The fstab format (https://wiki.archlinux.org/index.php/Fstab) may be
> appropriate for Linux, but may not be appropriate for other OS
> platforms.

50532d04 looks fine to me.  I can file a change to link the man page
[1] in a subsequent PR [2].

[1]: http://www.man7.org/linux/man-pages/man5/fstab.5.html
[2]: https://github.com/opencontainers/runtime-spec/pull/432#discussion_r63112750

jlb13:
I don't think we should call out an explicit must for Solaris to use a format defined for Linux. My comment was meant to indicate something like the existing format would be fine for Solaris. This wasn't meant to map explicitly to a requirement related to fstab format (which seems to be related to vfstab in Solaris). I'd be happy just leaving the bit about it for Linux, and other platforms can define their own formats.

Speaking of, should each OS have its own definition in this space? It seems like we could either use the platform section or have a platform label as part of the mount specification.

wking:
On Wed, May 18, 2016 at 12:25:37PM -0700, Jesse Butler wrote:
> … which seems to be related to vfstab in Solaris…

Can you give a link to the vfstab docs?  Or whatever the spec is
behind Solaris' support for this field?

> Speaking of, should each OS have its own definition in this space?
> It seems like we could either use the platform section or have a
> platform label as part of the mount specification.

This came up briefly again in today's meeting.  Here's my recording of
@mrunalp's position [1].  I expect this falls (with process.uid and
such) under the “don't namespace for unspecified asthetic reasons”.
Although I'm probably biased, since I'd rather not namespace any
settings.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-101

jlb13:
I mentioned vfstab in passing, we would likely not use that. We would likely use the filesystem (or 'fs') resource in zonecfg for Solaris. This is discussed in our documentation in various places (This[1] seems to be about the best place to hang a hat that I can find at the moment).

If it's helpful, the abstract configuration looks something like this for a given mount:

  fs <identifier>:
	dir: <guest mnt point>
	special: <host path>
	type: <fs type>
	options: [<fs options, per type>]

[1] http://docs.oracle.com/cd/E53394_01/html/E54752/z.admin.ov-2.html#VLZSOz.admin.ov-26

wking:
On Wed, May 18, 2016 at 04:59:40PM -0700, Jesse Butler wrote:
> This[1] seems to be about the best place to hang a hat that I can find at the moment
> …
> [1] http://docs.oracle.com/cd/E53394_01/html/E54752/z.admin.ov-2.html#VLZSOz.admin.ov-26

That doesn't talk much about which (filesystem-dependent) options are
supported.  Stumbling around blindly for a but turned up mount(1m)
[1], which lists a few option ((no)devices, (no)exec, …).  This seems
similar to fstab(5)'s list of basic filesystem options [2], which
lists ro, rw, defaults, noauto, user, owner, … and points at mount(8)
and swapon(8) for more details, and mount(8) has the meat in [3,4].

[1]: http://docs.oracle.com/cd/E53394_01/html/E54764/mount-1m.html#REFMAN1Mmount-1m
[2]: http://www.man7.org/linux/man-pages/man5/fstab.5.html
[3]: http://www.man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS
[4]: http://www.man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-SPECIFIC_MOUNT%20OPTIONS

tianon:
Is this PR still relevant?

In current `master`, this appears to already be resolved sufficiently to cover the goals of this PR:

```markdown
* **`options`** (list of strings, OPTIONAL) Mount options of the filesystem to be used.
  Linux: [supported][mount.8-filesystem-independent] [options][mount.8-filesystem-specific] are listed in [mount(8)][mount.8].
```

wking:
On Fri, Nov 04, 2016 at 03:12:53PM -0700, Tianon Gravi wrote:
> Is this PR still relevant?

+1 to closing this as “fixed by #443”.

hqhq:
Fixed by #443

PR #440, opened 2016-05-20
Fix make docs

hqhq:
`vbatts/pandoc` is the image name, not path of pandoc command,
so I don't think it's necessary to be configurable.

cc @wking

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,9 @@ EPOCH_TEST_COMMIT := 78e6667ae2d67aad100b28ee9580b41b7a24e667
 default: docs
 
 .PHONY: docs
-docs: output/docs.pdf output/docs.html
+docs: output/runtime-spec.pdf output/runtime-spec.html
 
-output/docs.pdf: $(DOC_FILES)
+output/runtime-spec.pdf: $(DOC_FILES)
 	mkdir -p output/ && \
 	$(DOCKER) run \
 	-it \
@@ -34,7 +34,7 @@ output/docs.pdf: $(DOC_FILES)
 	-u $(shell id -u) \
 	vbatts/pandoc -f markdown_github -t latex -o /$@ $(patsubst %,/input/%,$(DOC_FILES))
 
-output/docs.html: $(DOC_FILES)
+output/runtime-spec.html: $(DOC_FILES)
 	mkdir -p output/ && \
 	$(DOCKER) run \
 	-it \
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
 
 DOCKER ?= $(shell command -v docker)
-PANDOC ?= $(shell command -v pandoc)
 # These docs are in an order that determines how they show up in the PDF/HTML docs.
 DOC_FILES := \
 	version.md \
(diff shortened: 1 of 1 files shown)


laijs:
LGTM

Makefile
```diff
@@ -1,6 +1,5 @@
 
 DOCKER ?= $(shell command -v docker)
-PANDOC ?= $(shell command -v pandoc)
```
laijs: it is better to keep PANDOC
`PANDOC ?= vbatts/pandoc`

wking:
On Thu, May 19, 2016 at 07:42:43PM -0700, Qiang Huang wrote:
> `vbatts/pandoc` is the image name, not path of pandoc command,
> so I don't think it's necessary to be configurable.

Ah, I had missed all the line continuations, thanks.  Why are we using
a pandoc image instead of a local command?  That local command could
always be ‘docker run … vbatts/pandoc’.

Makefile
vbatts: geez this is now a little messy. 
Perhaps this $(PANDOC) change best be reverted. Otherwise the whole command for the `docs` target will have to be overhauled to allow for a local pandoc binary.

wking: On Mon, May 23, 2016 at 12:23:18PM -0700, Vincent Batts wrote:
> Otherwise the whole command for the `docs` target will have to be
> overhauled to allow for a local pandoc binary.

How difficult would this be?  It looks like it's just moving
‘$(DOCKER) … -u $(shell id -u)’ into a PANDOC fallback (if the shell
can find a local ‘docker’ but not a local ‘pandoc’).

vbatts: and also making the path to the input and output files

wking: On Mon, May 23, 2016 at 12:57:53PM -0700, Vincent Batts wrote:
> > @@ -1,6 +1,5 @@
> >  
> >  DOCKER ?= $(shell command -v docker)
> > -PANDOC ?= $(shell command -v pandoc)
> 
> and also making the path to the input and output files

Ah, yes.  So easiest way to support local pandoc is probably leave
PANDOC variable here and make the docs rules conditional.  If PANDOC
is non-empty, use:

  $(PANDOC) -f markdown_github -t latex -o $@ $(DOC_FILES)

and if PANDOC is empty use the current $(DOCKER) invocation with the
hard-coded vbatts/pandoc image.

The alternative is to make variables for the input and output
directories and set them conditionally depending on whether PANDOC is
empty.

hqhq: Is it necessary to offer such flexibility? Or does pandoc version matters a lot?
I don't think we should make the Makefile complex to meet such needs, I think people who really want to use their own local pandoc can use it directly without Makefile.

wking: On Mon, May 23, 2016 at 08:21:25PM -0700, Qiang Huang wrote:
> Is it necessary to offer such flexibility? Or does pandoc version
> matters a lot?

The flexibility is just “I want to build the docs easily without
installing Docker”.  I don't think pandoc versions matter much for us.

> I don't think we should make the Makefile complex to meet such
> needs, I think people who really want to use their own local pandoc
> can use it directly without Makefile.

Sure, but supporting them in the Makefile isn't all that much more
complex.  See #448, which also makes the Docker fallback case drier
and adds error handling for folks who have neither Docker nor pandoc
installed locally.

hqhq: Fair enough, #448 looks good to me, I'll close this PR, thanks.

PR #443, opened 2016-05-23
Rewrite mounts description in config.md

hqhq:
Also apply one sentence per line rule for list items.

### DIFF
--- a/config.md
+++ b/config.md
@@ -24,8 +24,9 @@ For example, if an implementation is compliant with version 1.0.1 of the spec, i
 
 Each container has exactly one *root filesystem*, specified in the *root* object:
 
-* **`path`** (string, required) Specifies the path to the root filesystem for the container. A directory MUST exist at the path declared by the field.
-* **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container. Defaults to false.
+* **`path`** (string, required) Specifies the path to the root filesystem for the container.
+  A directory MUST exist at the path declared by the field.
+* **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container, defaults to false.
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


wking:
On Mon, May 23, 2016 at 12:06:27AM -0700, Qiang Huang wrote:
> 1. Apple one sentence per line rule

“Apple” → “Apply”.

And I'm not sure there's consensus about whether the rule applies to
list items.  I've been using single-sentence-lines in list items, but
unless there is a consensus it's probably not worth changing existing
instances unless you're also tweaking the content.

config.md
```diff
@@ -25,8 +25,9 @@ NOTE that there is no guarantee for forward or backward compatibility for versio
 
 Each container has exactly one *root filesystem*, specified in the *root* object:
 
-* **`path`** (string, required) Specifies the path to the root filesystem for the container. A directory MUST exist at the path declared by the field.
-* **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container. Defaults to false.
+* **`path`** (string, required) Specifies the path to the root filesystem for the container.
+  A directory MUST exist at the path declared by the field.
+* **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container, defaults to false.
```
wking: I liked “Defaults to false” better as its own sentence, but I expect folks will get the idea either way ;).

wking: And defaults are a general enough idea that it may make more sense to mark them up (e.g. “**`readonly`** (bool, default: false) If true, then …”).  So settings would either be required or have a default value.

config.md
```diff
@@ -44,10 +45,14 @@ The runtime MUST mount entries in the listed order.
...
-* **`options`** (list of strings, optional) in the fstab format [https://wiki.archlinux.org/index.php/Fstab](https://wiki.archlinux.org/index.php/Fstab).
+  For the Windows operating system, one mount destination MUST NOT be nested within another mount.  (Ex: c:\\foo and c:\\foo\\bar).
+* **`type`** (string, required) The filesystem type of the filesystem to be mounted.
+  Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  Windows: ntfs.
+* **`source`** (string, required) A device name, but can also be a directory name or a dummy.
+  Windows: the volume name that is the target of the mount point. \\?\Volume\{GUID}\ (on Windows source is called target).
+* **`options`** (list of strings, optional) Mount options of the filesystem to be used.
+  Linux: see the [mount man page](http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS) for more information.
```
wking: I'd rather link both the filesystem-independent and filesystem-specific sections to avoid giving the impression that only the filesystem-independent options are supported.  How about something like:

    Linux: [supported][mount.8-filesystem-independent] [options][mount.8-filesystem-specific] are listed in [mount(8)][mount.8].

config.md
hqhq: > I liked “Defaults to false” better as its own sentence

I think it makes sense to combine them in one sentence, and it looks simpler, I'll change if more people think we should keep its own sentence.

> And defaults are a general enough idea that it may make more sense to mark them up (e.g. “readonly (bool, default: false) If true, then …”). So settings would either be required or have a default value.

Since it's a general change for all such defaults usage in runtime-spec, I'll leave it as a separate discussion if needed.

config.md
hqhq: That's a fair point, I'll change as you suggested.

hqhq:
Updated.

wking:
1895e071d looks good to me (with the exception of the default issue
[1], and I'm fine spinning that out into a separate discussion [2]).
I like your new, specific commit-summary a lot more too :).

[1]: https://github.com/opencontainers/runtime-spec/pull/443#discussion_r64248005
[2]: https://github.com/opencontainers/runtime-spec/pull/443#discussion_r65162290

wking:
On Tue, May 31, 2016 at 11:10:29AM -0700, W. Trevor King wrote:
> 1895e071d looks good to me (with the exception of the default issue
> [1], and I'm fine spinning that out into a separate discussion [2]).

Is there any chance of this landing soon?  I want to extend the spec
to talk about the ‘bind’ type and ‘rbind’ options [1], neither of
which are covered by the mount(8) FILESYSTEM-* MOUNT OPTIONS sections.
They *are* covered as long options in mount(8), but I think we need
some wording around them (and maybe --make-shared, --make-slave,
--make-rshared, etc.?), and I don't want to workup a PR for that
before we've dropped the Arch wiki as a lower layer of this spec :p.

[1]: https://github.com/opencontainers/runtime-spec/blame/bb6925ea99f0e366a3f7d1c975f6577475ca25f0/config.md#L62

hqhq:
Rebased, ping @vbatts @mrunalp @crosbymichael

wking:
On Thu, Jul 21, 2016 at 11:40:59PM -0700, Qiang Huang wrote:
> Rebased…

1895e071 → 0e8e0690 looks fine to me.  Besides the context changes,
there's some ‘:’ → ‘,’ and a sentence break that look minor to me.

hqhq:
@wking I consisted some ':' and ',' usage.

wking:
On Fri, Jul 22, 2016 at 12:13:29AM -0700, Qiang Huang wrote:
> @wking I consisted some ':' and ',' usage.

0e8e069 → 1752ce8 looks good to me.

vbatts:
LGTM

crosbymichael:
LGTM

PR #444, opened 2016-05-23
Use RFC_2119 style for config.md

hqhq:

### DIFF
--- a/config.md
+++ b/config.md
@@ -213,9 +213,9 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
 [**`platform.os`**](#platform) is used to lookup further platform-specific configuration.
 
 * **`linux`** (object, optional) [Linux-specific configuration](config-linux.md).
-  This should only be set if **`platform.os`** is `linux`.
+  This SHOULD only be set if **`platform.os`** is `linux`.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -213,9 +213,9 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
 [**`platform.os`**](#platform) is used to lookup further platform-specific configuration.
 
 * **`linux`** (object, optional) [Linux-specific configuration](config-linux.md).
-  This should only be set if **`platform.os`** is `linux`.
+  This SHOULD only be set if **`platform.os`** is `linux`.
 * **`solaris`** (object, optional) [Solaris-specific configuration](config-solaris.md).
-  This should only be set if **`platform.os`** is `solaris`.
+  This SHOULD only be set if **`platform.os`** is `solaris`.
```
wking: Maybe the pattern here should be “This MUST NOT be set if **`platform.os`** is not `$OS`”?  If you're writing a Linux config, you should not be setting `solaris`, but whether you set `linux` or not is up to you.

vbatts: I can't say that yet. Solaris has some linux compatibility modes. It could actually be feasible to have a Linux container that provides solaris configurations as well, IIUC

vbatts:
LGTM

crosbymichael:
LGTM

PR #446, opened 2016-05-23
Remove one JSON related MUST requirement

hqhq:
JSON is not a mandatory format, we should not use MUST
requirement on this.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -65,8 +65,6 @@ Unless otherwise stated, generating an error MUST leave the state of the environ
 
 This operation MUST generate an error if it is not provided the ID of a container.
 This operation MUST return the state of a container as specified in the [State](#state) section.
-In particular, the state MUST be serialized as JSON.
-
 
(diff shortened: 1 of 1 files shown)


wking:
On Mon, May 23, 2016 at 02:09:14AM -0700, Qiang Huang wrote:
> JSON is not a mandated format, we should not use MUST requirement on
> this.

For some background on this, see [1].  My main concern is “how will
callers handle this if they don't know what format the response is
in?”  But looking back on it now, that's a more general “What API does
the runtime expose to its callers?” question, so I'm fine punting this
requirement to [2] or similar.

[1]: https://github.com/opencontainers/runtime-spec/pull/225#discussion_r50301059
[2]: https://github.com/wking/oci-command-line-api

crosbymichael:
LGTM

It makes sense in the context of the actions that are being discussed in that section.

mrunalp:
LGTM

PR #447, opened 2016-05-23
Make path to bundle optional

hqhq:
Make it optional looks reasonable to me, and it's what
we do in runc now.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -70,9 +70,10 @@ In particular, the state MUST be serialized as JSON.
 
 ### Start
 
-`start <container-id> <path-to-bundle>`
+`start <container-id> [<path-to-bundle>]`
 
-This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
+This operation MUST generate an error if it is not provided the ID of a container.
+The path to the bundle is OPTIONAL, the implementation can define it's default path when path to the bundle is absent.
 If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error.
 Using the data in `config.json`, that are in the bundle's directory, this operation MUST create a new container.
 This includes creating the relevant namespaces, resource limits, etc and configuring the appropriate capabilities for the container.


wking:
On Mon, May 23, 2016 at 02:31:56AM -0700, Qiang Huang wrote:
> Make it optional looks reasonable to me, and it's what we do in runc
> now.

These sketches are not intended to be a command-line API [1].  I think
this change would be useful in a command-line API document, but is
probably too detailed for the operation overview here (where “the
runtime needs to know which config to use” is all you need to
express).

[1]: https://github.com/opencontainers/runtime-spec/pull/225#issuecomment-172626408

crosbymichael:
@hqhq i think this is just an error in the docs, it is a flag on the runtime and has always been a flag.  I think its safe to just remove this for now as it was never the case.

wking:
On Mon, May 23, 2016 at 12:57:55PM -0700, Michael Crosby wrote:
> I think its safe to just remove this for now as it was never the
> case.

What was never the case?  The runtime does need to know which
bundle/config it's starting.  For runC, it can get a reasonable
default by using getcwd(3), but that won't be a reasonable default for
all runtimes (e.g. if your runtime API is via HTTP web requests, the
web-server's CWD is probably not interesting).  The ‘start’ operation
needs more information here than, say, the ‘state’ operation, so I
think it makes sense to list that extra information here.

crosbymichael:
Oh, I see.  This is not specifying a CLI but just specifying the inputs to the start action.  If we look at it that way then bundle path is required even though most runtimes will imply the bundle path is the `cwd` but for the start action it is a required input.

wking:
On Mon, May 23, 2016 at 01:15:23PM -0700, Michael Crosby wrote:
> This is not specifying a CLI but just specifying the inputs to the
> start action.  If we look at it that way…

That's how I've been looking at it based on [1], but a few folks have
interpreted these as a CLI API now, so it's probably worth explicitly
saying that they aren't in the spec.  I can file a separate PR to that
effect tonight.

[1]: https://github.com/opencontainers/runtime-spec/pull/225#issuecomment-172626408

hqhq:
@wking It make sense if it's input for general operation, I already opened a PR https://github.com/opencontainers/runtime-spec/pull/450 to clarify this, PTAL, thanks.

PR #448, opened 2016-05-24
Makefile: Fix native pandoc invocation

wking:
In #428 I'd misunderstood vbatts/pandoc as a call to a locally-installed pandoc, when it's actually the name of a [Docker][1] [image][2].  With this series, we prefer a local pandoc if one exists, fall back to Docker and vbatts/pandoc if a local `docker` exists, and raise an error if neither `pandoc` nor `docker` exist.

This is an alternative to #440, which reverts dc9daf9e and doesn't allow for a locally-installed pandoc.

[1]: https://github.com/opencontainers/runtime-spec/pull/440
[2]: https://github.com/opencontainers/runtime-spec/pull/428#discussion_r63987603

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,20 @@
 
 DOCKER ?= $(shell command -v docker 2>/dev/null)
 PANDOC ?= $(shell command -v pandoc 2>/dev/null)
+ifeq "$(strip $(PANDOC))" ''
+	ifneq "$(strip $(DOCKER))" ''
+		PANDOC = $(DOCKER) run \
+			-it \
+			--rm \
+			-v $(shell pwd)/:/input/:ro \
+			-v $(shell pwd)/output/:/output/ \
+			-u $(shell id -u) \
+			vbatts/pandoc
+		PANDOC_SRC := /input/
+		PANDOC_DST := /
+	endif
+endif
+
 # These docs are in an order that determines how they show up in the PDF/HTML docs.
 DOC_FILES := \
 	version.md \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,25 +39,18 @@ default: docs
...
 output/docs.pdf: $(DOC_FILES)
 	mkdir -p output/ && \
-	$(DOCKER) run \
-	-it \
-	--rm \
-	-v $(shell pwd)/:/input/:ro \
-	-v $(shell pwd)/output/:/output/ \
-	-u $(shell id -u) \
-	$(PANDOC) -f markdown_github -t latex -o /$@ $(patsubst %,/input/%,$(DOC_FILES))
+	$(PANDOC) -f markdown_github -t latex -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES))
```
hqhq: If we have local pandoc, should also set PANDOC_DST and PANDOC_SRC, otherwise they'll be empty.

Makefile
```diff
@@ -25,25 +39,18 @@ default: docs
 .PHONY: docs
 docs: output/docs.pdf output/docs.html
 
+ifeq "$(strip $(PANDOC))" ''
+output/docs.pdf output/docs.html:
```
hqhq: Can you pickup my second patch in #440 ? I think it would be better to name it as `runtime-spec.pdf/html`, it's what I have to do now before send it to someone else who don't have source code, WDYT?

Makefile
wking: Empty strings are the values we want for a native pandoc, but I can explicitly set them to empty strings if you like.  Let me know if that's what you want.

Makefile
wking: Sure.  Do you want oci-runtime-spec-v{version}.{extension}?  If we're adding context in the filename, we might as well be thorough ;).

vbatts:
Fixes #454

vbatts:
LGTM

Makefile
vbatts: lets do that in another PR. I have mulled on a better name and a way to do include the version (git tag + abbreviated commit?) but haven't arrived at a clean enough conclusion yet.

laijs:
I think the default should be vbatts/pandoc.
we should produce the host-independent output in the default cases.

vbatts:
@laijs oh good point. I merged it because it's presently broken. Though for consistency sake that is a good point.

wking:
On Wed, May 25, 2016 at 05:19:53AM -0700, Lai Jiangshan wrote:
> I think the default should be vbatts/pandoc.
> we should produce the host-independent output in the default cases.

Unless we specify a specific version of vbatts/pandoc, I don't see how
it would be any more host-independent than a local pandoc.

PR #449, opened 2016-05-24
Add dev symbolic links in runtime_linux.md

hqhq:
Some fixes in runtime_linux.md, and added dev symbolic links.

It's adapted from:
https://github.com/opencontainers/runc/blob/master/libcontainer/SPEC.md

Addresses: opencontainers/runc#760

### DIFF
--- a/runtime-linux.md
+++ b/runtime-linux.md
@@ -1,8 +1,7 @@
 # Linux Runtime
 
 ## File descriptors
+
 By default, only the `stdin`, `stdout` and `stderr` file descriptors are kept open for the application by the runtime.
-
-The runtime may pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
-
-Some of the file descriptors may be redirected to `/dev/null` even though they are open.
+The runtime MAY pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
+Some of the file descriptors MAY be redirected to `/dev/null` even though they are open.
--- a/runtime-linux.md
+++ b/runtime-linux.md
@@ -5,3 +5,14 @@
 By default, only the `stdin`, `stdout` and `stderr` file descriptors are kept open for the application by the runtime.
 The runtime MAY pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
 Some of the file descriptors MAY be redirected to `/dev/null` even though they are open.
+
+## Dev symbolic links
+
+After the container has `/proc` mounted, the following standard symlinks MUST be setup within `/dev/` for the io.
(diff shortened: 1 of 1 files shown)


vbatts:
SGTM, though i'll defer to others

runtime-linux.md
```diff
@@ -1,8 +1,18 @@
...
 ## File descriptors
+
 By default, only the `stdin`, `stdout` and `stderr` file descriptors are kept open for the application by the runtime.
+The runtime MAY pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
+Some of the file descriptors MAY be redirected to `/dev/null` even though they are open.
+
+## Dev symbolic links
 
-The runtime may pass additional file descriptors to the application to support features such as [socket activation](http://0pointer.de/blog/projects/socket-activated-containers.html).
+After the container has `/proc` mounted, the following standard symlinks MUST be setup within `/dev/` for the io.
```
wking: I'd recommend wording like the existing [“MUST be made available”][1] or [“MUST … supply”][2] instead of “MUST be setup”.  Config authors and runtime callers should only care about the links being there, and shouldn't care about whether it was the runtime that set them up.  And runtimes that bind-mount `/dev` (or similar) into the container may already have these links and not need to create them themselves.  I'd also like to make the timing for these required objects explicit (see the third bullet [here][3]).  And cross-links between these “paths the runtime must setup if missing” sections would be good too.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config-linux.md#default-file-systems
[2]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config-linux.md#default-devices
[3]: https://github.com/opencontainers/runtime-spec/pull/164#issuecomment-138984824

runtime-linux.md
```diff
@@ -1,8 +1,18 @@
...
 
-Some of the file descriptors may be redirected to `/dev/null` even though they are open.
+|    Source       | Destination |
```
wking: [ln(1)][1] calls these “target” and “link name” respectively.

[1]: http://www.man7.org/linux/man-pages/man1/ln.1.html

wking:
On Tue, May 24, 2016 at 01:13:45AM -0700, Qiang Huang wrote:
>   * Add dev symbolic links in runtime_linux.md

We probably want to cite proc(5) here, which has:

  Most Linux MAKEDEV scripts symbolically link /dev/fd to
  /proc/self/fd, in fact.

and:

  Most systems provide symbolic links /dev/stdin, /dev/stdout, and
  /dev/stderr, which respectively link to the files 0, 1, and 2 in
  /proc/self/fd.

crosbymichael:
LGTM

mrunalp:
LGTM but needs rebase.

crosbymichael:
LGTM

PR #435, opened 2016-05-13
ROADMAP: Remove stale targets (landed PRs, image-spec, ocitools, etc.)

wking:
This PR cherry-picks the first two from #423 and adds additional
target-removal commits with the reasoning I posted [to the list][1].
I'm happy to pull out any contentious and/or obvious commits into
separate PRs if that makes reviewing and landing this easier.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!original/dev/pghr1WFU5KI/NDFTvMDiFAAJ

PR #451, opened 2016-05-24
specs-go/config: Drop platform-independent comment

wking:
This has been stale since cb2da543 (config: Single, unified config file, 2015-12-28, #284), when we dropped the attempt to distinguish between platform-independent and platform-dependent configuration.

This PR nibbles another commit off from #423 (which worked fairly well for #435).

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -2,9 +2,7 @@ package specs
 
 import "os"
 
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #452, opened 2016-05-24
config-linux: Make “don't modify filesystem permissions” generic

wking:
The user-namespace restriction isn't about the root filesystem in particular.  For example, if you bind mount in a second filesystem, the runtime shouldn't adjust ownership on that filesystem either.

I've also adjusted the old “permissions” to “ownership”, since that more clearly reflects the fields (user and group) that you would modify if you wanted to adjust for user namespacing.

This nibbles some more ground-work off from #423, since that approach worked well in #435.


crosbymichael:
LGTM

mrunalp:
LGTM

PR #453, opened 2016-05-24
schema: Replace "bundle" with "runtime configuration"

wking:
The bundle contains `config.json` and other things, but this schema is just for `config.json`.  This is mostly:

    $ sed -i 's|/bundle|/runtime/config|' schema/*.json

for the `id` fields, but I also tweaked `root.description` and `config.json`'s `description`.

### DIFF
--- a/schema/schema-linux.json
+++ b/schema/schema-linux.json
@@ -1,11 +1,11 @@
 {
     "linux": {
         "description": "Linux platform-specific configurations",
-        "id": "https://opencontainers.org/schema/bundle/linux",
+        "id": "https://opencontainers.org/schema/runtime/config/linux",
         "type": "object",
         "properties": {
             "devices": {
-                "id": "https://opencontainers.org/schema/bundle/linux/devices",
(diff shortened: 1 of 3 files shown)


wking:
Another PR nibbled off of #423 (like #435, #451, and #452).

mrunalp:
Hmm, not sure @vbatts WDYT?

crosbymichael:
-1

We are NOT getting rid of the bundle concept at this point.

crosbymichael:
We already discussed this on the call and agreed on keeping it.

wking:
On Wed, May 25, 2016 at 10:54:06AM -0700, Michael Crosby wrote:
> We are NOT getting rid of the bundle concept at this point. 

This commit was not about getting rid of the bundle concept (although
it helps with that).  It was “these schemas are for the runtime's
config.json”.  I'm pretty sure we aren't using them to validate the
bundle as a whole.

To make the distinction more clear, I'll file a PR to add JSON Schema
for the state JSON (also part of the runtime spec, but clearly not
part of a bundle).

PR #456, opened 2016-05-25
schema/schema.json: Update path.root description

wking:
Use wording from config.md, since the JSON Schema doesn't seem like a
good place to be picking new words.

A tiny piece pulled out of #453, with wording that is copy/pasted from
config.md.


hqhq:
LGTM

mrunalp:
LGTM

PR #459, opened 2016-05-26
[ Runtime ] Allow for excluding pid from state

RobDolinMS:
There may be cases where the pid may not be visible to the host.  
Windows Hyper-V Containers is an example of this.  

The proposed additional clause allows for pid to not be include (or set to an appropriate null/empty value) if this is the case.  

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -16,6 +16,7 @@ There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
 This allows the hooks to perform cleanup and teardown logic after the runtime destroys its own state.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
+The pid MAY be excluded or set to a value indicating empty or null if it is not seen by the host.
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
 


runtime.md
```diff
@@ -16,6 +16,7 @@ There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
 This allows the hooks to perform cleanup and teardown logic after the runtime destroys its own state.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
+The pid MAY be excluded or set to a value indicating empty or null if it is not seen by the host.
```
wking: I'd rather phrase this as:

> * **`pid`** (int, optional) is the ID ….
>   This property is REQUIRED on the following platforms: Linux, …

although that's tied up in [specifying platforms for VM-based containers][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-220721379

wking:
On Thu, May 26, 2016 at 03:55:44PM -0700, Rob Dolin (MSFT) wrote:
> There may be cases where the pid may not be visible to the host.  
> Windows Hyper-V Containers is an example of this.  

There's a bit of earlier discussion of this issue (in the runV
context) here [1].

Stepping back to the original motivation for the ‘pid’ entry, it
mostly revolved around looking up additional information about a
container so you could manipulate the container environment with hooks
[2,3].  If you don't expose something similar for Windows, does that
mean all container manipulation needs to pass through the runtime
bottleneck?  How do callers access and manipulate container features
like networking that aren't covered by the spec?  Maybe by putting
them into the spec (like [4])?

[1]: https://github.com/opencontainers/runtime-spec/pull/188#issuecomment-140296423
[2]: https://github.com/opencontainers/runtime-spec/issues/41#issue-92227918
[3]: https://github.com/opencontainers/runtime-spec/wiki/MeetingMinutes:-2015-08-26#minutes
[4]: https://github.com/opencontainers/runtime-spec/pull/411/files#diff-54e4235aaffe082955499e11eebda54cR71

tianon:
(definitely needs a rebase though)

runtime.md
cyphar: This phrasing looks wrong to me as well (especially "seen by the host"). However, I would also want that we specify what an "empty or null" value means for a PID. Should we make it a pointer? Or just make it `0` to indicate "empty"?

wking: Pointers are [our usual pattern for optional fields][1].  I don't think a zero PID is clearly enough “no-op in the spec” to qualify for the non-pointer-ok exception listed there.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/style.md#optional-settings-should-have-pointer-go-types

mrunalp:
@RobDolinMS needs rebase.

runtime.md
```diff
@@ -16,6 +16,7 @@ There is no requirement that it be unique across hosts.
 The ID is provided in the state because hooks will be executed with the state as the payload.
 This allows the hooks to perform cleanup and teardown logic after the runtime destroys its own state.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
```
hqhq: We could change all state fields phrased like:
**`pid`**: (int, OPTIONAL) ...
like we did in all other places.

wking: > We could change all state fields phrased like: pid: (int, OPTIONAL)...

I have that in flight with #507.  I'll pull that apart to isolate it from the `creating` change.

wking: > I have that in flight with #507.  I'll pull that apart to isolate it from the creating change.

Pulled off into #664 (ended up being just “cherry-pick the tip commit from #507”).

crosbymichael:
@RobDolinMS when you have time please rebase your PRs and make sure your sign off is correct.  Feel free to reopen when they have been updated.

PR #462, opened 2016-05-27
Add a 'status' field to our state struct

duglin:
When thinking about kill() vs delete(), a client needs to know (gracefully - w/o errors) when to call each. Which means it needs to be able to query the container to know what its current state is (running vs stopped).  This defines a property in our 'state' struct to convey that info.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -13,6 +13,13 @@ The state of a container MUST include, at least, the following properties:
 * **`id`**: (string) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
+* **`status`**: (string) is the runtime state of the container.
+The value MAY be one of:
+  * `created` : the container has been created but the user-specified code has not yet been executed
+  * `running` : the container has been created and the user-specified code is running
+  * `stopped` : the container has been created and the user-specified code has been executed but is no longer running
+
+  Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
@@ -23,6 +30,7 @@ When serialized in JSON, the format MUST adhere to the following pattern:
 {
     "ociVersion": "0.2.0",
     "id": "oci-container1",
+    "status": "running",
     "pid": 4422,
     "bundlePath": "/containers/redis"
 }
@@ -72,7 +80,9 @@ This operation MUST return the state of a container as specified in the [State](
 This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
 If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error and a new container MUST not be created.
 Using the data in [`config.json`](config.md), this operation MUST create a new container.
-This means that all of the resources associated with the container MUST be created, however, the user-specified process MUST NOT be run at this time.
+This means that all of the resources associated with the container MUST be created, however, the user-specified code MUST NOT be run at this time.
+
+Upon successful completion of this operation the `status` property of this container MUST be `created`.
 
 The runtime MAY validate `config.json` against this spec, either generically or with respect to the local system capabilities, before creating the container ([step 2](#lifecycle)).
 Runtime callers who are interested in pre-create validation can run [bundle-validation tools](implementations.md#testing--tools) before invoking the create operation.
@@ -88,12 +98,16 @@ Attempting to start an already started container MUST have no effect on the cont
 This operation MUST run the user-specified code as specified by [`process`](config.md#process-configuration).
 If the runtime fails to run the code as specified, an error MUST be generated.
 
+Upon successful completion of this operation the `status` property of this container MUST be `running`.
+
 ### Kill
 `kill <container-id> <signal>`
 
 This operation MUST generate an error if it is not provided the container ID.
 Attempting to send a signal to a container that is not running MUST have no effect on the container and MUST generate an error.
 This operation MUST send the specified signal to the process in the container.
+
+When the process in the container is stopped, irrespective of it being as a result of a `kill` operation or any other reason, the `status` property of this container MUST be `stopped`.
 
 ### Delete
 `delete <container-id>`


runtime.md
```diff
@@ -23,6 +30,7 @@ When serialized in JSON, the format MUST adhere to the following pattern:
 {
     "ociVersion": "0.2.0",
     "id": "oci-container1",
+	"status": "running",
```
crosbymichael: formatting issue

duglin: tabs... the root of all evil

crosbymichael:
LGTM

wking:
I commented via email last night, but GitHub seems to have eaten my
comment.  Here it is again:

On Thu, May 26, 2016 at 05:32:14PM -0700, Doug Davis wrote:
> When thinking about kill() vs delete(), a client needs to know
> (gracefully - w/o errors) when to call each.

-1.

There is no race-free way to do this gracefully without errors, since
“‘state’, process changes, take action based on stale state” is always
possible.  If a client wants to start a container, just run ‘start’
and catch errors.  The same approach works for the other commands,
with the additional complication of a wait-before-delete dance like
@julz outlined in [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/384#discussion_r64802554

duglin:
@wking the intent isn't to remove any chance of an error, the intent is to make it so that an error isn't part of the normal workflow just to figure out the state of a container.

wking:
On Fri, May 27, 2016 at 02:08:03PM -0700, Doug Davis wrote:
> @wking the intent isn't to remove any chance of an error, the intent
> is to make it so that an error isn't part of the normal workflow
> just to figure out the state of a container.

So “decrease the likelyhood of generating errors” (and I agree that
this change would do that).  Why is that useful?  Is there a problem
with errors?

I think that the *runtime* needs to be careful in this case (e.g. I
don't like this risk of sending the ‘start’ signal to user-specified
code [1]), but at the level of abstraction exposed by the spec
operations I see no such risks.  That doesn't mean that the risk isn't
there, but I think we should identify it (or some other use-case) more
clearly before landing a new field in the state.

[1]: https://github.com/opencontainers/runc/pull/827#discussion_r63461012

mrunalp:
I would prefer a `status` operation over a field in the state file.

duglin:
@mrunalp we already have one:  https://github.com/duglin/runtime-spec/blob/master/runtime.md#query-state
The problem is that it doesn't return enough info and I doubt we want 2 ops that return almost the same info.

mrunalp:
I meant status as a separate op that just returns the current run status of the container.

dqminh:
LGTM.

@duglin @wking AFAIK this is most useful when use for things like `inspect`, not really for `When thinking about kill() vs delete(), a client needs to know (gracefully - w/o errors) when to call each.` though,  as runtime should try to make those operations idempotent.

crosbymichael:
@mrunalp why would another op that just returned the status be better than the state op that returns status along with the other info?

wking:
On Tue, May 31, 2016 at 10:02:56AM -0700, Daniel, Dao Quang Minh wrote:
> @duglin @wking AFAIK this is most useful when use for things like
> `inspect`…

This is getting into very fuzzy territory to me.  “I want to know
$PROPERTY because I want to know $PROPERTY” is hard to argue
for/against ;).  I'd rather keep the spec restricted to “I want to
know $PROPERTY because I neet it to take $ACTION” is much easier to
reason about (do we want to support $ACTION?  Do you really need
$PROPERTY to do it?).  So, what sort of actions does this information
let you take?

> … not really for `When thinking about kill() vs delete()…` though,
> as runtime should try to make those operations idempotent.

At least with the current spec, delete is not idempotent [1]:

  Attempting to delete a container that does not exist MUST generate
  an error.

[1]: https://github.com/opencontainers/runtime-spec/blob/44f22774ce193c1362c9612ceef3d71f03dd82eb/runtime.md#delete

mrunalp:
@crosbymichael `state` is static whereas the running status of a process is more dynamic. The information in the state file could get stale.

mrunalp:
I think it will be hard to keep the status field in the state file correct and queries race free.

crosbymichael:
@mrunalp we are not talking about cat'ing a file.  The state response is dynamic as well and the status is correct it is returned.  Nothing static about it in the spec or in the runc implementation.

duglin:
the spec doesn't mention state files any more - we put it all behind an op.

mrunalp:
@crosbymichael Agree if the state isn't persisted in a file in a implementation (but anyway that is up to the implementation).

mrunalp:
@duglin Needs rebase.

duglin:
I think git is acting up again - there were not conflicts at all so the rebase shouldn't have been necessary.  Very odd.

runtime.md
```diff
@@ -13,6 +13,13 @@ The state of a container MUST include, at least, the following properties:
 * **`id`**: (string) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
+* **`status`**: (string) is the runtime state of the container.
+The value MAY be one of:
+  * `created` : the container has been created but the user-specified process has not yet been executed
+  * `running` : the container has been created and the user-specified process is running
+  * `stopped` : the container has been created and the user-specified process has been executed but is no longer running
```
wking: If we land a `status` property (I'm still [opposed][1]), I think we want to use `dead` here to mirror proc(5) and ps(1).  More details on why I don't like “stopped” in #465.

[1]: https://github.com/opencontainers/runtime-spec/pull/462#issuecomment-222249696

runtime.md
```diff
@@ -13,6 +13,13 @@ The state of a container MUST include, at least, the following properties:
 * **`id`**: (string) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
+* **`status`**: (string) is the runtime state of the container.
+The value MAY be one of:
+  * `created` : the container has been created but the user-specified process has not yet been executed
```
wking: I'd rather have this be “user-specified code” (#466).

runtime.md
```diff
@@ -13,6 +13,13 @@ The state of a container MUST include, at least, the following properties:
...
+  * `running` : the container has been created and the user-specified process is running
```
wking: I'd rather have this be “the container process has executed user-specified code has not exited”.  In Linux, in any case “running” is a technical term for one possible process state, and the status you're shooting for here also covers “sleeping”, “waiting”, and “stopped”.

runtime.md
duglin: I changed all "user-specific process" to "user-specific code" so the doc is now consistent (with itself anyway).

duglin:
oh man - had to rebase again.  Git is really acting up.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #450, opened 2016-05-24
Clarify the operation is not for command-line api

hqhq:
Replace: https://github.com/opencontainers/runtime-spec/pull/447

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>

PR #463, opened 2016-05-27
runtime: Drop create-specific API caveat from lifecycle

wking:
This wording is descended from 7117ede7 (Expand on the definition of
our ops, 2015-10-13, #225), but the idea is covered generically by
e53a72b (Clarify the operation is not for command-line api,
2016-05-24, #450), so we no longer need a create-specific note.
Especially in the lifecycle docs, where there's already enough going
on without this low-level detail.


crosbymichael:
W/E

LGTM

hqhq:
LGTM

PR #464, opened 2016-05-27
runtime: Consistent indent for "However, some actions..."

wking:
The shorter-than-normal (for the rest of this list) indent landed with
the line in be594153 (Split create and start, 2016-04-01, #384).

PR #465, opened 2016-05-27
runtime: Replace "process is stopped" with "process exits"

wking:
[proc(5)](http://man7.org/linux/man-pages/man5/proc.5.html) describes the following state entries in
`proc/[pid]/stat` (for modern kernels):
- R Running
- S Sleeping in an interruptible wait
- D Waiting in uninterruptible disk sleep
- Z Zombie
- T Stopped (on a signal)
- t Tracing stop
- X Dead

and [ps(1)](http://man7.org/linux/man-pages/man1/ps.1.html) has a bit more context (for modern kernels):
- D uninterruptible sleep (usually IO)
- R running or runnable (on run queue)
- S interruptible sleep (waiting for an event to complete)
- T stopped by job control signal
- t stopped by debugger during the tracing
- X dead (should never be seen)
- Z defunct ("zombie") process, terminated but not reaped by its
  parent

So I expect "stopped" to mean "process still exists but is paused,
e.g. by SIGSTOP".  And I expect "exited" to mean "process has finished
and is either a zombie or dead".

After this commit, `git grep -i stop` only turns up poststop-hook
stuff, a reference in principles.md, a "stoppage" in LICENSE, and some
ChangeLog entries.

This also replaces "container's process" with "container process" to
match usage in the rest of the repository.  After this commit:

```
$ git grep -i "container process" | wc -l
13
$ git grep -i "container's process" | wc -l
1
```

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -15,9 +15,10 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-    * `created`: the container has been created but the user-specified program has not yet been executed
-    * `running`: the container has been created and the user-specified program is running
-    * `stopped`: the container has been created and the user-specified program has been executed but is no longer running
+
+    * `created`: the container process has neither exited nor executed the user-specified program
+    * `running`: the container process has executed the user-specified program but has not exited
+    * `stopped`: the container process has exited
 
     Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the container process, as seen by the host.
@@ -55,8 +56,8 @@ The lifecycle describes the timeline of events that happen from when a container
    However, some actions might only be available based on the current state of the container (e.g. only available while it is started).
 4. Runtime's [`start`](runtime.md#start) command is invoked with the unique identifier of the container.
    The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
-5. The container's process is stopped.
-   This MAY happen due to them erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
+5. The container process exits.
+   This MAY happen due to erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
 6. Runtime's [`delete`](runtime.md#delete) command is invoked with the unique identifier of the container.
    The container MUST be destroyed by undoing the steps performed during create phase (step 2).
 


hqhq:
LGTM

mrunalp:
Looks fine. Needs rebase.

wking:
On Tue, May 31, 2016 at 09:41:40AM -0700, Mrunal Patel wrote:
> Looks fine. Needs rebase. 

Rebased with 23009bb → a95eb3b, although again [1] I don't see any changes
to the patch.

[1]: https://github.com/opencontainers/runtime-spec/pull/464#issuecomment-222759966

wking:
Maybe needs a reroll with my [suggested][1] [updates][2] to #462.

[1]: https://github.com/opencontainers/runtime-spec/pull/462#discussion_r65228341
[2]: https://github.com/opencontainers/runtime-spec/pull/462#discussion_r65229229

wking:
Rebased and adjusted along the lines of my #464 comments with a95eb3b
→ 5816f31.

runtime.md
```diff
@@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-  * `created` : the container has been created but the user-specified code has not yet been executed
-  * `running` : the container has been created and the user-specified code is running
-  * `stopped` : the container has been created and the user-specified code has been executed but is no longer running
+  * `created` : the container process has neither exited nor executed the user-specified code
+  * `running` : the container process has executed the user-specified code but has not exited
```
dqminh: meaning of `user-specified code` vs `container process` here seems confusing to me. For example, if `container process` is some init piece of the runtime, does this mean that the init piece has to be run for the lifetime of the user process because `stopped` only happens when the container process has exited ? Or did I understand this incorrectly ?

What do you think about replacing `user-specified code` with `container process` completely. Something like

```
* `created` : the container has been created but the container process has not yet been executed.
* `running`: the container process has been executed but has not exited.
* `stopped`: the container process has exited.
```

wking: On Wed, Jul 13, 2016 at 09:43:58AM -0700, Daniel, Dao Quang Minh wrote:
> > @@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
> >  There is no requirement that it be unique across hosts.
> >  * **`status`**: (string) is the runtime state of the container.
> >  The value MAY be one of:
> > -  * `created` : the container has been created but the user-specified code has not yet been executed
> > -  * `running` : the container has been created and the user-specified code is running
> > -  * `stopped` : the container has been created and the user-specified code has been executed but is no longer running
> > +  * `created` : the container process has neither exited nor executed the user-specified code
> > +  * `running` : the container process has executed the user-specified code but has not exited
> 
> meaning of `user-specified code` vs `container process` here seems
> confusing to me.

This terminology issue is being hashed out in #466.  Maybe we should
handle that first?

wking:
Rebased onto master with 5816f31 → 7797b33.

runtime.md
hqhq: @wking Since #466 is closed, I think most of the maintainers like `container process` more than `user-specified code`, but I think we can handle that in a separate PR.

runtime.md
```diff
@@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-    * `created`: the container has been created but the user-specified code has not yet been executed
-    * `running`: the container has been created and the user-specified code is running
-    * `stopped`: the container has been created and the user-specified code has been executed but is no longer running
+    * `created`: the container process has neither exited nor executed the user-specified code
+    * `running`: the container process has executed the user-specified code but has not exited
+    * `stopped`: the container process has exited
```
hqhq: Why remove `the container has been created` phrase? I don't think that redundant, especially for the `created` status.

wking: Maybe the container process crashed during creation, in which case we're `stopped` before the container has been fully created.  And see #507, which builds on this PR and adds `creating`.  In the absence of a `creating` status (this PR before #507), I think we're `created` even before creation has finished.

runtime.md
```diff
@@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-    * `created`: the container has been created but the user-specified code has not yet been executed
-    * `running`: the container has been created and the user-specified code is running
-    * `stopped`: the container has been created and the user-specified code has been executed but is no longer running
+    * `created`: the container process has neither exited nor executed the user-specified code
```
crosbymichael: Its not `user-specified code`, we went over this many times.

runtime.md
```diff
@@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
...
+    * `running`: the container process has executed the user-specified code but has not exited
```
crosbymichael: container process not `user-specified code`

runtime.md
crosbymichael: This is not a good description of the created state.  The container is in a created state when it has been fully initialized but is not yet running the containers main process

runtime.md
crosbymichael: It would be better to state that in the stopped state the container has ran the container's main process and it has exited leaving the container in the stopped state

crosbymichael: If create fails the container should be destroyed and will be by the runtimes.

runtime.md
```diff
@@ -55,8 +55,8 @@ The lifecycle describes the timeline of events that happen from when a container
    However, some actions might only be available based on the current state of the container (e.g. only available while it is started).
 4. Runtime's [`start`](runtime.md#start) command is invoked with the unique identifier of the container.
    The runtime MUST run the user-specified code, as specified by [`process`](config.md#process-configuration).
-5. The container's process is stopped.
-   This MAY happen due to them erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
+5. The container process exits.
+   This MAY happen due to erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
```
crosbymichael: This is not correct as the runtime's kill command can sent non terminating signals.

runtime.md
wking: > Its not user-specified code, we went over this many times.

As [I pointed out][1] after #466 was closed, all of the changes proposed by #466 are already live.  So I'd rather keep the currently-live “user-specified code” language.  If you feel like that language is unclear, adjusting to something else seems orthogonal to the stopped → exists fix that this PR is focused on.  Maybe open a different PR?

[1]: https://github.com/opencontainers/runtime-spec/pull/466#issuecomment-258569376

wking: >  The container is in a created state when it has been fully initialized but is not yet running the containers main process.

#507 addresses this by adding a `creating` state, and that seems orthogonal to the stopped → exists fix which I'm focusing on here.  Until then, `created` seems like the best stand in (a container being created is certainly not `running` or `stopped`).

runtime.md
wking: > This is not correct as the runtime's kill command can sent non terminating signals.

I agree, but this MAY does not list all possible triggers nor does it require listed triggers to cause container exit.  And it's the existing language and is orthogonal to the stopped → exited I'm focusing on here.  Maybe file a new PR with your suggested wording?

runtime.md
wking: > It would be better to state that in the stopped state the container has ran the container's main process…

This is not necessarily true, since the container process may have died during creation or between creation and the `start` call.

>  If create fails the container should be destroyed and will be by the runtimes.

Agreed, but we need to define the `status` passed to any post-stop hooks, which run [after the container process exits][1].  That happens even if the container process dies during creation.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#hooks

mrunalp:
@crosbymichael PTAL

crosbymichael:
LGTM

wking:
@hqhq, @mrunalp, now that @crosbymichael has LGTMed this, do either of you want to re-approve it?

hqhq:
LGTM

PR #466, opened 2016-05-27
runtime: Replace "user-specified process" with "user-specified code" in 'create'

wking:
Because we _do_ want to create the process at create-time, we just
want to idle it running runtime-specified code until start-time.  The
language I'm changing to already exists in the lifecycle and start
sections.

Since there were a number of links to the process spec, I replaced the
inline targets with a reference-style target.  And while I was doing
that, I added a missing close-paren to lifecycle step 2.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -35,12 +35,12 @@ The lifecycle describes the timeline of events that happen from when a container
 1. OCI compliant runtime's `create` command is invoked with a reference to the location of the bundle and a unique identifier.
    How these references are passed to the runtime is an implementation detail.
 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
-   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`](config.md#process-configuration) MUST NOT be run at this time.
+   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`][process]) MUST NOT be run at this time.
    Any updates to `config.json` after this step MUST NOT affect the container.
 3. Once the container is created additional actions MAY be performed based on the features the runtime chooses to support.
   However, some actions might only be available based on the current state of the container (e.g. only available while it is started).
 4. Runtime's `start` command is invoked with the unique identifier of the container.
-   The runtime MUST run the user-specified code, as specified by [`process`](config.md#process-configuration).
+   The runtime MUST run the user-specified code, as specified by [`process`][process].
 5. The container's process is stopped.
    This MAY happen due to them erroring out, exiting, crashing or the runtime's `kill` operation being invoked.
 6. Runtime's `delete` command is invoked with the unique identifier of the container.
@@ -72,7 +72,7 @@ This operation MUST return the state of a container as specified in the [State](
 This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
 If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error and a new container MUST not be created.
 Using the data in [`config.json`](config.md), this operation MUST create a new container.
-This means that all of the resources associated with the container MUST be created, however, the user-specified process MUST NOT be run at this time.
+This means that all of the resources associated with the container MUST be created, however, the user-specified code (from [`process`][process]) MUST NOT be run at this time.
 
 The runtime MAY validate `config.json` against this spec, either generically or with respect to the local system capabilities, before creating the container ([step 2](#lifecycle)).
 Runtime callers who are interested in pre-create validation can run [bundle-validation tools](implementations.md#testing--tools) before invoking the create operation.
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -72,7 +72,7 @@ This operation MUST return the state of a container as specified in the [State](
 This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
 If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error and a new container MUST not be created.
 Using the data in [`config.json`](config.md), this operation MUST create a new container.
-This means that all of the resources associated with the container MUST be created, however, the user-specified process MUST NOT be run at this time.
+This means that all of the resources associated with the container MUST be created, however, the user-specified code (from [`process`][process]) MUST NOT be run at this time.
```
hqhq: `user-specified code` seems a bit abstract to me, and I don't see the problem with the original one.

hqhq: ok I just found we are using `use-specified code` in `Start` operation, I think they need to be consistent, my preference would be `use-specified process` but I don't have strong opinion on this, I'll leave it to others.

wking: On Fri, May 27, 2016 at 06:50:14PM -0700, Qiang Huang wrote:
> > -This means that all of the resources associated with the container MUST be created, however, the user-specified process MUST NOT be run at this time.
> > +This means that all of the resources associated with the container MUST be created, however, the user-specified code (from [`process`][process]) MUST NOT be run at this time.
>
> `user-specified code` seems a bit abstract to me, and I don't see
> the problem with the original one.

The distinction between “process” and “code” is that fork(2) and
similar create a new process running the caller's code, while
execve(2) and similar execute new code in the caller's existing
process.  In this spec, the container *process* is created at
create-time (with fork(2) or similar) and then that process is
switched to user-specified code at start-time (with execve(2) or
similar).  I'm dropping the “process MUST NOT be run” phrasing here,
because we *do* want to create the container *process* here.  We just
don't want it running the user's code yet.

crosbymichael:
I think its a mistake to call it users code.  we are not running any "code".  we are not a jit'd runtime that takes code and runs it.  it is all processes and the key in the spec is called "process" so calling it code is just adding more confusion.

wking:
On Wed, Jun 01, 2016 at 02:37:55PM -0700, Michael Crosby wrote:
> I think its a mistake to call it users code.

execve(2) says “program” and unpacks that to “a binary executable, or
a script starting with a [shebang]” [1].  proc(5) documents
/proc/[pid]/exe talks about “the executed command” [2].  Create-time
*is* about creating a new container process, and start-time is just
about changing what that process is doing.

[1]: http://man7.org/linux/man-pages/man2/execve.2.html
[2]: http://man7.org/linux/man-pages/man5/proc.5.html

crosbymichael:
We are here to make meaningful abstractions for defining a container, not specify how the kernel executes code

wking:
On Wed, Jun 01, 2016 at 03:10:06PM -0700, Michael Crosby wrote:
> We are here to make meaningful abstractions for defining a
> container, not specify how the kernel executes code

To me, “user-specified process MUST NOT be run” sounds like “‘create’
should not create the container process”.  But it MUST create that
process.  So I'm suggesting we adopt wording that makes it clear that
the container process is created by ‘create’ and that ‘start’ just
tells the pre-existing container process to run the user-specified
code/program/executable/script/command/….  If you prefer a word beside
“code” from that list, let me know.  But I think the current spec
wording is needlessly conflating it with “process”.

tianon:
Rejected

I'm with @crosbymichael on this one -- I think `process` is a more appropriate word here than `code`.

wking:
On Fri, Nov 04, 2016 at 03:29:12PM -0700, Tianon Gravi wrote:
> I'm with @crosbymichael on this one -- I think `process` is a more
> appropriate word here than `code`.

In that case, we may need to revisit the current master and add a
longer description of the distinction.  Because:

  $ git grep 'user-specified code' origin/master -- runtime.md
  origin/master:runtime.md:    * `created`: the container has been created but the user-specified code has not yet been executed
  origin/master:runtime.md:    * `running`: the container has been created and the user-specified code is running
  origin/master:runtime.md:    * `stopped`: the container has been created and the user-specified code has been executed but is no longer running
  origin/master:runtime.md:   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`](config.md#process-configuration)) MUST NOT be run at this time.
  origin/master:runtime.md:   The runtime MUST run the user-specified code, as specified by [`process`](config.md#process-configuration).
  origin/master:runtime.md:This means that all of the resources associated with the container MUST be created, however, the user-specified code MUST NOT be run at this time.
  origin/master:runtime.md:This operation MUST run the user-specified code as specified by [`process`](config.md#process-configuration).
  $ git grep 'user-specified process' origin/master -- runtime.md
  …no hits…

It looks like the changes I was proposing here have made it in through
other PRs:

* #462, using “user-specified code” with the initial status values and
  changing ‘create’ from “user-specified process” to “user-specified
  code”.
* #603, closing the paren.

Does someone in the ‘process’ camp want to open an issue about adding
clarifying language?

PR #469, opened 2016-05-27
bundle: Remove rootfs requirement

wking:
Recently b2e9154a (Remove requirement for rootfs path to be relative,
2016-04-22, #394) relaxed the relative-path requirement on root.path
to allow absolute paths (see also discussion [here](https://github.com/opencontainers/runtime-spec/issues/389)).  The
motivation being that, while most folks will want a relative-path
pointing into the bundle directory, there are valid [use](https://github.com/opencontainers/runtime-spec/issues/389#issue-149584747)
[cases](https://github.com/opencontainers/runtime-spec/pull/394#issuecomment-213550294) that don't do that, and we want to support them too.

However, allowing absolute paths but requiring them to point to a
directory lying next to config.json seems to defeat the purpose.  This
commit removes the "root.path must point to a directory next to
config.json" restriction, so folks who want absolute paths in
root.path can use them without an unnecessary restriction.

And again, most folks will put their rootfs alongside config.json or
elsewhere inside the bundle directory.  But this list was about "This
MUST include the following artifacts", and we don't want to require
_everyone_ to do that.

This is the narrowly-scoped version of #423 with the minimum we need
to make #394 a useful change.

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -15,10 +15,5 @@ This REQUIRED file MUST reside in the root of the bundle directory and MUST be n
 When the bundle is packaged up for distribution, this file MUST be included.
 See [`config.json`](config.md) for more details.
 
-2. A directory representing the root filesystem of the container.
-While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
-When the bundle is packaged up for distribution, this directory MUST be included.
-This directory MUST be referenced from within the `config.json` file.
-
-While these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.
-In other words, a tar archive of a *bundle* will have these artifacts at the root of the archive, not nested within a top-level directory.
+While `config.json` MUST be present in a directory on the local filesystem, that directory itself is not part of the bundle.
+In other words, a tar archive of a *bundle* will have `config.json` at the root of the archive, not nested within a top-level directory.


wking:
Oh, and while I think the [`root`][1] entry [should be optional][2], this PR doesn't change that.  You still need to set `root.path`.  You just don't have to point it at a target lying next to `config.json` or within the bundle directory.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#root-configuration

crosbymichael:
Rejected

wking:
On Fri, May 27, 2016 at 11:53:02AM -0700, Michael Crosby wrote:
> Rejected

Reasoning?  How do you square that with #389 / #394?  Do you really
use absolute paths that always point to a directory next to your
config.json?

[1]: https://github.com/opencontainers/runtime-spec/issues/389#issue-149584747

crosbymichael:
Because the original sentences are correct and describe exactly what should happen/ it should be.

wking:
On Fri, May 27, 2016 at 03:51:05PM -0700, Michael Crosby wrote:
> Because the original sentences are correct and describe exactly what
> should happen/ it should be.

Following references from #389 takes me to docker/docker#22256.  I
haven't traced that back to figure out where spec.Root.Path is being
setup, but before the Docker PR, Docker's libcontainerd was
bind-mounting it to ${BUNDLE}/rootfs [1].  After the Docker PR it is
no longer doing that.  So I'm not clear on how the “these artifacts
MUST all be present in a single directory on the local filesystem…”
applies to that case.

Or more concretely, with /var/oci/config.json and the following
root.path entries:

a. "rootfs"
b. "./rootfs"
c. "my-rootfs"
d. "./my-rootfs"
e. "my/rootfs"
f. "./my/rootfs"
g. "/var/oci/rootfs"
h. "/var/oci/my-rootfs"
i. "/var/oci/my/rootfs"
j. "/tmp/rootfs

Which are valid?  Before #394, I think it was a, b, c, and d.  Since
#394 I think it also includes g and h.  The “MUST all be present in a
single directory” contraint I'm removing here blocks e, f, i, and j.
And the “MUST include the following” constraint from before the list
blocks j (as long as the rootfs is part of the artifact list).  So
after this PR, I expect all of those paths are legal.

And from my rough understanding of docker/docker#21969, I expect
Docker is now using paths like j (absolute paths pointing outside the
bundle directory).

Does that all sound accurate to you?  Once we agree on the facts, it
will be easier for me to understand why you think this PR is a bad
idea.  Because g and h sound much less useful to me than the
already-legal-before-#394 a, b, c, and d.

[1]: https://github.com/docker/docker/pull/22256/files#diff-15c000aa4859d8b3bae42251a9a13b15L171

wking:
On Fri, May 27, 2016 at 05:06:01PM -0700, W. Trevor King wrote:
> Or more concretely, with /var/oci/config.json and the following
> root.path entries…

ocitools adds another option to this list
(opencontainers/ocitools#113):

k. "."

I think that is currently illegal as well because of the “MUST all be
present in a single directory” constraint (the rootfs has a
config.json child, while the MUST requires a config.json sibling).
With this PR, k would also be legal.

Is my understanding of that incorrect?  Is k somehow legal with the
current spec wording?  Are there reasons for making it illegal?

wking:
On Fri, May 27, 2016 at 05:06:01PM -0700, W. Trevor King wrote:
> Or more concretely, with /var/oci/config.json and the following
> root.path entries:
> 
> …
> j. "/tmp/rootfs
> 
> Which are valid?  Before #394, I think it was a, b, c, and d.  Since
> #394 I think it also includes g and h.  The “MUST all be present in a
> single directory” contraint I'm removing here blocks e, f, i, and j.
> And the “MUST include the following” constraint from before the list
> blocks j (as long as the rootfs is part of the artifact list).  So
> after this PR, I expect all of those paths are legal.
> 
> And from my rough understanding of docker/docker#21969, I expect
> Docker is now using paths like j (absolute paths pointing outside the
> bundle directory).

I setup a container with Docker version 1.11.2 (build b9f10c9) and
confirmed the j-style path:

  # jq .bundle /run/containerd/a88d4b17c0f9b0347eb21c3f46ae124b2b38c7cf83495d3f3f60a08972bf1616/state.json 
  "/var/run/docker/libcontainerd/a88d4b17c0f9b0347eb21c3f46ae124b2b38c7cf83495d3f3f60a08972bf1616"
  # 

  # BUNDLE=$(jq -r .bundle /run/containerd/a88d4b17c0f9b0347eb21c3f46ae124b2b38c7cf83495d3f3f60a08972bf1616/state.json)
  # ROOT=$(jq -r .root.path $BUNDLE/config.json)
  # echo $BUNDLE
  /var/run/docker/libcontainerd/a88d4b17c0f9b0347eb21c3f46ae124b2b38c7cf83495d3f3f60a08972bf1616
  # echo $ROOT
  /var/lib/docker/btrfs/subvolumes/924b6be428d9dff1b206fb1ff6d4f217608860bfc367cd2052fbb0154e04fa0a
  # realpath --relative-to $BUNDLE $ROOT
  ../../../../var/lib/docker/btrfs/subvolumes/924b6be428d9dff1b206fb1ff6d4f217608860bfc367cd2052fbb0154e04fa0a

So I still think we want this PR to drop the requirement that
root.path is a sibling of config.json.

RobDolinMS:
@crosbymichael Do @wking's comments address your concern or do you still think this should be rejected?

vbatts:
after chatting about this: this is not behavior folks would expect. there is no sane assumption is there is no rootfs specified. It's a bad idea to assume it would default to '/' on the host.
I reject this too.

wking:
On Wed, Jan 18, 2017 at 02:23:56PM -0800, Vincent Batts wrote:
> this is not behavior folks would expect…

Having a root.path point at a non-sibling is what Docker does (or did
as of 1.11.2 [1]).

> there is no sane assumption is there is no rootfs specified.

With this PR, you still have to set root.path (because that's
specified in config.md, which I don't touch).  What this PR removes is
the requirement that the root.path be a sibling of the on-disk
config.json.

[1]: https://github.com/opencontainers/runtime-spec/pull/469#issuecomment-239631696

PR #397, opened 2016-04-25
*: add support for cgroup namespace

cyphar:
The cgroup namespace is a new kernel feature available in 4.6+ that
allows a container to isolate its cgroup hierarchy. This currently only
allows you to hide information from /proc/self/cgroup. But I'm currently
working with upstream on expanding it so that you can modify your
hierarchy inside a user namespace (even a rootless container).

Signed-off-by: Aleksa Sarai <asarai@suse.de>

PR #472, opened 2016-05-27
[Config | Hooks] Add "if supported by the platform"

RobDolinMS:
The spec currently suggests "Hooks MUST be called in the listed order."

If hooks are not supported by the platform, this would be a requirement that could not be met.

This PR adds "if supported by the platform" so that we're not imposing an impossible requirement on platforms that might not (currently) have support for hooks.  

### DIFF
--- a/config.md
+++ b/config.md
@@ -245,7 +245,7 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
 * [`Poststop`](#poststop) is a list of hooks to be run after the container process exits
 
 Hooks allow one to run code before/after various lifecycle events of the container.
-Hooks MUST be called in the listed order.
+Hooks MUST be called in the listed order if supported by the platform.
 The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
 
 Hook paths are absolute and are executed from the host's filesystem in the [runtime namespace][runtime-namespace].


wking:
I think we should just drop hooks (#473).

config.md
```diff
@@ -245,7 +245,7 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
 * [`Poststop`](#poststop) is a list of hooks to be run after the container process exits
 
 Hooks allow one to run code before/after various lifecycle events of the container.
-Hooks MUST be called in the listed order.
+Hooks MUST be called in the listed order if supported by the platform.
```
mrunalp: This is fine only if the bundle were to support different platforms else one could just not add hooks to the config at all.

wking: On Tue, May 31, 2016 at 09:33:38AM -0700, Mrunal Patel wrote:
> > -Hooks MUST be called in the listed order.
> > +Hooks MUST be called in the listed order if supported by the platform.
>
> This is fine only if the bundle were to support different platforms
> else one could just not add hooks to the config at all.

The new text here isn't clear, but based on [1], I think @RobDolinMS
means “Windows-based runtimes will error on configuration files with
meaningful content in ‘hooks’ (e.g. anything more than empty arrays
for the per-event properties)”.  So for Windows this PR is aiming for
“you *can't* have hooks” not “you have a choice to not have hooks”.
The latter is already true, although it is made more explicit by the
in-flight #427.

[1]: https://github.com/opencontainers/runtime-spec/pull/472#issue-157263212

mrunalp: @wking If that is the intent, then we should clarify further whether the platform will error out or just not call the hooks.

wking: On Tue, May 31, 2016 at 10:34:39AM -0700, Mrunal Patel wrote:
> > -Hooks MUST be called in the listed order.
> > +Hooks MUST be called in the listed order if supported by the platform.
>
> @wking If that is the intent, then we should clarify further whether
> the platform will error out or just not call the hooks.

I'm pretty sure “just not call the hooks” should not be valid runtime
behavior.  This is related to, but maybe not quite covered by, the “…
unable to create the environment specified…” language @cyphar has
in-flight with #397.

tianon: My preference here is definitely along the same lines (same as #476) -- if the runtime cannot support the functionality, then attempting to use it ought to result in an error (which stops users from scratching their heads wondering why their hooks aren't firing, for example).

tianon:
Did this one ever reach consensus?

wking:
On Fri, Nov 04, 2016 at 03:26:58PM -0700, Tianon Gravi wrote:
> Did this one ever reach consensus?

#559 means we'd need something like this PR if we really did want to
make hooks an optional part of compliant runtimes.  I'm not in favor
of that though.  If there really are systems where hooks are
unfeasible, the right solution is to make them a separate protocol [1]
and mint compliance certificates for the ($ARCH, $OS, $HOOK_SUPPORT,
$APIS) matrix (e.g. “certified OCI compliant for amd64 Linux with hook
support and the command line API”).  Then callers can check their
runtime's compliance certificate to see if they can rely on hooks or
not.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/README.md#L30

tianon:
Rejected

Indeed, this seems like the wrong way to express that Windows may not be able to support hooks in the same was as other platforms.

@RobDolinMS does Windows simply not _currently_ support hooks, or are there fundamental issues which make them impossible to support?

RobDolinMS:
Would a member of @opencontainers/runtime-spec-maintainers please assign this to me?

RobDolinMS:
@mrunalp asks: If hooks are not supported, can they just be not added to the config?

crosbymichael:
I believe this is covered by our wording that if a system cannot fulfill a field in the spec that the runtime should error out.  Closing this one.

PR #476, opened 2016-06-01
[ Config | Root Config ] Clarify readonly

RobDolinMS:
Clarifies how root config of readonly==true should be handled if not supported by the platform.  

### DIFF
--- a/config.md
+++ b/config.md
@@ -27,6 +27,7 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 
 * **`path`** (string, required) Specifies the path to the root filesystem for the container. A directory MUST exist at the path declared by the field.
 * **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container. Defaults to false.
+If readonly is present and set to true, but not supported by the platform, the platform MAY ignore.  
 
 ### Example
 


config.md
```diff
@@ -27,6 +27,7 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 
 * **`path`** (string, required) Specifies the path to the root filesystem for the container. A directory MUST exist at the path declared by the field.
 * **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container. Defaults to false.
+If readonly is present and set to true, but not supported by the platform, the platform MAY ignore.
```
mrunalp: I think that failing may be a better option.

philips: I agree failing seems better.

cyphar: In general with configs, failing is much better than silently ignoring errors. A readonly path means that the user has made certain assumptions about the setup and might be quite confused if suddenly things break underneath then.

tianon: +1; @RobDolinMS can you either update, close, or make a good argument for why you think ignoring is preferable? :smile:

wking: On Thu, Jun 09, 2016 at 10:25:55PM -0700, Aleksa Sarai wrote:
> For the Microsoft case, just make it so that your runtime errors out
> if a user specifies readonly (though we should make it clear
> somewhere in the spec whether or not all runtimes need to implement
> all optional options).

I've taken a stab at part of the general behavior in #559.  I'm not
quite sure how to flag options which a runtime can not support while
still being certified.  Some thoughts on that in #559 too.

tianon:
@RobDolinMS was this change covered by https://github.com/opencontainers/runtime-spec/pull/559? (ie, can this one be closed or updated? :smile: :pray:)

cyphar:
@tianon I think this needs to be updated to specify that not being able to implement `readonly` is fine (otherwise someone might think that some runtimes are not valid implementations of the OCI spec).

wking:
On Sat, Nov 05, 2016 at 08:35:58PM -0700, Aleksa Sarai wrote:
> I think this needs to be updated to specify that not being able to
> implement `readonly` is fine (otherwise someone might think that
> some runtimes are not valid implementations of the OCI spec).

If we add optional runtime support for a feature, we need to declare
that feature as a separate protocol in our protocol list [1].

But ideally we can just require runtime support.  It sounds like
Windows *can* support readonly mounts in general [2,3]; is there a
specific problem with the root filesystem?

[1]: https://github.com/opencontainers/runtime-spec/pull/472#issuecomment-258567371
[2]: https://github.com/opencontainers/runtime-spec/pull/504#r68169252
[3]: https://github.com/docker/docker/pull/26577

mrunalp:
Closing this. @RobDolinMS please re-open if you disagree.

PR #478, opened 2016-06-01
Makefile: filename docs.* -> oci-runtime-spec.*

vbatts:
When this repo was only 'specs', then the generic name was not so bad.
But now there is also the oci-image-spec, so this lines up it's unique
name as well.

This also variablizes the output filename so it will be easier for
release specific names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,17 @@
 
-SHELL  ?= $(shell command -v bash 2>/dev/null)
-DOCKER ?= $(shell command -v docker 2>/dev/null)
-PANDOC ?= $(shell command -v pandoc 2>/dev/null)
+EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
+OUTPUT_DIRNAME		?= output/
+DOC_FILENAME		?= oci-runtime-spec
+SHELL			?= $(shell command -v bash 2>/dev/null)
+DOCKER			?= $(shell command -v docker 2>/dev/null)
+PANDOC			?= $(shell command -v pandoc 2>/dev/null)
 ifeq "$(strip $(PANDOC))" ''
 	ifneq "$(strip $(DOCKER))" ''
 		PANDOC = $(DOCKER) run \
 			-it \
 			--rm \
 			-v $(shell pwd)/:/input/:ro \
-			-v $(shell pwd)/output/:/output/ \
+			-v $(shell pwd)/$(OUTPUT_DIRNAME)/:/$(OUTPUT_DIRNAME)/ \
 			-u $(shell id -u) \
 			vbatts/pandoc
 		PANDOC_SRC := /input/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,14 +1,17 @@
 
-SHELL  ?= $(shell command -v bash 2>/dev/null)
-DOCKER ?= $(shell command -v docker 2>/dev/null)
-PANDOC ?= $(shell command -v pandoc 2>/dev/null)
+EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
+OUTPUT_DIRNAME		?= output/
```
wking: This is pretty similar to `PANDOC_DST`.  It's probably better to combine them into a single variable.

vbatts: Well, no. They're similar, but not the same.

wking: On Wed, Jun 01, 2016 at 11:58:45AM -0700, Vincent Batts wrote:
> > @@ -1,14 +1,17 @@
> >  
> > -SHELL  ?= $(shell command -v bash 2>/dev/null)
> > -DOCKER ?= $(shell command -v docker 2>/dev/null)
> > -PANDOC ?= $(shell command -v pandoc 2>/dev/null)
> > +EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
> > +OUTPUT_DIRNAME		?= output/
> 
> Well, no. They're similar, but not the same.

Right.  I think they should be unified into a single variable, but
that unification will take a bit of thinking to figure out.

vbatts: so, not in this PR ;-)

wking: On Wed, Jun 01, 2016 at 12:09:19PM -0700, Vincent Batts wrote:
> > @@ -1,14 +1,17 @@
> >  
> > -SHELL  ?= $(shell command -v bash 2>/dev/null)
> > -DOCKER ?= $(shell command -v docker 2>/dev/null)
> > -PANDOC ?= $(shell command -v pandoc 2>/dev/null)
> > +EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
> > +OUTPUT_DIRNAME		?= output/
> 
> so, not in this PR ;-)

This PR is landing the new variable.  Whether two overlapping
variables is confusing enough to be worth blocking a merge is up to
the maintainers ;).  Still unifying them shouldn't be *that* hard.
Maybe just:

  CONTAINER_OUTPUT_DIRNAME = $(OUTPUT_DIRNAME)

here,

  CONTAINER_OUTPUT_DIRNAME = /$(OUTPUT_DIRNAME)

inside the ifneq DOCKER block, and PANDOC_DST →
CONTAINER_OUTPUT_DIRNAME in the rules?

crosbymichael:
LGTM

hqhq:
LGTM

PR #483, opened 2016-06-02
*: Remove hooks

### LINKED ISSUE #20: Support Lifecycle Hooks

### LINKED ISSUE #473: Drop hooks (now that we have a create/start split)
With #384, I expect we will be dropping hooks on all platforms.  The only uncertainty is around [pre-pivot mount manipulation][1], and I'm [not clear on why you need a hook for that][2].  In the absence of a spec-defined `run` ([added][3] at one point during #384 development but removed before merge), I see no reason to keep hooks around.

Can anyone explain why we need hooks for pre-pivot mount manipulations?  Or what other use-cases need hooks (vs. calling stuff between create and start)?  Otherwise we should drop them to avoid
complications like #256 and #472.

[1]: https://github.com/opencontainers/runtime-spec/pull/384#issuecomment-220038190
[2]: https://github.com/opencontainers/runtime-spec/pull/384#issuecomment-220740126
[3]: https://github.com/opencontainers/runtime-spec/pull/384#issue-147984345

wking:
This drops all instances (outside the ChangeLog) turned up by
case-insensitive searches for 'hook'.

The desire to drop hooks has been around [for a while now][1], but it
took a while to land a create/start split.  By the time that split did
land, dropping hooks was still [part of its motivation][2].  Hooks are
hard to [support robustly][3], and after the create/start split,
callers can do whatever they like at those times without having to [go
through the runtime][4].  There is still a use-case for folks who
prefer the old all-in-one UX, but we can support those folks with a
[higher-level wrapper built on hookless create/start primatives][5].
There was some last-minute discussion of [pre-pivot mount propagation
needing pre-start hooks][6], but that use case can be addressed by
[manipulating the mounts array][7].  With those arguments in place, the
consensus at today's meeting seemed in favor of [removing hooks from
the spec][8].

Fixes #473.
Fixes #472.
Fixes #427.
Fixes #395.
Fixes #20.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2015/opencontainers.2015-10-28-17.02.log.html#l-71
[2]: https://github.com/opencontainers/runtime-spec/pull/384#issue-147984345
[3]: https://github.com/opencontainers/runtime-spec/pull/265
[4]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-03-24.log.html#t2016-03-24T18:56:15
[5]: https://groups.google.com/a/opencontainers.org/d/msg/dev/Y7p6YW8zr4s/OVaAI_WDBAAJ
[6]: https://github.com/opencontainers/runtime-spec/pull/384#issuecomment-220038190
[7]: https://github.com/opencontainers/runtime-spec/pull/384#issuecomment-220740126
[8]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-06-01-17.01.log.html#l-83

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,8 @@ To provide context for users the following section gives example use cases for e
 #### Application Bundle Builders
 
 Application bundle builders can create a [bundle](bundle.md) directory that includes all of the files required for launching an application as a container.
-The bundle contains an OCI [configuration file](config.md) where the builder can specify host-independent details such as [which executable to launch](config.md#process-configuration) and host-specific settings such as [mount](config.md#mounts) locations, [hook](config.md#hooks) paths, Linux [namespaces](config-linux.md#namespaces) and [cgroups](config-linux.md#control-groups).
+The bundle contains an OCI [configuration file](config.md) where the builder can specify host-independent details such as [which executable to launch](config.md#process-configuration) and host-specific settings such as [mount](config.md#mounts) locations, Linux [namespaces](config-linux.md#namespaces) and [cgroups](config-linux.md#control-groups).
 Because the configuration includes host-specific settings, application bundle directories copied between two hosts may require configuration adjustments.
-
-#### Hook Developers
-
-[Hook](config.md#hooks) developers can extend the functionality of an OCI-compliant runtime by hooking into a container's lifecycle with an external application.
-Example use cases include sophisticated network configuration, volume garbage collection, etc.
 
 #### Runtime Developers
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -45,11 +45,3 @@ Systems:
 * Linux
 
 *Owner:* robdolinms as lead coordinator
-
-### Full Lifecycle Hooks
-
-Ensure that we have lifecycle hooks in the correct places with full coverage over the container lifecycle.
-
-Will probably go away with Vish's work on splitting create and start, and if we have exec.
-
-*Owner:*
--- a/config.md
+++ b/config.md
@@ -245,76 +245,6 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
 }
 ```
 
-## Hooks
-
-Lifecycle hooks allow custom events for different points in a container's runtime.
-Presently there are `Prestart`, `Poststart` and `Poststop`.
-
-* [`Prestart`](#prestart) is a list of hooks to be run before the container process is executed
-* [`Poststart`](#poststart) is a list of hooks to be run immediately after the container process is started
-* [`Poststop`](#poststop) is a list of hooks to be run after the container process exits
-
-Hooks allow one to run code before/after various lifecycle events of the container.
-Hooks MUST be called in the listed order.
-The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
-
-Hook paths are absolute and are executed from the host's filesystem in the [runtime namespace][runtime-namespace].
-
-### Prestart
-
-The pre-start hooks are called after the container process is spawned, but before the user supplied command is executed.
-They are called after the container namespaces are created on Linux, so they provide an opportunity to customize the container.
-In Linux, for e.g., the network namespace could be configured in this hook.
-
-If a hook returns a non-zero exit code, then an error including the exit code and the stderr is returned to the caller and the container is torn down.
-
-### Poststart
-
-The post-start hooks are called after the user process is started.
-For example this hook can notify user that real process is spawned.
-
-If a hook returns a non-zero exit code, then an error is logged and the remaining hooks are executed.
-
-### Poststop
-
-The post-stop hooks are called after the container process is stopped.
-Cleanup or debugging could be performed in such a hook.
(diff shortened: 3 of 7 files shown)


mrunalp:
I'm not sure about this and have started a thread on the mailing list.

wking:
On Thu, Jun 02, 2016 at 08:38:24AM -0700, Mrunal Patel wrote:
> I'm not sure about this and have started a thread on the mailing
> list.

Link to thread [1].

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Y7p6YW8zr4s
     Subject: Hooks and all-in-one operation
     Date: Wed, 1 Jun 2016 11:38:27 -0700
     Message-ID: <CANEZBD69Q4VV4UoBodWjtAR1PrJ1OysaMUHgA5cW_aGj2ZEhLQ@mail.gmail.com>

jamesodhunt:
This isn't strictly relevant to the OCI spec, but hooks are currently used by docker to configure the container network. If they go away, it's unclear to me how networking will be handled, so this change could well break docker.

Maybe hooks should be retained until there is some sort of OCI network spec since, like annotations, they are a useful facility to handle arbitrary container scenarios.

wking:
On Fri, Jun 17, 2016 at 01:53:59AM -0700, James Hunt wrote:
> If they go away, it's unclear to me how networking will be handled…

We should keep discussion in @mrunalp's thread [1] to avoid
fragmentation.  I think [2] addresses your concerns, but if not, post
to the list thread with why ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/483#issuecomment-223348774
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/Y7p6YW8zr4s/NtiVw9p2AwAJ
     Subject: Re: Hooks and all-in-one operation
     Date: Thu, 9 Jun 2016 14:50:49 -0700
     Message-ID: <20160609215049.GK8801@odin.tremily.us>

wking:
On Fri, Jun 17, 2016 at 09:09:31AM -0700, W. Trevor King wrote:
> We should keep discussion in @mrunalp's thread [1] to avoid
> fragmentation.
> …
> [1]: https://github.com/opencontainers/runtime-spec/pull/483#issuecomment-223348774

I posted to the thread on Tuesday asking for guidance now that the
consensus seems to have swung back around to “drop hooks”, but haven't
heard back.  Rebasing this into master turned out to be not
particularly bad, so I just went ahead and did it with ea41a14 →
0113d7a.

wking:
So consensus on *today's* meeting was to keep hooks [1].  To square
that against the DockerCon decision to drop them [2], the current
consensus on facts seem to be:

a. The create/start split lets you do any hook-like things you like
   without hooks.
b. There are significant existing users who are using hooks as they
   stand.
c. Specifying hooks will take some effort to get right.
d. The OCI will likely manage the hook spec if that gets punted to a
   higher layer.
e. Nobody wants folks relying on unspecified behavior for important
   stuff like network setup.
f. Once you put something into 1.0, you need to wait until 2.0 to take
   it back out.

Possible approaches:

i. Include both hooks and the create/start split in this spec.
ii. Require the create/start split in this spec, and define hooks but
    make them a separate protocol [3].
iii. Punt the hook specification to a higher layer (possibly still
     within the OCI, see (d)), and have a tool that reads a config
     containing hook information, wraps the base runtime, and calls
     the hook commands at the appropriate times.

The DockerCon decision seemed to revolve around (a) and (f), arguing
for removing hooks as a conservative step, and possibly adding them
back after 1.0 if there was a demonstrated need (although in the light
of (a), I'm not sure what that would have looked like).

Today's decision seemed to revolve around (b), (d), and (e).  Because
of (d), punting doesn't really effect OCI time, so (c) is less
important for the OCI.  It does slightly raise the cost for base
runtime implementers (who now have to handle hooks), but they aren't
that complicated to implement (they're more complicated to
specify/test ;).

The distinction between these three choices is small because of (d),
and the only real issue with carving hooks out into a separate
protocol/spec or not is (f).  The consensus on today's call was that
the backwards-compatibility inertia from (b) was sufficient that (f)
was unlikely to be an issue.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-08-17-20.59.log.html#l-61
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-08-03-21.01.log.html#l-21
[3]: https://github.com/opencontainers/runtime-spec/pull/527

Issue #480, opened 2016-06-01
What does an impl do with Annotations?


duglin:
The spec defines annotations in the config.json but it doesn't anything else about them.
For example, is that info suppose to be available via a `runc state id` type of operation?
If that info isn't retrievable/queriable then what is it for?

wking:
On Wed, Jun 01, 2016 at 10:40:35AM -0700, Doug Davis wrote:
> The spec defines annotations in the config.json but it doesn't
> anything else about them.

They are explicitly opaque (“arbitrary” [1]); the runtime doesn't care
what you do with them.

> For example, is that info suppose to be available via a `runc state
> id` type of operation?

You can get to the original config.json using the state's bundlePath
[2,3].

[1]: https://github.com/opencontainers/runtime-spec/blob/3b166cd0c2c1303e978754188a6e6c655cd6f422/config.md#annotations
[2]: https://github.com/opencontainers/runtime-spec/blob/3b166cd0c2c1303e978754188a6e6c655cd6f422/runtime.md#state
[3]: https://github.com/opencontainers/runtime-spec/pull/188#issuecomment-147026995

duglin:
Is that the workflow people are expecting?

Meaning: use "runc xxx" to manipulate and get the state of the container but then access the filesystem directly to retrieve the annotations/config.json?  

Would including config.json as part of the output of `state()` be so bad?

wking:
On Wed, Jun 01, 2016 at 11:00:41AM -0700, Doug Davis wrote:
> Is that the workflow people are expecting?
>
> Meaning: use "runc xxx" to manipulate and get the state of the
> container but then access the filesystem directly to retrieve the
> annotations/config.json?

For more background on this approach (but in the context of mounting),
see assorted rebuttals to my pushback in this thread [1].

> Would including config.json as part of the output of `state()` be so
> bad?

My problem with including it is that the config.json entries may be
stale by the time you find them in the state.  Folks should just ask
the kernel [2].  But the kernel will not know about annotations and
such.  I don't have a problem passing config.json's annotations
through to the state JSON [3].  But what's your use-case?  It may be
better to manage your own state-like registry with this sort of
metadata.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Zogdyb_hrsg
     Subject: Drop 'root' from state.json?
     Date: Thu, 3 Sep 2015 12:50:59 -0700
[...]

PR #484, opened 2016-06-02
Add annotations to the state json

duglin:
Closes: #480

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -23,6 +23,8 @@ The value MAY be one of:
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
+* **`annotations`**: (map) contains the list of annotations associated with the container.
+If no annotations were provided then this property MAY either be absent or an empty map.
 
 When serialized in JSON, the format MUST adhere to the following pattern:
 
@@ -32,7 +34,10 @@ When serialized in JSON, the format MUST adhere to the following pattern:
     "id": "oci-container1",
     "status": "running",
     "pid": 4422,
-    "bundlePath": "/containers/redis"
+    "bundlePath": "/containers/redis",
+    "annotations": {
+        "myKey": "myValue"
+    }
 }
 ```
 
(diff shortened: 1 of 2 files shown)


wking:
On Thu, Jun 02, 2016 at 01:25:37PM -0700, Doug Davis wrote:
>   * Add annotations to the state json

For previous work in this direction, see #188 (which was about adding
runtime-specified annotations to the state).  I think that was closed
prematurely after the config-author-specified annotations landed in
#331 [1].  Is there still interest in the runtime-specified
annotations use-case that lead to #188?  For example, runC apparently
injects a ‘bundle’ field into annotations to get it into the state
JSON [2].

When I suggested including config-author-specified content in the
state [3], @gao-feng pointed out that you could retrieve that
information via bundlePath [4].  If that route to the annotations is
not acceptable, should we just copy the whole config.json into the
state JSON [5]?  If not, how do we pick and choose what gets copied
over?

[1]: https://github.com/opencontainers/runtime-spec/issues/480#issuecomment-223405337
[2]: https://github.com/opencontainers/runc/pull/869/files#diff-e75bc7288fdc0ddc587b4152ecbd3933R105
[3]: https://github.com/opencontainers/runtime-spec/pull/188#issuecomment-146912525
[4]: https://github.com/opencontainers/runtime-spec/pull/188#issuecomment-147026995
[5]: https://github.com/opencontainers/runtime-spec/issues/480#issuecomment-223405337

crosbymichael:
LGTM

mrunalp:
While this looks fine, we need to be clear about what we copy from the configuration to the state as it is essentially redundant.

crosbymichael:
@mrunalp yes,  Since annotations are more user defined "labels" for a container not part of the configuration that create the container and its settings it makes since for something like this to be added to the state allow it to be a queried

mrunalp:
@crosbymichael They could also potentially be used to implement extensions that aren't yet in the spec.

crosbymichael:
@mrunalp yep, its something for external tooling, the runtime will just passthough and take no actions on the values.

mrunalp:
I am fine merging for now and we could maybe revisit if we see a need to move other values into the state. Thanks!

mrunalp:
LGTM

PR #486, opened 2016-06-03
Bump v1.0rc1

vbatts:
where ef1aa37acb7052e5b44948550fceb98ed94481a9 would be tagged `v1.0.0-rc1`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,74 @@
 OpenContainers Specifications
 
+Changes with v1.0.0-rc1:
+	Breaking changes:
+
+	* runtime: Split create and start, #384, #450, #463, #464, #467,
+	  #468
+	* runtime: Remove exec, #388
+	* runtime: Enviroment MUST match the configuration, #397
+	* config: Runtime MUST generate errors for unsupported platforms,
+	  #441
+	* config: Windows mount destinations MUST NOT be nested, #437
+
+	Additions:
+
+	* solaris: Added platform-specific configuration, #411, #424, #431,
+	  #436
+	* runtime: Add 'annotations' and 'status' to the state structure,
+	  #462, #484, #485
+	* runtime: State no longer needs to be serialized as JSON, #446
+	* runtime-linux: Add /dev symbolic links, #449
+	* config: Allow absolute paths for root.path (which previously
+	  required relative paths), #394
+	* config-linux: Add linux.mountLabel, #393
+	* config-linux: Add suport for cgroup namespace, #397
+	* config-linux: Runtime SHOULD NOT modify ownership of any
+	  referenced filesystem (previously the restriction only applied to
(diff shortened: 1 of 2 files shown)


wking:
On Fri, Jun 03, 2016 at 12:17:15PM -0700, Vincent Batts wrote:
>   * version: v1.0.0-rc1

The series through 41bfb585 looks fine to me, although I went with
“rc.1” in the ChangeLog to match SemVer's own versioning [1].  §9
talks about “dot separated identifiers” in the pre-release version.
But it probably only matters for sorting (e.g. rc10 probably sorts
before rc9, but rc.10 probably sorts after rc.9).

[1]: http://semver.org/v2.0.0.html

vbatts:
@wking the link is a 404.

I'm not really sure that I've run into `-rc.N` much if ever. While I don't entirely care, it is very common to see `-rcN`

wking:
On Fri, Jun 03, 2016 at 12:29:29PM -0700, Vincent Batts wrote:
> @wking the link is a 404.

Sorry, should have been: http://semver.org/spec/v2.0.0.html

> I'm not really sure that I've run into `-rc.N` much if ever. While I
> don't entirely care, it is very common to see `-rcN`

We're unlikely to hit rc.10 ;).  I think we should follow the SemVer
example, but yeah, it doesn't matter much.  Whichever way we go, the
ChangeLog, subsequent Go, and tag should be consistent.

crosbymichael:
rc1 over rc.1 is much better

crosbymichael:
LGTM

wking:
On Fri, Jun 03, 2016 at 12:58:28PM -0700, Michael Crosby wrote:
> rc1 over rc.1 is much better

I don't see *much* difference, and certainly nothing superior about
the short form, but feel free to reroll my ChangeLog commit to adjust
‘rc.1’ → ‘rc1’.

crosbymichael:
It already is adjusted

tianon:
LGTM

IIRC, rc.N is primarily a python thing, but IMO rcN looks better and is much more common

mrunalp:
LGTM

crosbymichael:
@vbatts let me know if you get a chance to push the tag or i can do it

crosbymichael:
06479209bdc0d4135911688c18157bd39bd99c22 should be the new commit to tag correct?

wking:
On Fri, Jun 03, 2016 at 03:03:17PM -0700, Michael Crosby wrote:
> 06479209bdc0d4135911688c18157bd39bd99c22 should be the new commit to tag correct?

Yup.

crosbymichael:
ok, pushed the tag

wking:
On Fri, Jun 03, 2016 at 03:23:14PM -0700, Michael Crosby wrote:
> ok, pushed the tag

Not signed?

PR #487, opened 2016-06-04
bundle: Remove distribution references

wking:
These distribution requirements belong in image-spec or similar.  They
don't apply to runtimes or filesystem bundles (the latter are covered
by the earlier "This MUST include the following artifacts"), which are
the two entities tested for compliance with this spec.

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -12,12 +12,10 @@ This MUST include the following artifacts:
 
 1. `config.json` : contains configuration data.
 This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
-When the bundle is packaged up for distribution, this file MUST be included.
 See [`config.json`](config.md) for more details.
 
 2. A directory representing the root filesystem of the container.
 While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
-When the bundle is packaged up for distribution, this directory MUST be included.
 This directory MUST be referenced from within the `config.json` file.
 
 While these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.


bundle.md
```diff
@@ -12,12 +12,10 @@ This MUST include the following artifacts:
 
 1. `config.json` : contains configuration data.
 This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
-When the bundle is packaged up for distribution, this file MUST be included.
 See [`config.json`](config.md) for more details.
 
 2. A directory representing the root filesystem of the container.
 While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
-When the bundle is packaged up for distribution, this directory MUST be included.
```
vbatts: this feels like a drive-by of MUST ...

wking: On Mon, Jun 06, 2016 at 08:16:10AM -0700, Vincent Batts wrote:
> >  See [`config.json`](config.md) for more details.
> >  
> >  2. A directory representing the root filesystem of the container.
> >  While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
> > -When the bundle is packaged up for distribution, this directory MUST be included.
> 
> this feels like a drive-by of MUST ...

My removal is a drive by?  Or the current sentences are drive-bys?  If
the latter, are you suggesting we keep the sentences (with “MUST” →
“must”) as toothless, informative context?  I'd rather just link to
image-spec's “what portions of the bundle do we transport” docs
(creating them if they don't exist) to provide that context.

hqhq:
These sentences are out of scope to me, and hard to be tested, so

LGTM

mrunalp:
LGTM

PR #488, opened 2016-06-04
Revert "Clarify backwards compatibility for major version 0"

wking:
This reverts commit 0f25f18b9b6dd315ba357d7f4f10d636474d7994 (#253).  Now that we're on to 1.0, we don't need to talk about 0.x.  And the lack of 0.x backwards compatability is covered by [SemVer 2.0 section 4][1]:

> Major version zero (0.y.z) is for initial development.  Anything may change at any time.  The public API should not be considered stable.

so removing the echo from our spec doesn't actually change anything.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #489, opened 2016-06-05
config: Make 'process.args' optional

wking:
Since be59415 (Split create and start, 2016-04-01, #384), it's possible for a container process to never execute user-specified code (e.g. you can call ‘create’, ‘kill’, ‘delete’ without calling ‘start’).  For folks who expect to do that, there’s no reason to define`process.args`.

The only other `process` property required for all platforms is `cwd`, but the runtime's idler code isn't specified in sufficient detail for the configuration author to have an opinion about what its working directory should be.

On Linux and Solaris, `uid` and `gid` are also required.  My preferred approach here is to [make those optional](https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/DWdystx5X3A) and [use platform defaults](https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216076069):

> If unset, the runtime will not attempt to manipulate the user ID (e.g. not calling setuid(2) or similar).

But the maintainer consensus is that they want those to be [explicitly](http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-04-17.00.log.html#l-44) [required](https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937010) [properties](https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937090).  With the current spec, one option could be to make process optional (with the idler's working directory unspecified) for OSes besides Linux and Solaris, but the main reason that Windows doesn't have a user property is that [we don't know how to specify it](https://github.com/opencontainers/runtime-spec/pull/96#issuecomment-128410585).  I expect all platforms will have some sort of required user field, which means they'll all have to define `process`.

While I'm indenting the sub-properties, also wrap them to [one line per sentence](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/style.md#one-sentence-per-line).

There was also some [related discussion](https://github.com/opencontainers/runc/pull/827#discussion_r63464201) in opencontainers/runc#827, although I'm not clear on the intended policy after that discussion.

### DIFF
--- a/config.md
+++ b/config.md
@@ -99,6 +99,8 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
   * **`args`** (array of strings, optional) executable to launch and any flags as an array.
     The executable is the first element and MUST be available at the given path inside of the rootfs.
     If the executable path is not an absolute path then the search $PATH is interpreted to find the executable.
+  * **`user`** (object, required) the process user.
+    The properties for this object are [platform dependent](#user).
 
   For Linux-based systems the process structure supports the following process specific fields:
 
--- a/config.md
+++ b/config.md
@@ -84,26 +84,35 @@ For the Windows operating system, one mount destination MUST NOT be nested withi
 See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [SetVolumeMountPoint](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365561(v=vs.85).aspx) in Windows.
 
 
-## Process configuration
+## Process
 
-* **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.
-* **`cwd`** (string, required) is the working directory that will be set for the executable. This value MUST be an absolute path.
-* **`env`** (array of strings, optional) contains a list of variables that will be set in the process's environment prior to execution. Elements in the array are specified as Strings in the form "KEY=value". The left hand side MUST consist solely of letters, digits, and underscores `_` as outlined in [IEEE Std 1003.1-2001](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html).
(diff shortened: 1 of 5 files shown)


wking:
I added a second commit that makes `process.user` required, since that was the reason I used above to make `process` required.

crosbymichael:
NOTLGTM

I don't like the proposal but I can be swayed if other maintainers feel strongly for it.

wking:
On Fri, Nov 04, 2016 at 3:37pm -0700, Michael Crosby wrote:
> I don't like the proposal…

Are you rejecting the “it's possible for a container process to never execute user-specified code (e.g. you can call ‘create’, ‘kill’, ‘delete’ without calling ‘start’)” use case?  Or do you accept that use case but feel they should insert a dummy `process.args`?  Or do you accept the use case *and* the optional `process.args` but have some technical quibbles with this PR?

mrunalp:
I am in favor of having the user specify the args. Even though the flow where the command is never executed is supported, that isn't a normal flow but more for scheduling or gc where one removes the container without ever starting the container after creating it.

crosbymichael:
Well its hard to tell what really changed in this PR because you changed an entire block instead of what you were talking about in the description.  There are more changes in this PR than just making args optional, therefore, I'm going to close it.

If you want to propose a single change that is easy for us to review and understand please open a new, small, and concise PR with only the change you are proposing and we can continue the discussion there for optional args.  You are making the process hard on the maintainers and yourself by your current strategy.  Multiple small, focused PRs are much better for everyone than big ones with unrelated changes.

wking:
On Fri, Nov 11, 2016 at 05:23PM -0800, Michael Crosby wrote:
> If you want to propose a single change that is easy for us to review and understand please open a new, small, and concise PR with only the change you are proposing and we can continue the discussion there for optional args.

Filed in #620.

PR #490, opened 2016-06-06
schema/validate: Allow schema identifiers to contain a URL scheme

wking:
And only fall back to assuming they're a local file if they don't
already contain :// (which is unlikely to show up in local paths).
This makes it easy to validate configurations against different
versions of the JSON Schema (e.g. v0.5.0 or v1.0.0-rc1).

Ping @vbatts.

### DIFF
--- a/schema/README.md
+++ b/schema/README.md
@@ -37,3 +37,9 @@ Then use it like:
 ```bash
 ./validate schema.json <yourpath>/config.json
 ```
+
+Or like:
+
+```bash
+./validate https://raw.githubusercontent.com/opencontainers/runtime-spec/v1.0.0-rc1/schema/schema.json <yourpath>/config.json
+```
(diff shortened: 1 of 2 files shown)


tianon:
@vbatts what're your thoughts on this?  I'm personally somewhere between neutral to +1 on the functionality since it seems like a nifty feature (although I'm not certain whether just testing whether the path contains `://` somewhere is sufficient or whether it'd make sense to do something with testing for file existence / valid URL parsing?)

wking:
On Fri, Nov 04, 2016 at 03:39:26PM -0700, Tianon Gravi wrote:
> … although I'm not certain whether just testing whether the path
> contains `://` somewhere is sufficient or whether it'd make sense to
> do something with testing for file existence / valid URL parsing?…

I think :// is unlikely in a genric filesystem path.  And folks who
want to make it explicit can always use file://yes://I://am://crazy://
;).

mrunalp:
@vbatts PTAL

vbatts:
LGTM

mrunalp:
LGTM

PR #493, opened 2016-06-08
config-linux: clarify cgroup requirements

cyphar:
This helps clear up some of the vagueness around cgroupsPath in the spec. It's
loosely modeled on what runC does (as well as allowing for custom cgroup
managers in runtimes). I also updated the cgroup section to make it clear that
lazy cgroup setup (where you only attach to the bare minimum of requested
cgroups) is allowed under the spec. This all originated from an IRC discussion
on #opencontainers.

This possibly obsoletes #270 (although that has some renaming going on).

/cc @opencontainers/runtime-spec-maintainers @wking

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -175,6 +175,9 @@ You can configure a container's cgroups via the `resources` field of the Linux c
 Do not specify `resources` unless limits have to be updated.
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
+A runtime MUST at least use the minimum set of cgroup controllers required to fulfill the `resources` settings.
+However, a runtime MAY attach the container process to additional cgroup controllers supported by the system.
+
 ###### Example
 
 ```json
--- a/config-linux.md
+++ b/config-linux.md
@@ -154,28 +154,27 @@ In addition to any devices configured with this setting, the runtime MUST also s
 ## Control groups
 
 Also known as cgroups, they are used to restrict resource usage for a container and handle device access.
-cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
+cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
+`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
 `cgroupsPath` is expected to be relative to the cgroups mount point.
 If `cgroupsPath` is not specified, implementations can define the default cgroup path.
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
-The Spec does not support [split hierarchy][cgroup-v2].
+The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
 The cgroups will be created if they don't exist.
+
+You can configure a container's cgroups via the `resources` field of the Linux configuration.
+Do not specify `resources` unless limits have to be updated.
+For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
 ###### Example
 
 ```json
    "cgroupsPath": "/myRuntime/myContainer"
 ```
-
-`cgroupsPath` can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
-
-You can configure a container's cgroups via the `resources` field of the Linux configuration.
-Do not specify `resources` unless limits have to be updated.
-For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
 #### Device whitelist
 
--- a/config-linux.md
+++ b/config-linux.md
@@ -159,8 +159,13 @@ For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
 `cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+If `cgroupsPath` is:
+* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
+* ... a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
+* ... not specified, the runtime MAY define the default cgroup path.
+Runtimes MAY consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
+If a `cgroupsPath` value is specified, the runtime MUST consistently attach to the same place in the cgroup hierarchy given the same value of `cgroupsPath`.
+
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
 The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
--- a/config-linux.md
+++ b/config-linux.md
@@ -181,7 +181,19 @@ However, a runtime MAY attach the container process to additional cgroup control
 ###### Example
 
 ```json
-   "cgroupsPath": "/myRuntime/myContainer"
+   "cgroupsPath": "/myRuntime/myContainer",
+   "resources": {
+      "memory": {
+         "limit": 100000,
+         "reservation": 200000
+      },
+      "devices": [
+         {
+            "allow": false,
+            "access": "rwm"
+         }
+      ]
+   }
 ```
 
 #### Device whitelist


config-linux.md
```diff
@@ -158,13 +158,20 @@ cgroups provide controls to restrict cpu, memory, IO, pids and network for the c
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+If `cgroupsPath` is
+* ... an absolute path (starting with `/`), the implementation MUST take the path to be relative to the cgroup mount point.
+* ... a relative path (not starting with `/`), implementations SHOULD take the value to be the suffix of the resulting cgroup path.
```
vishh: nit: What does `resulting cgroup path` mean in this context? Did you mean `relative to the cgroups of the runtime process`?

config-linux.md
```diff
@@ -158,13 +158,20 @@ cgroups provide controls to restrict cpu, memory, IO, pids and network for the c
...
+* ... not specified, implementations CAN define the default cgroup path.
+Implementations CAN consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
```
vishh: Instead, can we define what the format here? #270 attempts to address the format issue.

config-linux.md
```diff
@@ -158,13 +158,20 @@ cgroups provide controls to restrict cpu, memory, IO, pids and network for the c
...
+
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
 The Spec does not support [split hierarchy][cgroup-v2].
 The cgroups will be created if they don't exist.
 
+A runtime MUST at least use the minimum set of cgroups required to fulfil the resource restrictions specified by `config.json`.
+However, a runtime CAN decide to not attach to all cgroups supported by the system.
```
vishh: How about the case when no resource restrictions are specified? Are the runtimes expected to attach to specific cgroups subsystems by default? `devices` is an example here.

wking: On Wed, Jun 08, 2016 at 10:46:14AM -0700, Vish Kannan wrote:
> > +A runtime MUST at least use the minimum set of cgroups required to fulfil the resource restrictions specified by `config.json`.
> > +However, a runtime CAN decide to not attach to all cgroups supported by the system.
> 
> How about the case when no resource restrictions are specified?

And while we're overhauling this language, we may want to address
opting-out of runtime-managed cgroups (#237).

wking: On Wed, Jun 08, 2016 at 11:34:24AM -0700, W. Trevor King wrote:
> And while we're overhauling this language, we may want to address
> opting-out of runtime-managed cgroups (#237).

Actually, we may be able to avoid pulling that in here.  Your:

  * ... not specified, implementations CAN define the default cgroup path.

and:

  However, a runtime CAN decide to not attach to all cgroups supported
  by the system.

allow runtimes to handle unset cgroupsPath by joining whatever
controllers they want or none at all.  Since an explicit opt-out has
been contentious (although I still don't understand why), that
potential opt-out (at the runtime's discretion) is good enough for
now.

config-linux.md
cyphar: I'm trying to make it not-too restrictive. There's no reason that a runtime has to make it relative to the cgroups of the runtime process. "Should be the suffix" means that it should be the last component in the path (but component is the wrong word since it can container `/`).

config-linux.md
cyphar: The reason I wrote it like this is so that implementations have freedom when implementing cgroup managers. We already had this issue with systemd, where you have very odd restrictions on naming. Rather, I think that "you can come up with your own formatting rules if it's not an absolute path" is an entirely valid (and IMO better way than explicitly restricting things) way of doing things.

config-linux.md
cyphar: @vishh The devices cgroup thing is honestly an implementation detail of `runC`. Since users can drop `CAP_MKNOD` (and use user namespaces) there isn't a security issue with runtimes not joining devices -- as far as I can see (users can always set the device restrictions for the container anyway).

cyphar: @vishh The devices cgroup thing is honestly an implementation detail of `runC`. Since users can drop `CAP_MKNOD` (and use user namespaces) there isn't a security issue with runtimes not joining devices -- as far as I can see (users can always set the device restrictions for the container anyway).

>  How about the case when no resource restrictions are specified?

Then the runtime doesn't need to join any cgroups. However, we should probably add an "update" step to the lifecycle so that we can specify that a runtime MUST join any cgroups required to fulfill the resource restriction updates (using the same `cgroupsPath` rules as with setup).

config-linux.md
wking: On Wed, Jun 08, 2016 at 10:07:28PM -0700, Aleksa Sarai wrote:
> > +* ... a relative path (not starting with `/`), implementations SHOULD take the value to be the suffix of the resulting cgroup path.
>
> I'm trying to make it not-too restrictive. There's no reason that a
> runtime has to make it relative to the cgroups of the runtime
> process. "Should be the suffix" means that it should be the last
> component in the path (but component is the wrong word since it can
> container `/`).

How about:

  … MAY interpret the path relative to a runtime-determined location
  in the cgroup hierarchy.

config-linux.md
wking: On Wed, Jun 08, 2016 at 10:12:52PM -0700, Aleksa Sarai wrote:
> >  How about the case when no resource restrictions are specified?
> 
> Then the runtime doesn't need to join any cgroups. However, we
> should probably add an "update" step to the lifecycle so that we can
> specify that a runtime MUST join any cgroups required to fulfill the
> resource restriction updates (using the same `cgroupsPath` rules as
> with setup).

There is currently no update operation in the spec [1], and (for
namespaces anyway) join-and-tweak is forbidden [2]:

  Also, when a path is specified, a runtime MUST assume that the setup
  for that particular namespace has already been done and error out if
  the config specifies anything else related to that namespace.

(although I'm fine relaxing that restriction [3]).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#operations
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/config-linux.md#namespaces
[3]: https://github.com/opencontainers/runtime-spec/pull/158#issuecomment-138687129

config-linux.md
cyphar: I prefer SHOULD, since I believe a runtime should have a good reason for it to not do that way. But yes, I like that wording better.

wking: On Thu, Jun 09, 2016 at 05:42:57PM -0700, Aleksa Sarai wrote:
> > +* ... a relative path (not starting with `/`), implementations SHOULD take the value to be the suffix of the resulting cgroup path.
> 
> I prefer SHOULD, since I believe a runtime should have a good reason
> for it to not do that way…

Is which leaf of a random runtime-generated directory you're at all
that important to you?  I don't see a lot of space between MAY and
SHOULD here.

wking: Another thing worth clarifying about these relative paths is whether
they always create a new leaf in the hierarchy.  Given the uncertainty
of the MAY/SHOULD wording for interpreting relative paths, I'd say
that they MUST create a new leaf in the hierarchy (so folks who want
to join an existing node MUST use an absolute cgroupsPath).

cyphar: But then you have the systemd cgroup manager, where you specify a slice (which may not be a new leaf). While this might all change in cgroupv2, currently I'm trying to allow for custom cgroup managers to be compliant as long as they allow "absolute paths mode".

config-linux.md
cyphar: Yeah, I understand. So we'll just implement lazy cgroup handling the right way (TM) in runC and then if we later decide to update the spec to allow for updating we can borrow the base semantics from runC.

config-linux.md
wking: On Thu, Jun 09, 2016 at 10:23:11PM -0700, Aleksa Sarai wrote:
> > +* ... a relative path (not starting with `/`), implementations SHOULD take the value to be the suffix of the resulting cgroup path.
>
> But then you have the systemd cgroup manager, where you specify a
> slice (which may not be a new leaf).

Then maybe require a runtime servicing multiple ‘create’ calls with
identical cgroupsPaths to always join the same cgroups (regardless of
the absolute/relative/unset-ness)?  That way folks using relative
paths for systemd (or whatever) have some idea of whether they'll be
joining an existing node or creating a new leaf in the hierarchy.

cyphar: I've tried to resolve this in the latest version (saying that it must be consistent if a value of cgroupsPath is specified). I don't like the wording though, maybe you could help me reword it?

config-linux.md
```diff
@@ -154,15 +154,20 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
+* ... an absolute path (starting with `/`), the implementation MUST take the path to be relative to the cgroup mount point.
+* ... a relative path (not starting with `/`), implementations SHOULD take the value to be the suffix of the resulting cgroup path.
+* ... not specified, implementations CAN define the default cgroup path.
+Implementations CAN consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
+If a `cgroupsPath` value is specified, an implementation MUST consistently attach to the same place in the cgroup hierarchy given the same value of `cgroupsPath`.
+
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
-The Spec does not support [split hierarchy][cgroup-v2].
+The Spec does not support the [unified or "default" hierarchy][cgroup-v2].
```
wking: The old wording is [“we don't support per-controller paths” for the reasons discussed in cgroup-v2.txt][1]. It's not about limiting the cgroup version.  Although you're now the [second person][2] I've heard interpret the old wording that way, so we probably want to adjust the wording.  I think we *do* want to support v2 cgroups though, and it may be more clear [than I thought][3] if the runtime joins both v1 and v2 cgroups as needed to collect the controllers it wants to join (because a given controller can only be exposed in the v1 or v2 hierarchy, and not in both at once).  So if you want to join the `devices` controller and that controller's in the v2 hierarchy, you join the appropriate v2 cgroup.

[1]: https://github.com/opencontainers/runtime-spec/pull/137#discussion_r38601118
[2]: https://github.com/opencontainers/ocitools/pull/93#discussion_r65658258
[3]: https://github.com/opencontainers/ocitools/pull/93#discussion_r65658513

config-linux.md
```diff
@@ -154,15 +154,20 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
 Also known as cgroups, they are used to restrict resource usage for a container and handle device access.
-cgroups provide controls to restrict cpu, memory, IO, pids and network for the container.
+cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+If `cgroupsPath` is:
+* ... an absolute path (starting with `/`), the implementation MUST take the path to be relative to the cgroup mount point.
```
wking: nit: the spec is not terribly consistent on this (a number of entries probably predate the glossary), but I think we prefer “[runtime][1]” to “implementation”.  Using “runtime” helps distinguish runtime implementations from config-generating implementations and such.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/glossary.md#runtime

config-linux.md
```diff
@@ -154,15 +154,20 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
+* ... a relative path (not starting with `/`), implementations SHOULD take the value to be the suffix of the resulting cgroup path.
+* ... not specified, implementations CAN define the default cgroup path.
+Implementations CAN consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
+If a `cgroupsPath` value is specified, an implementation MUST consistently attach to the same place in the cgroup hierarchy given the same value of `cgroupsPath`.
```
wking: How about:

> Runtimes MUST consistently map the same `cgroupsPath` value to the same nodes in the cgroup hierarchy.  For example, launching two containers that both use `some/path` MUST result in both containers belonging to the same nodes.

config-linux.md
```diff
@@ -171,12 +176,15 @@ The cgroups will be created if they don't exist.
...
 
-`cgroupsPath` can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
 
 You can configure a container's cgroups via the `resources` field of the Linux configuration.
 Do not specify `resources` unless limits have to be updated.
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
+A runtime MUST at least use the minimum set of cgroup controllers required to fulfil the `resources` settings.
+However, a runtime CAN decide to not attach to any additional cgroup controllers supported by the system.
```
wking: I don't like having all of this normative stuff down in the “Example” subsection.  Can we move it all (from “`cgroupsPath` can be used to either…” through this paragraph) up into the “Control groups” section?

wking: And “fulfil” → “fulfill”.  And I'd drop “supported by the system”, since you don't have any choice about joining them or not, and the previous line's MUST covers the “you'll have to error if a necessary controller isn't supported” case.

config-linux.md
crosbymichael: is it important to have this?  Even if it was supported there would be no changes to the spec and this is more of a runtime concern right?

wking: On Fri, Jun 10, 2016 at 11:45:55AM -0700, Michael Crosby wrote:
> > -The Spec does not support [split hierarchy][cgroup-v2].
> > +The Spec does not support the [unified or "default" hierarchy][cgroup-v2].
>
> is it important to have this?  Even if it was supported there would
> be no changes to the spec and this is more of a runtime concern
> right?

It was added to motivate a single cgroup-path property
vs. per-controller paths [1].  I think it's useful to keep the
(informative) motivation documented here (although I think we can
clarify the wording).

[1]: https://github.com/opencontainers/runtime-spec/pull/137#discussion_r38597501

cyphar: But the thing is that we _do_ support split hierarchies (that's the default on any modern GNU/Linux system). The reason I made this change was because the link text was wrong (cgroupv2 is the exact opposite of a split hierarchy).

I understand the point (cgroupsPath is a single value), but there's something wrong about this sentence when I read it.

config-linux.md
cyphar: The reason I moved it here is because the `resources` section isn't mentioned before the examples section. I can move both parts above the examples though.

wking: On Fri, Jun 10, 2016 at 09:07:45PM -0700, Aleksa Sarai wrote:
> The reason I moved it here is because the `resources` section isn't
> mentioned before the examples section. I can move both parts above
> the examples though.

Yeah, I think you should move all this non-example stuff up out of the
example subsection.

config-linux.md
wking: On Fri, Jun 10, 2016 at 09:06:48PM -0700, Aleksa Sarai wrote:
> I understand the point (cgroupsPath is a single value), but there's
> something wrong about this sentence when I read it.

The wording could be improved.  My suggestion is [1]:

  This specification does not support per-controller paths for the
  reasons discussed in the [v2 cgroup docs][cgroup-v2].

[1]: https://github.com/opencontainers/runtime-spec/pull/493/files#r66658407

Mashimiao:
@cyphar  any updates?

cyphar:
@Mashimiao Thanks for pinging me, I completely forgot about this PR. :wink:

@wking There, I've fixed up all of the things you mentioned. PTAL.

config-linux.md
```diff
@@ -154,29 +154,48 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
-The Spec does not support [split hierarchy][cgroup-v2].
+The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
 The cgroups will be created if they don't exist.
 
+You can configure a container's cgroups via the `resources` field of the Linux configuration.
+Do not specify `resources` unless limits have to be updated.
+For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
+
+A runtime MUST at least use the minimum set of cgroup controllers required to fulfil the `resources` settings.
+However, a runtime CAN decide to not attach to any additional cgroup controllers supported by the system.
```
wking: I [still][1] think we want “fulfil” → “fulfill” (although it looks like [that's UK vs. US spelling][2], so not a big deal) and “supported by the system“ → “”.

[1]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r66660766
[2]: https://en.wiktionary.org/wiki/fulfill

config-linux.md
```diff
@@ -154,29 +154,48 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
+
 ###### Example
 
 ```json
    "cgroupsPath": "/myRuntime/myContainer"
+   "resources": {
```
wking: You'll need a comma after the `cgroupsPath` line now.

config-linux.md
```diff
@@ -154,29 +154,48 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
+cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
+If `cgroupsPath` is:
+* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
+* ... a relative path (not starting with `/`), the runtime SHOULD take the value to be the suffix of the resulting cgroup path.
```
wking: This still has the “resulting” language which is unclear.  Previous discussion starting [here][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r66303330

config-linux.md
cyphar: *ahem* UK spelling is clearly superior. :wink:

cyphar:
@wking There, I've fixed up all of your concerns I *think*.

config-linux.md
```diff
@@ -154,29 +154,48 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
+cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
+If `cgroupsPath` is:
+* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
+* ... a relative path (not starting with `/`), the runtime MUST create a new leaf in the cgroup hierarchy and MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
```
wking: “MUST create a new leaf” conflicts with the later “…MUST consistently attach to the same place …” (see also [here][1]).  Now that we have the consistently-attach wording, I'm ok with weaker wording like [this][2], and don't really care about [MAY vs. SHOULD][3].

[1]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r66563235
[2]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r66475713
[3]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r66557226

cyphar: What about something like:

```
* ... a relative path (not starting with `/`), the runtime MUST create a new leaf in the cgroup unless the same lexical path is used by another existing container [and MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy].
```

Where "existing" means that it has been at least created. I'm not sure what wording to use there. I can drop the "must create a new leaf" -- but you're the one who suggested it and I do agree with the sentiment somewhat. But maybe we should drop it -- I think consistency is far more important than whether it's a new leaf.

wking: On Fri, Jul 22, 2016 at 02:15:25AM -0700, Aleksa Sarai wrote:
> I can drop the "must create a new leaf" -- but you're the one who
> suggested it and I do agree with the sentiment somewhat. But maybe
> we should drop it -- I think consistency is far more important than
> whether it's a new leaf.

I think we should drop the new-leaf wording.  I'd suggested it before
we had the consistency wording.

cyphar: Sure. Fixed.

config-linux.md
```diff
@@ -154,29 +154,48 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
 
 The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` is expected to be relative to the cgroups mount point.
-If `cgroupsPath` is not specified, implementations can define the default cgroup path.
+`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
+If `cgroupsPath` is:
+* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
+* ... a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
+* ... not specified, the runtime CAN define the default cgroup path.
+Runtimes CAN consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
```
wking: “CAN” → “MAY”.  RFC 2119 does not define CAN.  And you probably want to drop the comma.

config-linux.md
```diff
@@ -154,29 +154,48 @@ In addition to any devices configured with this setting, the runtime MUST also s
...
-The Spec does not support [split hierarchy][cgroup-v2].
+The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
 The cgroups will be created if they don't exist.
 
+You can configure a container's cgroups via the `resources` field of the Linux configuration.
+Do not specify `resources` unless limits have to be updated.
+For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
+
+A runtime MUST at least use the minimum set of cgroup controllers required to fulfill the `resources` settings.
+However, a runtime CAN decide to not attach to any additional cgroup controllers supported by the system.
```
wking: “CAN” → “MAY” again.  And maybe shuffle this line to avoid “MAY decide to not”:

> The runtime MAY attach the container process to additional controllers.

cyphar:
@wking Addressed the rest of your comments.

wking:
19c2a92e → 4ed839e7 looks good to me.

There's still a comma [1] and “supported by the system” [2] that I
think we don't need, but neither of those have semantic impact.

[1]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r71904790
[2]: https://github.com/opencontainers/runtime-spec/pull/493#discussion_r66660766

crosbymichael:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runtime-spec/pull-request/493/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runtime-spec)

hqhq:
LGTM

PR #495, opened 2016-06-09
runtime.md: Fix sub-bullet indentation

grahamwhaley:
The indentation of the sub bullets for the status entries under State were not
indented by the 4 spaces required by markdown. They did not render well in
the pdf, at least with my native build with pandoc v1.13.2. Change them to 4
spaces, which now renders well.

Signed-off-by: Graham Whaley <graham.whaley@linux.intel.com>

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-  * `created` : the container has been created but the user-specified code has not yet been executed
-  * `running` : the container has been created and the user-specified code is running
-  * `stopped` : the container has been created and the user-specified code has been executed but is no longer running
+    * `created` : the container has been created but the user-specified code has not yet been executed
+    * `running` : the container has been created and the user-specified code is running
+    * `stopped` : the container has been created and the user-specified code has been executed but is no longer running
 
   Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.


vbatts:
Oh wait, adding spaces makes it better in the PDF?

grahamwhaley:
It does for me - in the html also. Without the spaces it was much less clear that those were sub-points to the previous point. I'm having trouble finding a definitive markdown reference (anybody?), but from the random pages I have found I think the requirement is for four spaces or a tab character to increase the level of indentation.

vbatts:
Yeah. Definitive documentation on stuff lack that is lacking

wking:
On Thu, Jun 09, 2016 at 04:45:16AM -0700, Graham Whaley wrote:
> I'm having trouble finding a definitive markdown reference (anybody?)…

From [1]:

  List items may consist of multiple paragraphs. Each subsequent
  paragraph in a list item must be indented by either 4 spaces or one
  tab…

I expect that's intended to be read with “block element” instead of
“paragraph”, in which case it's the reference you need.

[1]: http://daringfireball.net/projects/markdown/syntax#list

grahamwhaley:
I did a bit more digging. It is slightly more convoluted...
Looking at the Makefile, pandoc is invoked with the 'from' type as 'markdown_github'. I guess that definitively means the spec is formatted in github flavour markdown (I don't think I've seen that specifically noted anywhere else?). If we go find the github markdown reference:
https://help.github.com/articles/basic-writing-and-formatting-syntax/#lists
then it says:
  "You can create nested lists by indenting lines by two spaces."

Thus, I checked how the sub-list in runtime.md looked when viewed:
 - on github web interface (github rendered)
 - in the generated pdf
 - in the generated html
for both the master and the submitted patched version.

In the master version, it looks fine when rendered on github, but both the pdf and html render the sub-list as a 'non-bold' list at the same level of indent using the same style of bullets. On my screen/system at least it is hard to tell the difference from the top level list.
In the patched branch the sub-list is indented and uses a different bullet style, which is much clearer.

So, although technically we could stick to the two spaces as per the github_markdown spec, the docs seem to render much better with four spaces.

vbatts:
LGTM
that does make a big difference for the pdf and html docs. Thanks!

wking:
Should we file follow-up PRs for the other instances?  For example
[1,2] (and I imagine there are more).

[1]: https://github.com/opencontainers/runtime-spec/blame/83ec31b19342aa45628af9b8c07b156ab1fe235f/README.md#L9
[2]: https://github.com/opencontainers/runtime-spec/blame/83ec31b19342aa45628af9b8c07b156ab1fe235f/README.md#L157

vbatts:
I imagine this PR could add more. And image-spec needs to be double checked.

crosbymichael:
LGTM

wking:
On Fri, Jun 10, 2016 at 10:41:16AM -0700, Michael Crosby wrote:
> Merged #495.

Follow-up PRs it is [1] ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/495#issuecomment-225199683

PR #502, opened 2016-06-18
specs-go/config: Make Linux and Solaris omitempty (again)

wking:
I'd added some omitempties in 5c2193f3 (specs-go/config: Make Linux and Solaris omitempty, 2016-05-06, #431), but it turns out to not have the intended effect unless the field is also a pointer type (even after I shifted the `omitempty` from the `platform` tag to the `json` tag ;).  Before this commit:

    $ ./ocitools generate --template <(echo '{}')
    $ jq . config.json
    {
      "ociVersion": "1.0.0-rc1-dev",
      "platform": {
        "os": "linux",
        "arch": "amd64"
      },
      "process": {
        "user": {
          "uid": 0,
          "gid": 0
        },
        "args": [],
        "cwd": "/"
      },
      "root": {
        "path": "rootfs"
      },
      "hooks": {},
      "linux": {
        "cgroupsPath": ""
      },
      "solaris": {
        "cappedCPU": {},
        "cappedMemory": {}
      }
    }

And after this commit:

    $ ./ocitools generate --template <(echo '{}')
    $ jq . config.json
    {
      "ociVersion": "1.0.0-rc1-dev",
      "platform": {
        "os": "linux",
        "arch": "amd64"
      },
      "process": {
        "user": {
          "uid": 0,
          "gid": 0
        },
        "args": [],
        "cwd": "/"
      },
      "root": {
        "path": "rootfs"
      },
      "hooks": {},
    }

The remaining useless properties are addressed by other in-flight pull requests:

* 5ca74df (config: Make 'process.args' optional, 2016-06-04, #489)
* ad33f9c (config: Explicitly list 'hooks' as optional, 2016-05-06, #427)

So I've left them alone here.

Issue #494, opened 2016-06-08
Clarify `process.terminal` semantics.


cyphar:
Currently we just say the following:

> - **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.

While I understand that the spec should be vague on what a "terminal" is, I'm having trouble understanding what runC should do with this flag (maybe @crosbymichael can help clarify this). From my understanding (in the Linux parlance) it means that you don't create a new `pty` for the container. However, it's not clear if "don't create a new `pty`" means that we should `dup(2)` `/dev/null` over all of the process's stdio file descriptors -- or if it's something more subtle.

I'm tagging this as 1.0 because I feel like it's one of the smaller issues that we probably should clarify a bit (or at least agree on whether or not it even needs clarification).

wking:
On Wed, Jun 08, 2016 at 06:24:07AM -0700, Aleksa Sarai wrote:
> While I understand that the spec should be vague on what a
> "terminal" is, I'm having trouble understanding what runC should do
> with this flag…

I agree that the current wording is unclear [1].  And I see no reason
to be vague about what “terminal” is, unless someone can explain how
this setting is useful for systems that don't have a psuedoterminal
analog.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/hcbwL4T9lGs
     Subject: More detail for process.terminal?
     Date: Thu, 14 Jan 2016 14:44:08 -0800
     Message-ID: <20160114224408.GM6362@odin.tremily.us>

crosbymichael:
While we are talking about this, should this be an option in the spec or just something that is dynamic at runtime if the user passes a --tty and/or --console flag?

wking:
On Fri, Jun 10, 2016 at 10:55:18AM -0700, Michael Crosby wrote:
> While we are talking about this, should this be an option in the
> spec or just something that is dynamic at runtime if the user passes
> a --tty and/or --console flag?

For third-party config authors, that's “does this matter inside the
container?”.  I can't think of a reason why it would, so I'm fine
punting the option from the spec.  I would like to see a command-line
API [1] or some such so runtime-callers had a standardized way to
enable the functionality (which I think we still want to require
runtimes to support so folks can build web frontends for terminal
sessions and such).  And it's probably worth getting feedback on the
migration from anyone who's actually using the feature now (start a
thread on dev@ to raise awareness?  Ping folks in next week's
meeting?).

[...]

PR #504, opened 2016-06-21
[Proof of concept] Add Windows fields

jhowardmsft:
Signed-off-by: John Howard <John.Howard@microsoft.com>

Would like some feedback on this approach. This is the first step in removing the 'frankenstein' version of the spec used by Docker on Windows (https://github.com/docker/docker/blob/master/libcontainerd/windowsoci/oci_windows.go), which was based on a slightly earlier version of the OCI runtime spec (circa March 2016). 

In particular, I'd like to get feedback on the best approach to have common structure names (eg Resources, CPU, Memory, Network etc) which are actually quite different in their contents between Windows and other platforms. I noticed that since I first took a cut of the OCI spec for use in Docker, the Solaris additions have been made, so I followed the same pattern for adding a Windows structure to the main `Spec` structure. 

However, for those common structures, I modified it to be platform specific by prefixing the structure name with the platform. Hence `Resources` is renamed `LinuxResources`, and a new `WindowsResources` structure has been added, as an example. 

For structures which are very similar, I just added Windows specific fields, which is a pattern which already exists in the spec.

Hopefully that's there is concurrence in the above approach :smile:

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -25,6 +25,8 @@ type Spec struct {
 	Linux Linux `json:"linux" platform:"linux,omitempty"`
 	// Solaris is platform specific configuration for Solaris containers.
 	Solaris Solaris `json:"solaris" platform:"solaris,omitempty"`
+	// Windows is the platform specific configuration for Windows based containers.
+	Windows *Windows `json:"windows,omitempty" platform:"windows"`
 }
 
 // Process contains information to start a specific application inside the container.
@@ -51,10 +53,12 @@ type Process struct {
 	ApparmorProfile string `json:"apparmorProfile,omitempty" platform:"linux"`
 	// SelinuxLabel specifies the selinux context that the container process is run as. (this field is platform dependent)
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
+
+	// InitialConsoleSize specifies the initial [h,w] of the console. (this field is platform dependent)
+	InitialConsoleSize [2]int `json:"initialConsoleSize,omitempty" platform:"windows"`
 }
 
-// User specifies Linux/Solaris specific user and group information for the container's
-// main process.
+// User specifies the user information for the container
 type User struct {
 	// UID is the user id. (this field is platform dependent)
 	UID uint32 `json:"uid" platform:"linux,solaris"`
@@ -62,6 +66,8 @@ type User struct {
 	GID uint32 `json:"gid" platform:"linux,solaris"`
 	// AdditionalGids are additional group ids set for the container's process. (this field is platform dependent)
 	AdditionalGids []uint32 `json:"additionalGids,omitempty" platform:"linux,solaris"`
+	//  User is the user name. (this field is platform dependent)
+	User string `json:"user" platform:"windows"`
 }
 
 // Root contains information about the container's root filesystem on the host.
@@ -79,6 +85,8 @@ type Platform struct {
 	OS string `json:"os"`
 	// Arch is the architecture
 	Arch string `json:"arch"`
+	// OSVersion is the version of the operating system. (this field is platform dependent)
+	OSVersion string `json:"os.version,omitempty" platform:"windows"`
 }
 
 // Mount specifies a mount for a container.
@@ -90,8 +98,10 @@ type Mount struct {
 	// Source specifies the source path of the mount.  In the case of bind mounts on
 	// Linux based systems this would be the file on the host.
 	Source string `json:"source"`
-	// Options are fstab style mount options.
-	Options []string `json:"options,omitempty"`
+	// Options are fstab style mount options. (this field is platform dependent)
+	Options []string `json:"options,omitempty" platform:"linux,solaris"`
+	// Readonly specifies if the mount should be read-only. (this field is platform dependent)
+	Readonly bool `json:"readonly" platform:"windows"`
 }
 
 // Hook specifies a command that is run at a particular event in the lifecycle of a container
@@ -123,7 +133,7 @@ type Linux struct {
 	Sysctl map[string]string `json:"sysctl,omitempty"`
 	// Resources contain cgroup information for handling resource constraints
 	// for the container
-	Resources *Resources `json:"resources,omitempty"`
+	Resources *LinuxResources `json:"resources,omitempty"`
 	// CgroupsPath specifies the path to cgroups that are created and/or joined by the container.
 	// The path is expected to be relative to the cgroups mountpoint.
 	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
@@ -252,7 +262,7 @@ type BlockIO struct {
 }
 
 // Memory for Linux cgroup 'memory' resource management
-type Memory struct {
+type LinuxMemory struct {
 	// Memory limit (in bytes).
 	Limit *uint64 `json:"limit,omitempty"`
 	// Memory reservation or soft_limit (in bytes).
@@ -268,7 +278,7 @@ type Memory struct {
 }
 
 // CPU for Linux cgroup 'cpu' resource management
-type CPU struct {
+type LinuxCPU struct {
 	// CPU shares (relative weight (ratio) vs. other cgroups with cpu shares).
 	Shares *uint64 `json:"shares,omitempty"`
 	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
@@ -291,16 +301,16 @@ type Pids struct {
 	Limit *int64 `json:"limit,omitempty"`
 }
 
-// Network identification and priority configuration
-type Network struct {
+// LinuxNetwork identification and priority configuration
+type LinuxNetwork struct {
 	// Set class identifier for container's network packets
 	ClassID *uint32 `json:"classID"`
 	// Set priority of network traffic for container
 	Priorities []InterfacePriority `json:"priorities,omitempty"`
 }
 
-// Resources has container runtime resource constraints
-type Resources struct {
+// LinuxResources has container runtime resource constraints for Linux based containers
+type LinuxResources struct {
 	// Devices are a list of device rules for the whitelist controller
 	Devices []DeviceCgroup `json:"devices"`
 	// DisableOOMKiller disables the OOM killer for out of memory conditions
@@ -308,9 +318,9 @@ type Resources struct {
 	// Specify an oom_score_adj for the container.
 	OOMScoreAdj *int `json:"oomScoreAdj,omitempty"`
 	// Memory restriction configuration
-	Memory *Memory `json:"memory,omitempty"`
+	Memory *LinuxMemory `json:"memory,omitempty"`
 	// CPU resource restriction configuration
-	CPU *CPU `json:"cpu,omitempty"`
+	CPU *LinuxCPU `json:"cpu,omitempty"`
 	// Task resource restriction configuration.
 	Pids *Pids `json:"pids,omitempty"`
 	// BlockIO restriction configuration
@@ -318,7 +328,7 @@ type Resources struct {
 	// Hugetlb limit (in bytes)
 	HugepageLimits []HugepageLimit `json:"hugepageLimits,omitempty"`
 	// Network restriction configuration
-	Network *Network `json:"network,omitempty"`
+	Network *LinuxNetwork `json:"network,omitempty"`
 }
 
 // Device represents the mknod information for a Linux special device file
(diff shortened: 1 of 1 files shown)


specs-go/config.go
```diff
@@ -25,6 +25,8 @@ type Spec struct {
 	Linux Linux `json:"linux" platform:"linux,omitempty"`
 	// Solaris is platform specific configuration for Solaris containers.
 	Solaris Solaris `json:"solaris" platform:"solaris,omitempty"`
+	// Windows is the platform specific configuration for Windows based containers.
+	Windows Windows `json:"windows" platform:"windows,omitempty"`
```
wking: This is matching a broken pattern ;).  See #502 for what I think we should be using instead.

specs-go/config.go
```diff
@@ -79,6 +85,8 @@ type Platform struct {
 	OS string `json:"os"`
 	// Arch is the architecture
 	Arch string `json:"arch"`
+	// OSVersion is the version of the operating system. (this field is platform dependent)
+	OSVersion string `json:"os.version,omitempty" platform:"windows"`
```
wking: How do you expect this field to be used?  The consensus for config settings that aren't supported by all versions of the Linux kernel (all the config settings?) seems to be:

> Just return kernel errors in a reasonable way.  You don't have to check for this sort of thing ahead of time.

See [here][1] and #418.

[1]: https://github.com/opencontainers/runtime-spec/pull/397#issuecomment-223062579

specs-go/config.go
```diff
@@ -90,8 +98,10 @@ type Mount struct {
 	// Source specifies the source path of the mount.  In the case of bind mounts on
 	// Linux based systems this would be the file on the host.
 	Source string `json:"source"`
-	// Options are fstab style mount options.
-	Options []string `json:"options,omitempty"`
+	// Options are fstab style mount options. (this field is platform dependent)
+	Options []string `json:"options,omitempty" platform:"linux,solaris"`
```
wking: This looks like it should be part of #439.

specs-go/config.go
```diff
@@ -252,7 +262,7 @@ type BlockIO struct {
 }
 
 // Memory for Linux cgroup 'memory' resource management
-type Memory struct {
+type LinuxMemory struct {
```
wking: Doing this sort of rename on a per-collision basis (e.g. “we now have a Windows memory structure, so rename `Memory` → `LinuxMemory`”) seems like it will cause stability headaches for downstream consumers of this package.  I suggest [aggressively namespacing now][1] (and when new types are added), because “yeah, Solaris can also use `LinuxInterfacePriority`” is much easier on downstream users than “We're renaming `InterfacePriority` to `LinuxInterfacePriority` in runtime-spec v1.2”.

[1]: https://github.com/opencontainers/runtime-spec/pull/405/files/cc226d717a2eccf19dc81aee1c8a298a1ffe211c#r63821329

specs-go/config.go
```diff
@@ -469,3 +479,77 @@ type Syscall struct {
...
 }
+
+// Windows contains platform specific configuration for Windows based containers.
+type Windows struct {
+	// Resources contains constraints for containers.
+	Resources *WindowsResources `json:"resources,omitempty"`
+	// Networking contains the network connection settings for containers.
+	Networking *Networking `json:"networking,omitempty"`
+	// FirstStart is an optimization on first boot of a container.
+	FirstStart bool `json:"first_start,omitempty"`
```
wking: This property is likely best delegated to a runtime flag.  See previous discussion on flags vs. config settings [here][1] and [here][2], but the threshold seems to be close to “can you tell that was set from inside the container?”.

[1]: https://github.com/opencontainers/runtime-spec/issues/494#issuecomment-225251219
[2]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-223488103

specs-go/config.go
```diff
@@ -469,3 +479,77 @@ type Syscall struct {
...
+	// LayerFolder is the path to the current layer folder.
+	LayerFolder string `json:"layer_folder,omitempty"`
+	// Layer contains the filepaths of the parent layers.
+	LayerPaths []string `json:"layer_paths,omitempty"`
```
wking: These two layer settings seem like they belong to [image-spec][1] or similar.  Are they really a runtime concern?  Perhaps this will become clearer once this PR grows Markdown docs.

[1]: https://github.com/opencontainers/image-spec

specs-go/config.go
```diff
@@ -469,3 +479,77 @@ type Syscall struct {
...
+	// Settings relating to Hyper-V containers, omitted otherwise.
+	HvRuntime *HvRuntime `json:"hv_runtime,omitempty"`
```
wking: This is probably best delegated to `containerd` or some higher layer.  See previous discussion about the very-similar `HypervisorPath` (since dropped from #405) [here][1] and [here][2].

[1]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61262329
[2]: https://github.com/opencontainers/runtime-spec/pull/405#discussion_r61473810

specs-go/config.go
```diff
@@ -469,3 +479,77 @@ type Syscall struct {
...
+// WindowsNetwork has network runtime resource constraings for Windows based containers.
+type WindowsNetwork struct {
+	// Bandwidth is the maximum egress bandwidth in bytes per second.
+	Bandwidth *uint64 `json:"bandwidth,omitempty"`
+}
+
+// WindowsMemory has memory runtime resource constraints for Windows based containers.
+type WindowsMemory struct {
+	// Memory limit (in bytes).
+	Limit *int64 `json:"limit,omitempty"`
```
wking: Is this intentionally signed?  If so, please explain what negative values mean when you land the Markdown docs ;).

specs-go/config.go
```diff
@@ -469,3 +479,77 @@ type Syscall struct {
...
+	// Specifies maximum bytes per second for the system drive.
+	Bps *uint64 `json:"bps,omitempty"`
+	// Sandbox size indicates the size to expand the system drive to if it is currently smaller.
+	SandboxSize *uint64 `json:"sandbox_size,omitempty"`
+}
+
+// Networking contains the network configuration for Windows based containers.
+type Networking struct {
+	// List of endpoints to be attached to the container.
+	EndpointList []string `json:"endpoints,omitempty"`
```
wking: I'd expect `Endpoints` instead of `EndpointList` for the Go parameter behind and `endpoints` JSON parameter.

wking:
On Tue, Jun 21, 2016 at 04:47:42PM -0700, John Howard wrote:
> Would like some feedback on this approach.

I left a few nits as inline comments, but the general approach you
have here looks pretty good to me (and it matches our current sketch
of a policy [1]).  My main concern is about future type-name churn
impacting downstream Go users, and I left suggestions on addressing
that here [2].

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-101
[2]: https://github.com/opencontainers/runtime-spec/pull/504/files#r67990890

specs-go/config.go
jhowardmsft: Seems sensible. Updated to match #502.

specs-go/config.go
jhowardmsft: We use it today in docker/docker for difference in functionality between versions of Windows as opposed to whether things are supported or not supported. Hence it's not a case of us being able to return an error in the way described above.

specs-go/config.go
jhowardmsft: I'm not sure - that PR is documentation. The `Options` are very different between *nix and Windows - Windows obviously has no concept of fstab. I added the `platform` tag to make this field not applicable to Windows, and added a separate field which is for Windows.

wking: On Wed, Jun 22, 2016 at 09:29:38AM -0700, John Howard wrote:
> > -	// Options are fstab style mount options.
> > -	Options []string `json:"options,omitempty"`
> > +	// Options are fstab style mount options. (this field is platform dependent)
> > +	Options []string `json:"options,omitempty" platform:"linux,solaris"`
> 
> I'm not sure - that PR is documentation.

The pattern so far in this repo has been to try and land the
(canonical, #342) Markdown docs in the same PR that updates the Go
types.  I think the change you're making to Mount.Options here makes
sense, and think that #439 has the associated Markdown updates.  Why
not make both of those changes in one atomic pivot PR?

specs-go/config.go
wking: On Wed, Jun 22, 2016 at 09:25:09AM -0700, John Howard wrote:
> > +	// OSVersion is the version of the operating system. (this field is platform dependent)
> > +	OSVersion string `json:"os.version,omitempty" platform:"windows"`
> 
> We use it today in docker/docker for difference in functionality
> between versions of Windows as opposed to whether things are
> supported or not supported.

Grepping turned up [1,2].  I still think that is a runtime-flag
concern (especially if process.terminal is a runtime-flag concern
[3]).  Can't the code that currently sets OSVersion just switch over
to setting a runtime flag?  It doesn't seem like that would be a big
change.

[1]: https://github.com/docker/docker/blob/v1.12.0-rc2/libcontainerd/container_windows.go#L125-L126
[2]: https://github.com/docker/docker/blob/v1.12.0-rc2/libcontainerd/process_windows.go#L34-L51
[3]: https://github.com/opencontainers/runtime-spec/issues/494#issuecomment-225251219

specs-go/config.go
mrunalp: What is this expected to be used for? Can you provide an example?

specs-go/config.go
```diff
@@ -469,3 +479,77 @@ type Syscall struct {
...
+	// Specifies maximum Iops for the system drive.
+	Iops *uint64 `json:"iops,omitempty"`
+	// Specifies maximum bytes per second for the system drive.
+	Bps *uint64 `json:"bps,omitempty"`
+	// Sandbox size indicates the size to expand the system drive to if it is currently smaller.
+	SandboxSize *uint64 `json:"sandbox_size,omitempty"`
+}
+
+// Networking contains the network configuration for Windows based containers.
+type Networking struct {
```
mrunalp: Should probably be renamed once we determine if this is necessary.

specs-go/config.go
mrunalp: Curious why can't we use the existing `OS` field?

specs-go/config.go
```diff
@@ -90,8 +98,10 @@ type Mount struct {
 	// Source specifies the source path of the mount.  In the case of bind mounts on
 	// Linux based systems this would be the file on the host.
 	Source string `json:"source"`
-	// Options are fstab style mount options.
-	Options []string `json:"options,omitempty"`
+	// Options are fstab style mount options. (this field is platform dependent)
+	Options []string `json:"options,omitempty" platform:"linux,solaris"`
+	// Readonly specifies if the mount should be read-only. (this field is platform dependent)
+	Readonly bool `json:"readonly" platform:"windows"`
```
mrunalp: Could you parse a constant `readonly` from existing `options` to set this boolean for Windows?

specs-go/config.go
wking: On Wed, Jun 22, 2016 at 07:55:51PM -0700, Mrunal Patel wrote:
> > +	// OSVersion is the version of the operating system. (this field is platform dependent)
> > +	OSVersion string `json:"os.version,omitempty" platform:"windows"`
> 
> Curious why can't we use the existing `OS` field?

Would you use ‘old-windows’ or some such for “Windows before build
14350” [1]?  I think it's better to keep ‘windows’ there and pass that
build information through as a separate thing.  I just think the
separate thing should be a runtime flag instead of a config option.

[1]: https://github.com/docker/docker/blob/v1.12.0-rc2/libcontainerd/process_windows.go#L34-L51

specs-go/config.go
wking: On Wed, Jun 22, 2016 at 07:57:58PM -0700, Mrunal Patel wrote:
> > -	// Options are fstab style mount options.
> > -	Options []string `json:"options,omitempty"`
> > +	// Options are fstab style mount options. (this field is platform dependent)
> > +	Options []string `json:"options,omitempty" platform:"linux,solaris"`
> > +	// Readonly specifies if the mount should be read-only. (this field is platform dependent)
> > +	Readonly bool `json:"readonly" platform:"windows"`
>
> Could you parse a constant `readonly` from existing `options` to set
> this boolean for Windows?

Or stay compatible with Linux by using ‘ro’ and ‘rw’ [1].  Matching
Linux isn't a big deal, but I'm surprised there's no Windows analog to
[1].  Is “readonly” really the only knob?

[1]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS

jhowardmsft: Indeed. In the process of updating the docker engine to do exactly that at https://github.com/docker/docker/pull/26577.

specs-go/config.go
```diff
@@ -51,17 +53,21 @@ type Process struct {
...
+// User specifies the user information for the container
 type User struct {
 	// UID is the user id. (this field is platform dependent)
 	UID uint32 `json:"uid" platform:"linux,solaris"`
 	// GID is the group id. (this field is platform dependent)
 	GID uint32 `json:"gid" platform:"linux,solaris"`
 	// AdditionalGids are additional group ids set for the container's process. (this field is platform dependent)
 	AdditionalGids []uint32 `json:"additionalGids,omitempty" platform:"linux,solaris"`
+	//  User is the user name. (this field is platform dependent)
+	User string `json:"user" platform:"windows"`
```
anuthan: Any specific reason why it's not username, and can we expect a groupname in the future ?

jhowardmsft: This is a proof-of-concept PR. In the actual PR (now in master), it's username. https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L65. There are no plans for groups in the future (it's not really something that makes sense on Windows).

anuthan: Cool, was just curious. Thanks.

specs-go/config.go
jhowardmsft: Agreed. PR submitted to docker https://github.com/docker/docker/pull/26650

jhowardmsft:
Closing this proof-of-concept PR now there are multiple PRs out to add support for Windows and address other comments in this PR.

PR #508, opened 2016-06-23
runtime: Add an 'event' operation for subscribing to pushes

wking:
The current [state](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#query-state) operation allows callers to poll for the state, but for some workflows polling is inefficient (how frequently do you poll to balance the cost of polling against the timeliness of the response?) and push notifications make more sense.

The runtime's [create](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#create) process is in a unique position to detect these status transitions.
- As the actor carrying out container creation, it should have a clear idea of when that creation completes (for the `created` event).
- It may setup a communication channel with the container process to orchestrate creation, and that channel may be used to report the start event.
- It knows (or is) the parent of the container process, and POSIX's [wait(3)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html), [waitpid(3)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html), and [waitid(3)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html) only work for child processes.  From [wait(3)](http://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html):
  
  > Nothing in this volume of POSIX.1-2008 prevents an
  > implementation from providing extensions that permit a process
  > to get status from a grandchild or any other process, but a
  > process that does not use such extensions must be guaranteed to
  > see status from only its direct children.
  
  So the runtime can setup (or be) the parent waiting on the container process, and arrange for the 'stopped' event to be published on container exit.

I've tried to phrase the requirements conservatively to allow for runtimes that have to poll their kernel or some such to notice these changes.  I see the following runtime-support cases:
1. The runtime can easily supply a push-based event operation.  In this case, exposing that operation to callers faciliates push-based workflows without much cost.
2. The runtime cannot supply a push-based event operation, and has to emulate it by polling.  In this case, the runtime can pick a polling strategy that makes sense to its maintainers, and callers who aren't satisfied with that strategy can roll their own state poller without a big efficiency hit (in a lesser of two evils way).
   
   The requirement is currently worded so weakly that a runtime would be compliant with:
   1. Container process dies at noon.
   2. User calls [state](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#query-state) at 5pm.
   3. Runtime checks kernel, and sees that the container process is dead.
   4. Runtime publishes `stopped` event with a 5pm (and some microseconds) timestamp.
   5. Runtime returns state to the user with `stopped` in `status`.
   
   which is a pretty low bar.
3. The runtime could supply a push-based event operation, but it would be a lot of work.  These runtimes can use polling (2) as a quick-and-dirty solution until someone has time to implement a push-based solution (1).  The improvement is transparent to users, who can use the same event operation throughout and passively reap the benefits of the implementation improvements.

Without an operation like this, higher levels that need to trigger on these transitions without polling need to exercise a lot of control over the system:
- They must be the parent of (or on Linux, the nearest [`PR_SET_CHILD_SUBREAPER`](http://man7.org/linux/man-pages/man2/prctl.2.html) ancestor of) the [create](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#create) process, and the [create](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#create) process needs to exit after creation completes, if they want to block on the `created` event.
- They must proxy all [start](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#start) and [delete](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#delete) requests if they want to block on the `started` or `deleted` events.
- On Linux, they must be the nearest ancestor of the [create](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#create) operation to set [`PR_SET_CHILD_SUBREAPER`](http://man7.org/linux/man-pages/man2/prctl.2.html) if they want to block on the `stopped` event.

By making `event` a runtime requirement, we allow for efficient cross-platform push-based workflows while avoiding the need for tight orchestrator gate-keeping.  Runtimes that have implementation difficulties have an easy out that allows their callers to benefit from future implemenation improvements.  And callers that are not satisfied can always fall back to polling state or the proxy/`waitid`/`PR_SET_CHILD_SUBREAPER` approaches.

The buffering requirement is very generic, and I expect more details to land in the runtime API specification.  Requirements around the buffer-request semantics (e.g. “buffer for 15 seconds” or “buffer the last 5 events” or “buffer the `created` event”) seemed out of place at the level of detail in this specification.

The goal is to allow for:

```
$ funC create --event-buffer created ID &
$ funC event --event created ID && hook1 && hook2 && funC start ID
$ fg
```

To support blocking on the event without racing on “maybe the `created` event happened before the event operation attached”.

Given the small number of required events, this buffering should not be a large resource concern, and `--event-buffer created` would only ever require a single event to be buffered per container.

There is still a race on “maybe the container has already been destroyed and a second container has been created with the same ID before the event operation attached”, but that seems much less likely (especially since the caller is free to pick UUIDs).

This builds on #507, so that should be reviewed first.

**UPDATE**: note that this pub/sub model _does not_ require an additional long-running process.  For an example implementation based on shared access to a directory (like runC uses for [state](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime.md#query-state)) and inotify, see [wking/inotify-pubsub](https://github.com/wking/inotify-pubsub).

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -15,9 +15,10 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-  * `created` : the container process has neither exited nor executed the user-specified code
-  * `running` : the container process has executed the user-specified code but has not exited
-  * `stopped` : the container process has exited
+  * `creating` : the container is being created (step 2 in the [lifecycle](#lifecycle))
+  * `created` : the runtime has finished the [create operation](#create) (after step 2 in the [lifecycle](#lifecycle)), and the container process has neither exited nor executed the user-specified code
+  * `running` : the container process has executed the user-specified code but has not exited (after step 4 in the [lifecycle](#lifecycle))
+  * `stopped` : the container process has exited (step 5 in the [lifecycle](#lifecycle))
 
   Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
--- a/runtime.md
+++ b/runtime.md
@@ -89,6 +89,8 @@ This operation MUST return the state of a container as specified in the [State](
 This operation MUST generate an error if it is not provided the ID of a container.
 Attempting to query a container that does not exist MUST generate an error.
 This operation MUST subscribe the caller to push-notification about future events in chronological order.
+If the container's [create operation](#create) requested an event buffer, the buffered events MUST be published in chronological order before any future events are published.
+If the runtime could not perform the requested buffering, it MUST generate an error.
 Events MUST be published when the container's [`status`](#state) changes, and MAY be published for additional events.
 Events MUST include, at least, the following properties:
 
@@ -111,7 +113,7 @@ So the `timestamp` and event publication may not exactly match the associated ke
 
 ### Create
 
-`create <container-id> <path-to-bundle>`
+`create <container-id> <path-to-bundle> <event-buffer>`
 
 This operation MUST generate an error if it is not provided a path to the bundle and the container ID to associate with the container.
 If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST generate an error and a new container MUST not be created.
@@ -124,6 +126,8 @@ The runtime MAY validate `config.json` against this spec, either generically or 
 Runtime callers who are interested in pre-create validation can run [bundle-validation tools](implementations.md#testing--tools) before invoking the create operation.
 
 Any changes made to the [`config.json`](config.md) file after this operation will not have an effect on the container.
+
+Runtime callers MAY request an event buffer, in which case the runtime MUST buffer events associated with the container.
 
 ### Start
 `start <container-id>`
--- a/runtime.md
+++ b/runtime.md
@@ -7,13 +7,13 @@ Whether other entities using the same, or other, instance of the runtime can see
 
 ## State
 
-The state of a container MUST include, at least, the following properties:
+The state of a container includes the following properties:
 
-* **`ociVersion`**: (string) is the OCI specification version used when creating the container.
-* **`id`**: (string) is the container's ID.
+* **`ociVersion`** (string, required) is the OCI specification version used when creating the container.
+* **`id`** (string, required) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
-* **`status`**: (string) is the runtime state of the container.
+* **`status`** (string, required) is the runtime state of the container.
 The value MAY be one of:
   * `creating` : the container is being created (step 2 in the [lifecycle](#lifecycle))
   * `created` : the runtime has finished the [create operation](#create) (after step 2 in the [lifecycle](#lifecycle)), and the container process has neither exited nor executed the user-specified code
@@ -21,11 +21,13 @@ The value MAY be one of:
   * `stopped` : the container process has exited (step 5 in the [lifecycle](#lifecycle))
 
   Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
-* **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
-* **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
+* **`pid`** (int, required when `status` is `created` or `running`) is the ID of the main process within the container, as seen by the host.
+* **`bundlePath`**: (string, required) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
-* **`annotations`**: (map) contains the list of annotations associated with the container.
+* **`annotations`**: (map, required) contains the list of annotations associated with the container.
 If no annotations were provided then this property MAY either be absent or an empty map.
+
+The state MAY include additional properties.
 
 When serialized in JSON, the format MUST adhere to the following pattern:
 
--- a/runtime.md
+++ b/runtime.md
@@ -15,9 +15,9 @@ This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
 * **`status`**: (string) is the runtime state of the container.
 The value MAY be one of:
-  * `created` : the container has been created but the user-specified code has not yet been executed
-  * `running` : the container has been created and the user-specified code is running
-  * `stopped` : the container has been created and the user-specified code has been executed but is no longer running
+  * `created` : the container process has neither exited nor executed the user-specified code
+  * `running` : the container process has executed the user-specified code but has not exited
+  * `stopped` : the container process has exited
 
   Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the main process within the container, as seen by the host.
@@ -55,8 +55,8 @@ The lifecycle describes the timeline of events that happen from when a container
    However, some actions might only be available based on the current state of the container (e.g. only available while it is started).
 4. Runtime's `start` command is invoked with the unique identifier of the container.
    The runtime MUST run the user-specified code, as specified by [`process`](config.md#process-configuration).
-5. The container's process is stopped.
-   This MAY happen due to them erroring out, exiting, crashing or the runtime's `kill` operation being invoked.
+5. The container process exits.
+   This MAY happen due to erroring out, exiting, crashing or the runtime's `kill` operation being invoked.
 6. Runtime's `delete` command is invoked with the unique identifier of the container.
    The container MUST be destroyed by undoing the steps performed during create phase (step 2).
 
--- a/runtime.md
+++ b/runtime.md
@@ -82,6 +82,33 @@ This operation MUST generate an error if it is not provided the ID of a containe
 Attempting to query a container that does not exist MUST generate an error.
 This operation MUST return the state of a container as specified in the [State](#state) section.
 
+### Event
+
+`event <container-id>`
+
+This operation MUST generate an error if it is not provided the ID of a container.
+Attempting to query a container that does not exist MUST generate an error.
+This operation MUST subscribe the caller to push-notification about future events in chronological order.
+Events MUST be published when the container's [`status`](#state) changes, and MAY be published for additional events.
+Events MUST include, at least, the following properties:
+
+* **`type`** (string, required) The type of the event.
+  For the following [`status`](#state) transitions, the `type` values MUST be:
+
+    * `creating` → `created`: `created`
+    * `created` → `running`: `started`
+    * * → `stopped`: `stopped`
+
+  A `deleted` event MUST be published after a successful [delete operation](#delete), after which further events MUST NOT be generated.
+
+* **`id`** (string, required) The ID of the container which experienced the event.
+* **`timestamp`** (string, required) The time at which the event took place in [`date-time` format as specified by RFC 3339][rfc3339-s5.6].
+
+The `started` and `stopped` transitions happen in the kernel, and there may be a lag before the runtime notices.
+For example, if the container process dies, a runtime which is [waiting][waitid.3p] on it will take some time in the `SIGCHLD` handler before adjusting the [state](#state).
+A runtime that detects transitions by polling the kernel may trail the associated kernel transition by an even longer period.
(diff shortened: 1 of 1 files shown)


wking:
Some discussion with @mikebrow around this starting [here][1].

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/latest.log.html#t2016-06-23T20:09:09

cyphar:
I don't like this. For runC, it would mean that we'd have to make a long-running daemon to implement this functionality. IMO that defeats the whole point of runC -- it's a single binary you can use without needing any setup. If we make it so that runC doesn't exit after starting a container, then I'm still iffy about it.

wking:
On Mon, Jul 11, 2016 at 07:12:09PM -0700, Aleksa Sarai wrote:
> For runC, it would mean that we'd have to make a long-running daemon
> to implement this functionality.

Why is that?  In ccon, where I'm happy with “the start-socket exists”
as the “created” event, I'll implement this operation with a second
long-running process.  So:

  event-handling-process
  `-- ccon host process
      `-- ccon container process

where each parent outlives its child.  But runtimes that value process
efficiency but still want a server side for their ‘event’ clients
could do:

  event-handling host process
  `-- container process

If you don't want extra server-side stuff for runC, you could put
events in a subdirectory of your state directory and have ‘event’
clients use inotify to track additions to that directory.

And if you want to invest the minimum dev time, you can always fall
back to polling ‘state’ calls (approach 3 in my runtime-support cases
[1]).

> If we make it so that runC doesn't exit after starting a container,
> then I'm still iffy about it.

I agree that a system without this ‘event’ operation is lighter
weight.  But I think supporting ‘event’ is a lesser evil than the
“Without an operation like this…” conditions I list in [1].  Do you
have alternative, portable solutions to those problems?

[1]: https://github.com/opencontainers/runtime-spec/pull/508#issue-162046810

mikebrow:
I'm good with the idea of an event model.. but agree we should keep runc lower in the stack.  RunC is not an orchestrator its a low level component.  Just thinking out loud.. can't we map the event notifications to Mrunal's hooks and store event registratoins as annotations?  IOW maybe the spec should avoid describing how the implementation of the event model should be provided.

wking:
On Tue, Jul 12, 2016 at 07:05:15AM -0700, Mike Brown wrote:
> … can't we map the event notifications to Mrunal's hooks and store
> event registrations as annotations?

Can you unpack “event registrations”?  Do you just mean “the event
JSON”, e.g.:

  "annotations": {
    "opencontainers.org/events": "[{\"type\": \"created\", \"id\": \"oci-container1\", \"timestamp\": \"2016-07-12T09:33:10-0700\"}, …]"
  }

I like the state subdir approach better [1], since you don't have to
worry about parallel writes to a single state file, but runtimes
should be free to implement the operation however they like.

> IOW maybe the spec should avoid describing how the implementation of
> the event model should be provided.

I'm not sure where you're going with “should be provided”.  I think
the implementation should be up to the runtime, and this PR doesn't go
into detail about the exposed API (since none of the operations in
runtime.md do).

But if this PR lands, I think the API spec (#511 / #513) should
specify a required command-line API along the lines of the funC
example in [2].  Otherwise the compliance test suite and other generic
runtime-callers won't know how to invoke the operation.

[1]: https://github.com/opencontainers/runtime-spec/pull/508#issuecomment-231941035
[2]: https://github.com/opencontainers/runtime-spec/pull/508#issue-162046810

mikebrow:
/s/event registrations/event subscriptions/

crosbymichael:
+1 to what @cyphar said on why we shouldn't do this

wking:
On Tue, Jul 12, 2016 at 01:02:21PM -0700, Michael Crosby wrote:
> +1 to what @cyphar said on why we shouldn't do this

Responding to @cyphar, I pushed back on this requiring a long-running
daemon, waving my hands in a number of directions and mentioning
inotify briefly [1] ;).  To make that daemonless inotify reference a
bit more concrete, I've mocked it up in shell [2].  Following
@mikebrow's advice [3], I've added an UPDATE to my topic post to link
to that project too.

Does that address your concerns about long-running processes?  If not,
can you be specific about why?

[1]: https://github.com/opencontainers/runtime-spec/pull/508#issuecomment-231919042
[2]: https://github.com/wking/inotify-pubsub
[3]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-07-12.log.html#t2016-07-12T19:49:45

wking:
On Sat, Jul 16, 2016 at 04:03:17PM -0700, W. Trevor King wrote:
> To make that daemonless inotify reference a bit more concrete, I've
> mocked it up in shell [2].
> …
> [2]: https://github.com/wking/inotify-pubsub

And to make the OCI integration more concrete too, I have a ccon
branch where ccon-oci implements an event command along these lines
[1].  You can do (with process.terminal true):

  $ echo 'echo hi; exit' | ccon-oci create --id container-1 >output &
  $ ccon-oci event -He created container-1 && ccon-oci start container-1
  {"type": "created", "timestamp": "2016-07-23T04:16:24.503950Z", "id": "container-1"}
  $ cat /tmp/output 
  hi

If you use a second terminal for the event/start call, you can skip
the standard stream redirection:

  term1 $ ccon-oci create --id container-1
  …blocks after creation…

  term2 $ ccon-oci event -He created container-1 && ccon-oci start container-1
  {"timestamp": "2016-07-23T04:21:20.675097Z", "id": "container-1", "type": "created"}

  …create unblocks on term1…
  /root # echo hi
  hi
  /root # exit 17
  term1 $ echo $?
  17

That seems pretty clean to me, and doesn't require much implementation
magic to distribute events or collect the container-process's exit
code.

[1]: https://github.com/wking/ccon/tree/ccon-oci

wking:
Today I stumbled across some higher-level work discussion polling
vs. subscription: kubernetes/kubernetes#8756,
kubernetes/kubernetes#16831.

mrunalp:
This should be post-1.0

hqhq:
I agree, I'm tagging it to v1.1.0.

crosbymichael:
REJECT

Reading over this again I'm putting in my formal reject for this.  It is totally about implementation and if something wants to do this it can but it has no place specifying this because it places too many implantation constrains on the runtimes and is basically telling runtime devs that they need daemons or other fsnotify logic to build an events system.  

Keep the spec small and useful and build higher level things on top or else we will just have a bloated spec that it unusable.

wking:
> [It] is basically telling runtime devs that they need daemons or other fsnotify logic to build an events system.

You can also implement this by polling state (approach 3 in my initial post).  And the runtime portion doesn't have to use inotify (it just writes events to a directory).  The inotify watcher can be a [completely separate project][1].

> Keep the spec small and useful and build higher level things on top or else we will just have a bloated spec that it unusable.

But as the spec stands there is *no portable way to collect the container's exit code*.  We need something like this PR as a base for that; you can't build it into a higher level.  Of course, it could be part of a parallel spec, in which case folks who wanted a portable way to collect the container's exit code would look for both runtime-spec compliance and event-operation compliance, but I think we should have a clearer reason for splitting those apart when the bar is as low as “polling state” or “writing events to a directory on a platform that supports inotify”.

[1]: https://github.com/wking/inotify-pubsub

cyphar:
@wking Can you explain how this race can be handled with nothing but a directory full of files?

> To support blocking on the event without racing on “maybe the created event happened before the event operation attached”.

As far as I can tell, there's no nice way of removing events from the head of the event queue without having an actual queue structure (which the VFS is not). And reading/writing to a single file is not a good idea. The only solution I could come up with is basically having a single-file database, which is **way** overkill...

wking:
On Thu, Feb 02, 2017 at 07:15:59PM -0800, Aleksa Sarai wrote:
> @wking Can you explain how this race can be handled with nothing but
> a directory full of files?
>
> > To support blocking on the event without racing on “maybe the
> > created event happened before the event operation attached”.

Setup your inotify watcher and then read out the contents of the
directory (shell implementation here [1]).

[1]: https://github.com/wking/inotify-pubsub/blob/83fd18ddd88979e43f256182ffe70314cde9180b/inotify-sub#L104-L128

mikebrow:
From my post back in July.. "Just thinking out loud.. can't we map the event notifications to Mrunal's hooks and store event subscriptions as annotations?"  

@wking given Michael's strong Reject, what do you think of turning this idea into a code sample for using the hooks to show how runc can be extended, tested, monitored...?

mikebrow:
@wking does that mean you don't want to make this a code sample? The inability to collect exit codes as referenced by your #305 should be addressed. @mrunalp are we missing a hook here? If we're going to have hooks to cover the lifecycle of containers that would be an obvious one one to have. But that's separate from should we specify these events in the runtime spec.

vbatts:
folks are generally in agreement that this is not needed in at the runtime layer, but would be a fine feature to happen higher in the stack

wking:
On Wed, Mar 08, 2017 at 02:38:51PM -0800, Vincent Batts wrote:
> folks are generally in agreement that this is not needed in at the
> runtime layer, but would be a fine feature to happen higher in the
> stack

The problem is that, short of polling state, there's no way to add
this higher up the stack.  For more on why, see “The runtime's create
process is in a unique position to detect these status transitions” in
[1].

If you have some insider knowledge about the runtime implementation
(e.g. you know the container process will be an ancestor of the
runtime ‘create’ process and you are on Linux) the ‘create’ caller can
reach around the runtime by setting PR_SET_CHILD_SUBREAPER or similar
(see “Without an operation like this, higher levels that need to
trigger on these transitions without polling need to exercise a lot of
control over the system” in [1]).

And again, as I pointed out in the “I see the following
runtime-support cases” section of [1], runtimes can always supply this
operation by polling state themselves if that's really the best they
can do.  And polling state clearly needs no daemons or fsnotify logic
which @crosbymichael seems strongly against [2].

So I'm really not seeing where this is such an issue, and I think it's
unfortunate that the OCI has no plans to provide a portable way to
collect the container's exit code or block on container process exit.
In the meantime, I expect runtime-tools to continue avoiding the
create/start split [3] opencontainers/runtime-tools#305.  But I guess
we'll see if/when we actually start compliance-testing runtimes…

[1]: https://github.com/opencontainers/runtime-spec/pull/508#issue-162046810
[2]: https://github.com/opencontainers/runtime-spec/pull/508#issuecomment-277098707
[3]: https://github.com/opencontainers/runtime-tools/blob/8392e59c7c11b146462e20701c912242f83cf110/cmd/oci-runtime-tool/runtime_validate.go#L74

wking:
> If you have some insider knowledge about the runtime implementation (e.g. you know the container process will be an ancestor of the runtime `create` process and you are on Linux) the `create` caller can reach around the runtime by setting `PR_SET_CHILD_SUBREAPER` or similar …

And this is what containerd does, with [subreaper set here][1], the `create` wrapper adding a [`SIGCHLD`-based][2] [reaper][3], and the reaping [triggering][4] an [exit event][5].  There are currently plans to allow public access to these events (containerd/containerd#785, containerd/containerd#956), but the implementation is based on the assumption that the container process is an ancestor of the `create` call, and that the container process has no closer subreaper ancestor.  Neither of those assumptions are backed by the spec, which means containerd needs a runtime that is similar enough to runC to work, and cannot be safely used with a generic runtime-spec-compliant Linux runtime.

[1]: https://github.com/containerd/containerd/blob/f2963f75626abe1377c18425fc7cbfe4cc2d59b6/cmd/containerd-shim/shim_linux.go#L22-L23
[2]: https://github.com/containerd/containerd/blob/f2963f75626abe1377c18425fc7cbfe4cc2d59b6/cmd/containerd-shim/main_unix.go#L108-L109
[3]: https://github.com/containerd/containerd/blob/f2963f75626abe1377c18425fc7cbfe4cc2d59b6/linux/shim/service.go#L64-L67
[4]: https://github.com/containerd/containerd/blob/f2963f75626abe1377c18425fc7cbfe4cc2d59b6/linux/shim/service.go#L73
[5]: https://github.com/containerd/containerd/blob/f2963f75626abe1377c18425fc7cbfe4cc2d59b6/linux/shim/service.go#L340-L346

PR #510, opened 2016-06-26
Add text about extensions

duglin:
Mimic https://github.com/opencontainers/image-spec/pull/164
and they should be kept in-sync

### DIFF
--- a/config.md
+++ b/config.md
@@ -324,14 +324,23 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 **`annotations`** (object, optional) contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be a key-value map where both the key and value MUST be strings.
+While the value MUST be present, it MAY be an empty string.
+Keys MUST be unique within this map, and best practice is to namespace the keys.
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
+If there are no annotations then this property MAY either be absent or an empty map.
+Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown annotation key.
 
 ```json
 "annotations": {
-    "key1" : "value1",
-    "key2" : "value2"
+    "com.example.gpu-cores" : "2"
 }
 ```
+
+## Extensibility
+Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unkown property.
+Instead they MUST ignore unknown properties.
 
 ## Configuration Schema Example
 
@@ -683,8 +692,8 @@ Here is a full example `config.json` for reference.
         "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
     },
     "annotations": {
-        "key1": "value1",
-        "key2": "value2"
+        "com.example.key1": "value1",
+        "com.example.key2": "value2"
     }
 }
 ```


config.md
```diff
@@ -317,6 +320,10 @@ Annotations are key-value maps.
 }
 ```
 
+## Extensibility
+The `annotations` property MAY be used as an extensibility point to include additional information that is not defined as part of this specification.
+Excluding the `annotations` property, implementations that are reading/processing this configuration file MUST generate an error if they encounter an unkown property.
```
wking: This seems unnecessarily strict.  The `ociVersion` field uniquely (ish, #371) declares semantics for fields defined by this spec.  Maybe the additional properties are optional features (folks have talked about [having both Solaris and Linux sections in the same config][1], although that particular use-case doesn't make much sense to me).  In any case, I think this should be relaxed to SHOULD, and would recommend runtimes provide an `--unknown-properties=(ignore|warn|error)` flag or some such to allow callers to adjust the behavior.

More generally, I think JSON-LD is a [good][2] [way][3] to assign semantics to all properties in the config, whether they are stock OCI properties or extensions.  That would give callers a better chance of figuring out whether a given extension was ignorable or not by making it easier to track down its semantics.

[1]: https://github.com/opencontainers/runtime-spec/pull/411#discussion_r61623026
[2]: https://github.com/opencontainers/runtime-spec/pull/371#issuecomment-209684002
[3]: https://github.com/opencontainers/image-spec/pull/111#discussion_r65619280

duglin: I'd prefer we either ignore or fault unknown properties - no reason to open up interoperability issues with SHOULDs/MAYs/etc...   I picked the fault path because that's where more people seemed to be headed during the OCI F2F last week.  So this issue (and https://github.com/opencontainers/image-spec/pull/164 ) were designed to have the discussion - once we pick one we just need to make the same decision in both spots.

wking: On Sun, Jun 26, 2016 at 05:37:21PM -0700, Doug Davis wrote:
> > +## Extensibility
> > +The `annotations` property MAY be used as an extensibility point to include additional information that is not defined as part of this specification.
> > +Excluding the `annotations` property, implementations that are reading/processing this configuration file MUST generate an error if they encounter an unkown property.
>
> I'd prefer we either ignore or fault unknown properties - no reason
> to open up interoperability issues with SHOULDs/MAYs/etc...

It's hard to know whether a property should be optional or not without
details of the property itself.  For example, runC feels like the
version check is unnecessary [1], and ccon feels like the platform
check is unnecessary [2].  In the absence of detail about the property
in question, the conservative course is to error out (so I'm in favor
of SHOULDing that approach).  But “the runtime doesn't know what that
property is for” doesn't mean “the caller doesn't know what that
property is for”, and I'd rather allow users to be able to say “don't
worry about unrecognized properties, they're not important for this
container” (so I'm not in favor of MUSTing that approach).  And a
SHOULD would avoid name migration issues as new features are
standardized [3].

On the other hand, I'm generally in favor of for moving non-standard
things out of the config (e.g. hooks.json, if we drop hooks [4]), and
a MUST here would certainly help with that ;).

[1]: https://github.com/opencontainers/runc/pull/521
[2]: https://github.com/wking/ccon/tree/v0.4.0#configuration
[3]: https://tools.ietf.org/html/rfc6648#appendix-B
[4]: https://groups.google.com/a/opencontainers.org/d/msg/dev/Y7p6YW8zr4s/NtiVw9p2AwAJ
     Subject: Re: Hooks and all-in-one operation
     Date: Thu, 9 Jun 2016 14:50:49 -0700
     Message-ID: <20160609215049.GK8801@odin.tremily.us>

config.md
```diff
@@ -308,7 +308,10 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 This OPTIONAL property contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be key-value maps.
+Keys MUST be unique, and best practice is to namespace the keys.
```
mrunalp: Should we give an example and mention that "opencontainers.org" namespace is reserved for use in the spec?

duglin: I'm ok with that, I'm just not creative enough to think of a good example :-)   Suggestions welcome.

mrunalp: How about "acme.com/gpu-cores" ? :)

wking: On Tue, Jun 28, 2016 at 02:16:22PM -0700, Mrunal Patel wrote:
> > +Keys MUST be unique, and best practice is to namespace the keys.
> 
> How about "acme.com/gpu-cores" ? :)

example.com/gpu-cores?  acme.com is a real thing:

  $ whois acme.com | grep -A1 'Registrant Name'
  Registrant Name: JEFFREY POSKANZER
  Registrant Organization: ACME LABORATORIES

mrunalp: Yeah, it would be better to use `example.com/gpu-cores`

duglin: Modified the example to use example.com/gpu-cores

config.md
```diff
@@ -308,15 +308,21 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 This OPTIONAL property contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be key-value maps.
```
vishh: nit: Specify explicitly that the key and value are of type `string` and that the specification is not expecting any structure like DNS naming, etc..

duglin: @vishh what do you mean by not expecting any structure? People can put json in the value string if they want right?

duglin:
Updated based on today's conf call

config.md
```diff
@@ -308,15 +308,23 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 This OPTIONAL property contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be key-value maps where both the key and value are strings.
+Keys MUST be unique, and best practice is to namespace the keys.
+Keys starting with the `opencontainers.org` namespace are reserved and MUST NOT be used.
+If there are no annotations then this property MAY either be absent or an empty map.
+Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown annotation key.
```
wking: “MUST NOT generate an error” doesn't say how they *should* handle unknown entries.  I think we should [follow HTML's][1]:

> The [`HTMLUnknownElement`][2] interface must be used for HTML elements that are not defined by this specification (or other applicable specifications).

In our case, maybe wording like:

> Implementations consuming this configuration MAY log unrecognized annotation keys but MUST otherwise ignore them.

[1]: https://www.w3.org/TR/html5/dom.html#elements-in-the-dom
[2]: https://www.w3.org/TR/html5/dom.html#htmlunknownelement

wking: And while I abbreviated the “Implementations that are reading/processing this configuration” wording in my suggestion above, I think we should probably just go with “[Runtimes][1]”.  Do you have other consumers in mind with your more general wording?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/glossary.md#runtime

config.md
```diff
@@ -308,15 +308,23 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
...
-    "key1" : "value1",
-    "key2" : "value2"
+    "example.com/gpu-cores" : "2"
 }
 ```
 
+## Extensibility
+The `annotations` property MAY be used as an extensibility point to include additional information that is not defined as part of this specification.
+Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unkown property.
+Instead they MUST ignore unknown properties.
```
wking: I think we can combine these last two sentences along the same lines as my [annotation suggestion][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/510/files#r69845853

config.md
```diff
@@ -308,15 +308,23 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
...
 ```json
 "annotations": {
-    "key1" : "value1",
-    "key2" : "value2"
+    "example.com/gpu-cores" : "2"
 }
 ```
 
+## Extensibility
+The `annotations` property MAY be used as an extensibility point to include additional information that is not defined as part of this specification.
```
wking: With the same rules (runtimes ignore unknown keys) for both annotation keys and properties in general, I think we need more than this to explain what annotations is for.  I'd recommend shifting the normative stuff below to the top of this section, and then turning this line into a paragraph talking about how this spec will not clobber annotations outside of the reserved `opencontainers.org` key, while this spec MAY clobber any property outside of annotations.

config.md
```diff
@@ -308,15 +308,23 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 This OPTIONAL property contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be key-value maps where both the key and value are strings.
```
vishh: are `keys` optional? If yes, mention that explicitly.

wking: On Thu, Jul 07, 2016 at 11:05:09AM -0700, Vish Kannan wrote:
> > @@ -308,15 +308,23 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
> >  
> >  This OPTIONAL property contains arbitrary metadata for the container.
> >  This information MAY be structured or unstructured.
> > -Annotations are key-value maps.
> > +Annotations MUST be key-value maps where both the key and value are strings.
> 
> are `values` optional? If yes, mention that explicitly. 

Probably best to be explicit either way.  The Go type uses
map[string]string [1], not map[string]*string, so I expect the
intention was to require a string value and not support null values
(although I'm personally a fan of not typing the content at all [2],
so I may not be the best judge of the typing intentions).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/specs-go/config.go#L21-L22
[2]: https://github.com/opencontainers/runtime-spec/pull/331#issuecomment-192477931

duglin: updated

config.md
duglin: I don't think we can mandate they ignore them because that's how people may turn on/off extensions.

config.md
```diff
@@ -308,15 +308,26 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 This OPTIONAL property contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be key-value maps where both the key and value MUST be strings.
+While the key and value MUST be present, they MAY be empty strings.
```
wking: Empty-string values are fine, but empty-string keys sounds like a bad idea to me.

config.md
```diff
@@ -308,15 +308,26 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
...
+Keys MUST be unique, and best practice is to namespace the keys.
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myK
+ey`.
```
wking: Accidental newline?

config.md
```diff
@@ -308,15 +308,26 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
...
+Keys starting with the `org.opencontainers` namespace are reserved and MUST NOT be used.
```
wking: What if someone registers my-opencontainers.org?  I expect we want to claim `org.opencontainers` (which we'd never use, but just to be clear), and anything that starts with `org.opencontainers.`.

config.md
```diff
@@ -308,15 +308,26 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
...
+If there are no annotations then this property MAY either be absent or an empty map.
+Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown annotation key.
 
 ```json
 "annotations": {
-    "key1" : "value1",
-    "key2" : "value2"
+    "example.com/gpu-cores" : "2"
```
wking: This needs to get reversed too, and I guess it would be `com.example.gpu-cores`.  The slash approach doesn't seem as appropriate with the reversed domain names.

config.md
duglin: fixed

config.md
duglin: fixed

config.md
duglin: fixed

config.md
duglin: fixed

duglin:
FYI the PR was written to use reverse-DNS as was agreed in the last f2f meeting

vbatts:
Thanks @duglin.
Do y'all think we ought to have a sentence regarding _preference_ of key notation (reverse vs. domain-style)? But not a strict requirement.

duglin:
if people are ok with it I can add a sentence about how they SHOULD use reverse-DNS notation.

vbatts:
@duglin I know @RobDolinMS is wishing to avoid SHOULDs as they make for wider interpretation. I'm not sure how best to write it as an encouragement but not an invalidating factor

wking:
On Tue, Aug 30, 2016 at 10:21:09AM -0700, Vincent Batts wrote:
> @duglin I know @RobDolinMS is wishing to avoid SHOULDs as they make
> for wider interpretation. I'm not sure how best to write it as an
> encouragement but not an invalidating factor

Ignoring a SHOULD doesn't make your implementation invalid, it just
makes it complaint (where following the SHOULD can make you
unconditionally compliant, although see #543).

However, this key-namespacing suggestion is targetted at external spec
authors (folks documenting annotations for a platform not managed by
runtime-/image-spec).  That's not an audience we currently talk about
much [1], so I'm fine avoiding a SHOULD to avoid confusing our
existing formal audiences (runtime developers and configuration
authors).

[1]: https://github.com/opencontainers/runtime-spec/tree/v1.0.0-rc1#use-cases

duglin:
hmm, I'm not sure why a SHOULD would be bad since its pretty much what we want - an encouragement (we can use RECOMMEND if we want too) - w/o making it a hard requirement. @RobDolinMS what's the issue with SHOULDs?

wking:
On Tue, Aug 30, 2016 at 05:38:07PM -0700, Doug Davis wrote:
> hmm, I'm not sure why a SHOULD would be bad…

To make [1] clearer, the line you recently removed from the image-spec
PR was:

  Keys SHOULD be named using a reverse domain notation -
  e.g. `com.example.myKey`.

Assuming you add a line like that here, how would you test compliance
with the SHOULD in ocitools?  You'd need a new command to read an
external spec, check for the annotation keys it defines, determine
which DNS names were controlled by the spec authors, and ensure the
annotation keys started with reverse-domain versions of those names.

Ocitools currently tests configurations and runtime implementations
for spec compliance.  I think that's enough for formal testing, and we
can address external spec authors with non-RFC-2119 language.

[1]: https://github.com/opencontainers/runtime-spec/pull/510#issuecomment-243516406

duglin:
> Assuming you add a line like that here, how would you test compliance with the SHOULD in ocitools?

two options:
1 - we don't check it since its just a SHOULD
2 - we issue an "informational" message as a warning about it, but not fail - unless the compliance tool wants to have a "be really strict" mode that treats all SHOULDs as MUSTs - which I kind of like as an *option*.

wking:
On Tue, Aug 30, 2016 at 08:59:26PM -0700, Doug Davis wrote:
> > Assuming you add a line like that here, how would you test
> > compliance with the SHOULD in ocitools?
>
> two options:
> 1 - we don't check it since its just a SHOULD

This is #543, which I'm not in favor of, but that's not the point…

> 2 - we issue an "informational" message as a warning about it…

This (and your (1)) are generic SHOULD stuff.  I'm more concerned
about how you'd implement the test *at all*.  Issues from [1]:

* “read an external spec”,
* “check for the annotation keys it defines”, and
* “determine which DNS names were controlled by the spec authors”.

Do you have implementation suggestions for those (I don't)?  If not,
how could you test the SHOULD?

[1]: https://github.com/opencontainers/runtime-spec/pull/510#issuecomment-243648546

duglin:
I feel like I'm missing something. Generic statements:
- The test tools shouldn't be reading external specs.
- There may be some things in our spec that we just can't test, but that doesn't mean we shouldn't have a requirement about it.
- Specs should not be changed because of testing limitations. That's like modifying your product to match the testing tools, rather than the end-user requirements.  Crazy talk  :-)

Testing a SHOULD is no different than testing a MUST - either you can or can't do it with the test tools. The only difference is how you deal with the detection of them not following the rule. MUST always results in an error ; SHOULDs are either ignored, generate warnings, or generate an error if there's some kind of "strict" option.

wking:
On Wed, Aug 31, 2016 at 04:43:28AM -0700, Doug Davis wrote:
> I feel like I'm missing something. Generic statements:
> - The test tools shouldn't be reading external specs.

The line I think you're considering adding [1]:

  Keys SHOULD be named using a reverse domain notation -
  e.g. `com.example.myKey`.

is a requirement placed on external specs.  It is not a requirement
placed on runtimes, nor is it a requirement placed on configurations.
This is the third-audience point I tried to explain in [2].

If you see it as a runtime or configuration requirement, could you
explain how you'd test runtimes or configurations for compliance?

> - There may be some things in our spec that we just can't test, but
>   that doesn't mean we shouldn't have a requirement about it.

We certainly *could* have a SHOULD requirement for external specs.
But if that's the only requirement and testing tools are impossibly
difficult to write?  I don't see the point.  On the other hand,
runtime and configuration authors reading this spec might see the
SHOULD and mistakenly worry that it somehow applies to them.  So I'd
rather avoid the RFC-2119 language for external-spec requirements
until we have a plan for testing them and/or a handful (at least > 1)
such requirements.

[1]: https://github.com/opencontainers/runtime-spec/pull/510#issuecomment-243648546
[2]: https://github.com/opencontainers/runtime-spec/pull/510#issuecomment-243516406

duglin:
upated

config.md
```diff
@@ -324,15 +324,24 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 
 **`annotations`** (object, optional) contains arbitrary metadata for the container.
 This information MAY be structured or unstructured.
-Annotations are key-value maps.
+Annotations MUST be a key-value map where both the key and value MUST be strings.
+While the value MUST be present, it MAY be an empty string.
+Keys MUST be unique within this map, and best practice is to namespace the keys.
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
```
wking: Missing the trailing dot for `org.opencontainers.`.

“subsequent specifications” is unclear to me.  Can we [follow image-spec][1] and use “… are reserved for future versions of the specification.”?

[1]: https://github.com/opencontainers/image-spec/blame/08ef0b1d662ac889d3bf2e99957acddad8e5804f/descriptor.md#L37

config.md
```diff
@@ -324,15 +324,24 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
...
+If there are no annotations then this property MAY either be absent or an empty map.
+Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown annotation key.
 
 ```json
 "annotations": {
-    "key1" : "value1",
-    "key2" : "value2"
+    "com.example.gpu-cores" : "2"
```
wking: While touching these lines, we have an opportunity to follow `jq` (and every other JSON pretty-printer I've seen, and drop the space before the colon:

    "com.example.gpu-cores": "2"

vbatts:
LGTM

RobDolinMS:
The use of SHOULD in line 330 of this PR looks good to me.  /cc @duglin, @wking, @vbatts 

I also agree with @duglin's comment about how test tools interpret:
https://github.com/opencontainers/runtime-spec/pull/510#issuecomment-243652186

mrunalp:
LGTM

PR #511, opened 2016-07-05
Add initial pass at a cmd line spec

duglin:
Gives us something to shoot arrows at  :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ Table of Contents
 - Runtime and Lifecycle
   - [General Runtime and Lifecycle](runtime.md)
   - [Linux-specific Runtime and Lifecycle](runtime-linux.md)
+  - [Runtime Command Line Interface](cmd.md)
 - Configuration
   - [General Configuration](config.md)
   - [Linux-specific Configuration](config-linux.md)
--- a/cmd.md
+++ b/cmd.md
@@ -1,0 +1,96 @@
+# Runtime Command Line Interface
+
+This document specifies the command line syntax for runtime operations defined in the [Runtime and Lifecycle](runtime.md) specification.
+It is STRONGLY RECOMMENDED that implementations adhere to the command line definition defined below to ensure interoperability.
+This document will not duplicate the information/semantics specified in the [Runtime and Lifecycle](runtime.md) document, rather it just focuses on syntax of the command line.
+
+## Convention
+
+For the purpose of this document, the following syntactical conventions are used:
+
+* `[ ... ]` denotes an optional field
+* `< ... >` denotes a substitution field. 
+The word(s), or phrase, within the `<>` describes the information to be passed-in.
+
+## General Format and Behavior
+
+The general format of all commands MUST be:
+```
+runtime [global-options] action [action-specific-options] [arguments]
+```
+
+Unknown options (global and action specific) MUST generate an error and exit with a non-zero exit code, without changing the state of the environment.
+
+Upon successful running of an action, the exit code MUST be zero.
+
+If there is an error during the running of an action, then:
+* the exit code MUST be non-zero
+* any error text MUST be displayed on stderr
+* the state of the environment SHOULD be the same as if the action was never attempted (modulo any possible trivial ancillary changes such as logging)
+
+### Global Options
+
+Global options are ones that apply to all actions.
+This specification doesn't define any global options.
+Implementation MAY define their own.
+
+### Action Specific Options
+
+All actions MUST support the `--help` action-specific-option, which:
+* MUST display some help text to stdout
+* MUST NOT perform the `action` itself
+* MUST have an exit code of zero if the help text is successfully displayed
+
+Implementations MAY define their own action-specific options.
+
+## Actions
+
+This section defines the actions defined by this specification.
+
+### State
+
+Format: `runtime [global-options] state [options] <container-id>`
+
+Options: None
+
+This action MUST display the state of the specific container to stdout.
+Unless otherwise specified by the user, the format of the state MUST be in JSON.
+
+See [Query State](runtime.md#query-state).
+
+### Create
+
+Format: `runtime [global-options] create [options] <container-id>`
+
+Options:
+* `-b <dir>`,  `--bundle <dir>` The path to the root of the bundle directory. If not specified the default value MUST be the current working directory.
+* `--console <path>` The PTY slave path for the newly created container.
+* `--pid-file <path>` The file path into which the process ID is written. If not specified then a file MUST NOT be created.
+
+See [Create](runtime.md#create).
+
+### Start
+
+Format: `runtime [global-options] start [options] <container-id>`
+
+Options: None
+
+See [Start](runtime.md#start).
+
+### Kill
+
+Format: `runtime [global-options] kill [options] <container-id> <signal>`
+
+Options: None
+
+The `signal` MUST either be the signal's numerical value (e.g. `15`) or the signal's name (e.g. `SIGTERM`).
+
+See [Kill](runtime.md#kill).
+
+### Delete
+
+Format: `runtime [global-options] delete [options] <container-id>`
+
+Options: None
+
+See [Delete](runtime.md#delete).
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ DOC_FILES := \
 	bundle.md \
 	runtime.md \
 	runtime-linux.md \
+	cmd.md \
 	config.md \
 	config-linux.md \
 	config-solaris.md \


wking:
On Tue, Jul 05, 2016 at 03:10:31PM -0700, Doug Davis wrote:
> Gives us something to shoot arrows at  :-)

There's also existing work in [1], linked from [2].

[1]: https://github.com/wking/oci-command-line-api
[2]: https://groups.google.com/a/opencontainers.org/d/msg/dev/BIxya5eSNLo/8mO8rU7gBwAJ
     Subject: Re: Specifying the runtime's command-line interface
     Date: Wed, 2 Dec 2015 12:40:09 -0800
     Message-ID: <20151202204009.GA23518@odin.tremily.us>

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+
+See [Query State](runtime.md#query-state).
+
+### Create
+
+Format: `runtime [global-options] create [options] <container-id>`
+
+Options:
+* `-b <dir>`,  `--bundle <dir>` The path to the root of the bundle directory. If not specified the default value MUST be the current working directory.
+* `--console <path>` The PTY slave path for the newly created container.
```
cyphar: We still need to figure out how we should deal with consoles in the spec. I'm personally still wondering what the semantics of consoles should be.

/cc @wking

duglin: I think that kind of thing would be in the runtime.md doc no? I tried really hard to keep this doc focused just on the syntax and not semantics, where possible.

cyphar: I'm including whether or not the console options should even be a command-line option. Right now it feels wrong to me, especially since `terminal` is in the config.

wking: On Wed, Jul 06, 2016 at 07:16:56AM -0700, Doug Davis wrote:
> > +* `--console <path>` The PTY slave path for the newly created container.
> 
> I think that kind of thing would be in the runtime.md doc no?

I agree, and think it's not worth worrying about how terminal
information fits into the API before we have a clearer non-API spec
around what the setting means or have punted it from the non-API spec
[1] (which is why wking/oci-command-line-api does not talk about
process.terminal-related stuff).

[1]: https://github.com/opencontainers/runtime-spec/issues/494#issuecomment-225251219

duglin: For now I tried to make this doc align with what `runc` does so at least we're consistent. If we decide to change `runc` then that PR can cover both the code and this doc.

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+
+This section defines the actions defined by this specification.
+
+### State
+
+Format: `runtime [global-options] state [options] <container-id>`
+
+Options: None
+
+This action MUST display the state of the specific container to stdout.
```
wking: #513 has some wording around the [expected encoding][1] for the output from these commands.  Background on this in wking/oci-command-line-api#2 and wking/oci-command-line-api#16.

[1]: https://github.com/opencontainers/runtime-spec/pull/513/files#diff-0c1e5a5baa5649945697a530e09acbf5R18

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+### Create
+
+Format: `runtime [global-options] create [options] <container-id>`
+
+Options:
+* `-b <dir>`,  `--bundle <dir>` The path to the root of the bundle directory. If not specified the default value MUST be the current working directory.
+* `--console <path>` The PTY slave path for the newly created container.
+* `--pid-file <path>` The file path into which the process ID is written. If not specified then a file MUST not be created.
+
+See [Create](runtime.md#create).
```
wking: Also relevant is [the file-descriptor wording][1], which I brought over in #513 (see 50fd47a9).

I think it's also worth pointing out that (with the current runC approach) this command exits immediately after container-creation completes, with an exit code reflecting succesful (or not) creation.  In the event that the standard streams (inherited by the container process) were not something persistent (like a pseudo-TTY), I'm not sure where that leaves the container process (probably setup for SIGPIPE and similar).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/runtime-linux.md#file-descriptors

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Kill
+
+Format: `runtime [global-options] kill [options] <container-id> <signal>`
+
+Options: None
+
+The `signal` MUST either be the signal's numerical value (e.g. `15`) or the signal's name (e.g. `SIGTERM`).
+
+See [Kill](runtime.md#kill).
```
wking: I think this needs more wording around allowed signals, since Windows does not support POSIX signals.  I have some in #513 (see 24246184).

README.md
```diff
@@ -16,6 +16,7 @@ Table of Contents
 - Runtime and Lifecycle
   - [General Runtime and Lifecycle](runtime.md)
   - [Linux-specific Runtime and Lifecycle](runtime-linux.md)
+  - [Runtime Command Line Interface](cmd.md)
```
wking: You want a `Makefile` entry too, see #513's 4a257255.

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+
+* `[ ... ]` denotes an optional field
+* `< ... >` denotes a substitution field. 
+The word(s), or phrase, within the `<>` describes the information to be passed-in.
+
+## General Format and Behavior
+
+The general format of all commands MUST be:
+```
+runtime [global-options] action [action-specific-options] [arguments]
```
wking: #513 [places limits on the option and action formats][1] so runtimes can identify unrecognised actions as such.  See wking/oci-command-line-api#8 for more discussion.

[1]: https://github.com/opencontainers/runtime-spec/pull/513/files#diff-0c1e5a5baa5649945697a530e09acbf5R10

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+Upon successful running of an action, the exit code MUST be zero.
+
+If there is an error during the running of an action, then:
+* the exit code MUST be non-zero
+* any error text MUST be displayed on stderr
+* the state of the environment MUST be the same as if the action was never attempted (modulo any possible trivial ancillary changes such as logging)
```
wking: I think this last line is an unachievable goal, and since #418 the spec has cut runtimes a bit more slack.  Attempting cleanup (as required by #418) doesn't mean “successfully cleaned up” (see [here][1]).

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc1/runtime.md#L124

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+Unless otherwise specified by the user, the format of the state MUST be in JSON.
+
+See [Query State](runtime.md#query-state).
+
+### Create
+
+Format: `runtime [global-options] create [options] <container-id>`
+
+Options:
+* `-b <dir>`,  `--bundle <dir>` The path to the root of the bundle directory. If not specified the default value MUST be the current working directory.
```
wking: I don't think we want to bother with short-form options in this spec.  The goal is to provide a minimal, generic API for driving a compliant runtime (which the long-form gives us).  Generic runtime-callers (like the compliance test suite) can use that long form, and leave runtimes free to pick whatever short-form abbreviations they like (if any).

cmd.md
```diff
@@ -0,0 +1,94 @@
...
+* `--console <path>` The PTY slave path for the newly created container.
+* `--pid-file <path>` The file path into which the process ID is written. If not specified then a file MUST not be created.
```
wking: I'm guessing this is to enable the [`PR_SET_CHILD_SUBREAPER` magic][1]?  That approach is not portable (and maybe container PIDs are not even portable, #459).  I'd rather see a [portable way to collect the child exit code][2], and #508 is my attempt at groundwork for that.

[1]: https://github.com/opencontainers/runc/pull/827#discussion_r63598816
[2]: https://github.com/opencontainers/runtime-spec/pull/513#issuecomment-230978790

wking: And an even less controversial reason for dropping this requirement is “you can currently get that by calling `state`”.

runcom:
@duglin what about `--version`? and its output? I found runc has a 3 line output from `--version`. Should it be part of this? Should the output follow some convention? (sorry if I'm out of mind saying this)

duglin:
@runcom I would actually prefer for this to be left out of the spec for now. I like the idea of our cmd line spec focusing on just the corresponding features mentioned in the runtime spec. 

While getting the version info is interesting, I don't think its critical for interop right now. If we want it then, yes as you said, we'd need to include the definition of the output (in machine readable format too), and should probably be in the runtime spec.

runcom:
Yeah, I didn't mean to have that included in this PR, I'll open another issue to track my question/issue

vbatts:
/me on my phone
Is there presently a way to determine version of the spec implemented by a
cli?

wking:
On Tue, Jul 19, 2016 at 02:46:10PM -0700, Vincent Batts wrote:
> Is there presently a way to determine version of the spec
> implemented by a cli?

No, and there's a general lack of clarity in how ociVersion should be
applied.  More details in opencontainers/tob#16 and discussion in [1].
@duglin was going to file something about this [2].

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-07-06-17.00.log.html#l-75
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-07-06-17.00.log.html#l-82

README.md
duglin: done

cmd.md
duglin: softened it to a SHOULD

cmd.md
duglin: What do people think about dropping this? If we don't need it in the cmd line spec (because people should just use state() to get it) then I'm inclined to think we should drop it from runc as well.

PR #512, opened 2016-07-06
Add a working call-in number

duglin:

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ When in doubt, start on the [mailing-list](#mailing-list).
 ## Weekly Call
 
 The contributors and maintainers of all OCI projects have a weekly meeting Wednesdays at 10:00 AM (USA Pacific.)
-Everyone is welcome to participate via [UberConference web][UberConference] or audio-only: 646-494-8704 (no PIN needed.)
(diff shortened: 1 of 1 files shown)


wking:
UberConference has been working when @mrunalp is around; are we sure
we want to completely replace it?  I like being able to see who's
talking when I can't recognize someone's voice.

If we do remove it, we'll want to remove the remove the reference link
too [1].

[1]: https://github.com/duglin/runtime-spec/blame/e43c75b33f24eea1076ab53515bf5b9207025c7d/README.md#L165

caniszczyk:
Sorry I should have got back sooner.

At the LF we created a line for OCI now, you're welcome to use that one:

Join the call: https://www.uberconference.com/opencontainers
Optional dial in number: 888-587-9088
Alternate number: 860-706-8529
No PIN needed

duglin:
@caniszczyk thanks! I updated the PR to align with the info you provided.

mrunalp:
Thanks!

mrunalp:
LGTM

vbatts:
LGTM

vbatts:
this should be mirrored in the image-spec and templates

duglin:
@vbatts will try to do so later today....

duglin:
found a ref in image-spec but not templates

PR #513, opened 2016-07-06
Add Runtime CLI Spec

wking:
In today's meeting [I said I'd PR this](http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-07-06-17.00.log.html#l-24), so here it is.  I've kept the history to preserve previous review and contribution information.  I've tried to summarize any discussion that happened elsewhere in the commit messages, but there may be additional background in [the old repo's PRs](https://github.com/wking/oci-command-line-api/pulls?q=is%3Apr+is%3Aclosed).

Fixes #434.

PR #518, opened 2016-07-18
Clarify wording for terminal setting and /dev/console

mrunalp:
Signed-off-by: Mrunal Patel mrunalp@gmail.com

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -148,7 +148,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 * [`/dev/random`][random.4]
 * [`/dev/urandom`][random.4]
 * [`/dev/tty`][tty.4]
-* [`/dev/console`][console.4]
+* [`/dev/console`][console.4] is setup if terminal is enabled in the config by bind mounting the pseudoterminal slave to /dev/console.
 * [`/dev/ptmx`][pts.4].
   A [bind-mount or symlink of the container's `/dev/pts/ptmx`][devpts].
 
--- a/config.md
+++ b/config.md
@@ -94,7 +94,8 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 
 **`process`** (object, REQUIRED) configures the container process.
 
-* **`terminal`** (bool, OPTIONAL) specifies whether you want a terminal attached to that process, defaults to false.
+* **`terminal`** (bool, optional) specifies whether a terminal is attached to that process, defaults to false.
+  On Linux, a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
 * **`consoleSize`** (object, OPTIONAL) specifies the console size of the terminal if attached, containing the following properties:
   * **`height`** (uint, REQUIRED)
   * **`width`** (uint, REQUIRED)
@@ -744,3 +745,4 @@ Here is a full example `config.json` for reference.
 [mount.8-filesystem-independent]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS
 [mount.8-filesystem-specific]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-SPECIFIC_MOUNT%20OPTIONS
 [mount.8]: http://man7.org/linux/man-pages/man8/mount.8.html
+[stdin.3]: http://man7.org/linux/man-pages/man3/stdin.3.html


crosbymichael:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runtime-spec/pull-request/518/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runtime-spec)

config.md
```diff
@@ -85,7 +85,7 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 
 ## Process configuration
 
-* **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.
+* **`terminal`** (bool, optional) specifies whether a terminal is allocated for and attached to that process. Defaults to false.
```
wking: I think this is still too vague.  Who allocates the terminal?  It sounds like runC expects the runtime-caller to already have a slave, although ccon uses this setting to have the container process opens the pair with `posix_openpt`.  See also #494.

config-linux.md
```diff
@@ -147,7 +147,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 * [`/dev/random`][random.4]
 * [`/dev/urandom`][random.4]
 * [`/dev/tty`][tty.4]
-* [`/dev/console`][console.4]
+* [`/dev/console`][console.4] is setup if terminal is enabled in the config
```
wking: Is “this device is missing” the usual way to handle “you don't have a current console”?  It seems like processes that expect `/dev/console` would expect something besides ENOENT for that (although maybe they handle both ENOENT and whatever “you don't have a current console” returns, and maybe “you don't have a current console” is ENOENT).

config.md
mrunalp: @wking I think that both approaches are valid and we shouldn't add language to support one but not the other.

wking: On Wed, Jul 20, 2016 at 12:10:26PM -0700, Mrunal Patel wrote:
> > -* **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process. Defaults to false.
> > +* **`terminal`** (bool, optional) specifies whether a terminal is allocated for and attached to that process. Defaults to false.
>
> @wking I think that both approaches are valid and we shouldn't add
> language to support one but not the other.

I'm ok with being vague in config.md as long as it gets spelled out in
the command-line API specs.  Otherwise this is going to be hard to
compliance-test.  If we're trying to focus just on the
inside-the-container view here, how about:

  If true, the container process's standard streams will all be
  attached to a terminal.  Defaults to false.

If we're not going into detail about who allocates the terminal, I'd
rather not mention “allocate” at all.

wking:
I didn't record an entry for this in the meeting log, but I remember
saying (around [1]?) that I'd post notes here to a ccon implementation
of opening the psuedoterminal pair inside the container (which ccon
has done for a while) and passing it out to the caller via a Unix
socket (ccon previously only passed master from its container process
to its host process).  I've just pushed an implementaiton in [2],
where:

  $ ccon … --pseudoterminal-socket PATH

will connect to the Unix socket listening at PATH, and pass the master
file descriptor through.  The current implementation isn't fully
polished:

* The message should probably carry both the master file descriptor
  and an identifying token (e.g. the container ID) so that a single
  listener can collect masters from multiple containers.

* There's currently no way to have a ccon invocation exit early.  The
  host process waits for the container process to collect and return
  its exit code, regardless of whether it passes the master file
  descriptor out or not.

But it should be sufficient for folks who want to kick the tires of
the master pass-out approach.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-07-27-21.00.log.html#l-173
[2]: https://github.com/wking/ccon/tree/pseudoterminal-master-socket

mrunalp:
Rebased and updated.

config.md
```diff
@@ -90,7 +90,9 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 
 ## Process configuration
 
-* **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process, defaults to false.
+* **`terminal`** (bool, optional) specifies whether a terminal is attached to that process, defaults to false.
+  On Linux, a psuedo-terminal pair is allocated for the container process and the standard streams are duped to the pty slave.
```
wking: For extra clarity (and to match “Atomically duplicate 'oldfd' on 'newfd'” wording in [dup(2)][1] and “pseudoterminal” in [pts(4)][2]), I'd recommend:

> On Linux, a psuedoterminal pair is allocated for the container process and the psuedoterminal slave is duplicated on the container process's [standard streams][stdin.3].
>
> [stdin.3]: http://man7.org/linux/man-pages/man3/stdin.3.html

[1]: http://man7.org/linux/man-pages/man2/dup.2.html#NOTES
[2]: http://man7.org/linux/man-pages/man4/pts.4.html

mrunalp: Updated.

config.md
```diff
@@ -90,7 +90,9 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 
 ## Process configuration
 
-* **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process, defaults to false.
+* **`terminal`** (bool, optional) specifies whether a terminal is attached to that process, defaults to false.
+  On Linux, a psuedoterminal pair is allocated for the container process and the psuedoterminal slave is duplicated on the container process's [standard streams][stdin.3].
+  Also, the psuedoterminal slave is bind mounted at /dev/console.
```
wking: I think “bind mounted” is overly specific.  Ccon does not bind mount `/dev/console`, it relies on the container's [`posix_openpt`][1] and the kernel/libc implementation to get the appropriate console device behavior.  In this case, I suggest we lead with the ocitools implementation of the test (how do you validate proper /dev/console handling?) and then figure out the wording after we have ocitools tests we like.  I expect the test will be something like “container writes to /dev/console and test harness ensures it reads that string out of the pseudoterminal master”, but we may want other tests in addition (I'm not familiar enough with /dev/console to know).

[1]: http://man7.org/linux/man-pages/man3/posix_openpt.3.html

wking: Actually, /dev/console looks like a different beast from the psuedoterminal.  Why link them at all?

config-linux.md
wking: 437ff74 → d789654 dropped the earlier link between `process.terminal` and `/dev/console`.  Do we want to keep this conditional here?  I'd rather drop the change and require the runtime to always provide `/dev/console` (pointing to whatever console it feels is appropriate).

config-linux.md
```diff
@@ -147,7 +147,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 * [`/dev/random`][random.4]
 * [`/dev/urandom`][random.4]
 * [`/dev/tty`][tty.4]
-* [`/dev/console`][console.4]
+* [`/dev/console`][console.4] is setup if terminal is enabled in the config by bind mounting the psuedoterminal slave to /dev/console.
```
hqhq: I think `/dev/console` can be a common terminal and not just pseudoterminal, no?

config.md
```diff
@@ -90,7 +90,8 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 
 ## Process configuration
 
-* **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process, defaults to false.
+* **`terminal`** (bool, optional) specifies whether a terminal is attached to that process, defaults to false.
+  On Linux, a psuedoterminal pair is allocated for the container process and the psuedoterminal slave is duplicated on the container process's [standard streams][stdin.3].
```
hqhq: s/psuedoterminal/pseudoterminal

config-linux.md
mrunalp: @hqhq For container runtimes, it is the pseudoterminal slave. I haven't seen a use case or setup otherwise.

config.md
mrunalp: Pushed fix. Thanks!

mrunalp:
This is ready for review. ping @opencontainers/runtime-spec-maintainers

config-linux.md
```diff
@@ -147,7 +147,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 * [`/dev/random`][random.4]
 * [`/dev/urandom`][random.4]
 * [`/dev/tty`][tty.4]
-* [`/dev/console`][console.4]
+* [`/dev/console`][console.4] is setup if terminal is enabled in the config by bind mounting the pseudoterminal slave to /dev/console.
```
wking: I still think:

* We [want `/dev/console` unconditionally][1].
* We [don't want to require bind-mounting the slave over this][2].  Let the runtime point `/dev/console` to a place it considers appropriate for [logging system messages][3], which may not be the pseudoterminal slave.  In fact, I'd expect the validation for `/dev/console` to be “[Is the major 5, the minor 1, and the type ‘char’?][4]”, which is not going to be satisfied by bind-mounting the slave.

[1]: https://github.com/opencontainers/runtime-spec/pull/518#issuecomment-240494625
[2]: https://github.com/opencontainers/runtime-spec/pull/518#discussion_r73214575
[3]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7.1#n3248
[4]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7.1#n210

hqhq:
@cyphar According to the new console rewrite in runc, any thoughts about how should we define `/dev/console` in runtime-spec? I'm afraid bind mount pseudoterminal slave to /dev/console would be too  implementation detailed for spec, and it should not be the only way for a container.

config-linux.md
cyphar: >     We don't want to require bind-mounting the slave over this. Let the runtime point /dev/console to a place it considers appropriate for logging system messages, which may not be the pseudoterminal slave. In fact, I'd expect the validation for /dev/console to be “Is the major 5, the minor 1, and the type ‘char’?”, which is not going to be satisfied by bind-mounting the slave.

That validation does not work inside containers. That particular major and minor number is reserved for consoles that are created as part of initial bootup (think **physical** console) on Linux. I think that it is entirely sufficient to just state that `/dev/console` should be a valid place to log system messages (which is what you suggested).

wking:
On Thu, Sep 29, 2016 at 01:49:00AM -0700, Aleksa Sarai wrote:
> -* [`/dev/console`][console.4]
> +* [`/dev/console`][console.4] is setup if terminal is enabled in the config by bind mounting the pseudoterminal slave to /dev/console.
>
> > We don't want to require bind-mounting the slave over this. Let
> > the runtime point /dev/console to a place it considers appropriate
> > for logging system messages, which may not be the pseudoterminal
> > slave. In fact, I'd expect the validation for /dev/console to be
> > “Is the major 5, the minor 1, and the type ‘char’?”, which is not
> > going to be satisfied by bind-mounting the slave.
>
> That validation does not work inside containers. That particular
> major and minor number is reserved for consoles that are created as
> part of initial bootup (think **physical** console) on Linux.

It *can* work inside containers:

  $ unshare -Urm file /dev/console 
  /dev/console: character special (5/1)

All of the logged messages will end up in the host console, but maybe
that's the semantics we want.

> I think that it is entirely sufficient to just state that
> `/dev/console` should be a valid place to log system messages (which
> is what you suggested).

If we allow devices other than c5/1 at that path, I think we need to
be very clear about the API they should support.  For example, screen
uses TIOCCONS to redirect the console output.  Is that required to
work inside OCI containers (assuming the caller has sufficient
permission [1])?  You can test this with ‘ctrl-a :console on’ in a
screen.  Does that catch output written to /dev/console if it isn't
c5/1?

[1]: http://man7.org/linux/man-pages/man4/tty_ioctl.4.html

cyphar:
@wking 

Sure, that validation /could/ work. But what you've done in your example is just use the parent's `/dev/console`. If we just did `mknod("/dev/console", makedev(5,1))` then a container could start modifying what the host console sees (read: the console that people trying to administer the system are going to log into). So we **definitely** do not want to make that a **restriction** (which is what I meant when I said "validation"), because it could be seen as a security issue. If you want to have any restriction for what we support then we should simply say that it must be "a console" (which on Linux means that it's the slave of a `ptmx` pair).

hqhq:
Yeah I think `/sys/dev/char/5:1/` is not isolated, if you mknod /dev/console with c(5, 1) in the container and change the backend device, it'll change `/sys/dev/char/5:1/` and affect host.

wking:
On Thu, Sep 29, 2016 at 08:15:03PM -0700, Qiang Huang wrote:
> Yeah I think `/sys/dev/char/5:1/` is not isolated, if you mknod
> /dev/console with c(5, 1) in the container and change the backend
> device, it'll change `/sys/dev/char/5:1/` and affect host.

I'm fine with that (although I'd also be fine with a console device
that was namespace-able like devpts is).  If you don't want your
containers redirecting /dev/console, don't give them CAP_SYS_ADMIN.
And if you don't want your containers writing to /dev/console, adjust
its permissions so they can't touch it (which may include mapping
containers from non-root host users).  If you launch containers from
root host users and let them keep CAP_SYS_ADMIN, then presumably
you're ok with them mucking about with the host /dev/console (maybe
you have a containerized console monitor?).

If you don't care about TIOCCONS you can mount another pseudoterminal
slave (or whatever) at /dev/console, and tools that just write to
/dev/console won't care.  But you can do that with explicit settings
in your container config.  Of course, you can create a c5:1 device
with explicit settings in the container config too.  It's unclear to
me if an entry in linux.devices will clobber or error if the target
path already exists (e.g. because the runtime has bind-mounted the
pseudoterminal slave there).  It's also unclear to me if a the
pseudoterminal is opened early enough for a mounts entry to bind-mount
/dev/pts/1 over /dev/console.  A pre-start hook or post-create
manipulation would work for either approach, regardless of what the
runtime sets up during the create call.  So whichever way we pick, we
should be very clear about it, so runtime callers can take steps to
get their desired /dev/console if it doesn't match what the spec
requires the runtime to create.

cyphar:
@wking Again, I never said that using `/dev/console` should be **forbidden** (though I think such use would be A Bad Idea (tm)). What I'm saying is that we **must not** restrict `/dev/console` to always be `5:1` because of the issues I've discussed.

I also maintain that "somewhere to dump system messages" is sufficient restriction / definition for users. If someone wants to use something specialised like `TIOCCONS` modify it then they should use the tools we provide to do so -- it doesn't make sense to cripple the spec to pander to people who want to use a very odd (and host-centric so it doesn't even make much sense in the context of containers) API.

wking:
On Fri, Sep 30, 2016 at 04:04:14AM -0700, Aleksa Sarai wrote:
> I also maintain that "somewhere to dump system messages" is
> sufficient restriction / definition for users. If someone wants to
> use something specialised like `TIOCCONS` modify it then they should
> use the tools we provide to do so…

So the spec entry should be something like:

* [`/dev/console`][console.4] MUST be a terminal and MAY be the [system console][system-console].
    Only the system console supports [`TIOCCONS`][tty_ioctl.4], so containers relying on `TIOCCONS` SHOULD check the runtime-provided `/dev/console` and if neccessary replace it with a [system console][] device.

[system-console]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7.1#n210
[tty_ioctl.4]: http://man7.org/linux/man-pages/man4/tty_ioctl.4.html

cyphar:
@wking I think this is what we should put:

```
* [`/dev/console`][console.4] MUST be the preferred location for system messages to be written by the container (aside from the standard streams). It MAY be a terminal and MAY be the [system console][system-console]. Only the system console supports [`TIOCCONS`][tty_ioctl.4], so containers relying on `TIOCCONS` SHOULD check the runtime-provided `/dev/console` and if neccessary replace it with a [system console][] device.
```

Footnotes the same as your comment. I'm not sure about the terminal restriction to be completely honest. But I'd consider using your wording if you can give a good example of how a misreading of the spec could result in something we shouldn't allow happening.

wking:
On Fri, Sep 30, 2016 at 08:12:24AM -0700, Aleksa Sarai wrote:
> * [`/dev/console`][console.4] MUST be the preferred location for
>   system messages to be written by the container (aside from the
>   standard streams).

I didn't want to MUST it as a logging location, because that's
untestable without a standard API for retrieving its contents.  I'd
rather just link to [1] at some point and let users decide what it
should be used for on their own.  On the other hand, what it *is*
(definitely a terminal, possibly c5/1) is testable.

> It MAY be a terminal and MAY be the [system console][system-console].

The system console is a terminal:

  $ file /dev/console 
  /dev/console: character special (5/1)
  $ sudo stty -F /dev/console 
  speed 38400 baud; line = 0;
  min = 1; time = 0;
  ignbrk -brkint -icrnl -imaxbel
  -opost -onlcr
  -isig -icanon -iexten -echo -echoe -echok -echoctl -echoke

So I think it MUST be a terminal and MAY be c5/1.

> I'm not sure about the terminal restriction to be completely
> honest. But I'd consider using your wording if you can give a good
> example of how a misreading of the spec could result in something we
> shouldn't allow happening.

I'd rather have the spec require c5/1, and runC is currenly
bind-mounting the pseudoterminal slave here.  Both of those are
terminals, and I haven't heard anyone suggest putting a non-terminal
here.  So I'd like to require a terminal to keep the provided
/dev/console as close to c5/1 as possible while still forming a
consensus ;).  I don't know how many people are mucking about with the
terminal settings on /dev/console though.  If you put a named pipe
there, I'd guess most people would still be fine.  But that's just a
guess, and I'd rather be conservative in the v1.0 spec.  We can always
relax this to “something you can open and write to, which MAY be a
terminal” later if a runtime really does want to use a named pipe or
some such and can make a convincing argument for nobody actually
needing this to be a terminal.

[1]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7.1#n3248

cyphar:
@wking 

> I didn't want to MUST it as a logging location, because that's untestable without a standard API for retrieving its contents.

But then you run into "where does that terminal pipe to" territory and so on. At some point you have to accept that there's certain things we cannot test and are out-of-scope. Where `/dev/console` feeds to is something that is out of scope IMO.

> The system console is a terminal:

The system console **in the host**, which is **not namespaced in the Linux kernel**. We should be asking ourselves "what is the **purpose** of `/dev/console`" -- not "what does it actually look like on my host". Not to mention that the wacky `ioctl` you referenced will also influence the host on Linux so it's still a problem. This is not true (from my understanding) in illumos, which is why I'm fine with keeping it as a `MAY`. But requiring `c5/1` is just insanity.

> So I'd like to require a terminal to keep the provided /dev/console as close to c5/1 as possible while still forming a consensus ;). 

I'm fine with saying that it `MUST be a terminal` so long as we don't require `c5/1` -- we should not be requiring un-namespaced stuff. However, I'm just trying to not be insanely restrictive. `/dev/console` is a very odd historical artifact and we should not be assuming that people want the technically most accurate translation in the specification -- especially since the translation is something as weird as "you can modify how the host's console setup works inside a container".

wking:
On Sat, Oct 01, 2016 at 08:14:05AM -0700, Aleksa Sarai wrote:
> > I didn't want to MUST it as a logging location, because that's
> > untestable without a standard API for retrieving its contents.
>
> But then you run into "where does that terminal pipe to" territory
> and so on. At some point you have to accept that there's certain
> things we cannot test and are out-of-scope. Where `/dev/console`
> feeds to is something that is out of scope IMO.

Agreed, which is why I would rather not “MUST be the preferred
location for system messages”.  Make that untestable suggestion
informative instead of normative.

> > The system console is a terminal:
>
> The system console **in the host**, which is **not namespaced in the
> Linux kernel**. We should be asking ourselves "what is the
> **purpose** of `/dev/console`" -- not "what does it actually look
> like on my host". Not to mention that the wacky `ioctl` you
> referenced will also influence the host on Linux so it's still a
> problem. This is not true (from my understanding) in illumos, which
> is why I'm fine with keeping it as a `MAY`. But requiring `c5/1` is
> just insanity.

I think c5/1 is reasonable, but I doubt I can build a consensus around
it.  My previous proposal only required a terminal [1], which is a
much larger class of devices (although it contains c5/1).  And while
asking what /dev/console is used for is nice, it's a hard question to
answer without a broader audit of existing software than I have time
for ;).  We know that if we provide all of the c5/1 semantics we're
ok, because Linux software consuming /dev/console is used to
interacting with c5/1.  If we want to deviate from c5/1, because of
security concerns in containers where the user is root on the host
and/or retains CAP_SYS_ADMIN on the host but is for some reason not
supposed to be accessing the non-namespaced c5/1, then I think we
should be very careful about the deviations we allow.  Explicitly
breaking TIOCCONS is fine, because you *have* to break TIOCCONS if you
leave c5/1.  Explicitly breaking terminal-ness is… maybe ok?  I don't
know.  And without someone pitching a strong “here's why I need it,
and here's why it's fine…”, I'd rather avoid it.  We can always relax
the terminal restriction later if/when someone does make that pitch.

> > So I'd like to require a terminal to keep the provided
> > /dev/console as close to c5/1 as possible while still forming a
> > consensus ;).
> 
> I'm fine with saying that it `MUST be a terminal` so long as we
> don't require `c5/1` -- we should not be requiring un-namespaced
> stuff. However, I'm just trying to not be insanely restrictive.

I think we have a fair amount of common ground here.  Taking another
stab at spec wording:

* [`/dev/console`][console.4] MUST be a terminal and MAY be the
  [system console][system-console-device]. `/dev/console` is usually
  [used for logging system messages][system-console-usage].  Only the
  [system console][system-console-device] supports
  [`TIOCCONS`][tty_ioctl.4], so containers relying on `TIOCCONS`
  SHOULD check the runtime-provided `/dev/console` and if neccessary
  replace it with a [system console device][system-console-device].

[system-console-device]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7#n210
[system-console-usage]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7#n3248
[tty_ioctl.4]: http://man7.org/linux/man-pages/man4/tty_ioctl.4.html

[1]: https://github.com/opencontainers/runtime-spec/pull/518#issuecomment-250769468

cyphar:
@wking I'm fine with your proposed wording. But just a quick point about this:

> If we want to deviate from c5/1, because of security concerns in containers where the user is root on the host and/or retains CAP_SYS_ADMIN on the host but is for some reason not supposed to be accessing the non-namespaced c5/1, then I think we should be very careful about the deviations we allow.

The devices restrictions and several other tools can be used to give CAP_SYS_ADMIN to a process (which is a requirement for an insane number of operations as it is the catch-all capability) without allowing for modifications to the host that would be unexpected by people operating the host.

mrunalp:
We do try to restrict access to non-namespaced locations where possible in runc. For e.g. we validate the sysctls that are namespaced and only allow those. I think /dev/console probably falls under the same category. 
I think we have these options:
1. Bind mount /dev/console to the pty slave  to ensure that writes by software that write to that location don't leak out to the host.
2. Don't do anything at all with /dev/console when terminal is true. If the user wants then they could add a device node entry knowing that it is going to leak out or they could bind mount /dev/console over /dev/pts/1 adding a mount to the config.
3. Add a boolean that enables the bind mounting of /dev/console to pty slave.

wking:
On Wed, Oct 05, 2016 at 09:45:57AM -0700, Mrunal Patel wrote:
> We do try to restrict access to non-namespaced locations where
> possible in runc.

I'm completely fine with that as a runC policy.  I just don't think we
want to *require* it in the spec.

> 1. Bind mount /dev/console to the pty slave to ensure that writes by
>    software that write to that location don't leak out to the host.

I'm fine with this as an implementation, but don't think we want to
require it as the only legal implementation.

> 2. Don't do anything at all with /dev/console when terminal is
>    true.

I'm not sure what “don't do anything” is here.  Are you saying “use
c5/1”?  That would be my favorite choice, but…

> If the user wants then they could add a device node entry knowing
> that it is going to leak out or they could bind mount /dev/console
> over /dev/pts/1 adding a mount to the config.

… this may not work if you do it from a ‘mounts’ entry, because the
spec does not require an order between “mount all the ‘mounts’” and
“setup /dev/pts/1”.  You can use a pre-start hook or post-create
manipulation [1].

> 3. Add a boolean that enables the bind mounting of /dev/console to
>    pty slave.

I doubt many people care particularly about where /dev/console points
or whether it's a c5/1 or bind-mounted pseudoterminal slave.  For
folks who do care, it's easy enough to do this in a pre-start hook or
post-create manipulation.  I'd rather not add another terminal-related
config option for something that's used rarely and easy to handle
directly when you do need it.

With the wording I float in [2], both (1) and using c5/1 would be
valid implementation choices, so I still think that wording makes
sense.

[1]: https://github.com/opencontainers/runtime-spec/pull/518#issuecomment-250653763
[2]: https://github.com/opencontainers/runtime-spec/pull/518#issuecomment-250940348

hqhq:
>  [`/dev/console`][console.4] MUST be a terminal and MAY be the
  [system console][system-console-device]. `/dev/console` is usually
  [used for logging system messages][system-console-usage].  Only the
  [system console][system-console-device] supports
  [`TIOCCONS`][tty_ioctl.4], so containers relying on `TIOCCONS`
  SHOULD check the runtime-provided `/dev/console` and if neccessary
  replace it with a [system console device][system-console-device].

Is TIOCCONS the only case that requires `/dev/console` to be a system console? I doubt it so maybe we can twist it a bit as:
* [`/dev/console`][console.4] MUST be a terminal and MAY be the [system console][system-console-device]. `/dev/console` is usually [used for logging system messages][system-console-usage].  For operations which require `/dev/console` to be [system console][system-console-device] such as
[`TIOCCONS`][tty_ioctl.4], containers SHOULD check the runtime-provided `/dev/console` and if necessary replace it with a [system console][system-console-device].

[system-console-device]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7#n210
[system-console-usage]: https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/Documentation/devices.txt?id=refs/tags/v4.7#n3248
[tty_ioctl.4]: http://man7.org/linux/man-pages/man4/tty_ioctl.4.html

wking:
On Wed, Oct 05, 2016 at 12:59:48PM -0700, Qiang Huang wrote:
> Is TIOCCONS the only case that requires `/dev/console` to be a
> system console?

I am not aware of any others, but I like your more conservative
wording better.

wking:
In today's meeting there was more discussion on whether /dev/console
should be required when process.terminal isn't true [1].  I argued for
always requiring /dev/console, since it's for system logs, and that
seems orthogonal to process.terminal.  @crosbymichael pointed out
resource-consumption issues if the runtime opens a new pseudoterminal
pair for each container (if it tries to source that /dev/console by
bind-mounting a pseudoterminal slave).  And he pointed out that runC
currently only creates /dev/console when process.terminal is true and
that works pretty well.  I suggested not requiring /dev/console at all
then (and users who need it can add a devices entry for c5/1 or
bind-mount their stdin over /dev/console as they see fit), and
@crosbymichael cited past concerns with tmux and sshd.  However, I see
no references to /dev/console in either codebase [2,3].  The only code
I've been able to find that actually uses /dev/console is screen, and
then only for TIOCCONS [4,5].  If we are going to require /dev/console
at all, I think we should either always require it (regardless of
process.terminal) or build a better case for why it's related to
process.terminal.  For example, with explict step to generate
/dev/console-is-missing errors in tmux or OpenSSH, or links to issues
that describe those steps.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-05-21.01.log.html#l-20
[2]: https://github.com/tmux/tmux/search?utf8=%E2%9C%93&q=%2Fdev%2Fconsole
[3]: https://github.com/openssh/openssh-portable/search?utf8=%E2%9C%93&q=%2Fdev%2Fconsole
[4]: https://github.com/opencontainers/runtime-spec/pull/518#issuecomment-250521182
[5]: http://git.savannah.gnu.org/cgit/screen.git/tree/src/tty.sh?h=v.4.4.0#n1067

cyphar:
@wking 

> I suggested not requiring /dev/console at all then (and users who need it can add a devices entry for c5/1 or bind-mount their stdin over /dev/console as they see fit), and @crosbymichael cited past concerns with tmux and sshd. However, I see no references to /dev/console in either codebase [2,3].

If you're willing to test a **VERY** large amount of software with that change, then I would accept removing the requirement. However, that would just be time consuming and i would prefer that we have the requirement and then drop it in later versions of the spec. This PR (and all of the other PRs for terminal) have dragged on for months and months and it's starting to get quite aggravating TBH.

wking:
On Tue, Oct 18, 2016 at 01:51:25AM -0700, Aleksa Sarai wrote:
> > I suggested not requiring /dev/console at all then (and users who
> > need it can add a devices entry for c5/1 or bind-mount their stdin
> > over /dev/console as they see fit), and @crosbymichael cited past
> > concerns with tmux and sshd. However, I see no references to
> > /dev/console in either codebase [2,3].
>
> If you're willing to test a **VERY** large amount of software with
> that change, then I would accept removing the requirement. However,
> that would just be time consuming and i would prefer that we have
> the requirement and then drop it in later versions of the spec. This
> PR (and all of the other PRs for terminal) have dragged on for
> months and months and it's starting to get quite aggravating TBH.

I think that “provide something we're not sure we need, since we can
remove it later” is going the wrong way ;).  Can we *not* provide it
until we're sure we need it?  I think the burden of proof for a
feature should be to show why it's needed (which you can actually do),
not to show that nobody needs it (which seems impossible).  That said,
if you have suggestions on what you'd like to see for demonstrating
that it's not needed, I'm willing to give it a go.

While we're working out when /dev/console is actually needed, folks
who know they need it (because they're planning on using TIOCCONS?) or
suspect they need it can easily add a linux.devices entry or do
post-create bind-mounting.

cyphar:
@wking 
> I think that “provide something we're not sure we need, since we can remove it later” is going the wrong way ;). Can we *not* provide it until we're sure we need it?

Wouldn't that be non-backwards-compatible? Since we'd be adding a new restriction to runtimes? Or would that be fine for 1.1?

>  I think the burden of proof for a feature should be to show why it's needed (which you can actually do), not to show that nobody needs it (which seems impossible).

`/dev/console` was added to Docker/runC for reasons of "legacy software". The status quo for containers is that it exists -- so the burden of proof is on you (that it is no longer a concern).

> That said, if you have suggestions on what you'd like to see for demonstrating that it's not needed, I'm willing to give it a go. 

Remove the `/dev/console` handling from runC (and Docker) and run as many different images using Docker to see whether or not everything works in both instances. I can't give you a list of software to test, because I'm not sure what software uses `/dev/console`. Maybe some GitHub searching can help you here.

> While we're working out when /dev/console is actually needed, folks who know they need it (because they're planning on using TIOCCONS?) or suspect they need it can easily add a linux.devices entry or do post-create bind-mounting.

Again, `c5/1` is not going to be a requirement for `/dev/console`. I will NACK any patch that makes this a requirement. So if someone wants to use `c5/1` they will have to either check their runtime or explicitly make it themselves. For the "general logging" usecase, that's what we're testing here.

wking:
On Tue, Oct 18, 2016 at 03:43:25AM -0700, Aleksa Sarai wrote:
> > I think that “provide something we're not sure we need, since we
> > can remove it later” is going the wrong way ;). Can we *not*
> > provide it until we're sure we need it?
>
> Wouldn't that be non-backwards-compatible? Since we'd be adding a
> new restriction to runtimes? Or would that be fine for 1.1?

Both major and minor bumps add new restrictions to runtimes.  The
backwards compatibility is for configs (so a 1.0 config still works
with a 1.1 runtime).  The wording for this landed in #523.

> > I think the burden of proof for a feature should be to show why
> > it's needed (which you can actually do), not to show that nobody
> > needs it (which seems impossible).
>
> `/dev/console` was added to Docker/runC for reasons of "legacy
> software". The status quo for containers is that it exists -- so the
> burden of proof is on you (that it is no longer a concern).

“Somebody else has already picked this direction” isn't a good
argument.  If a solid argument around legacy software needing
/dev/console mostly when process.terminal is true exists in the Docker
/ runC archives, then folks who are pushing for runtime-spec to follow
that example should be able to link to that previous discussion.
*Then* it would be on me to try and poke holes in anything I didn't
find convincing.  But saying that the discussion happened without
giving a reference to it makes it hard to have a rational discussion
here.

> > That said, if you have suggestions on what you'd like to see for
> > demonstrating that it's not needed, I'm willing to give it a go.
> 
> Remove the `/dev/console` handling from runC (and Docker) and run as
> many different images using Docker to see whether or not everything
> works in both instances. I can't give you a list of software to
> test, because I'm not sure what software uses `/dev/console`. Maybe
> some GitHub searching can help you here.

GitHub searching turns up many, many links to kernel docs ;).  I
didn't get deep enough into the results to find any non-kernel
software using it.  I'll start translating Docker images to OCI and
see if they run without /dev/console, although “see whether or not
everything works” sounds like an impossibly high bar, even for a
container as simple as a SSH daemon process.  There are too many
possibilities to handle as a black box without grepping around for
/dev/console.

> > While we're working out when /dev/console is actually needed,
> > folks who know they need it (because they're planning on using
> > TIOCCONS?) or suspect they need it can easily add a linux.devices
> > entry or do post-create bind-mounting.
> 
> Again, `c5/1` is not going to be a requirement for `/dev/console`.

Agreed.  Mounting c5/1 is just one of several options for providing
/dev/console.  The post-create bind-mounting is for folks who want to
bind /proc/${CONTAINER_PID}/fd/1 over /dev/console.

wking:
Poking around trying to find /dev/console consumers turned up syslog
(unsurprisingly, since /dev/console is for system logging).  syslog-ng
doesn't hard-code /dev/console anywhere, but there are commented-out
entries in their Fedora [1,2] and Red Hat [3,4] examples.  So folks
running syslog in their containers who have configured it to write to
/dev/console will need something terminal-like there.

rsyslogd has _PATH_CONSOLE defaulting to /dev/console (at
compile-time) [5].  It is only used for TTY “detection” [6].  There is
also a commented-out entry in the RedHat example config [7].  So this
seems like the same situation as syslog-ng: the logger doesn't hit
/dev/console by default, but a user might configure it to write there.

The “syslog may be configured to write to /dev/console” seems
orthogonal to process.terminal to me.  And if we don't require
runtimes to supply /dev/console at all, those users can either
configure syslog to write somewhere else or use linux.devices or
post-create manipulation to put a terminal at /dev/console.

[1]: https://github.com/balabit/syslog-ng/blob/v3.5.6/contrib/fedora-packaging/syslog-ng.conf#L50
[2]: https://github.com/balabit/syslog-ng/blob/v3.5.6/contrib/fedora-packaging/syslog-ng.conf#L70
[3]: https://github.com/balabit/syslog-ng/blob/v3.5.6/contrib/syslog-ng.conf.RedHat#L41
[4]: https://github.com/balabit/syslog-ng/blob/v3.5.6/contrib/syslog-ng.conf.RedHat#L62
[5]: https://github.com/rsyslog/rsyslog/blob/v8.22.0/runtime/rsyslog.h#L150-L152
[6]: https://github.com/rsyslog/rsyslog/blob/v8.22.0/runtime/stream.c#L264
[7]: https://github.com/rsyslog/rsyslog/blob/v8.22.0/platform/redhat/rsyslog.conf#L45

crosbymichael:
systemd uses it

https://github.com/systemd/systemd/pull/4262

mrunalp:
Rebased.

wking:
On Wed, Oct 19, 2016 at 10:09AM -0700, Michael Crosby wrote:
> systemd/systemd#4262

Ah, thanks :).  In that issue, Lennart links to [1] which has:

  If you write a container solution, please consider supporting the
  following interfaces.

  Execution Environment

  …

  3. Pre-mount /dev as (container private) tmpfs for the container and
     bind mount some suitable TTY to /dev/console.  Also, make sure to
     create device nodes for /dev/null, /dev/zero, /dev/full,
     /dev/random, /dev/urandom, /dev/tty, /dev/ptmx in /dev…

  …

  5. If systemd detects it is run in a container it will spawn a
     single shell on /dev/console, and not care about VTs or multiple
     gettys on VTs. (But see $container_ttys below.)

  …

  Environment Variables

  …

  3. systemd can automatically spawn login gettys on additional
     ptys. A container manager can set the $container_ttys=
     environment variable for the container's PID 1 to tell it on
     which ptys to spawn gettys. The variable should take a space
     seperated list of pty names, without the leading "/dev" prefix,
     but with the "pts/" prefix included. Note that despite the
     variable's name you may only specify ptys, and not other types of
     ttys. Also you need to specify the pty itself, a symlink will not
     suffice. This is implemented in systemd-getty-generator(8). Note
     that this variable should not include the pty that /dev/console
     maps to if it maps to one (see below). Example: if the container
     receives "container_ttys=pts/7 pts/8 pts/14" it will spawn three
     additionally login gettys on ptys 7, 8 and 14.

  …

  What You Shouldn't Do

  …

  3. Do not cross-link /dev/kmsg with /dev/console. They are different
     things, you cannot link them to each other.

The “single shell on /dev/console” bit is similar to the kernel's “on
which logins should be permitted in single-user mode” [2].  This usage
also couples /dev/console to process.terminal.  I don't think that
coupling is strong enough to lean on it here though, because the
syslog use-case is *not* coupled to terminal stdio for the container
process (when configured to use /dev/console, syslog would be
explicitly opening it for logging).  I'm not familiar enough with
systemd, and there are too many /dev/console hits in its code for me
to find out quickly, but it's possible that you'd want to run systemd
with or without terminal stdio too.

So I recommend we add a new console boolean (linux.console?  Maybe
also solaris.console?) to tell the runtime whether or not it needs to
provide /dev/console.  The runtime would be free to provide that
/dev/console however it likes, as long as it provides a terminal.
runC would probably be:

* Creating a single pseudoterminal pair if either process.terminal or
  linux.console was set.
* Exposing the slave at /dev/console if linux.console was set (via a
  bind mount).
* Duping the slave onto the container process's stdio if
  process.terminal was set.

Does that sound workable?

[1]: https://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/
[2]: https://www.kernel.org/doc/Documentation/devices.txt

wking:
On Fri, Oct 21, 2016 at 10:19:14AM -0700, Aleksa Sarai wrote:
> @wking You mentioned in the weekly call that you'd put a summary of the discussion here? Would you mind doing so?

Sorry, GitHub seems to be eating my emailed comments.  Posted summary was for #513, and I just re-posted via the web-UI [here][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/513#issuecomment-255464428

tianon:
@cyphar is the wording in this PR accurate in your opinion, or should it be adjusted?

IMO it sounds like it's an improvement over current `master` as-is

cyphar:
Yeah, this is overall an improvement. This overall LGTM, we can deal with whether we should remove some of the restrictions later.

tianon:
LGTM

Thanks for confirming! :heart:

hqhq:
LGTM

Issue #69, opened 2015-07-12
Expand the list of valid instruction set specifiers


petrosagg:
Currently the spec states that the `arch` property in `config.json` should conform with the Go language spec. However, this spec only covers the following architectures: `amd64`, `386` and `arm`. 

Since one of the goals of the Open Container Project is to support multiple architectures, and given the fact that, at least for ARM [1], it is very common to have different versions and different extensions depending on the SoC, the current list is insufficient.

Common architectures include `armv5`, `armv6` and `armv7a`. The list could be further expanded by adding  `armv5`, `mips` and `powerpc`. Another dimention that should be put into consideration is whether or not the architecture features a floating point unit, which will require architecture specifiers such as `armv7ahf`, `armv6hf`, `mipsel`etc.

A good source of conventions can be found in the source code of the Open Embedded project [2]. Doing a quick grep for available architecture specifiers (`AVAILTUNES`) yields the following list:

[...]

Issue #39, opened 2015-06-30
byte order info


glevand:
I didn't see any way to know the byte order (endian) of the container image other than inspecting binaries in the image.  Byte order info should be in the config file so that a system can determine compatibility with only the config, and not need to download the entire container image when testing.

Maybe this is a shortcoming of using GOARCH as the set of values for platform:arch.

stevvooe:
@glevand Could you give an example of where this may matter but wouldn't be covered by the platform/architecture? For example, is `ppc64le` not sufficient for the architecture?

glevand:
@stevvooe I think the issue is that the spec requires a platform have the go language ported to it (to have a GOARCH value).  Is that what we want, a dependency on the go language? If so, that should be stated in the spec since it is a significant dependency.  Looking at the [FAQ] (https://www.opencontainers.org/faq), it states 'Container formats and runtime should not be bound to clients, to higher level frameworks...'.  I think the go language could be considered a high level framework.

What prompted me to open this issue was support for aarch64_be, but I now see the problem is that the spec requires go language support, which aarch64_be does not have.

stevvooe:
@glevand Understood.

I agree that we shouldn't be dependent on go language support. However, we should follow the conventions for `GOARCH`/`GOOS` since they have done a thorough, modern job of capturing this variation. The ease of cross-compilation with Go is evidence of this.

Either way, my question still stand. At this point, you've only pointed out an inconsistency but we really need to understand why it matters. __Could you give an example of where this may matter but wouldn't be covered by the platform/architecture?__ What specific problem are you trying to solve that is not covered by architecture?

jonboulle:
FWIW we went through a similar discussion in appc and ended up deciding against using raw GOARCH values (`arm` or `arm64`) for ARM architectures since they don't offer the necessary granularity (you need to combine it with GOARM but even that doesn't get you all the way). We instead corresponded with the folks at ARM Holdings and used their suggestions for the most appropriate values - which for the most part turns out to be the equivalent of `utsname.machine` on Linux - https://github.com/appc/spec/blob/master/schema/types/labels.go#L24

stevvooe:
[...]

PR #527, opened 2016-08-10
Add language for compliance requirements around platforms

mrunalp:

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,13 @@ Table of Contents
 
 In the specifications in the above table of contents, the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 
-An implementation is not compliant if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
-An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols is said to be "unconditionally compliant".
+An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
+An implementation is compliant for a given CPU architecture if it satisfies all the MUST and REQUIRED requirements for the protocols it implements.
+An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols on a given CPU architecture is said to be "unconditionally compliant" with those protocols and architectures.
+
+Protocols defined by this specification are:
+* Linux containers: [runtime.md](runtime.md), [config.md](config.md), [config-linux.md](config-linux.md), and [runtime-linux.md](runtime-linux.md).
+* Solaris containers: [runtime.md](runtime.md), [config.md](config.md), and [config-solaris.md](config-solaris.md).
 
 # Use Cases
 
--- a/config-solaris.md
+++ b/config-solaris.md
@@ -1,7 +1,6 @@
 # Solaris Application Container Configuration
 
 Solaris application containers can be configured using the following properties, all of the below properties have mappings to properties specified under zonecfg(1M) man page, except milestone.
-The Solaris specification is entirely optional.
 
 ## milestone
 The SMF(Service Management Facility) FMRI which should go to "online" state before we start the desired process within the container.


wking:
I'd rather not use both “one of the platforms” and “for the protocols
it implements”.  If we want to split this spec into “protocols” we
should stick to the protocols language with the new addition.  If we
want to switch to “platforms” language, we should update the existing
compliance language to use “platforms”.

The “platforms” phrasing is nice and specific for the current spec,
but discussion in #513 included “what if we support ‘at least one of
the following APIs’?” [1], so we may want to stick with the
less-specific “protocols” phrasing to cover both platforms and APIs.
In that case I think it makes sense to make the mapping between
protocols and MUST/SHOULD/… requirements clear.  For example:

  An implementation is not compliant if it fails to satisfy one or
  more of the MUST or REQUIRED requirements for the protocols it
  implements.  An implementation that satisfies all the MUST or
  REQUIRED and all the SHOULD requirements for its protocols is said
  to be "unconditionally compliant".  Protocols defined by this spec
  are:

  * Linux containers: runtime.md, config.md, config-linux.md, and
    runtime-linux.md.
  * Solaris containers: runtime.md, config.md, and config-solaris.md.

And once we land an API spec, it would look something like:

  * Linux containers: runtime.md, config.md, config-linux.md,
    runtime-linux.md, and an API.
  * Solaris containers: runtime.md, config.md and config-solaris.md,
    and an API.

  APIs include at least one of:

  * The [command line API](command-line-api.md) 

[1]: https://github.com/opencontainers/runtime-spec/pull/513/files#r72701237

mrunalp:
@wking I like your wording but we want to avoid repeating the line for each platform if we can. The only thing that will be different is config-platform.md.

wking:
On Wed, Aug 10, 2016 at 01:40:45PM -0700, Mrunal Patel wrote:
> The only thing that will be different is config-platform.md. 

That's not true now, although it would be if you can roll
runtime-linux.md into other places.  #513 moves the standard-streams
stuff into the command line API and you could move the dev links into
config-linux.md.  If we do get a consistent pattern for which files
are part of which “protocols”, I'm fine just explaining the pattern
instead of listing them all out.

mikebrow:
I like most of @wking 's wording but not the part about conditionally vs unconditionally compliant.

philips:
As discussed on the call @mrunalp please add language around "system architectures (ARM, x86-64, etc)" as well as OSes.

hqhq:
OSes are platforms, by architecture, we usually mean x86_64, ARM etc.. I think we need language about architectures.

mrunalp:
Updated.

README.md
```diff
@@ -25,7 +25,10 @@ Table of Contents
 In the specifications in the above table of contents, the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 
 An implementation is not compliant if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
-An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols is said to be "unconditionally compliant".
+An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols is said to be "unconditionally compliant" on one or more CPU architectures.
```
wking: As it stands, your CPU qualification only covers “unconditionally compliant”, and we want to qualify “compliant” as well.  How about:

> An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
> An implementation is compliant for a given CPU architecture if it satisfies all the MUST and REQUIRED requirements for the protocols it implements.
> An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols on a given CPU architecture is said to be "unconditionally compliant" with those protocols and architectures.

which also covers “compliant” explicitly  (it was previously implied to be “all cases that aren't explicitly non-compliant”).

README.md
```diff
@@ -25,7 +25,10 @@ Table of Contents
...
+Protocols defined by this specification are:
+* Linux containers: runtime.md, config.md, config-linux.md, and runtime-linux.md.
+* Solaris containers: runtime.md, config.md, and config-solaris.md.
```
wking: nit: link to the files, e.g. `[runtime.md](runtime.md)`?

README.md
mrunalp: Sounds good. Thanks!

mrunalp:
Pushed update.

wking:
On Wed, Aug 17, 2016 at 10:45:36AM -0700, Mrunal Patel wrote:
> Pushed update. 

6a5b144 looks good to me.

We probably want to remove wiggles like “The Solaris specification is
entirely optional” [1] now that we have this centralized definition of
what requirements belong to which protocols.  I'm fine discussing that
as part of this PR if folks want, or punting it to follow up work.

[1]: https://github.com/opencontainers/runtime-spec/commit/7c9daebaa7abf5d6df2a6b2e6dae55254efbd6d9#diff-54e4235aaffe082955499e11eebda54cR4

mrunalp:

wking:
de3f1af looks good to me too.  I'm not aware of anything else that
should change to match this new compliance language, but we can find
and fix other issues (if any) later on.

hqhq:
LGTM, I thought we might need some CPU architecture specific fields in spec, but we need to figure out what fields and if they exist.

wking:
On Thu, Aug 18, 2016 at 08:14:54PM -0700, Qiang Huang wrote:
> LGTM, I thought we might need some CPU architecture specific fields
> in spec, but we need to figure out what fields and if they exist.

The current PR has “for a given CPU architecture” and similar language
and the spec has a platform.arch field.  What more are you looking
for?  Are you talking about #39 and #69?

README.md
```diff
@@ -24,8 +24,13 @@ Table of Contents
...
 
-An implementation is not compliant if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
-An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols is said to be "unconditionally compliant".
+An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
+An implementation is compliant for a given CPU architecture if it satisfies all the MUST and REQUIRED requirements for the protocols it implements.
+An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols on a given CPU architecture is said to be "unconditionally compliant" with those protocols and architectures.
+
+Protocols defined by this specification are:
+* Linux containers: [runtime.md](runtime.md), [config.md](config.md), [config-linux.md](config-linux.md), and [runtime-linux.md](runtime-linux.md).
+* Solaris containers: [runtime.md](runtime.md), [config.md](config.md), and [config-solaris.md](config-solaris.md).
```
wking: Thinking over the tangentially-related #536, we probably want to include bundle.md and glossary.md in both lists.

I think including glossary.md is unfortunate; perhaps we can revisit the [decision to put the UTF-8 requirement for configuration JSON there][1]?

[1]: https://github.com/opencontainers/runtime-spec/pull/146#issuecomment-138970417

hqhq:
> The current PR has “for a given CPU architecture” and similar language
and the spec has a platform.arch field.  What more are you looking
for?  Are you talking about #39 and #69?

Yeah, something like that, and I think the problem described in #39 is not resolved yet, GOARCH from golang is not sufficient for all architectures, maybe we should reopen it and discuss more.

wking:
On Fri, Aug 26, 2016 at 08:12:09AM -0700, Qiang Huang wrote:
> … I think the problem described in #39 is not resolved yet, GOARCH
> from golang is not sufficient for all architectures…

#441 brought over some language from opencontainers/image-spec#60 that
says “If you need something that's not in GOARCH, please let us know”.
I think that allows us to patch holes in GOARCH as they arise.

hqhq:
@wking Maybe for architecture itself, GOARCH would be sufficient, but we might need some information like byte order, variant, abi etc.. which might be considered out of architecture name, so argue with GOARCH may not be a wise move.

wking:
On Fri, Aug 26, 2016 at 08:45:34AM -0700, Qiang Huang wrote:
> … so argue with GOARCH may not be a wise move.

Maybe not, I don't know.  But I still think the “for a given CPU
architecture” covers the current non-OS portion of ‘platform’.
If/when we land language for platform.byte-order (or whatever), we can
amend the architecture-ish scoping here.

vbatts:
LGTM

PR #530, opened 2016-08-13
config: Document 'recursive' and 'bind' mount options extensions

wking:
The are not filesystem types, so they don't belong in `type`.  The specs claim mount(2) as inspiration for this modeling (which makes sense, since that's the syscall Linux runtimes will make to implement it), and there (recursive) bind is represented by mountflags (`MS_REC | MS_BIND`).  Currently the `options` property handles both mount(2)'s mountflags and data, so `options` is a good spot for these two settings.

We may eventually add entries for other mount(8) command-line options (e.g. `--move`, `--make-shared`, …), but I've left those off until someone pitches a motivational use case for them.

The example I'm touching used:

    "type": "bind",
    …
    "options": ["rbind", ...]

but I don't see a point to putting `bind` in `type` when it's not a filesystem type and you already have `rbind` in `options`.  We could have stuck closer to that example with an unset type and:

    "options": ["rbind", ...]

and that would have been closer to mount(8)'s `--rbind` API, but I've gone with `recursive` here to stay closer to mount(2).  This will scale better if/when we eventually add options for `MS_SLAVE`, etc.

Since there are existing consumers using the old example format and similar things like:

    $ git log --oneline -1 | cat
    03e8b89 Merge pull request #176 from hmeng-19/set_oom_score_adj
    $ ./ocitools generate --template <(echo '{}') --bind ab:cd:ro | jq '.mounts[0]'
    {
      "destination": "cd",
      "type": "bind",
      "source": "ab",
      "options": [
        "bind",
        "ro"
      ]
    }

this may be a breaking change for some spec consumers (although that ocitools example will still work, because `options` contains `bind`, which means the `type` will be ignored).  But even if there are broken consumers, we're still pre-1.0, the spec never explained what `bind`/`rbind` meant before this commit, and consolidating the Linux mount spec around mount(2) now will make life less confusing in the future.

Spun off from [here][1] and [here][2].

[1]: https://github.com/opencontainers/ocitools/pull/119#discussion_r68279630
[2]: https://github.com/opencontainers/runtime-spec/pull/443#issuecomment-234463900

### DIFF
--- a/config.md
+++ b/config.md
@@ -41,17 +41,24 @@ Each container has exactly one *root filesystem*, specified in the *root* object
 
 You MAY add array of mount points inside container as `mounts`.
 The runtime MUST mount entries in the listed order.
-The parameters are similar to the ones in [the Linux mount system call](http://man7.org/linux/man-pages/man2/mount.2.html).
+The parameters are similar to the ones in [the Linux mount system call][mount.2].
 
 * **`destination`** (string, required) Destination of mount point: path inside container.
   For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
-* **`type`** (string, required) The filesystem type of the filesystem to be mounted.
-  Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+* **`type`** (string, optional) The type of filesystem to mount.
+  If `type` is unset, the runtime MAY ask the kernel to guess the desired type.
+  Depending on the mount `options`, the `type` field MAY be ignored.
+  For example, `type` is ignored when `options` contains `bind`; see the `MS_BIND` description in [mount(2)][mount.2].
+  Linux: [*filesystemtype*][mount.2] values supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
   Windows: ntfs.
 * **`source`** (string, required) A device name, but can also be a directory name or a dummy.
   Windows: the volume name that is the target of the mount point, \\?\Volume\{GUID}\ (on Windows source is called target).
 * **`options`** (list of strings, optional) Mount options of the filesystem to be used.
   Linux: [supported][mount.8-filesystem-independent] [options][mount.8-filesystem-specific] are listed in [mount(8)][mount.8].
+  Linux runtimes MUST also support the following options:
+
+    * `recursive`, with the semantics of [`MS_REC`][mount.2].
+    * `bind`, with the semantics of [`MS_BIND`][mount.2].
 
 ### Example (Linux)
 
@@ -65,9 +72,8 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
     },
     {
         "destination": "/data",
-        "type": "bind",
         "source": "/volumes/testing",
-        "options": ["rbind","rw"]
+        "options": ["recursive", "bind", "rw"]
     }
 ]
 ```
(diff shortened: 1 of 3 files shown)


hqhq:
LGTM

config.md
```diff
@@ -65,9 +72,8 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
     },
     {
         "destination": "/data",
-        "type": "bind",
         "source": "/volumes/testing",
-        "options": ["rbind","rw"]
+        "options": ["recursive", bind", "rw"]
```
cyphar: Quote missing on `bind`.

config.md
```diff
@@ -41,17 +41,24 @@ Each container has exactly one *root filesystem*, specified in the *root* object
...
+The parameters are similar to the ones in [the Linux mount system call][mount.2].
 
 * **`destination`** (string, required) Destination of mount point: path inside container.
   For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
-* **`type`** (string, required) The filesystem type of the filesystem to be mounted.
-  Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+* **`type`** (string, optional) The type of filesystem to mount.
+  If `type` is unset, the runtime MAY ask the kernel to guess the desired type.
+  Depending on the mount `options`, the `type` field MAY be ignored.
+  For example, `type` is ignored when `options` contains `bind` or `rbind`; see the `MS_BIND` description in [mount(2)][mount.2].
```
cyphar: Shouldn't this be `recursive` and `bind`? Or am I misunderstanding what the purpose of the change is?

wking:
I've pushed 1ef5938 → 05b3277 which:

* Fixed a missing quote around ‘bind’ [1].
* Removed a stale ‘rbind’ reference [2].

[1]: https://github.com/opencontainers/runtime-spec/pull/530#discussion_r76999330
[2]: https://github.com/opencontainers/runtime-spec/pull/530#discussion_r76999467

config.md
wking: On Wed, Aug 31, 2016 at 07:37:32AM -0700, Aleksa Sarai wrote:
> > +  For example, `type` is ignored when `options` contains `bind` or `rbind`; see the `MS_BIND` description in [mount(2)][mount.2].
>
> Shouldn't this be `recursive` and `bind`? Or am I misunderstanding
> what the purpose of the change is?

It shouldn't be mentioning ‘rbind’, so I've removed that with 1ef5938
→ 05b3277.  I'm sure ‘type’ is ignored for MS_BIND.  I'm less sure
it's always ignored for MS_REC, so I don't mention ‘recursive’ here in
the 05b3277 version of this line.  I can't think of a case where you'd
specify MS_REC *and* a filesystem type, but we only need a single
example of ignoring ‘type’ here and don't need to be exhaustive.

crosbymichael:
NOT LGTM

`recursive` is not even an option, @wking just made it up and added it to a doc.  `MS_REC` makes no sense what-so-ever outside of a bind mount and that is why we, `mount`, and everyone else in the world supports `rbind` and `bind` not `recursive` and `recursive,bind`. 

Checking the mount options for `bind` or `rbind` is fine if the `type` is empty and is a good idea.  But this does not require a spec change but can be updated in the runtime code.  Its not critical to spec the change because passing bind as the fstype/device in the mount syscall does not cause any issues and is ignored.

This is getting old, maintainers, please watch what you approve.

wking:
On Fri, Oct 28, 2016 at 2:06PM -0700, Michael Crosby wrote:
> MS_REC makes no sense what-so-ever outside of a bind mount and that
> is why we, mount, and everyone else in the world supports rbind and
> bind not recursive and recursive,bind.

And mount(8) also has --make-rshared, --make-rslave, --make-rprivate,
and --make-runbindable, all of which use MS_REC [1] (and I pointed out
the MS_SLAVE case in my topic post).  So I still think a separate
‘recursive’ option is better.

However, if the maintainer consensus is that separate recursive
versions (rbind now and rshared, rslave, … later) are the way to go,
I'll reroll this PR to make that happen.  Were there any other
concerns with this PR?  Should I open a new one that goes down the r*
path?  What I want to avoid is widespread use (including in
runtime-tools, opencontainers/runtime-tools#119) of an undefined
(in the nasal demons sense) mount option.

[1]: https://git.kernel.org/cgit/utils/util-linux/util-linux.git/tree/libmount/src/optmap.c?h=stable/v2.28#n122

PR #532, opened 2016-08-17
Restore hook language removed by create/start split

wking:
Details in the commit messages.

Instead of making these changes in a fresh commit off master:

```
o---o---o  master
         \
          o  tk/restore-hook-lifecycle
```

I've made them in two merges so `git blame` can tell that some lines come from historical commits, and that I'm just pulling them out of the history.  The structure of this branch is:

```
a---c---o---o  master
 \ x         \
  b-d---------e  tk/restore-hook-lifecycle
```

where `b` is #384, `c` is #384 landing, and `d` and `e` are new for this PR.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -52,22 +52,33 @@ The lifecycle describes the timeline of events that happen from when a container
 
 1. OCI compliant runtime's [`create`](runtime.md#create) command is invoked with a reference to the location of the bundle and a unique identifier.
 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
-   If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST generate an error.
+   If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST [generate an error](#errors).
    While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified program (from [`process`](config.md#process)) MUST NOT be run at this time.
    Any updates to [`config.json`](config.md) after this step MUST NOT affect the container.
 3. Once the container is created additional actions MAY be performed based on the features the runtime chooses to support.
    However, some actions might only be available based on the current state of the container (e.g. only available while it is started).
 4. Runtime's [`start`](runtime.md#start) command is invoked with the unique identifier of the container.
-   The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
-5. The container process exits.
+5. The [prestart hooks](config.md#prestart) MUST be invoked by the runtime.
+   If any prestart hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 10.
+6. The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
+7. The [poststart hooks](config.md#poststart) MUST be invoked by the runtime.
+   If any poststart hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
+8. The container process exits.
    This MAY happen due to erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
-6. Runtime's [`delete`](runtime.md#delete) command is invoked with the unique identifier of the container.
-   The container MUST be destroyed by undoing the steps performed during create phase (step 2).
+9. Runtime's [`delete`](runtime.md#delete) command is invoked with the unique identifier of the container.
+10. The container MUST be destroyed by undoing the steps performed during create phase (step 2).
+11. The [poststop hooks](config.md#poststop) MUST be invoked by the runtime.
+    If any poststop hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
 
 ## <a name="runtimeErrors" />Errors
 
 In cases where the specified operation generates an error, this specification does not mandate how, or even if, that error is returned or exposed to the user of an implementation.
 Unless otherwise stated, generating an error MUST leave the state of the environment as if the operation were never attempted - modulo any possible trivial ancillary changes such as logging.
+
+## <a name="runtimeWarnings" />Warnings
+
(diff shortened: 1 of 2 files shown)


runtime.md
```diff
@@ -54,17 +56,28 @@ The lifecycle describes the timeline of events that happen from when a container
 3. Once the container is created additional actions MAY be performed based on the features the runtime chooses to support.
    However, some actions might only be available based on the current state of the container (e.g. only available while it is started).
 4. Runtime's `start` command is invoked with the unique identifier of the container.
-   The runtime MUST run the user-specified code, as specified by [`process`](config.md#process-configuration).
-5. The container's process is stopped.
+5. The [prestart hooks](config.md#prestart) MUST be invoked by the runtime.
```
wking: On IRC, @mrunal [suggested switching steps 4 and 5][1], so the `pre-start` hooks fire at the end of `create` instead of the beginning of `start`.  I think that would be surprising considering the naming (621bdb4 talks about a possible `postcreate` anchor in that position).  But we already have the confusing `poststop` running after `delete` (621bdb4 talks about this too).

Personally, I'd rather drop hooks (#483).  But if we keep them for backwards compat, I can see that updating consumers to go from runtime-executed hooks to orchestrator executed hooks would be a largish change.  However, renaming config fields (`prestart` → `postcreate`) seems like a trivial change, and I'd much rather force pre-1.0 users to pay that cost now (they understood pre-1.0 was unstable) than have to explain to all new users why our hook names make no sense ;).

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-08-17.log.html#t2016-08-17T22:52:49

wking: So to be clear on options for resolving this, we have:

a. Keep this line as it is, with `prestart` firing after `start` was called but before the meat of the `start` action (execing the user-specified code).
b. Shift `prestart` to fire at the end of `create`.
    i. And still call it `prestart`.
    ii. And rename it to `postcreate`.

I'm ok with (a) and (b.ii).  I think (b.i) will have a more impact with users resolving confusion, while (b.ii) will have a very small impact on existing hook consumers who have to rename their config settings.

iamanush:
@wking I'm curious if bringing back the hooks will then enable the VM-based runtime configuration details [1] as a hook?

[1] https://github.com/opencontainers/runtime-spec/pull/405

wking:
On Thu, Aug 18, 2016 at 07:46:01PM -0700, Anush Krishnamurthy wrote:
> @wking I'm curious if bringing back the hooks will then enable the
> VM-based runtime configuration details [1] as a hook?

#405 is currently (38cbfc9d0) adding vm.imagePath, vm.kernel.path,
vm.kernel.parameters, and vm.kernel.initrd.  I don't see how you'd
address any of those via hooks; can you explain the connection you see
in more detail?

iamanush:
@wking I was wondering if #405 can be enabled alternately via hooks.

wking:
On Tue, Aug 23, 2016 at 07:00:32PM -0700, Anush Krishnamurthy wrote:
> @wking I was wondering if #405 can be enabled alternately via hooks.

I still don't see how [1].  Can you be more specific about how you see
it working?  Without something like #405 landing, I think you'd have
to address the #405 use cases with annotations [2] or runtime options
[3] (and not hooks).

[1]: https://github.com/opencontainers/runtime-spec/pull/532#issuecomment-241068054
[2]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-223964664
[3]: https://github.com/opencontainers/runtime-spec/pull/405#issuecomment-224000254

iamanush:
@wking: Sorry about the churn. Got it now. I confused myself.

mrunalp:
Needs rebase.

wking:
On Thu, Jan 05, 2017 at 04:01:22PM -0800, Mrunal Patel wrote:
> Needs rebase. 

Recreated the origin/master merge with 8e8454c → 31d8883.

runtime.md
```diff
@@ -13,6 +13,8 @@ The state of a container MUST include, at least, the following properties:
 * **`id`**: (string) is the container's ID.
 This MUST be unique across all containers on this host.
 There is no requirement that it be unique across hosts.
+The ID is provided in the state because hooks will be executed with the state as the payload.
```
crosbymichael: we don't need to hijack the id's description to tell someone why its here because of another feature.  It is enable to document that in the hooks description that the container's state is given to the hook as it's payload.

wking: This line was just [recovered from the graveyard][1].  It sounds like you want it to stay dead, so I'll drop it when I rebase.

[1]: https://github.com/opencontainers/runtime-spec/commit/be594153b522f52bebd0500cef6fe0f1d77f6a59#diff-b84a8d65d8ed53f4794cd2db7e8ea731L16

wking:
I've pushed 31d8883 → f0835e9, dropping the state ‘id’ context [1],
removing the merges that enabled ‘git blame’ [2], picking up
@jlbutler's suggested MUST additions [3], and tightening the lifecycle
wording in the config.md entries.

[1]: https://github.com/opencontainers/runtime-spec/pull/532#discussion_r99232480
[2]: https://github.com/opencontainers/runtime-spec/pull/532#pullrequestreview-19904062
[3]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-02.log.html#t2017-02-02T22:56:45

crosbymichael:
@wking can you rebase this today?

wking:
Rebased onto master with f0835e9 → 17165f8.

crosbymichael:
LGTM

mrunalp:
Needs rebase again :/

wking:
On Fri, Mar 03, 2017 at 09:42:53AM -0800, Mrunal Patel wrote:
> Needs rebase again :/

Rebased again (and made some “generate an error” instances links) with 17165f8 → f636808.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #533, opened 2016-08-18
schema/config-linux: add min/max limit for resource memory and cpu

Mashimiao:

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -129,11 +129,28 @@
                             },
                             "period": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/cpu/period",
-                                "$ref": "defs.json#/definitions/uint64Pointer"
+                                "oneOf": [
+                                    {
+                                        "type": "integer",
+                                        "minimum": 1000,
+                                        "maximum": 1000000
+                                    },
+                                    {
+                                        "type": "null"
+                                    }
+                                ]
                             },
                             "quota": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/cpu/quota",
-                                "$ref": "defs.json#/definitions/int64Pointer"
+                                "oneOf": [
+                                    {
+                                        "type": "integer",
+                                        "minimum": 1000
+                                    },
+                                    {
+                                        "type": "null"
+                                    }
+                                ]
                             },
                             "realtimePeriod": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/cpu/realtimePeriod",
@@ -198,7 +215,16 @@
                             },
                             "swappiness": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/swappiness",
-                                "$ref": "defs.json#/definitions/uint64Pointer"
+                                "oneOf": [
+                                    {
+                                        "type": "integer",
+                                        "minimum": 0,
+                                        "maximum": 100
+                                    },
+                                    {
+                                        "type": "null"
+                                    }
+                                ]
                             }
                         }
                     },


schema/config-linux.json
```diff
@@ -170,11 +170,14 @@
                             },
                             "period": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/cpu/period",
-                                "$ref": "defs.json#/definitions/uint64Pointer"
+                                "type": "integer",
+                                "minimum": 1000,
+                                "maximum": 1000000
```
cyphar: Where do these magic numbers come from? If they come from the kernel, please document them in the `config-linux.md` file.

schema/config-linux.json
```diff
@@ -239,7 +242,9 @@
                             },
                             "swappiness": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/swappiness",
-                                "$ref": "defs.json#/definitions/uint64Pointer"
+                                "type": "integer",
+                                "minimum": 0,
+                                "maximum": 100
```
cyphar: The minimum is `-1` for `swappiness`.

schema/config-linux.json
```diff
@@ -170,11 +170,14 @@
                             },
                             "period": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/cpu/period",
-                                "$ref": "defs.json#/definitions/uint64Pointer"
```
cyphar: These absolutely need to be `uint64Pointer`s. The reason for this is explained (in great detail) in #233.

schema/config-linux.json
dqminh: https://github.com/torvalds/linux/blob/01ea443982203fcdee1250ab630ab6516f0a16e6/kernel/sched/core.c#L8326-L8327

schema/config-linux.json
wking: On Thu, Aug 18, 2016 at 02:41:21AM -0700, Aleksa Sarai wrote:
> > @@ -239,7 +242,9 @@
> >                              },
> >                              "swappiness": {
> >                                  "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/swappiness",
> > -                                "$ref": "defs.json#/definitions/uint64Pointer"
> > +                                "type": "integer",
> > +                                "minimum": 0,
> > +                                "maximum": 100
> 
> The minimum is `-1` for `swappiness`.

The Go type is unsigned, so we're blocking -1 there.  More discussion
on this issue, whether we need to allow -1, and potential alternatives
in #499.

schema/config-linux.json
Mashimiao: Sorry, I read #233 in detail. But I think `uint64Pointer` can't satisfy the special requirements for swappiness, period and quota.
I updated PR to support `null` type for nil pointer.

vbatts:
@cyphar thoughts on this now?

tianon:
friendly ping @cyphar :heart: :pray:

cyphar:
I'm confused what this will look like as an actual Go structure.

wking:
On Fri, Nov 04, 2016 at 08:01:49PM -0700, Aleksa Sarai wrote:
> I'm confused what this will look like as an actual Go structure.

They'll be *uint64, just like they are now.  The current uint64Pointer
definition is just like what @Mashimiao is adding here [1], except
with different limits on the integer case [2].

[1]: https://github.com/opencontainers/runtime-spec/blob/41ef05e714f2e629796664aae640c777b2d34e9a/schema/defs.json#L69-L78
[2]: https://github.com/opencontainers/runtime-spec/blob/41ef05e714f2e629796664aae640c777b2d34e9a/schema/defs.json#L39-L43

crosbymichael:
How is this even used at the system level?

wking:
On Mon, Nov 07, 2016 at 09:16AM -0800,, Michael Crosby wrote:
> How is this going to used at the system level? 

Is that “who consumes the JSON Schema”?  There's currently `schema/validate.go` in this repo which lets anyone check their JSON against the official schema.  And I've floated opencontainers/runtime-tools#197 to lean on those JSON Schemas for runtime-tools' validation.  I'd expect folks building a new config or checking a config they've just received from someone else will want to validate that config before attempting to launch a container from it.

Mashimiao:

crosbymichael:
This would be validating system settings.  My feeling is that we don't need to spec out the kernel, its not our job.

REJECT 

from me

wking:
> This would be validating system settings. My feeling is that we don't need to spec out the kernel, its not our job.

I don't think we're ready for a hard line on that issue, since we currently have a fair amount of in-spec kernel documentation (e.g. [here][1], [here][2], [here][3], …).  But maybe @jlbutler will get us pointed more clearly in his reroll of #673.

And if we want to punt docs to the kernel, we should probably require that `linux.resources.cpu.period` have the same semantics as [`cpu.cfs_period_us`][4].  Otherwise it is not clear what we're punting to.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc3/config-linux.md#L87
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc3/config-linux.md#L115
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc3/config-linux.md#L216
[4]: https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt

Mashimiao:
@crosbymichael  we also validate range of blkio.weight in schema

mrunalp:
Agree with @crosbymichael, Let's not add limits for values.

PR #455, opened 2016-05-25
Makefile: prefer bash

vbatts:
https://github.com/opencontainers/runtime-spec/issues/454

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

Issue #454, opened 2016-05-25
Problem building specification


nuest:
Hi! I'm really interested to get involved and t ostart I wanted to build the spec locally. I am still learning to use make, so I'm having some issues running it. Apologies for the newbie-level questions!

First, on my machine (Ubuntu 16.04 with GNU make 4.1) I have to explicitly set the SHELL variable, e.g. `make docs -e SHELL=bash` (with `sh` i get some unexpected operator errors).
Is that expected? Should I have this in my environment already? I assume this is to allow (power) users to configure their environment, but I just want to be sure.

Second, when I make the docs, I get an error message that docker does not find the image:

```
~/git/runtime-spec$ make docs -e SHELL=bash
mkdir -p output/ && \
/usr/bin/docker run \
-it \
--rm \
-v /home/daniel/git/runtime-spec/:/input/:ro \
-v /home/daniel/git/runtime-spec/output/:/output/ \
-u 1000 \
/usr/bin/pandoc -f markdown_github -t latex -o /output/docs.pdf /input/version.md /input/README.md /input/code-of-conduct.md /input/principles.md /input/style.md /input/ROADMAP.md /input/implementations.md /input/project.md /input/bundle.md /input/runtime.md /input/runtime-linux.md /input/config.md /input/config-linux.md /input/config-solaris.md /input/glossary.md
/usr/bin/docker: Error parsing reference: "/usr/bin/pandoc" is not a valid repository/tag.
See '/usr/bin/docker run --help'.
Makefile:29: recipe for target 'output/docs.pdf' failed
```

There actually is no image name in the created call, and never has been. _What am I missing here?_

Once I got this working I'd be happy to contribute some introductory documentation, if you agree that would be useful.

vbatts:
The `SHELL=bash` is interesting. ubuntu must use `ash` as their `sh`.
As for the primary issue, that should be resolved now. Do a `git pull` and `make docs`.

nuest:
Thanks! Everything works now.

wking:
What operator errors were you seeing with your `sh`?  I don't think we need to do anything Bash-specific in the Makefile.

nuest:
My apologies for the delay. I could reproduce the issue probably by going back to the specific commit. Just let me know if this would be helpful still. Thanks!

PR #534, opened 2016-08-19
Makefile: comment on seemingly unused variable

vbatts:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 
 EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
 OUTPUT_DIRNAME		?= output/
+SHELL			?= $(shell command -v bash 2>/dev/null) # this is due to debain using `ash`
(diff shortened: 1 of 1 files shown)


wking:
SHELL was added in f3fdf03a (Makefile: prefer bash, 2016-05-25, #455)
referencing #454.  I am in favor of a Makefile that only needs a POSIX
shell, but I'd like to understand any POSIX-shell-incompatibilities.
I would have preferred to have that information before we added the
variable [1], but I guess I'm ok removing it, waiting for someone to
reopen #454, and then collecting that information before re-adding
SHELL (or ideally, fixing the POSIX-shell-incompatible command).

[1]: https://github.com/opencontainers/runtime-spec/issues/454#issuecomment-221598064

vbatts:
Oh oh oh
This commit should instead be.... A comment

vbatts:
updated PTAL

wking:
I'd rather just fix the POSIX-compatiblity issue, instead of working
around it.  Fix filed in #542.

Issue #537, opened 2016-08-24
[linux] Tweaking host namespaces?


wking:
Since #158 we've required runtimes to error out if a configuration joins an existing namespace and adjusts it somehow (e.g. joining an existing UTC namespace and setting `hostname`, #214).  However, the wording from #158 (which survives untouched in the current master) only talks about “when a path is specified”.  I expect we want to do one of two things for consistency:

a. Lift the OCI restriction and [allow join-and-tweak][1] where the kernel supports it.  When we landed the current restriction, the main issues seemed to be “[we don't have a clear use-case for join and tweak][2]” (although see #305) and “[this is a foot gun][3]” (I'd rather leave policy to higher-level config linters).

b. Extend the OCI restriction to all cases where the runtime does not create a new namespace.  Besides the already covered “namespace entry exists and includes `path`”, we'd also want to forbid configs that were missing the relevant namespace(s) entirely (in which case the container inherits the host namespace(s)).

I'm partial to (a) in the long run, but (b) is less of a shift from the current spec and likely a better choice for a pending 1.0.  I'm happy to PR text for either option, although the text itself will likely not be much trouble once we decide which course we want to take ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/158#issuecomment-138687129
[2]: https://github.com/opencontainers/runtime-spec/pull/158#issuecomment-138997548
[3]: https://github.com/opencontainers/runtime-spec/pull/158#issuecomment-139106987

crosbymichael:
No this is too split brained.  You either create a new one and fulfill the settings that the user requested or you join a pre-created namespace that is complete.  We don't want a mix of the two.

wking:
On Wed, Aug 24, 2016 at 09:45:51AM -0700, Michael Crosby wrote:
> No this is too split brained.  You either create a new one and
> fulfill the settings that the user requested or you join a
> pre-created namespace that is complete.  We don't want a mix of the
> two.

This sounds like “we should implement (b)” (which I agree makes more
sense for 1.0 [1]).  Or do you have some reason you do want to allow
folks to configure host namespaces?  I'll work up a (b) PR to help
make that change more concrete.

[1]: https://github.com/opencontainers/runtime-spec/issues/537#issue-172878558

crosbymichael:
No we should not touch host namespaces

wking:
On Wed, Aug 24, 2016 at 10:02:04AM -0700, Michael Crosby wrote:
[...]

PR #538, opened 2016-08-24
config-linux: Extend no-tweak requirement to runtime namespaces

wking:
Implement (b) from #537.  Also make it explicit that not listing a
namespace type will cause the container to inherit the runtime
namespace of that type.  More details in the commit message, although
that mostly summarizes the discussion from #537.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #540, opened 2016-08-26
config: Shift Linux-namespace no-tweaking rule to a generic rule

wking:
There are more background references for the Linux-namespaces no-tweaking rule in 01c2d55f (config-linux: Extend no-tweak requirement to runtime namespaces, 2016-08-24, #538).  But the old rule's:

> ... error out if the config specifies anything else related to that namespace.

was overly broad.  For example, it [arguably blocked you from setting network interface priorities for interfaces belonging to an old network namespace](https://github.com/opencontainers/ocitools/pull/200/files#r75907957) even if you were setting those priorities in a new cgroup (because the interfaces and therefore priorities for them are related to the old network namespace).

The new rule tries to apply the spirit of the old rule (“don't touch things that already exist”) more generally so we have a consistent approach that clearly _does_ allow you to configure a new cgroup without having to care about new/old namespaces.

I'm personally fine with join-and-tweak, but the maintainer consensus is that it's [too](https://github.com/opencontainers/runtime-spec/pull/158) [complicated](https://github.com/opencontainers/runtime-spec/issues/537#issuecomment-242132288) to allow (at least for now, some folks [want](https://github.com/opencontainers/runtime-spec/issues/305) [it](https://github.com/opencontainers/runtime-spec/issues/17)).

### DIFF
--- a/config.md
+++ b/config.md
@@ -8,6 +8,22 @@ This includes the process to run, environment variables to inject, sandboxing fe
 
 Below is a detailed description of each field defined in the configuration format.
 
+## Ownership
+
+This specification treats system attributes as belonging to a particular container.
+It does not support one container adjusting attributes that belong to another container.
+Runtimes MUST generate an error if the configuration sets an attribute that belongs to another container.
+
+For example, the following are valid:
+
+* A Linux configuration that creates a new [UTS namespace](config-linux.md#namespaces) and sets [`hostname`](#hostname).
+* A Linux configuration that sets [network limits][config-linux.md#network] with a new [control group][config-linux.md#control-groups], regardless of whether it creates a new [network namespace](config-linux.md#namespaces).
+
+While the following are invalid:
+
+* A Linux configuration that sets [`hostname`](#hostname) but does not create a new [UTS namespace](config-linux.md#namespaces).
+* A Linux configuration that sets [network limits][config-linux.md#network] with an existing [control group][config-linux.md#control-groups].
+
 ## Specification version
 
 * **`ociVersion`** (string, required) MUST be in [SemVer v2.0.0](http://semver.org/spec/v2.0.0.html) format and specifies the version of the Open Container Runtime Specification with which the bundle complies.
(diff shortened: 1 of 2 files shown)


config.md
```diff
@@ -8,6 +8,22 @@ This includes the process to run, environment variables to inject, sandboxing fe
...
+
+For example, the following are valid:
+
+* A Linux configuration that creates a new [UTS namespace](config-linux.md#namespaces) and sets [`hostname`](#hostname).
+* A Linux configuration that sets [network limits][config-linux.md#network] with a new [control group][config-linux.md#control-groups], regardless of whether it creates a new [network namespace](config-linux.md#namespaces).
+
+While the following are invalid:
+
+* A Linux configuration that sets [`hostname`](#hostname) but does not create a new [UTS namespace](config-linux.md#namespaces).
+* A Linux configuration that sets [network limits][config-linux.md#network] with an existing [control group][config-linux.md#control-groups].
```
Mashimiao: How to judge a cgroup is an existing one or a new one?

wking: On Fri, Aug 26, 2016 at 01:24:04AM -0700, Ma Shimiao wrote:
> > +* A Linux configuration that sets [network limits][config-linux.md#network] with an existing [control group][config-linux.md#control-groups].
> 
> How to judge a cgroup is an existing one or a new one?

If cgroupsPath is absolute, you can require no cgroups to exist at
that relative path from the controller mount points [1].  You can also
compare cgroupsPath with values set for other running containers [2],
if you happen to know which other containers are running and which
runtime was used to launch them.  In other cases, external tooling
like ocitools should probably skip this test.

Runtimes creating a container, on the other hand, have an easy way to
tell: they're using a new cgroup if they try to mkdir the leaf group
and it does not return EEXIST or other error.

[1]: https://github.com/opencontainers/runtime-spec/pull/493/files#diff-428eec4013a655816cdefafd5d3505f1R163
[2]: https://github.com/opencontainers/runtime-spec/pull/493/files#diff-428eec4013a655816cdefafd5d3505f1R167

PR #541, opened 2016-08-26
Remove code-of-conduct.md and version.md when clean

hqhq:
So we can use latest files after make clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,4 +114,5 @@ endif
 .PHONY: clean
 clean:
 	rm -rf $(OUTPUT_DIRNAME) *~
+	rm -f code-of-conduct.md version.md
 


wking:
c57a0238 looks good to me.

crosbymichael:
Where do these files come from?

wking:
On Fri, Aug 26, 2016 at 09:53:18AM -0700, Michael Crosby wrote:
> Where do these files come from?

They're built for the docs [1,2].

[1]: https://github.com/opencontainers/runtime-spec/blob/069e8e1c16765e792ddefb32a6ad52e0c70b05ae/Makefile#L58-L62
[2]: https://github.com/opencontainers/runtime-spec/blob/069e8e1c16765e792ddefb32a6ad52e0c70b05ae/Makefile#L58-L62

crosbymichael:
LGTM

vbatts:
LGTM

PR #544, opened 2016-08-29
add timeout for Hook

Mashimiao:

### DIFF
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -41,6 +41,16 @@
             "minimum": 0,
             "maximum": 18446744073709552000
         },
+        "intPointer" : {
+            "oneOf": [
+                {
+                    "type": "integer"
+                },
+                {
+                    "type": "null"
+                }
+            ]
+        },
         "uint16Pointer": {
             "oneOf": [
                 {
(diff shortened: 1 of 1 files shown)


vbatts:
LGTM

schema/defs.json
```diff
@@ -114,6 +114,16 @@
...
+                },
+                "timeout": {
+                    "oneOf": [
+                        {
+                            "type": "integer"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ]
```
wking: We may want to add this `oneOf` as `intPointer` to match our other `*Pointer` definitions.

Mashimiao: fine, fixed.

crosbymichael:
LGTM

crosbymichael:
@vbatts you have to do it again because of the push

wking:
Hmm, should timeout be unsigned?  I can't think of a use-case for
negative timeouts that isn't already covered by the field being
optional/nullable.

vbatts:
@wking integer is fine for now.

vbatts:
LGTM

PR #547, opened 2016-08-29
ci-tools: versions of golang

vbatts:
Add and update golang versions. Also fix install.tools target for
installing govet

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,6 @@
 EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
 OUTPUT_DIRNAME		?= output/
 DOC_FILENAME		?= oci-runtime-spec
-SHELL			?= $(shell command -v bash 2>/dev/null)
 DOCKER			?= $(shell command -v docker 2>/dev/null)
 PANDOC			?= $(shell command -v pandoc 2>/dev/null)
 ifeq "$(strip $(PANDOC))" ''
@@ -63,7 +62,7 @@ version.md: ./specs-go/version.go
 
 HOST_GOLANG_VERSION	= $(shell go version | cut -d ' ' -f3 | cut -c 3-)
 # this variable is used like a function. First arg is the minimum version, Second arg is the version to be checked.
-ALLOWED_GO_VERSION	= $(shell test '$(shell /bin/echo -e "$(1)\n$(2)" | sort -V | head -n1)' == '$(1)' && echo 'true')
+ALLOWED_GO_VERSION	= $(shell test '$(shell /bin/echo -e "$(1)\n$(2)" | sort -V | head -n1)' = '$(1)' && echo 'true')
 
 .PHONY: test .govet .golint .gitvalidation
 
--- a/Makefile
+++ b/Makefile
@@ -68,9 +68,7 @@ ALLOWED_GO_VERSION	= $(shell test '$(shell /bin/echo -e "$(1)\n$(2)" | sort -V |
 
 test: .govet .golint .gitvalidation
 
-# `go get golang.org/x/tools/cmd/vet`
 .govet:
-	@go tool | grep -qw vet || (echo "ERROR: 'go vet' not found. Consider 'make install.tools' target" && false)
 	go vet -x ./...
 
 # `go get github.com/golang/lint/golint`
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@
 EPOCH_TEST_COMMIT	:= 78e6667ae2d67aad100b28ee9580b41b7a24e667
 OUTPUT_DIRNAME		?= output/
 DOC_FILENAME		?= oci-runtime-spec
-SHELL			?= $(shell command -v bash 2>/dev/null)
+SHELL			?= $(shell command -v sh 2>/dev/null)
```
wking: I don't think we need this.  From the [GNU Make docs][1]:

> The program used as the shell is taken from the variable `SHELL`. If this variable is not set in your makefile, the program `/bin/sh` is used as the shell.

Although GNU Make does some dancing for DOS/Windows.  Are you including this line to skip that dance?  If so, why?

[1]: https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html

vbatts: fixed it

Makefile
```diff
@@ -98,17 +97,17 @@ install.tools: .install.golint .install.govet .install.gitvalidation
...
 ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
-	go get github.com/golang/lint/golint
+	go get -u github.com/golang/lint/golint
 endif
 
 # go vet is now included in >=go1.5, so no need to get it.
 .install.govet:
-ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
-	go get golang.org/x/tools/cmd/vet
+ifneq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
```
wking: This looks like it's allowing 1.4 again.  I think we want to check if the Go version is in a whitelisted set, or not in a blacklisted set.

vbatts: But we're not testing against 1.4 in Travis.

wking: On Mon, Aug 29, 2016 at 04:03:25PM -0700, Vincent Batts wrote:
> >  endif
> >  
> >  # go vet is now included in >=go1.5, so no need to get it.
> >  .install.govet:
> > -ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
> > -	go get golang.org/x/tools/cmd/vet
> > +ifneq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
> 
> But we're not testing against 1.4 in Travis.

Travis is not the only Makefile consumer.  If it was, you wouldn't
need this check at all.

vbatts: I'm not seeing where go1.4 is supported. What are you saying? that go1.4 ought to be allowed, but isn't? or that govet now support go1.4?

wking: On Tue, Aug 30, 2016 at 10:08:31AM -0700, Vincent Batts wrote:
> >  endif
> >  
> >  # go vet is now included in >=go1.5, so no need to get it.
> >  .install.govet:
> > -ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
> > -	go get golang.org/x/tools/cmd/vet
> > +ifneq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
>
> I'm not seeing where go1.4 is supported. What are you saying? that
> go1.4 ought to be allowed, but isn't? or that govet now support
> go1.4?

With your change, the Makefile will no longer error out when the host
Go version is 1.4.  But govet does not support Go 1.4, so we still
want to error out in that situation.  Travis will not invoke the
Makefile with a host Go 1.4, but a non-Travis user/developer might, so
we still want to cover the Go 1.4 situation.

vbatts: turns out that it's not needed. govet is not installable even with go1.4. I've removed the install line entirely

vbatts:
PTAL

crosbymichael:
LGTM

Makefile
```diff
@@ -93,22 +92,16 @@ endif
...
 # golint does not even build for <go1.5
 .install.golint:
 ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
-	go get github.com/golang/lint/golint
-endif
-
-# go vet is now included in >=go1.5, so no need to get it.
-.install.govet:
-ifeq ($(call ALLOWED_GO_VERSION,1.5,$(HOST_GOLANG_VERSION)),true)
-	go get golang.org/x/tools/cmd/vet
```
wking: If you drop this, you'll also want to rewrite/remove the line in the `.govet` recipe with the:

> ERROR: 'go vet' not found. Consider 'make install.tools' target

message.

vbatts:
PTAL

wking:
d4ede0d36 looks good to me.

crosbymichael:
LGTM

tianon:
LGTM

PR #548, opened 2016-08-30
unified stringPointer for *string

Mashimiao:

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -290,14 +290,8 @@
                 }
             },
             "cgroupsPath": {
-                "oneOf": [
-                    {
-                        "type": "null"
-                    },
-                    {
-                        "type": "string"
-                    }
-                ]
+                "id": "https://opencontainers.org/schema/bundle/linux/cgroupsPath",
+                "$ref": "defs.json#/definitions/stringPointer"
             },
             "rootfsPropagation": {
                 "id": "https://opencontainers.org/schema/bundle/linux/rootfsPropagation",


schema/config-linux.json
```diff
@@ -204,7 +204,7 @@
                                     "type": "object",
                                     "properties": {
                                         "pageSize": {
-                                            "type": "string"
+                                            "$ref": "defs.json#/definitions/stringPointer"
```
wking: This is a bit sticky, since the Markdown has [`pageSize` as required][1] but the Go [has it as a pointer][2].  I don't see the point of declaring `hugepageLimits` without setting both (in which case I don't think they should be pointers in either Go or JSON Schema), but I haven't read enough of the history to say for sure.

[1]: https://github.com/opencontainers/runtime-spec/blob/90451c4923edfa8162425707c8a496085277267e/config-linux.md#huge-page-limits
[2]: https://github.com/opencontainers/runtime-spec/blob/90451c4923edfa8162425707c8a496085277267e/specs-go/config.go#L198-L199

vbatts:
@Mashimiao do you have a case where the JSON value for the cgrouppath is `null`?

wking:
On Tue, Aug 30, 2016 at 10:33:23AM -0700, Vincent Batts wrote:
> @Mashimiao do you have a case where the JSON value for the
> cgrouppath is `null`?

cgroupsPath is optional [1], and for most (all?) of our optional
settings we treat “unset” and “set to ‘null’” identically.  I'm not
sure you can distinguish between those cases in Go without using
interface{}-based JSON parsing.  For an example of the Markdown spec
explicitly equating “unset” and “set to ‘null’”, see [2].

[1]: https://github.com/opencontainers/runtime-spec/blob/53917150dafa8e56e2aa20b8316103f2d1c4de76/config-linux.md#control-groups
[2]: https://github.com/opencontainers/runtime-spec/blob/53917150dafa8e56e2aa20b8316103f2d1c4de76/config-linux.md#device-whitelist

wking:
On Thu, Sep 29, 2016 at 01:43:56AM -0700, Ma Shimiao wrote:
>                                          "pageSize": {
> -                                            "type": "string"
> +                                            "$ref": "defs.json#/definitions/stringPointer"
> 
> @wking 
> Well, I'm a little confused about types in specs-go.  As you
> mentioned, pagesize is `required`, but it is declared as a pointer
> in GO.

If it's required, I think it should *not* be a pointer in Go, and that
it should stay a string in the JSON Schema here.  It should also be
added to a ‘required’ block in the JSON Schema (but it isn't at the
moment).

> As Namespace's property Path is `optional`, but it is declared as
> string type in GO.

This is less cut and dried.  The current policy is in [1], and for
Namespace.Path I think not having a pointer is fine.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/style.md#optional-settings-should-have-pointer-go-types

Mashimiao:
@wking  @opencontainers/runtime-spec-maintainers  PTAL

wking:
This is obsolete now that #653 has landed.  #656 is also removing the `cgroupsPath` `oneOf` (among other things), so I think we can close this PR.

PR #551, opened 2016-08-30
schema: Run 'make fmt'

wking:
This should probably be part of our CI testing.

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -202,8 +202,7 @@
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/cpu/shares",
                                 "$ref": "defs.json#/definitions/uint64Pointer"
                             }
-                        },
-                        "type": "object"
+                        }
                     },
                     "disableOOMKiller": {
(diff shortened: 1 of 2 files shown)


vbatts:
LGTM

Though CI would just do the formatting. It would need to check that it needs to be formatted

wking:
On Wed, Aug 31, 2016 at 08:17:33AM -0700, Vincent Batts wrote:
> Though CI would just do the formatting. It would need to check that
> it needs to be formatted

In an unrelated project and CI framework, I diff the original with the
auto-formatted version [1].  And there's an earlier test for “is the
syntax valid?” [2].  I can add something like that here (with a for
loop?) or put in in a separate PR if folks are interested.

[1]: https://github.com/azurestandard/api-spec/blob/87b72f8a5f7f31820d54a57be26dc0f7fc8cd434/circle.yml#L4
[2]: https://github.com/azurestandard/api-spec/blob/87b72f8a5f7f31820d54a57be26dc0f7fc8cd434/circle.yml#L3

crosbymichael:
LGTM

PR #553, opened 2016-09-01
[ReadMe] Just one level/tier of compliance (rebase)

RobDolinMS:

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,8 @@ Table of Contents
 
 In the specifications in the above table of contents, the keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 
-An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
-An implementation is compliant for a given CPU architecture if it satisfies all the MUST and REQUIRED requirements for the protocols it implements.
-An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols on a given CPU architecture is said to be "unconditionally compliant" with those protocols and architectures.
+An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST, REQUIRED, or SHALL requirements for the protocols it implements.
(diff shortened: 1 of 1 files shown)


RobDolinMS:
This is a rebase of #543 
/cc @wking @polvi @crosbymichael @mrunalp

README.md
```diff
@@ -24,9 +24,8 @@ Table of Contents
...
 
-An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
-An implementation is compliant for a given CPU architecture if it satisfies all the MUST and REQUIRED requirements for the protocols it implements.
-An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols on a given CPU architecture is said to be "unconditionally compliant" with those protocols and architectures.
+An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST, REQUIRED, or SHALL requirements for the protocols it implements.
+An implementation is compliant for a given CPU architecture if it satisfies all the MUST, REQUIRED, and SHALL requirements for the protocols it implements.
```
duglin: why are we saying the same thing twice, just backwards?
Also, why do we need to say "CPU architecture"? Why not just "An implementation is compliant if it ...."  ?

wking: On Thu, Sep 01, 2016 at 10:59:01AM -0700, Doug Davis wrote:
> > -An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements.
> > -An implementation is compliant for a given CPU architecture if it satisfies all the MUST and REQUIRED requirements for the protocols it implements.
> > -An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols on a given CPU architecture is said to be "unconditionally compliant" with those protocols and architectures.
> > +An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST, REQUIRED, or SHALL requirements for the protocols it implements.
> > +An implementation is compliant for a given CPU architecture if it satisfies all the MUST, REQUIRED, and SHALL requirements for the protocols it implements.
> 
> why are we saying the same thing twice, just backwards?

We could say:

  An implementation is compliant for a given CPU architecture if and only if …

to reduce duplication, but without the “and only if” we need to say it
both ways.

> Also, why do we need to say "CPU architecture"? Why not just "An
> implementation is compliant if it ...."  ?

Compliance is granted for a given architecture (more on this in #527).

RobDolinMS:
@duglin Yes, lines 27 and 28 are similar.

RobDolinMS:
@duglin I kept "CPU architecture" as it was added between my initial commit and this rebase and I wanted this change specifically focused on removing the two tiers/levels of compliance.

vbatts:
LGTM

crosbymichael:
LGTM

PR #555, opened 2016-09-06
schema/config-linux: fix namespace based on runtime spec

Mashimiao:

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -48,14 +48,17 @@
             },
             "namespaces": {
                 "id": "https://opencontainers.org/schema/bundle/linux/namespaces",
-                "type": "array",
-                "items": {
-                    "anyOf": [
-                        {
+                "oneOf": [
+                    {
+                        "type": "array",
+                        "items": {
                             "$ref": "defs-linux.json#/definitions/NamespaceReference"
                         }
-                    ]
-                }
+                    },
+                    {
+                        "type": "null"
+                    }
+                ]
             },
             "resources": {
                 "id": "https://opencontainers.org/schema/bundle/linux/resources",


wking:
I don't see a need to support ‘null’ when callers can use an empty
array or leave the property unset.  And the Go has:

  type Linux struct {
    …
    Namespaces []Namespace `json:"namespaces,omitempty"`
    …
  }

so I expect that will either generate non-empty array or
unset-property JSON.

However, my only concern with allowing ‘null’ is that it makes the
spec and JSON Schema slightly more complicated without providing more
power.  If folks are fine making any optional setting nullable, even
if we don't use a pointer for the Go property [1], I'll go along with
it.  And in that case, 8e592291 looks good to me.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc1/style.md#optional-settings-should-have-pointer-go-types

Mashimiao:
On 09/07/2016 02:40 AM, W. Trevor King wrote:
> ype Linux struct {
> …
> Namespaces []Namespace `json:"namespaces,omitempty"`
> …
> }
> 
> so I expect that will either generate non-empty array or
> unset-property JSON.

That's right.
But, as schema is used to check json file.
`null, []` both can be applied to namespaces in json file and
converted to GO code.
So, I think if setted null to namespaces in json file, we should accept it.

> However, my only concern with allowing ‘null’ is that it makes the
spec and JSON Schema slightly more complicated without providing more
power.  

I agree. 
But I think we should think based on writing json file directly.
If write json file directly, null is an possible case for namespaces.
Unless we appoint it to be invalid.

-- 
Ma Shimiao
Development Dept.I
Nanjing Fujitsu Nanda Software Tech. Co., Ltd.(FNST)

wking:
On Tue, Sep 06, 2016 at 10:46:43PM -0700, Ma Shimiao wrote:
> But I think we should think based on writing json file directly.  If
> write json file directly, null is an possible case for namespaces.
> Unless we appoint it to be invalid.

Right.  I'm saying I prefer making making ‘null’ invalid, because in
the namespaces case there's nothing you say with ‘null’ that you
aren't saying with ‘[]’.  Folks who want to say “use all the host
namespaces” can write ‘[]’.  Also allowing them to write ‘null’
doesn't matter to Go consumers, but makes the JSON Schema slightly
more complicated (this PR) and may also make non-Go implementations
slightly more complicated.

On the other hand, those are small slightlies, so I'm ok with allowing
‘null’ if there is a need to cater to folks who want two ways to say
the same thing ;).

Mashimiao:
any comments?

dqminh:
I agree with not allowing null for an array.

mrunalp:
We discussed this on the call and decided that having null doesn't make sense.

PR #556, opened 2016-09-07
Replace " : " with ": "

wking:
There's an outside change that these are intentional, since [I pointed one of these out earlier][1] and it wasn't fixed.  But I haven't seen ` : ` used intentionally outside of this project, and don't think we want to break ground in that direction ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/510#discussion_r77291554


crosbymichael:
LGTM

vbatts:
LGTM

Issue #509, opened 2016-06-24
Suggestion: clarify Masked and Readonly Paths


Mashimiao:
There is not clearly requirement for Masked or Readonly Paths.
I think we must definitely define they must be absolute path  or not in runtime spec.

wking:
On Fri, Jun 24, 2016 at 02:31:48AM -0700, Ma Shimiao wrote:
> I think we must definitely define they must be absolute path or not
> in runtime spec.

This is related to a general need to specify the anchor for relative
paths [1].  If there is a config.md-wide entry like:

  Relative paths are interpreted relative to the directory holding
  config.json.

Then having relative paths here (or in other places) seems
sufficiently well-defined to me.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-72
     and the next two lines.

hqhq:
Fixed by: https://github.com/opencontainers/runtime-spec/pull/587

PR #558, opened 2016-09-08
config.md: clarify the root filesystem path

Mashimiao:

### DIFF
--- a/config.md
+++ b/config.md
@@ -25,6 +25,8 @@ For example, if an implementation is compliant with version 1.0.1 of the spec, i
 **`root`** (object, required) configures the container's root filesystem.
 
 * **`path`** (string, required) Specifies the path to the root filesystem for the container.
+  The path can be an absolute path (starting with /) or a relative path (not starting with /), which is relative to the bundle.
+  For example (Linux), with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
   A directory MUST exist at the path declared by the field.
 * **`readonly`** (bool, optional) If true then the root filesystem MUST be read-only inside the container, defaults to false.
 


config.md
```diff
@@ -25,6 +25,8 @@ For example, if an implementation is compliant with version 1.0.1 of the spec, i
 **`root`** (object, required) configures the container's root filesystem.
 
 * **`path`** (string, required) Specifies the path to the root filesystem for the container.
+  The path MUST be an a relative path (not starting with /), which is relative to the bundle. e.g.
```
wking: We used to have a relative `path` restriction, but it was (wisely, I think) lifted in #394.  There is in-flight work in #469 to finish off the work that #394 started, but I don't think we want to go back to requiring relative paths here.

wking: However, I think it does make sense to [clarify the base from which relative paths are interpreted][1], in cases where someone does use a relative path here.  I'd recommend doing that with a stand-alone section early in this file, so other relative-path locations can benefit from it.  That would allow us to relax some absolute-path requirements we're still carrying around (e.g. for [namespace paths][2]).

[1]: https://github.com/opencontainers/runtime-spec/issues/509#issuecomment-228414738
[2]: https://github.com/opencontainers/runtime-spec/blob/22c29c18eddc75ccee9f6b202f61524f29653c5f/config-linux.md#namespaces

Mashimiao: Acutally, I think no matter absolute path or relative path is OK.
I just want to clarify which kind of value can be set  to path.

config.md
```diff
@@ -25,6 +25,8 @@ For example, if an implementation is compliant with version 1.0.1 of the spec, i
 **`root`** (object, required) configures the container's root filesystem.
 
 * **`path`** (string, required) Specifies the path to the root filesystem for the container.
+  The path can be an absolute path (starting with /) or a relative path (not starting with /), which is relative to the bundle. e.g.
+  If the absolute path of root filesystem is /to/bundle/rootfs, we can set it or rootfs to`path.
```
vbatts: looks like a dangling backtick `
and i'm not sure i follow the "we can set it or rootfs to path" sentence.

wking: On Thu, Sep 08, 2016 at 07:00:35AM -0700, Vincent Batts wrote:
> > @@ -25,6 +25,8 @@ For example, if an implementation is compliant with version 1.0.1 of the spec, i
> >  **`root`** (object, required) configures the container's root filesystem.
> >  
> >  * **`path`** (string, required) Specifies the path to the root filesystem for the container.
> > +  The path can be an absolute path (starting with /) or a relative path (not starting with /), which is relative to the bundle. e.g.
> > +  If the absolute path of root filesystem is /to/bundle/rootfs, we can set it or rootfs to`path.
> 
> looks like a dangling backtick `
> and i'm not sure i follow the "we can set it or rootfs to path" sentence.

How about:

  For example, with a bundle at `/to/bundle` and a root filesystem at
  `/to/bundle/rootfs`, the `path` value can be either
  `/to/bundle/rootfs` or `rootfs`.

Mashimiao: On 09/09/2016 12:39 AM, W. Trevor King wrote:
> How about: For example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
It looks better than mine. Fixed.

config.md
```diff
@@ -25,6 +25,8 @@ For example, if an implementation is compliant with version 1.0.1 of the spec, i
 **`root`** (object, required) configures the container's root filesystem.
 
 * **`path`** (string, required) Specifies the path to the root filesystem for the container.
+  The path can be an absolute path (starting with /) or a relative path (not starting with /), which is relative to the bundle.
+  For example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
```
hqhq: The example should be explicit for *unix since other platforms like windows don't take `/` as root.

hqhq:
LGTM

vbatts:
LGTM

PR #563, opened 2016-09-14
config: Add consoleSize to process

jhowardmsft:

Extracting pieces from the proof of concept PR for Windows OCI support at https://github.com/opencontainers/runtime-spec/pull/504. This adds the `consoleSize` platform specific field to the `Process` structure.

### DIFF
--- a/config.md
+++ b/config.md
@@ -95,6 +95,9 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 **`process`** (object, REQUIRED) configures the container process.
 
 * **`terminal`** (bool, OPTIONAL) specifies whether you want a terminal attached to that process, defaults to false.
+* **`consoleSize`** (object, OPTIONAL) specifies the console size of the terminal if attached, containing the following properties:
+  * **`height`** (uint, REQUIRED)
+  * **`width`** (uint, REQUIRED)
 * **`cwd`** (string, REQUIRED) is the working directory that will be set for the executable.
   This value MUST be an absolute path.
 * **`env`** (array of strings, OPTIONAL) contains a list of variables that will be set in the process's environment prior to execution.
@@ -139,6 +142,10 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
 ```json
 "process": {
     "terminal": true,
+    "consoleSize": {
+        "height": 25,
+        "width": 80
+    },
     "user": {
         "uid": 1,
         "gid": 1,
@@ -174,6 +181,10 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
 ```json
 "process": {
     "terminal": true,
+    "consoleSize": {
+        "height": 25,
+        "width": 80
+    },
     "user": {
         "uid": 1,
         "gid": 1,
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -31,6 +31,8 @@ type Spec struct {
 type Process struct {
 	// Terminal creates an interactive terminal for the container.
 	Terminal bool `json:"terminal,omitempty"`
+	// ConsoleSize specifies the size of the console.
+	ConsoleSize Box `json:"consoleSize,omitempty"`
 	// User specifies user information for the process.
 	User User `json:"user"`
 	// Args specifies the binary and arguments for the application to execute.
@@ -50,6 +52,14 @@ type Process struct {
 	ApparmorProfile string `json:"apparmorProfile,omitempty" platform:"linux"`
 	// SelinuxLabel specifies the selinux context that the container process is run as.
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
+}
+
+// Box specifies dimensions of a rectangle. Used for specifying the size of a console.
+type Box struct {
+	// Height is the vertical dimension of a box.
+	Height uint `json:"height"`
+	// Width is the horizontal dimension of a box.
+	Width uint `json:"width"`
 }
 
 // User specifies specific user (and group) information for the container process.


config.md
```diff
@@ -118,6 +118,10 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
 * **`noNewPrivileges`** (bool, optional) setting `noNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
 [The kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call.
 
+For Windows-based systems, the process structure supports the following process specific fields:
+
+* **`initialconsolesize`** (array of ints, optional) specifies the initial console size (height, width) of the terminal if attached.
```
philips: capitalization is significant

jhowardmsft: Yes, I deliberately did this as the capitalisation is consistent with other items such as rlimits (vs `rLimits`).  Are those incorrect as well? If so, I'll open another PR to address that, but fix the capitalisation to `initialConsoleSize` here.

wking:
On Wed, Sep 14, 2016 at 01:38:46PM -0700, John Howard wrote:
> Yes, I deliberately did this as the capitalisation is consistent
> with other items such as rlimits (vs `rLimits`).  Are those
> incorrect as well? If so, I'll open another PR to address that, but
> fix the capitalisation to `initialConsoleSize` here.

Touche ;).  I'm fine with rlimits, rLimits, and resourceLimits.  And
consistency is nice, but there's also a benefit to using the
platform's conventional casing [1].  I'm fine letting per-patform
maintainers go their own way on this sort of style decision.

jhowardmsft:
Updated it to camelCase. Given there's almost no `Windows` specific fields in the spec currently, I may as well start with a sensible pattern.

tianon:
LGTM

specs-go/config.go
```diff
@@ -51,6 +51,8 @@ type Process struct {
 	ApparmorProfile string `json:"apparmorProfile,omitempty" platform:"linux"`
 	// SelinuxLabel specifies the selinux context that the container process is run as. (this field is platform dependent)
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
+	// InitialConsoleSize contains the initial h,w of the console size. (this field is platform dependent)
```
wking: Drop “this field is platform dependent” (#568, #569).

jhowardmsft: Yes, of course :)

specs-go/config.go
```diff
@@ -51,6 +51,8 @@ type Process struct {
...
+	InitialConsoleSize [2]int `json:"initialConsoleSize" platform:"windows"`
```
jhowardmsft: Yup. Good catch. Updating - pushing shortly.

jhowardmsft: Done. And the matching doc change ;)

tianon:
LGTM redux

wking:
c86f32c looks good to me, although “of the console size” could
probably be trimmed to “of the console” ;).

anuthan:
LGTM

jhowardmsft:
Grammer police :oncoming_police_car: :smile: @wking - Fixed, thanks for catching.

wking:
On Thu, Sep 15, 2016 at 10:29:25AM -0700, John Howard wrote:
> Grammer police :oncoming_police_car: :smile: @wking - Fixed, thanks for catching.

2a7b7c6aa1 looks even better to me ;).

config.md
```diff
@@ -118,6 +118,10 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
 * **`noNewPrivileges`** (bool, optional) setting `noNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
 [The kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call.
 
+For Windows-based systems, the process structure supports the following process specific fields:
+
+* **`initialConsoleSize`** (array of uints, optional) specifies the initial console size (height, width) of the terminal if attached.
```
tianon: Just realized this is `(height, width)` -- isn't it more common for these particular types of geographic values to be ordered `(width, height)`?  Is this ordered this way arbitrarily, or is this how the underlying platform support orders them?

For example:

image

jhowardmsft: The ordering here is consistent with the way the platform API is implemented.

tianon: :+1: thanks for confirming :heart:

tianon:
LGTM redux

config.md
```diff
@@ -118,6 +118,10 @@ For more information about SELinux, see  [Selinux documentation](http://selinuxp
 * **`noNewPrivileges`** (bool, optional) setting `noNewPrivileges` to true prevents the processes in the container from gaining additional privileges.
 [The kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call.
 
+For Windows-based systems, the process structure supports the following process specific fields:
```
laijs: I think it is also needed in the linux.

jhowardmsft: @laijs Why do you think that? It's needed on Windows to init the conhost terminal, but it's only ever been in Windows to date.

laijs: In linux, we also use a default winsize for the tty console when starting the container, and then we resize it later, these is a gap between them.

jhowardmsft:
Updated to not make this a Windows-specific field after comment https://github.com/opencontainers/runtime-spec/pull/563/files#r79284892 from @laijs

specs-go/config.go
```diff
@@ -31,6 +31,8 @@ type Spec struct {
 type Process struct {
 	// Terminal creates an interactive terminal for the container.
 	Terminal bool `json:"terminal,omitempty"`
+	// InitialConsoleSize contains the initial h,w of the console.
+	InitialConsoleSize [2]uint `json:"initialConsoleSize"`
```
laijs: Hello, @maintainers, how about use a more verbose structure here?

```go
type ConsoleWindowSize struct {
	Height uint `json:"height"`
	Width  uint `json:"width"`
}
```

wking: +1 for explicit height/width.  Although probably no need for the specific name.  `type Box` would give us something we could use in other places, although I can't think of other consumers now.

hqhq: +1 for this change, @jhowardmsft WDYT?

jhowardmsft: Agreed. Updated using `Box`, and updated the docs accordingly

jhowardmsft:
@wking @laijs @hqhq Updated.

config.md
```diff
@@ -95,6 +95,7 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 **`process`** (object, required) configures the container process.
 
 * **`terminal`** (bool, optional) specifies whether you want a terminal attached to that process, defaults to false.
+* **`initialConsoleSize`** (box, optional) specifies the initial console size of the terminal if attached.
```
hqhq: I prefer naming as `consoleWindowSize` as @laijs suggested, because we might want to update process console size and it seems weird to assign new config to the field with `initial` prefixed.

jhowardmsft: @hqhq - @laijs explicitly said this was needed on Linux, so I don't understand your comment? As for `initial`, this is important - it's the how the console is initialised for the console start, so I don't agree with your other suggestion.

hqhq: Yes Linux has tty console size, but the `Window` in the name `consoleWindowSize` doesn't mean `Windows operate system`, it just reflect semantics from http://man7.org/linux/man-pages/man4/tty_ioctl.4.html (section `Get and set window size`).

And I don't think we need to specify `initial` for the usage of initialization, otherwise we'll call `InitialArgs`, `InitialRlimits`, etc.. Not to mention that they can also be used in non-initialization stages.

wking:
On Sat, Sep 17, 2016 at 11:24:02PM -0700, Qiang Huang wrote:
> Not to mention that they can also be used in non-initialization
> stages.

I'm fine dropping initial, but thought I'd point out that the current
spec is not supporter of join-and-tweak.  See the in-flight #540
trying to clarify that position, and the open #305 in favor of
allowing join-and-tweak.  I'm personally in favor of allowing
join-and-tweak, but filed #540 because I'd rather have clearer wording
on the no-tweaking restriction while we want for the consensus to
shift towards allowing join-and-tweak.

hqhq:
@wking Yeah I plan to put this back on table post-1.0, but we'll support it in runc whether it's in spec or not, so we better keep it looks reasonable for update.

jhowardmsft:
Are there not already many fields which are not configurable post create? I don't care though, if consensus is to remove 'initial', so be it. Imo it seems pretty descriptive of the purpose though.

wking:
On Sun, Sep 18, 2016 at 12:34:59AM -0700, John Howard wrote:
> Are there not already many fields which are not configurable post
> create?

It may be helpful to think about this in three classes:

a. Initial setup.  The resource (e.g. a console) was just created and
   needs to be configured.  That's what most/all of the current config
   settings are about.
b. Out-of-band tweaking.  Users can do whatever they want outside of
   the runtime, e.g. dragging their console around to resize it.
c. Runtime tweaking.  A subsequent runtime invocation updates an
   existing resource.  This is what #305 and #540 are talking about.

It's probably good to remind users that most settings are not set in
stone, and that (b) and maybe eventually (c) are possible (one example
where this isn't true is the user map [1], which is “write once and
you're stuck with it” [2]).  But I don't think prefixing properties
with “initial” is the most efficient way to remind them.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config-linux.md#user-namespace-mappings
[2]: http://man7.org/linux/man-pages/man7/user_namespaces.7.html
     > An attempt to write more than once to a uid_map file in a user
     > namespace fails with the error EPERM.  Similar rules apply for
     > gid_map files.

jhowardmsft:
So to be clear before I update. Drop `Initial`, that seems clear. But I'm not sure based on https://github.com/opencontainers/runtime-spec/pull/563#issuecomment-247833453 whether consensus is to have this as a Windows only field in this PR, and open it up to other platforms in the future as and when those platforms have runtimes that make use of the field.

wking:
On Mon, Sep 19, 2016 at 09:12:44AM -0700, John Howard wrote:
> But I'm not sure based on
> https://github.com/opencontainers/runtime-spec/pull/563#issuecomment-247833453
> whether consensus is to have this as a Windows only field in this
> PR, and open it up to other platforms in the future as and when
> those platforms have runtimes that make use of the field.

I wouldn't worry about current runtime support.  Other platforms will
need some size, and Linux currently uses a default that you can't
override from the OCI config [1].  So I'd expose the property on all
platforms and explicitly say that the size is unspecified if the
property is unset.  And we should probably pull in something like
opencontainers/image-spec#311 so we have a clear definition for
“unspecified”.  Runtimes that currently ignore this property will be
noncompliant with new runtime-spec releases until they add support,
but that's true of any new property.  Images that don't set this
property would still be compliant, because the unspecified default is
what they get now.  If we were versioning every change, adding this
property would be a minor-bump to ociVersion.

[1]: https://github.com/opencontainers/runtime-spec/pull/563#r79284892

jhowardmsft:
@wking. Thanks. The current commit f2bc43b63ee01701b130fa8d0c4aa22f6d7fe613 should reflect the changes based on all feedback to date.  

I would be inclined to say that adding something like https://github.com/opencontainers/image-spec/pull/311 belongs in a separate PR rather than polluting this one. I'm happy to put that together later today.

wking:
On Mon, Sep 19, 2016 at 11:02:46AM -0700, John Howard wrote:
> I would be inclined to say that adding something like
> https://github.com/opencontainers/image-spec/pull/311 belongs in a
> separate PR rather than polluting this one. I'm happy to put that
> together later today.

That's fine with me.  Leaving the consoleSize-unset behavior implicit
for an afternoon is fine.  And perhaps the
opencontainers/image-spec#311 port lands first, in which case we can
reroll this to explicitly make the consoleSize-unset case unspecified.

f2bc43b63 looks good to me, except the commit summary should now be:

  config: Add consoleSize to process

jhowardmsft:
Ah yes, of course. Commit summary and PR title updated.

wking:
55956b85 and a punt on porting opencontainers/image-spec#311 looks
good to me.

hqhq:
LGTM

jhowardmsft:
@tianon - sorry, it's gone through a few revisions since your last LGTMs. Any chance of a further review? Thanks ❤️

mrunalp:
LGTM

tianon:
Nice, I like the introduction of the `box` type to avoid having to just "know" what order the arguments go in. :+1:

LGTM
(although appears to now have a conflict in `config.md`)

wking:
On Tue, Sep 20, 2016 at 02:27:34PM -0700, Tianon Gravi wrote:
> (although appears to now have a conflict in `config.md`)

#574 is probably causing a lot of trivial conflicts.

jhowardmsft:
Rebased. Yes, it was #574 (optional --> OPTIONAL)

config.md
```diff
@@ -95,6 +95,7 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 **`process`** (object, REQUIRED) configures the container process.
 
 * **`terminal`** (bool, OPTIONAL) specifies whether you want a terminal attached to that process, defaults to false.
+* **`consoleSize`** (box, OPTIONAL) specifies the console size of the terminal if attached.
```
jhowardmsft: Updated using the first notation from your suggestions.

wking:
a502caf looks good to me (like most of the previous iterations of this
PR ;).

hqhq:
LGTM

mrunalp:
LGTM

PR #564, opened 2016-09-14
Platform annotations on `Process` struct

jhowardmsft:

Extracting pieces from the proof of concept PR for Windows OCI support at https://github.com/opencontainers/runtime-spec/pull/504. This PR tidies up the `Process` structure by putting annotations for platform support on `Capabilities`, `Rlimits` and `NoNewPrivileges`, none of which are used on `Windows`. In addition, it removes what appears to be an erroneous blank line that appeared to serve no purpose.

Note I'm assuming (probably incorrectly :innocent:) , that `Rlimits` and `NoNewPrivileges` are also relevant on `Solaris` too, but I'm not sure. If that is incorrect, I will update the PR to remove the `solaris` annotation.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -43,10 +43,9 @@ type Process struct {
 	// Capabilities are Linux capabilities that are kept for the container.
 	Capabilities []string `json:"capabilities,omitempty" platform:"linux"`
 	// Rlimits specifies rlimit options to apply to the process.
-	Rlimits []Rlimit `json:"rlimits,omitempty"`
+	Rlimits []Rlimit `json:"rlimits,omitempty" platform:"linux"`
 	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container.
-	NoNewPrivileges bool `json:"noNewPrivileges,omitempty"`
-
+	NoNewPrivileges bool `json:"noNewPrivileges,omitempty" platform:"linux"`
 	// ApparmorProfile specifies the apparmor profile for the container.
 	ApparmorProfile string `json:"apparmorProfile,omitempty" platform:"linux"`
 	// SelinuxLabel specifies the selinux context that the container process is run as.


specs-go/config.go
```diff
@@ -40,13 +40,12 @@ type Process struct {
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
 	Cwd string `json:"cwd"`
-	// Capabilities are Linux capabilities that are kept for the container.
+	// Capabilities are Linux capabilities that are kept for the container. (this field is platform dependent)
 	Capabilities []string `json:"capabilities,omitempty" platform:"linux"`
```
wking: Doesn't the `platform:"linux"` tag already cover the “this field is platform dependent” idea?

specs-go/config.go
```diff
@@ -40,13 +40,12 @@ type Process struct {
...
-	// Rlimits specifies rlimit options to apply to the process.
-	Rlimits []Rlimit `json:"rlimits,omitempty"`
-	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container.
-	NoNewPrivileges bool `json:"noNewPrivileges,omitempty"`
-
+	// Rlimits specifies rlimit options to apply to the process. (this field is platform dependent)
+	Rlimits []Rlimit `json:"rlimits,omitempty" platform:"linux,solaris"`
+	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container. (this field is platform dependent)
+	NoNewPrivileges bool `json:"noNewPrivileges,omitempty" platform:"linux,solaris"`
```
wking: I don't think that these `Rlimits` or `NoNewPrivileges` are on Solaris.  And we can probably wait for Solaris folks to pipe up if they are, instead of adding them here and waiting for Solaris folks to pipe up to re-remove them ;).

specs-go/config.go
jhowardmsft: Consistency. It's all over this file whenever a platform annotation is present.

wking:
On Wed, Sep 14, 2016 at 01:51:28PM -0700, John Howard wrote:
> -	// Capabilities are Linux capabilities that are kept for the container.
> +	// Capabilities are Linux capabilities that are kept for the container. (this field is platform dependent)
>  	Capabilities []string `json:"capabilities,omitempty" platform:"linux"`
> 
> Consistency. It's all over this file whenever a platform annotation is present. 

+1 for consistency.  How do folks feel about consistently removing
this from the comment when a platform field exists?  And there are
fields like mounts[]options with different rules depending on the
platform, but I expect that folks will recognize that without having a
Go comment here to remind them.  Are people really reading the Go
comments and skipping the Markdown spec?

jhowardmsft:
@wking - Yes, it all does seem more than a tad superfluous IMO. But that said, it's still a little orthogonal to this PR itself :innocent:

specs-go/config.go
jhowardmsft: @anuthan - Could you comment on the use (or lack of) of `Rlimits` and `NoNewCapabilities` on Solaris? Thanks :smile:

wking:
On Wed, Sep 14, 2016 at 02:44:53PM -0700, John Howard wrote:
> @wking - Yes, it all does seem more than a tad superfluous IMO. But
> that said, it's still a little orthogonal to this PR itself
> :innocent:

This PR does two things (as of 103926ff):

a. Add ‘solaris’ to some ‘platform’ tags.
b. Add “(this field is platform dependent)” to some Go comments.

I think we want to drop (a) [1], but am fine waiting on Solaris-dev
feedback there.  And I think we want to go the other way with (b).

If this PR is just about (a), I'm happy to file my other-(b) as a
separate PR.

[1]: https://github.com/opencontainers/runtime-spec/pull/564/files#r78837457

jhowardmsft:
@wking - Let's try and keep these two things independent, and handle the go comments as a separate issue/PR. I think that will be cleaner in the long run.

wking:
On Wed, Sep 14, 2016 at 04:11:27PM -0700, John Howard wrote:
> @wking - Let's try and keep these two things independent, and handle
> the go comments as a separate issue/PR. I think that will be cleaner
> in the long run.

Sounds good.  I spun off the comment removal into #568.

vbatts:
LGTM

vbatts:
_(side note: we ought to define these tags **somewhere**)_

wking:
On Thu, Sep 15, 2016 at 07:28:19AM -0700, Vincent Batts wrote:
> LGTM

Really?  Solaris does use these fields?

And I agree we should define the tags.  We already have the
enum here [1], just need to link it with the JSON tags.

[1]: https://github.com/opencontainers/runtime-spec/blame/b3ce1951cf83f10f5f0c1f9365522da3755a255f/README.md#L30

vbatts:
@wking chill.
Solaris supports more of the Linux API than I realize. I'll leave it up to folks like @jlb13 to correct me.

specs-go/config.go
anuthan: Solaris does have Rlimits and NoNewPrivileges . Safe to remove :)

jhowardmsft: @anuthan Thanks for confirming! Removed.

jhowardmsft:
@vbatts @wking Removed the annotation for `solaris`

wking:
On Thu, Sep 15, 2016 at 10:25:59AM -0700, John Howard wrote:
> @vbatts @wking Removed the annotation for `solaris`

ce3ac33d7 looks good to me :).

hqhq:
LGTM

jhowardmsft:
What's the process for full approval and merging in this repo? There's two LGTMs above, but I'm not sure if more are still required - it appears it does from above, but I'm not sure why. Forgive my ignorance 😇

hqhq:
@jhowardmsft We have a `.pullapprove.yml` which set `reset_on_push: true` and `reviewers: required: 2`, so after you force pushed, we still need two LGTMs before merging.

jhowardmsft:
Ah, I see. Thanks for confirming :smile:

mrunalp:
LGTM

PR #565, opened 2016-09-14
Windows: User struct changes

jhowardmsft:

Extracting pieces from the proof of concept PR for Windows OCI support at #504. This PR modified the `User` struct by fixing the description, adding a Windows-specific `user` field, and updating the documentation to include the changes, plus provide a sample JSON in the context of the `Process` structure.

### DIFF
--- a/config.md
+++ b/config.md
@@ -186,7 +186,31 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
     "cwd": "/root",
     "args": [
         "/usr/bin/bash"
+    ]
+}
+```
+
+#### Windows User
+
+For Windows based systems the user structure has the following fields:
+
+* **`username`** (string, optional) specifies the user name for the process.
+
+### Example (Windows)
+
+```json
+"process": {
+    "terminal": true,
+    "user": {
+        "username": "containeradministrator"
+    },
+    "env": [
+        "VARIABLE=1"
     ],
+    "cwd": "c:\\foo",
+    "args": [
+        "someapp.exe",
+    ]
 }
 ```
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -53,8 +53,7 @@ type Process struct {
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
 }
 
-// User specifies Linux/Solaris specific user and group information
-// for the container process.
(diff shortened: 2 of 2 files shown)


specs-go/config.go
```diff
@@ -53,15 +53,16 @@ type Process struct {
...
+// User specifies specific user (and group) information for the container process.
 type User struct {
 	// UID is the user id. (this field is platform dependent)
 	UID uint32 `json:"uid" platform:"linux,solaris"`
 	// GID is the group id. (this field is platform dependent)
 	GID uint32 `json:"gid" platform:"linux,solaris"`
 	// AdditionalGids are additional group ids set for the container's process. (this field is platform dependent)
 	AdditionalGids []uint32 `json:"additionalGids,omitempty" platform:"linux,solaris"`
+	// User is the user name. (this field is platform dependent)
+	User string `json:"user,omitempty" platform:"windows"`
```
wking: To avoid `User.User`, can we call this `Username` or something?  And in that case it could be either `Process.User.Username` or `Process.Username` instead of `Process.User.User`.  I like `Process.Username` best, as long as Windows doesn't plan on needing a more complicated structure in the future (in which case it's probably better to stay under `Process.User`).

jhowardmsft: Good idea. I prefer `Username`. Updated

Unfortunately though, I can't predict the future as well :smile_cat:

config.md
```diff
@@ -190,6 +190,30 @@ _Note: For Solaris, uid and gid specify the uid and gid of the process inside th
...
+    "user": {
+        "username": "containeradministrator"
+    },
+    "env": [
+        "VARIABLE=1"
+    ],
+    "cwd": "c:\\foo",
+    "args": [
+        "someapp.exe"],
+    ],
```
wking: Drop the trailing comma?

RobDolinMS: @jhowardmsft I think you may have duplicate instances of "]," on lines 212 and 213.

jhowardmsft: Good catch. That's what I get for copy/pasting from the Solaris example immediately above it. Fixed both :smile:

jhowardmsft: @RobDolinMS Yes, fixed.

tianon:
LGTM

vbatts:
LGTM

PR #570, opened 2016-09-15
config: Link platform:"…" JSON tags with ~~protocol~~ platform slugs

wking:
Spun off from [here](https://github.com/opencontainers/runtime-spec/pull/564#issuecomment-247344440).

More details on the individual pivots in the commit messages.  The
meat is the one-liner in the last commit, and the bit I'm not sure
about (also discussed in the commit message ;) is whether we want to
rename the JSON tag from `platform:"…"` to `protocol:"…"` to
consolidate around the “protocol” term used in #527 to [decouple
compliance from platforms](https://github.com/opencontainers/runtime-spec/pull/527#issuecomment-238979250).  For example, if we want to add
optional extensions (e.g. allowing Linux containers that use
namespacing but not cgroups), we could have a ‘linux’ protocol (with
the base stuff) and a ‘linux-cgroups’ protocol (with the cgroups stuff
and the base stuff).  Runtimes that implemented both would be
compliant with both ‘linux’ and ‘linux-cgroups’.

### DIFF
--- a/config.md
+++ b/config.md
@@ -2,7 +2,8 @@
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
 The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).
-Platform-specific configuration schema are defined in the [platform-specific documents](#platform-specific-configuration) linked below.
+[Platform](spec.md#platforms)-specific configuration schema are defined in the [platform-specific documents](#platform-specific-configuration) linked below.
+For properties that are only defined for some [platforms](spec.md#platforms), the Go property has a `platform` tag listing those protocols (e.g. `platform:"linux,solaris"`).
 
 The configuration file contains metadata necessary to implement standard operations against the container.
 This includes the process to run, environment variables to inject, sandboxing features to use, etc.
--- a/spec.md
+++ b/spec.md
@@ -3,9 +3,10 @@
 The [Open Container Initiative](http://www.opencontainers.org/) develops specifications for standards on Operating System process and application containers.
 
 Protocols defined by this specification are:
-* Linux containers: [runtime.md](runtime.md), [config.md](config.md), [config-linux.md](config-linux.md), and [runtime-linux.md](runtime-linux.md).
-* Solaris containers: [runtime.md](runtime.md), [config.md](config.md), and [config-solaris.md](config-solaris.md).
-* Windows containers: [runtime.md](runtime.md), [config.md](config.md), and [config-windows.md](config-windows.md).
+
+* `linux`: [runtime.md](runtime.md), [config.md](config.md), [config-linux.md](config-linux.md), and [runtime-linux.md](runtime-linux.md).
+* `solaris`: [runtime.md](runtime.md), [config.md](config.md), and [config-solaris.md](config-solaris.md).
+* `windows`: [runtime.md](runtime.md), [config.md](config.md), and [config-windows.md](config-windows.md).
 
 # Table of Contents
 
--- a/spec.md
+++ b/spec.md
@@ -2,7 +2,9 @@
 
 The [Open Container Initiative](http://www.opencontainers.org/) develops specifications for standards on Operating System process and application containers.
 
-Protocols defined by this specification are:
+# Platforms
+
+Platforms defined by this specification are:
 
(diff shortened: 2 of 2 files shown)


RobDolinMS:
@wking Please break-out into separate PRs

wking:
I pushed the header normalization out into #650 and rebased the remaining two commits onto master with eb82bad → c987b8f, so this is now focused exclusively on slugging protocols and [defining the Go binding's `platform` tag][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/564#issuecomment-247344440

config.md
```diff
@@ -2,6 +2,7 @@
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
 The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).
+For properties that are only defined for some [protocols](spec.md#protocols), the Go property has a `platform` tag listing those protocols (e.g. `platform:"linux,solaris"`).
```
crosbymichael: I'd call them platforms not protocols, because that is what they actually are and why the tag is named that way.

wking: > I'd call them platforms not protocols, because that is what they actually are and why the tag is named that way.

That's what they are now, but `protocols` [leaves the door open for future entries that are not platforms][1].  Are we comfortable closing that door (I'd rather leave it open, but I'm not a maintainer)?

If maintainers are comfortable closing the door, how thoroughly should I replace ‘protocol’ with ‘platform’?  Do you want me to change these too:

    $ git grep -i protocol origin/master
    origin/master:spec.md:Protocols defined by this specification are:
    origin/master:spec.md:An implementation is not compliant for a given CPU architecture if it fails to satisfy one or more of the MUST, REQUIRED, or SHALL requirements for the protocols it implements.
    origin/master:spec.md:An implementation is compliant for a given CPU architecture if it satisfies all the MUST, REQUIRED, and SHALL requirements for the protocols it implements.

For example,

> An implementation is compliant for a given CPU architecture if it satisfies all the MUST, REQUIRED, and SHALL requirements for the platforms it implements.

reads a bit strangely to me, because we are currently including the CPU architecture in `platform`.  Maybe the protocol-less version of that line would be:


> An implementation is compliant for a given CPU architecture if it satisfies all the MUST, REQUIRED, and SHALL requirements for the operating systems it implements.

[1]: https://github.com/opencontainers/runtime-spec/pull/570#issue-177220681

hqhq: I'd vote for `platform` as well, and I don't see why our spec would define such non-platform protocol such as `linux-cgroup` protocol.

hqhq: If we use `platform`, it should be consistent through all spec, I'm fine with your second replacement, I think in our spec, `platform` and `operating system` mean the same thing.

hqhq:
@wking Could your update and rebase since this PR has a `v1.0.0` milestone, thanks.

config.md
wking: > I don't see why our spec would define such non-platform protocol such as `linux-cgroup` protocol.

I'd floated [making `hooks` a protocol][1] to accommodate a Windows runtime that chose not to support then despite their being possible on Windows (although I'm not sure if hooks are [possible or not on Windows][2]).  And I think we want a `base` protocol to cover stuff that all compliant runtimes will need (e.g. `ociVersion`).  But having a maintainer reply to my counter-argument with a clear rejection of "protocol" is enough for a rebase here.  I'll get it pushed sometime today, and then file a follow-up PR adding `base`, which you can pick up if it looks useful.

[1]: https://github.com/opencontainers/runtime-spec/pull/472#issuecomment-258567371
[2]: https://github.com/opencontainers/runtime-spec/pull/472#issue-157263212

wking:
On Wed, Feb 08, 2017 at 12:37:06AM -0800, Qiang Huang wrote:
> @wking Could your update and rebase since this PR has a `v1.0.0`
> milestone, thanks.

Rebased with “protocol” → “platform” (although I still think that's
unnecessarily restricting future platform/protocol choices [1]) in
a78a538 → 4af0c72.  Despite that change, I think 4af0c72 is still a
step mostly-forward because it defines the semantics of the JSON tags.

[1]: https://github.com/opencontainers/runtime-spec/pull/570#discussion_r100095721

hqhq:
LGTM

vbatts:
LGTM

PR #554, opened 2016-09-01
Add references to Windows containers

RobDolinMS:
Signed-off-by: Rob Dolin <robdolin@microsoft.com>

PR #573, opened 2016-09-17
Add support for Windows based containers

jhowardmsft:

This adds Windows support to the OCI runtime-spec: code, JSON schema and markdown documentation. Note that the support has been added using 'aggressive namespacing', the matching changes of which for Linux and Solaris being currently out for review in https://github.com/opencontainers/runtime-spec/pull/567.

This is a formal follow-on the the previous proof-of-concept PR to add support for Windows, https://github.com/opencontainers/runtime-spec/pull/504

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -1,0 +1,99 @@
+# Windows-specific Container Configuration
+
+This document describes the schema for the [Windows-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+The Windows container specification uses APIs provided by the Windows Host Compute Service (HCS) to fulfill the spec.
+
+## Resources
+
+You can configure a container's resource limits via the OPTIONAL `resources` field of the Windows configuration.
+
+### Memory
+
+`memory` is an OPTIONAL configuration for the container's memory usage.
+
+The following parameters can be specified:
+
+* **`limit`** *(uint64, OPTIONAL)* - sets limit of memory usage in bytes.
+
+* **`reservation`** *(uint64, OPTIONAL)* - sets the guaranteed minimum amount of memory for a container in bytes.
+
+#### Example
+
+```json
+    "windows": {
+        "resources": {
+            "memory": {
+                "limit": 2097152,
+                "reservation": 524288
+            }
+        }
+    }
+```
+
+### CPU
+
+`cpu` is an OPTIONAL configuration for the container's CPU usage.
+
+The following parameters can be specified:
+
+* **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container.
+
+* **`shares`** *(uint16, OPTIONAL)* - specifies the relative weight to other containers with CPU shares. The range is from 1 to 10000.
+
+* **`percent`** *(uint, OPTIONAL)* - specifies the percentage of available CPUs usable by the container.
+
+#### Example
+
+```json
+    "windows": {
+        "resources": {
+            "cpu": {
+                "percent": 50
+            }
+        }
+    }
+```
+
+### Storage
+
+`storage` is an OPTIONAL configuration for the container's storage usage.
+
+The following parameters can be specified:
+
+* **`iops`** *(uint64, OPTIONAL)* - specifies the maximum IO operations per second for the system drive of the container.
+
+* **`bps`** *(uint64, OPTIONAL)* - specifies the maximum bytes per second for the system drive of the container.
+
+* **`sandboxSize`** *(uint64, OPTIONAL)* - specifies the minimum size of the system drive in bytes.
+
+#### Example
+
+```json
+    "windows": {
+        "resources": {
+            "storage": {
+                "iops": 50
+            }
+        }
+    }
+```
+
+### Network
+
+`network` is an OPTIONAL configuration for the container's network usage.
+
+The following parameters can be specified:
+
+* **`egressBandwidth`** *(uint64, OPTIONAL)* - specified the maximum egress bandwidth in bytes per second for the container.
+
+#### Example
+
+```json
+    "windows": {
+        "resources": {
+            "network": {
+                "egressBandwidth": 1048577
+            }
+        }
+   }
+```
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -1,0 +1,75 @@
+{
+    "windows": {
+        "description": "Windows platform-specific configurations",
+        "id": "https://opencontainers.org/schema/bundle/windows",
+        "type": "object",
+        "properties": {
+            "resources": {
+                "id": "https://opencontainers.org/schema/bundle/windows/resources",
+                "type": "object",
+                "properties": {
+                    "memory": {
+                        "id": "https://opencontainers.org/schema/bundle/windows/resources/memory",
+                        "type": "object",
+                        "properties": {
+                            "limit": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/memory/limit",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "reservation": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/memory/reservation",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            }
+                        }
+                    },
+                    "cpu": {
+                        "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu",
+                        "type": "object",
+                        "properties": {
+                            "count": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/count",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "shares": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/shares",
+                                "$ref": "defs-windows.json#/definitions/cpuSharesPointer"
+                            },
+                            "percent": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/percent",
+                                "$ref": "defs.json#/definitions/percentPointer"
+                            }
+                        }
+                    },
+                    "storage": {
+                        "id": "https://opencontainers.org/schema/bundle/windows/resources/storage",
+                        "type": "object",
+                        "properties": {
+                            "iops": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/storage/iops",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "bps": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/storage/bps",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "sandboxSize": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/storage/sandboxSize",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            }
+                        }
+                    },
+                    "network": {
+                        "id": "https://opencontainers.org/schema/bundle/windows/resources/network",
+                        "type": "object",
+                        "properties": {
+                            "egressBandwidth": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/network/egressBandwidth",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
--- a/schema/defs-windows.json
+++ b/schema/defs-windows.json
@@ -1,0 +1,20 @@
+{
+    "definitions": {
+        "cpuShares": {
+            "description": "Relative weight to other containers with CPU Shares defined",
+            "type": "integer",
+            "minimum": 1,
+            "maximum": 10000
+        },
+        "cpuSharesPointer": {
+            "oneOf": [
+                {
+                    "$ref": "#/definitions/cpuShares"
+                },
+                {
+                    "type": "null"
+                }
+            ]
+        }
+    }
+}
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -25,6 +25,8 @@ type Spec struct {
 	Linux *Linux `json:"linux,omitempty" platform:"linux"`
 	// Solaris is platform specific configuration for Solaris containers.
 	Solaris *Solaris `json:"solaris,omitempty" platform:"solaris"`
+	// Windows is platform specific configuration for Windows based containers, including Hyper-V containers.
+	Windows *Windows `json:"windows,omitempty" platform:"windows"`
 }
 
 // Process contains information to start a specific application inside the container.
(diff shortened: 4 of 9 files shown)


config-windows.md
```diff
@@ -0,0 +1,80 @@
...
+
+`storage` is used to set limits on the container's storage usage.
+
+The following parameters can be specified:
+
+* **`iops`** *(uint64, optional)* - specifies the maximum Iops for the system drive of the container.
+
+* **`bps`** *(uint64, optional)* - specifies the maximum bytes per second for the system drive of the container.
+
+* **`sandboxSize`** *(uint64, optional)* - specifies the size to expand the system drive of the container to if it is currently smaller.
```
wking: Maybe:

> specifies the minimum size of the system drive in bytes.

config-windows.md
```diff
@@ -0,0 +1,80 @@
...
+
+* **`egressBandwidth`** *(uint64, optional)* - specified the maximum egress bandwidth in bytes per second for the container.
+
+###### Example
+
+```json
+   "network": {
+        "egressBandwidth": 1048577
+   }
+```
```
wking: Trailing newline.

specs-go/config.go
```diff
@@ -25,6 +25,8 @@ type Spec struct {
 	Linux *Linux `json:"linux,omitempty" platform:"linux"`
 	// Solaris is platform specific configuration for Solaris containers.
 	Solaris *Solaris `json:"solaris,omitempty" platform:"solaris"`
+	// Windows is platform specific configuration for Windows based containers, including Hyper-V containers.
+	Windows *Windows `json:"solaris,omitempty" platform:"windows"`
```
wking: solaris -> windows.

config-windows.md
jhowardmsft: Agreed. Far more succinct. Will be in next push.

config-windows.md
jhowardmsft: Fixed in next push.

specs-go/config.go
jhowardmsft: Embarrassing copy/paste error! Yes, that one is definitely fixed in the upcoming push.

jhowardmsft:
@wking - comments addressed. And updated README from #554.

config-windows.md
```diff
@@ -0,0 +1,80 @@
+# Windows-specific Container Configuration
+
+This document describes the schema for the [Windows-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+The Windows container specification uses APIs provided by the Windows Host Compute Service (HCS) to fulfill the spec.
+
+
+#### Memory
+
+`memory` is used to set limits on the container's memory usage.
```
wking: The existing spec is not very consistent about this, but I'd like to see an OPTIONAL dropped into this sentence.  Maybe:

> `memory` is an OPTIONAL configuration for the container's memory usage.

specs-go/config.go
```diff
@@ -406,6 +408,58 @@ type Anet struct {
 	Macaddress string `json:"macAddress,omitempty"`
 }
 
+// Windows defines the runtime configuration for Windows based containers, including Hyper-V containers.
+type Windows struct {
+	// Resources contains information for handling resource constraints for the container.
+	Resources *WindowsResources `json:"resources,omitempty"`
+}
```
wking: You don't seem to define this in the Markdown.  If you really want:

```json
{
  …
  "windows": {
    "resources": {
      "memory": {…}
    }
  }
}
```

instead of:

```json
{
  …
  "windows": {
    "memory": {…}
  }
}
```

You should drop a `resources` note in the Markdown.

config-windows.md
jhowardmsft: No objection at all. Modified memory, CPU, storage, and networking to all match this pattern.

wking:
On Sat, Sep 17, 2016 at 09:21:03PM -0700, John Howard wrote:
> +#### Memory
> +
> +`memory` is used to set limits on the container's memory usage.
>
> No objection at all. Modified memory, CPU, storage, and networking
> to all match this pattern.

I think we want all-caps OPTIONAL to trigger the RFC 2119 semantics
[1].  I'll file a separate PR to update the existing instances.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/README.md#L25

specs-go/config.go
jhowardmsft: Yes, done.

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### Memory
+
+`memory` is an OPTIONAL configuration for the container's memory usage.
+
+The following parameters can be specified:
+
+* **`limit`** *(uint64, optional)* - sets limit of memory usage in bytes
+
+* **`reservation`** *(uint64, optional)* - sets soft limit of memory usage in bytes
```
wking: Linux has [resource limits with soft and hard settings][1]:

> The  soft  limit  is  the value that the kernel enforces for the corresponding resource.  The hard limit acts  as  a  ceiling for the soft limit:  an  unprivileged process may set only its soft limit to a value in the range from 0 up to the hard limit, and (irreversibly) lower its hard limit.  A privileged process (under Linux: one with the `CAP_SYS_RESOURCE` capability) may make arbitrary changes to either limit value.

So when you say “soft limit”, that's what I'm thinking.  However, “reservation” sounds more like “I need at least this much”.  Is this setting a minimum amount required by the container?  Or an initial limit on the amount that the container can consume?

[1]: http://man7.org/linux/man-pages/man2/getrlimit.2.html

jhowardmsft:
@wking And final update for now - change `optional` to `OPTIONAL` per your latest comment. Didn't realise you meant the capitalisation the previous time.

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+###### Example
+
+```json
+    "windows": {
+        "resources" {
```
wking: Missing a colon.

config-windows.md
jhowardmsft: Memory management is very different between Linux and Windows. Here, it means a guaranteed minimum amount of memory. I used the existing terminology though. Perhaps incorrectly?

config-windows.md
jhowardmsft: Fixed (all 4 occurances)

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+            "memory": {
+                "limit": 2097152,
+                "reservation": 524288
+	        }
+	    }
```
jhowardmsft: Fixed. Tab rather than 4 spaces. Blame editor :innocent:

config-windows.md
wking: I didn't realize we already had that language, and think it needs to be clarified for the Linux case too.  Grepping through runC shows that `reservation` is backed by `memory.soft_limit_in_bytes`.  [Kernel docs for that][1] have the following to say about soft limits:

> .7. Soft limits
>
> Soft limits allow for greater sharing of memory. The idea behind soft limits
> is to allow control groups to use as much of the memory as needed, provided
>
> a. There is no memory contention
> b. They do not exceed their hard limit
>
> When the system detects memory contention or low memory, control groups
> are pushed back to their soft limits. If the soft limit of each control
> group is very high, they are pushed back as much as possible to make
> sure that one control group does not starve the others of memory.
>
> Please note that soft limits is a best-effort feature; it comes with
> no guarantees, but it does its best to make sure that when memory is
> heavily contended for, memory is allocated based on the soft limit
> hints/setup. Currently soft limit based reclaim is set up such that
> it gets invoked from `balance_pgdat` (`kswapd`).

So in your case I think `reservation` makes a lot of sense for the property name, but you should reroll the description to say something like “guaranteed minimum amount of memory”.

[1]: https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt

config-windows.md
```diff
@@ -0,0 +1,99 @@
+# Windows-specific Container Configuration
+
+This document describes the schema for the [Windows-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+The Windows container specification uses APIs provided by the Windows Host Compute Service (HCS) to fulfill the spec.
+
+You can configure a container's resource limits via the `resources` field of the Windows configuration.
+Do not specify `resources` unless required.
```
wking: This last line seems overly strict.  How about dropping it and using “…the OPTIONAL `resources`…” in the preceding line?

wking: And probably worth a `## Resources` before this paragraph so you can link to it specifically after the Windows-specific file grows other types of configurations (which is presumably why you have a `resources` level at all ;).

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+#### Memory
```
wking: With `## Resources` above, this one should be an h3.

wking:
This is looking pretty good to me.  Still needs a JSON Schema, but it
may be better to wait on that until the Markdown/Go settle down.

config-windows.md
jhowardmsft: Re wording - I pulled that wording from the Linux spec, or so I thought, so I must have been a little over-keen modifying for the Windows version. So yes, have toned down the verbiage accordingly. Will be in next push.

jhowardmsft: Updated to include `## Resources` too. Update pushed.

config-windows.md
jhowardmsft: Indeed. Have fixed the h1..h4s to be consistent throughout.

jhowardmsft:
Now with JSON schema :smile:

wking:
On Mon, Sep 19, 2016 at 01:36:21PM -0700, John Howard wrote:
> +You can configure a container's resource limits via the `resources` field of the Windows configuration.
> +Do not specify `resources` unless required.
> 
> Re wording - I pulled that wording from the Linux spec…

Looks like you did ;).  I've filed #576 to adjust this language on
Linux.

schema/config-windows.json
```diff
@@ -0,0 +1,75 @@
...
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/count",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "shares": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/shares",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "percent": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/percent",
+                                "$ref": "defs.json#/definitions/uint8Pointer"
```
wking: Maybe we want to use `defs.json#/definitions/percentPointer` backed by something like:

```json
"percent": {
  "type": "integer",
  "minimum": 0,
  "maximum": 100
},
"percentPointer": {
  "oneOf": [
    {
      "$ref": "#/definitions/percent"
    },
    {
      "type": "null"
    }
  ]
},
```

jhowardmsft: Good catch - didn't think of that. Will update.

jhowardmsft: Updated in 40bf0b7619e079700d0c9c232264e8f7cbf9f71f

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+    }
+```
+
+### CPU
+
+`cpu` is an OPTIONAL configuration for the container's CPU usage.
+
+The following parameters can be specified:
+
+* **`count`** *(uint64, optional)* - specifies the number of CPUs available to the container.
```
wking: I'm fine with `uint64` here, but am enjoying a mental image of a containers that needs 4.3 billion CPUs ;).

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+* **`shares`** *(uint64, optional)* - specifies the relative weight to other containers with CPU shares. The range is from 1 to 10000.
```
wking: If we know the cap is 10000, we can use a `uint16`.

config-windows.md
jhowardmsft: Future proofing :)

schema/config-windows.json
```diff
@@ -0,0 +1,75 @@
...
+                        "type": "object",
+                        "properties": {
+                            "count": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/count",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
+                            "shares": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/shares",
+                                "$ref": "defs.json#/definitions/uint64Pointer"
+                            },
```
wking: This probably needs `defs-windows.json` for CPUShares so we can check the 1–10k range.

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+You can configure a container's resource limits via the OPTIONAL `resources` field of the Windows configuration.
+
+### Memory
+
+`memory` is an OPTIONAL configuration for the container's memory usage.
+
+The following parameters can be specified:
+
+* **`limit`** *(uint64, optional)* - sets limit of memory usage in bytes.
```
vishh: Is `limit` being over-committed?

schema/config-windows.json
jhowardmsft: Yup, will update.

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+`cpu` is an OPTIONAL configuration for the container's CPU usage.
+
+The following parameters can be specified:
+
+* **`count`** *(uint64, optional)* - specifies the number of CPUs available to the container.
+
+* **`shares`** *(uint64, optional)* - specifies the relative weight to other containers with CPU shares. The range is from 1 to 10000.
+
+* **`percent`** *(uint, optional)* - specifies the percentage of available CPUs usable by the container.
```
vishh: I assume the CPU object is essentially a union across three parameters?

jhowardmsft: @vishh Well yes, but bearing in mind any or all of them are optional.

schema/config-windows.json
jhowardmsft: @wking - Updated in d73cf3ec2f6691ffe4225dc91154ecc4d81e6287

config-windows.md
jhowardmsft: Yes. More information on overall memory management of containers on Windows (which is different between Windows Server and Hyper-V containers) will eventually make it into the documentation on microsoft.com, but for now there's nothing published as such.

schema/defs-windows.json
```diff
@@ -0,0 +1,20 @@
...
+        "cpuShares": {
+            "description": "Relative weight to other containers with CPU Shares defined",
+            "type": "integer",
+            "minimum": 0,
+            "maximum": 10000
+        },
+        "cpuSharesPointer": {
+            "oneOf": [
+                {
+                    "$ref": "#/definitions/CPUShares"
```
wking: I don't have much of a preference between `cpuShares` and `CPUShares`, but we should pick one and stick to it ;).

schema/defs-windows.json
```diff
@@ -0,0 +1,20 @@
+{
+    "definitions": {
+        "cpuShares": {
+            "description": "Relative weight to other containers with CPU Shares defined",
+            "type": "integer",
+            "minimum": 0,
```
wking: The Markdown docs say the minimum is 1.

schema/defs-windows.json
jhowardmsft: I'll stick with camelCase. Mea culpa

schema/defs-windows.json
jhowardmsft: And they're right, this is incorrect. Thanks for catching. Updated in 6c66d16ea27afcbcdbf2aa35242a497377d4198f

specs-go/config.go
```diff
@@ -406,6 +408,58 @@ type Anet struct {
 	Macaddress string `json:"macAddress,omitempty"`
 }
 
+// Windows defines the runtime configuration for Windows based containers, including Hyper-V containers.
+type Windows struct {
```
jhowardmsft: Oops, meant to update the Windows one. Reverted this and fixed the Windows one instead in d57e63f35e268110de046d50ae84157603e427da

config-windows.md
jhowardmsft: Fixed to `cpuShares` type.

wking:
d57e63f3 looks good to me :).

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Memory
+
+`memory` is an OPTIONAL configuration for the container's memory usage.
+
+The following parameters can be specified:
+
+* **`limit`** *(uint64, optional)* - sets limit of memory usage in bytes.
+
+* **`reservation`** *(uint64, optional)* - sets the guaranteed minimum amount of memory for a container in bytes.
```
mrunalp: optional -> OPTIONAL

mrunalp: Can you make similar changes throughout this PR?

config-windows.md
```diff
@@ -0,0 +1,99 @@
...
+    }
+```
+
+### Storage
+
+`storage` is an OPTIONAL configuration for the container's storage usage.
+
+The following parameters can be specified:
+
+* **`iops`** *(uint64, optional)* - specifies the maximum Iops for the system drive of the container.
```
mrunalp: Iops --> IO operations or iops?

config-windows.md
jhowardmsft: @mrunalp Yes, of course. Done and pushed.

config-windows.md
jhowardmsft: @mrunalp Updated to `IO operations per seconds`.

mrunalp:
LGTM

RobDolinMS:
@vbatts Would you please give this a look?

vbatts:
Good stuff. Thanks @jhowardmsft 

LGTM

PR #577, opened 2016-09-19
[Linux Config] Clarify which filesystem is being referenced

RobDolinMS:
Signed-off-by: Rob Dolin robdolin@microsoft.com

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -8,7 +8,7 @@ The Linux container specification uses various kernel features like namespaces, 
 The Linux ABI includes both syscalls and several special file paths.
 Applications expecting a Linux environment will very likely expect these files paths to be setup correctly.
 
-The following filesystems MUST be made available in each containerized application's filesystem
+The following filesystems MUST be made available in each container's filesystem
 
 |   Path   |  Type  |
 | -------- | ------ |
--- a/config-linux.md
+++ b/config-linux.md
@@ -8,7 +8,7 @@ The Linux container specification uses various kernel features like namespaces, 
 The Linux ABI includes both syscalls and several special file paths.
 Applications expecting a Linux environment will very likely expect these files paths to be setup correctly.
 
-The following filesystems MUST be made available in each application's filesystem
+The following filesystems MUST be made available in each containerized application's filesystem
 
 |   Path   |  Type  |
 | -------- | ------ |


config-linux.md
```diff
@@ -8,7 +8,7 @@ The Linux container specification uses various kernel features like namespaces,
 The Linux ABI includes both syscalls and several special file paths.
 Applications expecting a Linux environment will very likely expect these files paths to be setup correctly.
 
-The following filesystems MUST be made available in each application's filesystem
+The following filesystems MUST be made available in each containerized application's filesystem
```
wking: @crosbmichael pointed out a while back that [we do not define “application”][1].  Can we use “… in the container mount namespace”?  The glossary does define “[container namespace][2]”.

[1]: https://github.com/opencontainers/runtime-spec/pull/107#issuecomment-145615479
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/glossary.md#container-namespace

cyphar: ^ @crosbymichael *

crosbymichael: You can just say `in each container's filesystem.`

@wking you are using the term incorrectly when referring to namespaces.  A mount namespace is nothing more than a new mounttable for the container and is decoupled from the root filesystem.

wking:
On Tue, Sep 20, 2016 at 09:21:24AM -0700, Michael Crosby wrote:
> > @@ -8,7 +8,7 @@ The Linux container specification uses various kernel features like namespaces,
>  The Linux ABI includes both syscalls and several special file paths.
>  Applications expecting a Linux environment will very likely expect these files paths to be setup correctly.
>  
> -The following filesystems MUST be made available in each application's filesystem
> +The following filesystems MUST be made available in each containerized application's filesystem
> 
> You can just say `in each container's filesystem.`
>
> @wking you are using the term incorrectly when referring to
> namespaces.  A mount namespace is nothing more than a new mounttable
> for the container and is decoupled from the root filesystem.

And that mount table is where these filesystems must be available,
isn't it?  mount(8) calls the whole tree available from / the “file
hierarchy” and “the big file tree” ;).  I'm ok with something like
that, but “the container mount namespace” seems to clearly cover the
same idea.  Talking about “the container's filesystem”, on the other
hand, sounds more like a single mount, and less like the whole tree.

mrunalp:
@RobDolinMS Could you squash the commits?

hqhq:
ping @RobDolinMS

tianon:
LGTM

(repeated to appease PullApprove.....)

mrunalp:
Needs rebase :/

RobDolinMS:

RobDolinMS:
Replaced with #658.

PR #578, opened 2016-09-19
[Runtime] Fix formatting and clarify status values

RobDolinMS:
When reading the PDF, this line is not clear if it applies just to status or to all state values

Signed-off-by: Rob Dolin robdolin@microsoft.com

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -18,8 +18,8 @@ The value MAY be one of:
     * `created`: the container has been created but the user-specified code has not yet been executed
     * `running`: the container has been created and the user-specified code is running
     * `stopped`: the container has been created and the user-specified code has been executed but is no longer running
+    * Additional `status` values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 
-  Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -18,8 +18,8 @@ The value MAY be one of:
     * `created`: the container has been created but the user-specified code has not yet been executed
     * `running`: the container has been created and the user-specified code is running
     * `stopped`: the container has been created and the user-specified code has been executed but is no longer running
+    * Additional `status` values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 
-  Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
```
wking: Maybe if we use a four-space indent?  Possibly missed by #495.

RobDolinMS:
Tagging @vbatts to take a look.  Thanks in advance.

mrunalp:
LGTM

hqhq:
I think we should use a four-space indent here as suggested by @wking

hqhq:
@RobDolinMS Can you check if the four-space indent fixes your problem? I've tested and it looks good to me. Put that sentence with these states doesn't look reasonable to me.

hqhq:
@RobDolinMS I've carried this in #608 , so I'll close this one, thanks.

PR #579, opened 2016-09-19
[Runtime] Clarify what is being referenced as "process"

RobDolinMS:
Also fix missing ")"

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -49,12 +49,12 @@ The lifecycle describes the timeline of events that happen from when a container
 1. OCI compliant runtime's `create` command is invoked with a reference to the location of the bundle and a unique identifier.
 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
    If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST generate an error.
-   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`](config.md#process-configuration) MUST NOT be run at this time.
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -49,12 +49,12 @@ The lifecycle describes the timeline of events that happen from when a container
 1. OCI compliant runtime's `create` command is invoked with a reference to the location of the bundle and a unique identifier.
 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
    If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST generate an error.
-   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`](config.md#process-configuration) MUST NOT be run at this time.
+   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process configuration`](config.md#process-configuration)) MUST NOT be run at this time.
```
wking: I'm ok adding “configuration”, but:

1. I think we the link to the config section discussing the `process` setting is sufficient, so we don't *need* to say “configuration”, and
2. If we add it, “configuration” should go outside the backticks (so: “from the `process` configuration`”).

wking:
And apparently GitHub doesn't cross-link [issues referenced from review comments][1], so I'll mention #466 from a traditional comment ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/579#pullrequestreview-652575

hqhq:
I don't think we need to add `configuration` here, the link would be sufficient, like we don't have to explicitly say `process_config` as the name of field in https://github.com/opencontainers/runtime-spec/blob/master/config.md#process-configuration to specify the process configuration. And the missing ")" is fixed in #603 , so I'll close for now, feel free to comment afterward.

PR #581, opened 2016-09-21
specs-go/config: Make consoleSize a pointer

wking:
Otherwise we it will never be omitted when empty.  This is the same case as 6323157 (specs-go/config: Make Linux and Solaris omitempty (again), 2016-06-17, #502).

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -32,7 +32,7 @@ type Process struct {
 	// Terminal creates an interactive terminal for the container.
 	Terminal bool `json:"terminal,omitempty"`
 	// ConsoleSize specifies the size of the console.
-	ConsoleSize Box `json:"consoleSize,omitempty"`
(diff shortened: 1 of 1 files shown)


crosbymichael:
Can we finish the discussion in other PRs until you start opening up more PRs to change other unrelated things?

wking:
On Thu, Sep 22, 2016 at 09:14:50AM -0700, Michael Crosby wrote:
> Can we finish the discussion in other PRs until you start opening up
> more PRs to change other unrelated things?

I'd rather file the issue, and you can tag it priority/P3 if you don't
think it's important.  In this case, I'm fixing code that landed 21
hours ago (#563), so I don't think this one is out of the blue.

wking:
I'd forgotten about this one, but the change recently landed via #792.

PR #582, opened 2016-09-23
config-linux: add maskedDirPaths

AkihiroSuda:
maskedDirPaths is used to mask directories rather than files.
For example, the /sys/firmware directory should be masked because it can contain some sensitive files (https://github.com/docker/docker/pull/26618):
  - /sys/firmware/acpi/tables/{SLIC,MSDM}: Windows license information:
  - /sys/firmware/ibft/target0/chap-secret: iSCSI CHAP secret

IMO the issue should be fixed but no need to get this PR in v1.0

runc PR: https://github.com/opencontainers/runc/pull/1068

Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -562,6 +562,18 @@ Its value is either slave, private, or shared.
     ]
 ```
 
+## Masked Dir Paths
+
+`maskedDirPaths` is similar to `maskedPaths` but for directories rather than files.
+
+###### Example
+
+```json
+    "maskedPaths": [
+        "/sys/firmware"
+    ]
+```
+
 ## Readonly Paths
 
 `readonlyPaths` will set the provided paths as readonly inside the container.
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
Alternative idea is to make sure `maskedPaths` supports directories as well. 
RFC.

cyphar:
IMO we should just make `makedPaths` mask directories as well -- since it's called masked**Path**s not masked**File**s.

AkihiroSuda:
Note that we need consideration for non-existent paths if we support directories in `maskedPaths`.

For compatibility, we should just bind mount `/dev/null` for any non-existent path, because we cannot determine if the path points a file or a directory.
However, binding /dev/null looks strange for directories.

WDYT?

cyphar:
@AkihiroSuda I'm not sure. IMO these are implementation quirks -- what matters is that the path is not accessible (you could implement it purely with apparmor and IMO it would be "valid").

crosbymichael:
We already mask dirs with `maskedPaths` so we should not have this in the spec and fix any issues you are seeing with this.

PR #583, opened 2016-09-27
config: Convert process.rlimits from an array to an object

wking:
Rlimits do not need either ordering or repeat entries for a single type.  While come JSON libraries preserve object key order or allow repeats, there are many JSON libraries which do not (e.g. Python and JavaScript both parse JSON objects into hash tables).  Using objects here reinforces the unimportance of ordering or repeated entries.

Also add Solaris support.  I'm not entirely clear on this, because while Solaris is POSIX-certified system and there is a Solaris man page for setrlimit, @anuthan [claims no Solaris support for rlimits](https://github.com/opencontainers/runtime-spec/pull/564#discussion_r79016149).

The `additionalProperties` object bit comes from [here](https://spacetelescope.github.io/understanding-json-schema/reference/object.html#properties), although it is not documented in [draft 4 of the JSON Schema RFC](https://tools.ietf.org/html/draft-zyp-json-schema-04).

### DIFF
--- a/config.md
+++ b/config.md
@@ -107,13 +107,21 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
   The executable is the first element and MUST be available at the given path inside of the rootfs.
   If the executable path is not an absolute path then the search $PATH is interpreted to find the executable.
 
-For Linux-based systems the process structure supports the following process specific fields:
+For Linux and Solaris systems, the process structure supports the following process-specific fields:
+
+* **`rlimits`** (object, OPTIONAL) configures [rlimits][setrlimit.3] for the container process.
+    Valid keys are `RLIMIT_*` resources.
+    POSIX [defines several][setrlimit.3], and [Linux][setrlimit.2-linux] and [Solaris][setrlimit.2-solaris] add additional, platform-specific resources.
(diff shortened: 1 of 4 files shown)


crosbymichael:
uhh, just because order is not important does not mean you have to change the schema

wking:
On Tue, Sep 27, 2016 at 09:04:54AM -0700, Michael Crosby wrote:
> uhh, just because order is not important does not mean you have to
> change the schema

True, but using an array means entries are referenced by index
(e.g. ‘rlimits[0]’, but I think validation output would be easier to
read if they were referenced by resource (e.g.
‘rlimits["RLIMIT_CORE"]’).  opencontainers/runtime-tools#223 is
currently (with opencontainers/runtime-tools@f175fae0) using the
resource in validation error messages [1], which assumes there will be
no duplicates.  I'd rather make that an explicit requirement, and
converting to an object here sets us up for that (I've just filed #584
with the other half of making that explicit).

[1]: https://github.com/opencontainers/runtime-tools/pull/223/commits/f175fae04ac39c69be6f12f900129f06a62898d1#diff-5a3a1bbf65af88a77495090e40fb5dfcR238

mrunalp:
Closed by #607

PR #584, opened 2016-09-27
glossary: Make objects explicitly unordered and forbid duplicate names

wking:
Pin down our JSON definition to a particular RFC (which we can explicitly bump if neccessary), instead of referencing the floating JSON homepage.

Explicitly make objects unordered and forbid duplicate names to avoid relying on unportable behavior.  RFC 7159 is [a bit more relaxed][1]:

> The names within an object SHOULD be unique.

but [warns][1]:

> An object whose names are all unique is interoperable in the sense
> that all software implementations receiving that object will agree
> on the name-value mappings.  When the names within an object are not
> unique, the behavior of software that receives such an object is
> unpredictable.  Many implementations report the last name/value pair
> only.  Other implementations report an error or fail to parse the
> object, and some implementations report all of the name/value pairs,
> including duplicates.

The RFC also [warns about order portability][1]:

> JSON parsing libraries have been observed to differ as to whether or
> not they make the ordering of object members visible to calling
> software.  Implementations whose behavior does not depend on member
> ordering will be interoperable in the sense that they will not be
> affected by these differences.

And has some (informative?) language about [entries being unordered][2]:

> An object is an unordered collection of zero or more name/value pairs…

[1]: https://tools.ietf.org/html/rfc7159#section-4
[2]: https://tools.ietf.org/html/rfc7159#section-1

### DIFF
--- a/glossary.md
+++ b/glossary.md
@@ -20,6 +20,8 @@ On Linux, a leaf in the [namespace][namespaces.7] hierarchy in which the [config
 ## JSON
 
 All configuration [JSON][] MUST be encoded in [UTF-8][].
+JSON objects MUST NOT include duplicate names.
+The order of entries in JSON objects is not significant.
 
 ## Runtime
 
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM

hqhq:
LGTM

PR #586, opened 2016-09-30
specs-go/config: fix required items type

Mashimiao:

These items are defined as `required`, I'm not sure the history, but I think they should not be pointer type.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -208,9 +208,9 @@ type LinuxRlimit struct {
 // LinuxHugepageLimit structure corresponds to limiting kernel hugepages
 type LinuxHugepageLimit struct {
 	// Pagesize is the hugepage size
-	Pagesize *string `json:"pageSize,omitempty"`
+	Pagesize string `json:"pageSize"`
 	// Limit is the limit of "hugepagesize" hugetlb usage
-	Limit *uint64 `json:"limit,omitempty"`
+	Limit uint64 `json:"limit"`
 }
 
 // LinuxInterfacePriority for network interfaces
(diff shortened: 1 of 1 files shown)


wking:
125b879f looks good to me.  The associated JSON Schemas need
'required' entries for these fields as well, but that can happen in a
follow-up PR if you don't want to handle it here.

Mashimiao:

cyphar:
Looks fine to me, but doesn't this need a rebase after the aggressive namespacing patch?

mrunalp:
LGTM

hqhq:
Needs rebase but why there is no conflict detect?

Mashimiao:
rebased
@opencontainers/image-spec-maintainers  PTAL

hqhq:
LGTM
But you're pinging the wrong group :)

Mashimiao:
Ah,  that's my fault...  fixed.

specs-go/config.go
```diff
@@ -300,7 +300,7 @@ type LinuxCPU struct {
 // LinuxPids for Linux cgroup 'pids' resource management (Linux 4.3)
 type LinuxPids struct {
 	// Maximum number of PIDs. Default is "no limit".
-	Limit *int64 `json:"limit,omitempty"`
+	Limit int64 `json:"limit"`
```
crosbymichael: Isn't `nil` here meaning no limit?

stevvooe: What does `0` mean?

wking: @crosbymichael, `nil` is not allowed because [`limit` is required][1].

@stevvooe, I expect 0 should be illegal.  Can you file a separate PR to require strictly positive value?

[1]: https://github.com/opencontainers/runtime-spec/blob/e91c55e9a13d36200a1bb6a233ee9583761d1764/config-linux.md#pids

crosbymichael:
LGTM

specs-go/config.go
cyphar: From the kernel side, `0` is functionally identical to `1` (because you can join a cgroup even if it has a limit of `0`).

PR #591, opened 2016-10-18
Require the mount destination to be created before hand

mrunalp:
This will allow the rootfs to be truly read only (like a NFS read only mount for e.g.)

cc: @rhvgoyal @philips 

### DIFF
--- a/config.md
+++ b/config.md
@@ -47,6 +47,7 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
 
 * **`destination`** (string, REQUIRED) Destination of mount point: path inside container.
   For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
+  The destination MUST be present in the root filesystem. Runtime MUST NOT create the destination.
 * **`type`** (string, REQUIRED) The filesystem type of the filesystem to be mounted.
   Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
   Windows: ntfs.


mrunalp:
Another alternative would be to add a flag to CLI API to make this optional.

config.md
```diff
@@ -47,6 +47,7 @@ The parameters are similar to the ones in [the Linux mount system call](http://m
 
 * **`destination`** (string, REQUIRED) Destination of mount point: path inside container.
   For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
+  The destination MUST be present in the root filesystem. Runtime MUST NOT create the destination.
```
wking: “Runtime” → “The runtime”.

wking:
a4695bc0 is fine with me, although it will make populating tmpfs
mounts difficult (e.g. if you want a tmpfs at /tmp and a fuse mount at
at /tmp/stuff).  I don't have a use case for that sort of thing myself
though, and folks can always use hooks or post-create manipulation.

config.md
wking: Also, one sentence per line.

cyphar: Why can't we just make this a restriction which only applies if `readonly` is set to true for the `rootfs`. I get why you'd want a sweeping statement like this, but as @wking mentioned it does make nested `tmpfs` mounts quite hard -- and in fact would make runC's method of handling cgroups not really possible anymore.

cyphar: Also, the "MUST be present in the root filesystem" means that you can't really have nested `tmpfs` mounts anyway. Because `/tmp/a/b` must already have existed in the host root filesystem in order for you to mount it (which would require making the directory). Maybe we should make it something like:

> The destination MUST already be present if the destination mountpoint is on the root filesystem.
> Runtimes MUST NOT modify the container's root filesystem to fulfil destination requirements.

rhatdan:
I believe the goal here is that the base rootfs is readonly.  Mounting multiple directories on top of a tmpfs should be allowed.  Getting the wording correct on this is paramount.  
We want to use runc on the same rootfs for multiple containers, so it can not be scribbling all over the rootfs.

cyphar:
@rhatdan What do you think of my proposed wording in https://github.com/opencontainers/runtime-spec/pull/591#discussion_r84050973?

rhvgoyal:
May be make it per mount property where callers can specify whether to create destination file/dir or not if it is not present. That should be generic enough to take care of wide variety of use cases.

rhatdan:
@cyphar I like your wording.

wking:
Can we get more details on how this would break @runC's cgroup
handling [1]?  It would be unfortunate to completely break runC
containers where the rootfs is readonly and the only mount.

On Wed, Oct 19, 2016 at 05:33:48AM -0700, Vivek Goyal wrote:
> May be make it per mount property where callers can specify whether
> to create destination file/dir or not if it is not present.

I like this better than @cyphar's wording [2], because it lets folks
safely share filesystems besides their root one (e.g. maybe they want
to mount the same /etc into a bunch of containers).  Although even
Windows uses ‘ro’ to represent read-only for other mounts [3,4], so we
could lean on that if we do end up tying this restriction to
read-only-ness.

[1]: https://github.com/opencontainers/runtime-spec/pull/591#discussion_r84050612
[2]: https://github.com/opencontainers/runtime-spec/pull/591#discussion_r84050973
[3]: https://github.com/opencontainers/runtime-spec/pull/504#r68169631
[4]: https://github.com/docker/docker/pull/26577/files#diff-da772cf2abf2ae03bec24ef9b38f8200R131

mrunalp:
I think making it a per mount config may give most flexibility as @rhatdan suggested.

crosbymichael:
I don't like having the option.  Why can the runtime not just set to create it if it does not exist and if it cannot it errors.  That still leaves it up to the high level to create these directories if its going to have an RO rootfs and if it does not the runtime gives a error.

I think an additional field per mount just adds more complexity than just erroring if it does not exist and it cannot be created.

crosbymichael:
Besides, you are still not solving the real problem of having a shared rootfs with volumes when volumes are a container level config and the rootfs is a shared object across containers.  Even if you create these in some "init" layer, containers without a volume at the same path will still have these dirs/files showing up in other containers because the rootfs+init layer is shared.

rhvgoyal:
@crosbymichael I will not create directories for read-only containers and expect them to be part of the image. That way applications will not have any surprise. 

I think this requires little thought process shift. directories being mounted on need to be part of the image for read-only containers.

Alternatively, we users could think of mounting a tmpfs and creating directories in that tmpfs and doing all the volume mounting there for read-only containers?

crosbymichael:
@rhvgoyal ok, that makes more sense.  If that is the case I don't see why we need a spec change as it raises more questions about nested mounts.  

The runtimes can stay the same and when you mount from the filesystem's you just make it readonly.  If the path does not exist the runtime tries to create the dir and errors as a read-only filesystem error and the image author should know that they need to create these dirs or they need to remove a volume mount.  

Modifying the spec or adding a new field to tell the runtime to create the dir or not is just splitting the information in two places.  As an image/container author I have to both, create a dir in the image.  Tell the spec, dont create this dir.  It should just work as expected. I don't see a spec change adding any value to solving this problem.

rhvgoyal:
@crosbymichael so you are suggesting that caller can leave rootfs read-only so that if runc tries to create a file/dir it fails? We can try it. Hopefully it works with pivot_root(".", ".") stuff.

crosbymichael:
I mean, if we had this spec MUST how are you supposed to validate or what type of error are you going to return anyways?  It would basically be the same with or without.

crosbymichael:
@mrunalp ya, and doing a stat before a mkdirall.  I cannot remember if mkdirall returns an error if the underlaying filesystem is RO but the dir exists.  Just something to double check

mrunalp:
Closing this per the discussion on OCI weekly call today. We can handle this in the runtime (runc). Thanks!

PR #597, opened 2016-10-27
Forbid duplicated namespaces with same `type`.

WeiZhang555:
Related runc PR: https://github.com/opencontainers/runc/pull/1150

Signed-off-by: Zhang Wei zhangwei555@huawei.com


wking:
Instead of using an array for a structure where order is insignificant
and duplicates are not allowed, I'd rather use an object (since the
just-landed #584, all of our objects forbid duplicates and do not have
significant ordering).  I've filed #598 with that approach.  I
understand that migrating from an array to an object will be painful,
but I think it's worth it if we have no reason to use arrays for
portable ordering or duplication.  One benefit of using an object is
that we get much easier add-or-replace and removal code than hacks
like this [1].

[1]: https://github.com/opencontainers/runtime-tools/blob/389c73a0b20576b201a655c1ef37f5558f8ff64e/generate/generate.go#L777-L813

crosbymichael:
LGTM

hqhq:
LGTM

PR #598, opened 2016-10-27
config-linux: Convert linux.namespaces from an array to an object

wking:
Namespaces do not need repeated entries and the ordering is handled by the runtime regardless of the spec ordering (e.g. in opencontainers/runC#977).  Using an object leans on the new wording from #584 to make both of those points explicit.

An alternative to #597.

I feel like we've talked this idea over before, but the only reference I could dig up was my passing mention of “probably convert the array to an object keyed by name” [here](https://github.com/opencontainers/runtime-spec/pull/275#r48071357).

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -23,10 +23,9 @@ A namespace wraps a global system resource in an abstraction that makes it appea
 Changes to the global resource are visible to other processes that are members of the namespace, but are invisible to other processes.
 For more information, see [the man page](http://man7.org/linux/man-pages/man7/namespaces.7.html).
 
-Namespaces are specified as an array of entries inside the `namespaces` root field.
-The following parameters can be specified to setup namespaces:
+* **`namespaces`** (object, OPTIONAL) specifies the container namespaces.
+    Valid keys are:
 
-* **`type`** *(string, REQUIRED)* - namespace type. The following namespaces types are supported:
     * **`pid`** processes inside the container will only be able to see other processes inside the same container.
     * **`network`** the container will have its own network stack.
     * **`mount`** the container will have an isolated mount table.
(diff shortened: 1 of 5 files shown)


config-linux.md
```diff
@@ -35,39 +34,29 @@ The following parameters can be specified to setup namespaces:
...
-            "type": "user"
-        },
-        {
-            "type": "cgroup"
-        }
+        "mount": {},
+        "ipc": {},
+        "uts": {},
+        "user": {},
+        "cgroup": {}
```
hqhq: Then empty and absence have different meanings, I think that's some confusing we are trying to avoid.

wking: I'm ok making that distinction (like I'm ok [distinguishing unset and `null`][1]).  If the consensus is that we shouldn't make that distinction, alternatives include:

a. Using `anyOf` to allow boolean or namespace-object values and requiring `path` in namespace-objects.  E.g. `"pid": true` or `"pid": {"path": "..."}`.  This is clear in JSON Schema, but may be awkward to unmarshal in Go.

b. Replace namespace objects with the path string, and distinguish between unset, empty, and non-empty strings.  This makes namespace payload extension more difficult, but I don't hear anyone calling for namespace extension at the moment.

c. Keep the namespace object, require `path`, and distinguish between empty and non-empty strings.  This is just exchanging empty-string checking (one level deeper) for the empy-object check that currently concerns you.

[1]: https://groups.google.com/a/opencontainers.org/forum/m/#!topic/dev/nfuv8NnOsUg

wking: d. Adding a `new` boolean to the namespace object and erroring if `new` is true and `path` is set/non-empty.

WeiZhang555: I'm leaning towards `b`, as you said, as long as we don't have the need for namespace extension, it should be enough.
For better future extension, `c` looks better though.

`a` will be really disaster for json unmarshaling, and `d` looks redundant.

specs-go/config.go
```diff
@@ -141,7 +141,7 @@ type Linux struct {
 	// If resources are specified, the cgroups at CgroupsPath will be updated based on resources.
 	CgroupsPath *string `json:"cgroupsPath,omitempty"`
 	// Namespaces contains the namespaces that are created and/or joined by the container
-	Namespaces []LinuxNamespace `json:"namespaces,omitempty"`
+	Namespaces map[string]LinuxNamespace `json:"namespaces,omitempty"`
```
WeiZhang555: It should be `map[LinuxNamespaceType]LinuxNamespace`, what's more, maybe we need to rename `LinuxNamespace` to `LinuxNamespaceProperty`

config-linux.md
wking: Both (b) and (c) are really close to the PR's current approach.  Do folks really see a significant difference between switching on empty strings vs. empty objects?

cyphar:
This looks fairly unwieldy to me (given the path comments), why can't we just say in the spec that duplicate entries are forbidden and call it a day?

hqhq:
I don't see much benefit from it either, since Michael already approved #597 , I guess he don't like this PR either, so I think we can close this and merge #597 instead.

PR #599, opened 2016-10-27
Use IO instead of io

hqhq:
For consistency, while all other places use IO.

$ grep -rnIw IO * | wc -l
10


vbatts:
LGTM

wking:
2379be75c looks good to me, although Wikipedia prefers I/O [1].

[1]: https://en.wikipedia.org/wiki/Input/output

crosbymichael:
LGTM

PR #602, opened 2016-10-28
runtime.md: change exist to exit

coolljt0725:

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -44,7 +44,7 @@ When serialized in JSON, the format MUST adhere to the following pattern:
 See [Query State](#query-state) for information on retrieving the state of a container.
 
 ## Lifecycle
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -44,7 +44,7 @@ When serialized in JSON, the format MUST adhere to the following pattern:
 See [Query State](#query-state) for information on retrieving the state of a container.
 
 ## Lifecycle
-The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exist.
+The lifecycle describes the timeline of events that happen from when a container is created to when it ceases to exit.
```
wking: “ceases to exist” is correct.  The container process only exits once, and that's at step (5) of this lifecycle which then continues on to step (6).  After the `delete` in step (6), the container no longer *exists*.

coolljt0725: @wking   Got it, thanks.

PR #603, opened 2016-10-28
typo: fix missing `)`

coolljt0725:

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -49,7 +49,7 @@ The lifecycle describes the timeline of events that happen from when a container
 1. OCI compliant runtime's [`create`](runtime.md#create) command is invoked with a reference to the location of the bundle and a unique identifier.
 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
    If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST generate an error.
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -49,7 +49,7 @@ The lifecycle describes the timeline of events that happen from when a container
 1. OCI compliant runtime's [`create`](runtime.md#create) command is invoked with a reference to the location of the bundle and a unique identifier.
 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
    If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST generate an error.
-   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`](config.md#process-configuration) MUST NOT be run at this time.
+   While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified code (from [`process`](config.md#process-configuration)) MUST NOT be run at this time.
```
wking: This is also fixed in #466 and #579.  It would be nice if we could land at least one of those three PRs ;).

crosbymichael:
LGTM

hqhq:
@wking I don't think we can get a quick consensus in #466 , and I don't the major fix in #579 is needed, so I'll merge this instead, thanks.

LGTM

PR #604, opened 2016-10-28
config.md: clarify timeout of hooks is OPTIONAL

coolljt0725:

### DIFF
--- a/config.md
+++ b/config.md
@@ -355,7 +355,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 ```
 
 `path` is REQUIRED for a hook.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -355,7 +355,7 @@ If a hook returns a non-zero exit code, then an error is logged and the remainin
 ```
 
 `path` is REQUIRED for a hook.
-`args` and `env` are OPTIONAL.
+`args`, `env` and `timeout` are OPTIONAL.
```
wking: Also covered by #427 which cleans up a lot of other hook spec too.

coolljt0725: @wking  didn't find that :)

PR #606, opened 2016-11-02
Use rt instead of realtime

hqhq:
I don't have strong opinion on this, `rt` is widely used
as present of `realtime`, it makes the name shorter and
cleaner. And this is also the naming method which Linux
kernel use.

We can consider this before 1.0 when we still have chance.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -315,19 +315,19 @@ For more information, see [the cpusets cgroup man page][cgroup-v1-cpusets].
 
 The following parameters can be specified to setup the controller:
 
-* **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup
+* **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup.
 
-* **`quota`** *(uint64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+* **`quota`** *(uint64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below).
 
-* **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only)
+* **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only).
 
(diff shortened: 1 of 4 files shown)


config-linux.md
```diff
@@ -315,19 +315,19 @@ For more information, see [the cpusets cgroup man page][cgroup-v1-cpusets].
...
 
-* **`quota`** *(uint64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+* **`quota`** *(uint64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below).
 
-* **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only)
+* **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only).
 
-* **`realtimeRuntime`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources
+* **`rtRuntime`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources.
```
cyphar: This should mention the `realtime` scheduler as with `rtPeriod`.

config-linux.md
```diff
@@ -315,19 +315,19 @@ For more information, see [the cpusets cgroup man page][cgroup-v1-cpusets].
...
 
-* **`realtimePeriod`** *(uint64, OPTIONAL)* - same as **`period`** but applies to realtime scheduler only
+* **`rtPeriod`** *(uint64, OPTIONAL)* - same as **`period`** but applies to realtime scheduler only.
```
cyphar: I'm fine with the change, but I'd prefer if we wrote `realtime (RT) scheduler` to make it clear where `rt` comes from.

hqhq:
@cyphar Updated, thanks.

tianon:
I'm somewhere between neutral and -1 on this change -- IMO, the verbose name is more immediately obvious, but I don't feel strongly enough to block the PR if other maintainers disagree :+1:

mrunalp:
LGTM

RobDolinMS:
I agree with @tianon that the verbose "realtime" is better than "rt"  

"rt" could be confused with "runtime" (or possibly more obscurely, "retry"))

/cc @mrunalp

hqhq:
OK, let's keep it as it is, thanks guys.

PR #607, opened 2016-11-03
Forbid duplicated rlimits with same type

hqhq:
Alternative of #583 , as what we do for `namespaces`
in #597 .

### DIFF
--- a/config.md
+++ b/config.md
@@ -110,16 +110,23 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 For Linux-based systems the process structure supports the following process specific fields:
 
 * **`capabilities`** (array of strings, OPTIONAL) capabilities is an array that specifies Linux capabilities that can be provided to the process inside the container.
-Valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html)
-* **`rlimits`** (array of rlimits, OPTIONAL) rlimits is an array of rlimits that allows setting resource limits for a process inside the container.
-The kernel enforces the `soft` limit for a resource while the `hard` limit acts as a ceiling for that value that could be set by an unprivileged process.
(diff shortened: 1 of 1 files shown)


wking:
On Wed, Nov 02, 2016 at 07:15:57PM -0700, Qiang Huang wrote:
> Alternative of #583 , as what we do for `namespaces`
> in #597 .

The concern with using objects for namespaces was the path handling
[1] (and I *still* think it would have been better to use an object
there [2]).  But we require at least one limit for rlimit entries
(otherwise what's the point?), so I don't see a reason to *not* go
with an object for rlimits.

[1]: https://github.com/opencontainers/runtime-spec/pull/598#issuecomment-257481007
[2]: https://github.com/opencontainers/runtime-spec/pull/597#issuecomment-256678568

wking:
On Wed, Nov 02, 2016 at 08:36:27PM -0700, W. Trevor King wrote:
> But we require at least one limit for rlimit entries (otherwise
> what's the point?)…

Actually, we don't.  But an rlimit entry with neither hard nor soft is
a no-op, while in #598 we needed to distinguish between no-op and
“create a new namespace”.  So I think my initial “so I don't see a
reason to *not* go with an object for rlimits” still stands.

hqhq:
@wking I still don't see the problem of using array here after we fix the duplicated type issue.

wking:
It's nothing insurmountable, but add-or-replace and removal are much easier with objects.  In #597 I [linked][1] to [this hoop jumping][2] you need to support then for arrays.

[1]: https://github.com/opencontainers/runtime-spec/pull/597#issuecomment-256678568
[2]: https://github.com/opencontainers/runtime-tools/blob/389c73a0b20576b201a655c1ef37f5558f8ff64e/generate/generate.go#L777-L813

crosbymichael:
LGTM

dqminh:
LGTM

mrunalp:
LGTM

PR #608, opened 2016-11-03
Fix several format issues found by pdf and html

hqhq:
This carries #578 and fixes some other format issues.


wking:
661314a9 looks good to me.

tianon:
LGTM

(although I think review would probably benefit from some before/after samples or screenshots)

dqminh:
LGTM

PR #612, opened 2016-11-08
[Bundle] Adds anchor tags for headings and bullets

RobDolinMS:

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -1,6 +1,6 @@
-# Filesystem Bundle
+# <a name="filesystemBundle" />Filesystem Bundle
 
-## Container Format
+## <a name="containerFormat" />Container Format
 
 This section defines a format for encoding a container as a *filesystem bundle* - a set of files organized in a certain way, and containing all the necessary data and metadata for any compliant runtime to perform all standard operations against it.
 See also [OS X application bundles](http://en.wikipedia.org/wiki/Bundle_%28OS_X%29) for a similar use of the term *bundle*.
@@ -10,11 +10,11 @@ The definition of a bundle is only concerned with how a container, and its confi
 A Standard Container bundle contains all the information needed to load and run a container.
 This MUST include the following artifacts:
 
-1. `config.json`: contains configuration data.
+1. <a name="containerFormat01" />`config.json`: contains configuration data.
 This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
 See [`config.json`](config.md) for more details.
 
-2. A directory representing the root filesystem of the container.
+2. <a name="containerFormat02" />A directory representing the root filesystem of the container.
 While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
 This directory MUST be referenced from within the `config.json` file.
 


bundle.md
```diff
@@ -1,6 +1,6 @@
-# Filesystem Bundle
+# <a name="filesystemBundle" />Filesystem Bundle
 
-## Container Format
+## <a name="containerFormat" />Container Format
```
wking: I don't think we need to add anchors to the headers, since both [GitHub][1] and Pandoc agree on the auto-generated `container-format` anchor (and likewise for other headers):

    $ curl -Ls https://github.com/opencontainers/runtime-spec/releases/download/v1.0.0-rc2/oci-runtime-spec-1.0.0-rc2.html | grep 'Container Format'
    <h2 id="container-format">Container Format</h2>
[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/bundle.md#container-format

bundle.md
```diff
@@ -10,11 +10,11 @@ The definition of a bundle is only concerned with how a container, and its confi
 A Standard Container bundle contains all the information needed to load and run a container.
 This MUST include the following artifacts:
 
-1. `config.json`: contains configuration data.
+1. <a name="containerFormat01" />`config.json`: contains configuration data.
```
wking: These anchors look good to me, although I think we should stick to the lowercase, hyphenated form that GitHub and Pandoc are using already for headers.  In that case, the name here would be `container-format-01`.

bundle.md
RobDolinMS: The proposed PR separates the anchor tag from the header text.  

This follows the recommendation from the Open API Initiative (formerly Swagger) which.

I'm hesitant to use the auto-generated anchor links from GitHub since we might change the header text (ex: adding section numbers like "4.1 Container Format" and that would change the auto-generated anchor links.

RobDolinMS:
See related PR #614

bundle.md
wking: I don't think link stability is a major issue, because folks referencing the spec should be pinning their reference to a particular version ([like this][1]).  If we change the header later on, that link (pinned to v1.0.0-rc2) will still work.  Even if you do attempt to provide a stable anchor between spec releases, sometimes the section will go away, or get moved to a different file, or be split into two sections, or whatever.  In all of those cases the old unpinned link will no longer work regardless of any attempt at freezing the anchor.  Do we really expect lots of churn in header text but minimal churn in the section or file structure?

That said, if folks do expect significant churn in header text and are happy to nip that in the bud by adding these headers to all of our existing sections, more power to 'em ;).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/bundle.md#container-format

bundle.md
mrunalp: Does it make sense to actually not be hyphenated so we can distinguish them from github/pandoc anchors?

RobDolinMS:
@opencontainers/runtime-spec-maintainers Per discussion on today's OCI Dev ConCall, please merge this PR.  Thanks.

tianon:
LGTM

hqhq:
LGTM

PR #614, opened 2016-11-08
[Runtime] Add anchor tags for headings and bullets

RobDolinMS:

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -1,29 +1,29 @@
-# Runtime and Lifecycle
+# <a name="runtimeAndLifecycle" />Runtime and Lifecycle
 
-## Scope of a Container
+## <a name="scopeOfAContainer" />Scope of a Container
 
 Barring access control concerns, the entity using a runtime to create a container MUST be able to use the operations defined in this specification against that same container.
 Whether other entities using the same, or other, instance of the runtime can see that container is out of scope of this specification.
 
-## State
+## <a name="state" />State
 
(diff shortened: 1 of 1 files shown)


cyphar:
Is that syntax defined by the markdown spec? Because I would expect that it would result in you having lots of trailing `<a>` tags.

wking:
On Sat, Nov 12, 2016 at 06:03:32PM -0800, Aleksa Sarai wrote:
> Is that syntax defined by the markdown spec?

From the spec [1]:

  For any markup that is not covered by Markdown’s syntax, you simply
  use HTML itself…
  …
  Span-level HTML tags — e.g. <span>, <cite>, or <del> — can be used
  anywhere in a Markdown paragraph, list item, or header. If you want,
  you can even use HTML tags instead of Markdown formatting; e.g. if
  you’d prefer to use HTML <a> or  tags instead of Markdown’s
  link or image syntax, go right ahead.

So what @RobDolinMS has with e6fd589b is legal, although I'd prefer
moving to a markup language with better support for long-form docs
(#615).

[1]: http://daringfireball.net/projects/markdown/syntax#html

vbatts:
needs a rebase

mrunalp:
Needs rebase.

crosbymichael:
@RobDolinMS when you have time please rebase your PRs and make sure your sign off is correct.  Feel free to reopen when they have been updated.

PR #619, opened 2016-11-11
config: Change “Process configuration” → “Process” header

wking:
All of these sections are about configuration, and we don't usually use “{Whatever} configuration” in the headers.

Micro-pivot spun off from #489.


mrunalp:
LGTM

hqhq:
LGTM

PR #525, opened 2016-08-03
config: Synchronize comments between Markdown and Go

wking:
Spun off from [here][1], this unifies the Markdown and Go comments for all of `Spec` except the `Linux` and `Solaris` entries, because it's better to not have to maintain different descriptions for the same idea.  The series also makes a few knock-on changes for consistency.

There seemed like enough was going on after making `Spec` consistent, so I'm postponing consistency for other Go types until after this lands.

This PR will conflict with a number of long-open PRs that have more complete fixes for some of these issues (e.g. opencontainers/runtime-spec#427), but those commits have mostly sat unreviewed for months, so I'm hoping the more restricted pivots here have a better chance of landing.  I can rebase any of the conflicted PRs (that I've submitted anyway) around this if/when it lands.

[1]: https://github.com/opencontainers/ocitools/pull/175#discussion_r73217087

PR #620, opened 2016-11-11
config: Make 'process.args' optional

wking:
Since #384, it's possible for a container process to never execute user-specified code (e.g. you can call `create`, `kill`, `delete` without calling `start`).  For folks who expect to do that, there`s no reason to define `process.args`.

The main point of #489 without the surrounding cleanup, since most of the cleanup has since landed via:

* The one-sentence-per-line fixes in #443.
* The line calling out `process` as required from #525.

There was some [negative][2] [review][3] of this change in #489, although @mrunalp acknowledged [support for the “never calling `start`” workflow][3].  If there is a maintainer consensus around that workflow being supported, requiring `process.args` in all cases [seems like needless hoop-jumping][4].  I'm fine with *validation tooling* warning users “you might have forgotten to put something meaningful in `process.args`, so `start` calls will fail”.  But I'd rather not have *this spec* require dummy `process.args` from folks who will never call `start`.

[1]: https://github.com/opencontainers/runtime-spec/pull/489#issuecomment-259856833
[2]: https://github.com/opencontainers/runtime-spec/pull/489#issuecomment-258564522
[3]: https://github.com/opencontainers/runtime-spec/pull/489#issuecomment-259783257
[4]: https://github.com/opencontainers/runtime-spec/pull/489#issuecomment-259787865

### DIFF
--- a/config.md
+++ b/config.md
@@ -126,8 +126,9 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 * **`cwd`** (string, REQUIRED) is the working directory that will be set for the executable.
   This value MUST be an absolute path.
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
-* **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
+* **`args`** (array of strings, OPTIONAL) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
+  This property is REQUIRED when [`start`](runtime.md#start) is called.
 
 For Linux-based systems the process structure supports the following process specific fields:
 
--- a/runtime.md
+++ b/runtime.md
@@ -103,6 +103,7 @@ This operation MUST generate an error if it is not provided the container ID.
 Attempting to start a container that does not exist MUST generate an error.
 Attempting to start an already started container MUST have no effect on the container and MUST generate an error.
 This operation MUST run the user-specified program as specified by [`process`](config.md#process).
+This operation MUST generate an error if `process.args` was not set.
 
 Upon successful completion of this operation the `status` property of this container MUST be `running`.
 
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -74,8 +74,7 @@
             "id": "https://opencontainers.org/schema/bundle/process",
             "type": "object",
             "required": [
-                "cwd",
-                "args"
+                "cwd"
             ],
             "properties": {
                 "args": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -38,7 +38,7 @@ type Process struct {
 	// User specifies user information for the process.
 	User User `json:"user"`
 	// Args specifies the binary and arguments for the application to execute.
-	Args []string `json:"args"`
+	Args []string `json:"args,omitempty"`
 	// Env populates the process environment for the process.
 	Env []string `json:"env,omitempty"`
 	// Cwd is the current working directory for the process and must be


WeiZhang555:
Well, I think if 
>If `process.args` was not configured, the runtime MUST generate an error.
Then it should make more sense to keep it REQUIRED.

>although @mrunalp acknowledged support for the “never calling start” workflow.
If the user code will never be started, you can specify anything in `args` such as 'args: ["dummyCode"]', I don't see any benefits from removing the required `args` field from spec. On the other hand, we should detect missing `args` and report error earlier to user.

so -1 on this.

wking:
On Mon, Nov 14, 2016 at 01:09:22AM -0800, zhangwei_cs wrote:
> > although @mrunalp acknowledged support for the “never calling
> > start” workflow.
>
> If the user code will never be started, you can specify anything in
> `args` such as 'args: ["dummyCode"]'…

That's not legal, because we currently require the executable to exist
[1].  If you have an empty rootfs, you'd have to use "/" [2].

> On the other hand, we should detect missing `args` and report error
> earlier to user.

I'm fine having validation tooling that warns (or errors, as long as
it's not the compliance-testing invocation) on missing process.args.
I just don't see a point to requiring:

  {"process": {"args": ["/"], …}, …} 

for folks who will never call ‘start’.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/config.md#L102
[2]: https://github.com/opencontainers/runtime-spec/pull/489#issuecomment-259787865

runtime.md
```diff
@@ -103,6 +103,7 @@ This operation MUST generate an error if it is not provided the container ID.
 Attempting to start a container that does not exist MUST generate an error.
 Attempting to start an already started container MUST have no effect on the container and MUST generate an error.
 This operation MUST run the user-specified code as specified by [`process`](config.md#process-configuration).
+If `process.args` was not configured, the runtime MUST generate an error.
```
mikebrow: If the `process.args` array is not specified or if the user-specified code cannot be found or started the runtime MUST generate an error.

wking: I'd rather not make that change.  The execution will usually be by a call to `execve(2)` or similar, and while that returns errors like `ENOENT`, `ENOEXEC`, etc. on failure, there's no way to get it to return something on success (because by that point the process will be running the user-specified program).  And `start` is just signalling an external process to execute code (e.g. see opencontainers/runC#886, where the `start` process drains a fifo to trigger an `execve` in the container process).  So I don't see a way to block on potential `execve` errors in the `start` process short of `ptrace(2)` monitoring, and that seems like too high a cost.

I'd rather have `start` error out if it failed to *trigger* the start attempt (which the container process can confirm with “start received.  I'm about to `execve`, wish me luck”).  Folks who are curious about whether that `execve` succeeded can either `ptrace` the container process themselves,  or monitor `state`, etc.

[1]: https://github.com/opencontainers/runc/blob/v1.0.0-rc2/start.go#L30

mikebrow: Yeah, I wasn't looking for failures in the process started, I'm looking for failure to specify a runnable executable as arg0 of the process.args array. At least change it to:
```
If the process.args array does not include a first argument specifying the user-specified code the runtime MUST generate an error.
```

mikebrow: Otherwise we need some sort of default.. description. No?

wking: That sounds like a subset of the [current `process.args` requirements][1], so it would fall under the [“MAY validate” portion of `create`][2].  Do we need a “MUST validate `process`” (or just `process.args`?) clause in `start`?  My personal preference would be to leave validation up to the caller, who can run [bundle-validation tools][3] whenever they want.  I'm happy to make that an explicit part of the `start` spec, if folks think it would help.  I'll grudgingly make a MUST-validate-whatever clause a part of the `start` spec if that's the maintainer consensus ;).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config.md#process
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/runtime.md#create
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/implementations.md#testing--tools

mikebrow: I thought the point was start does not have to be called.. thus we only have this MUST when start is called. Otherwise this MUST would be over in the process.arg part of the spec. Alternatively could move this MUST to the process.args section and say when start is called...

wking: > Otherwise we need some sort of default.. description.

With optional validation, runtimes that choose not to validate in `start` will have this sort of miss-configuration result in the same sort of [error-workflow that applies to `execve` errors][1].  I don't see a need for a default, and am not sure what sort of default you'd pick in any case.

[1]: https://github.com/opencontainers/runtime-spec/pull/620#discussion_r95419303

mikebrow: Yeah I didn't see a good default either... certainly would not have picked "/" :)

wking: > I thought the point was start does not have to be called.

Yes.

> … thus we only have this MUST when start is called.

That's what I was going for with an [optional `process.args`][1] which still contains [a MUST for an initial argument][2].  That requirement (if `process.args` is set, it MUST contain that leading entry) applies to all configs, not just when `start` is called.

[1]: https://github.com/opencontainers/runtime-spec/pull/620/commits/35ee63bd00cee7da9401b9854620112c4d83d8e3#diff-c9c91c29b41257aea3a3403cc606ad99R131
[2]: https://github.com/opencontainers/runtime-spec/pull/620/commits/35ee63bd00cee7da9401b9854620112c4d83d8e3#diff-c9c91c29b41257aea3a3403cc606ad99R132

dqminh: changing `is not configured` to `is not set` is more understandable to me.
also maybe we should use the same set of phrases above ? maybe `This operation MUST generate an error if  process.args is not set`

wking: > maybe `This operation MUST generate an error if process.args is not set`

Done (and rebased onto master) with 35ee63b → 7071706, although I used “was” instead of “is” because what matters is [what was in the configuration at `create`-time][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc3/runtime.md#L97

runtime.md
```diff
@@ -103,6 +103,7 @@ This operation MUST generate an error if it is not provided the container ID.
 Attempting to start a container that does not exist MUST generate an error.
 Attempting to start an already started container MUST have no effect on the container and MUST generate an error.
 This operation MUST run the user-specified program as specified by [`process`](config.md#process).
+This operation MUST generate an error if `process.args` was not set.
```
mikebrow: The process.args array is currently "OPTIONAL", not REQUIRED. So if it's optional we need to explain why/when its REQUIRED. I know you know why, but it does not explicitly say why / when. The MUST requirement being in the START command paragraph "implies" that the process.args array (first element) is REQUIRED only when START is used but is not explicit or explained. It just says MUST be defined. The word "set" does not express explicitly what you mean. I only mention this because I believe MUST language should only have one interpretation.

wking: After some [discussion on IRC][1], @mikebrow recommended the following addition to the `process.args` docs:

> When start is used process.args is no longer OPTIONAL and MUST be set.

There's also a similar conditional requirement [in the device spec][2], but “REQUIRED unless you won't call `start`” and “OPTIONAL unless you'll call `start` in which case it's required” both feel too awkward.

With 7071706 → e45ddba I've added:

> When [`start`](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/runtime.md#errors) will be called, this property is REQUIRED.

which gets the cross-reference to the `start` command.  It replaces @mikebrow's “is no longer OPTIONAL and MUST be set” with “is REQUIRED”, because I like “REQUIRED” better than “MUST be set”.  I think “is no longer OPTIONAL” is fairly clear from the context, but I'm ok adding that back in if folks feel like it's necessary.  And I've replaced “is used” with “will be called” because you need to [set this field at `create` time][3] in order to use it when `start` is called.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-01-11.log.html#t2017-01-11T18:34:39
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc3/config-linux.md#L118
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc3/runtime.md#L97

mikebrow:
Looks good to me.

config.md
```diff
@@ -126,8 +126,9 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 * **`cwd`** (string, REQUIRED) is the working directory that will be set for the executable.
   This value MUST be an absolute path.
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
-* **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
+* **`args`** (array of strings, OPTIONAL) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
+  When [`start`](runtime.md#start) will be called, this property is REQUIRED.
```
dqminh: `This property is required when start is called` seems better for me.

RobDolinMS:

config.md
wking: Done with e45ddba → bebdd29.

mikebrow: when works..

crosbymichael:
If we want to be pedantic basically all of the process is not needed if you don't specify args or do start.

wking:
> If we want to be pedantic basically all of the process is not needed if you don't specify args or do start.

But there are affects to not specifying `process` even if you do not call `start`.  For example, `process.user` has no default so it's unclear what user the runtime's pre-`start` container process would be running as.  I'm fine defining a default user to cover that case, but there has been resistance to that [in][1] [the][2] [past][3].

However the `process.user` discussion works out, making `process.args` optional seems like a non-controversial first step.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-04-17.00.log.html#l-44
[2]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937010
[3]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937090

crosbymichael:
Overall, if we want to do this, I would just make process optional and be done with it.  The init is always launched as root or whatever the userns mapping is and we don't change.  It would be a fully privileged process in the container however and there is not a good way around that as we need to retain the caps for the finalizing step of the container.  

Its not really a use case that we want to support but it can be done for advanced users that want to fire off many execs without fully creating a container and it up to the runtime and implementer to secure things, we cannot guarantee anything. 

What do the other maintainers think?

crosbymichael:
After a brief chat with maintainers in IRC, we are going to close this specific modification.  Just having args as optional makes no sense at all.  Either the entire process section should be optional for what you want to do or it should be left as is but this change is not ideal for any usecase.

PR #623, opened 2016-11-14
config: Drop Solaris UID/GID note

wking:
The note is from #411, but as I [pointed out there][1], this is also true for Linux.  #412 landed in parallel with more explicit namepacing for these fields, so we no longer need the overly-specific Solaris note.

[1]: https://github.com/opencontainers/runtime-spec/pull/411#r61620322


wking:
CC @anuthan.

mrunalp:
LGTM

dqminh:
LGTM

anuthan:
Sorry for the delayed response, this is fine, probably doesn't matter now too.

Issue #615, opened 2016-11-09
Replace Markdown with reStructuredText or AsciiDoc or some such?


wking:
On Tue, Nov 08, 2016 at 01:54:21PM -0800, @RobDolinMS [wrote][1]:
> I'm hesitant to use the auto-generated anchor links from GitHub since we might change the header text (ex: adding section numbers like "4.1 Container Format" and that would change the auto-generated anchor links.

Given this and our ongoing issues with single-page HTML/PDF generation (most recently #562), I'd rather switch to a framework that provides better support for long-form technical docs than Markdown.  Both reStructuredText and AsciiDoc (among many other options) make it easy to automate this sort of section numbering, cross referencing, etc., and have good support for publishing into HTML, PDF, and other formats.  Instead of repeating that work by hand (and renumbering all later sections when you remove an early one?), I'd rather just switch to a more structured language.  And GitHub [supports both formats][2] with it's web UI.

To ease the cost of manually inserting anchors (#612, #613, #614, and more to come), it would be nice to automatically add anchors to finer-grained components as well (like [Purple Numbers][3]).  I haven't found any modern tooling doing that yet though.  Still, I think auto-numbered sections and more robust HTML/PDF generation are useful enough goals in their own right.

If this sort of transition sounds useful, I'm happy to put together a PR.  That would also help turn up any limitations in either approach.

[1]: https://github.com/opencontainers/runtime-spec/pull/612#discussion_r87092493
[2]: https://github.com/github/markup#markups
[3]: https://en.wikipedia.org/wiki/Purple_Numbers

RobDolinMS:
I'm not deeply familiar reStructuredText or AsciiDoc.  

@wking Do you think this would be an easy conversion or something that might take non-trivial work?  

If this would be non-trivial, maybe we move this for consideration to post v1?

wking:
On Thu, Nov 10, 2016 at 01:13:29PM -0800, Rob Dolin (MSFT) wrote:
> @wking Do you think this would be an easy conversion or something
> that might take non-trivial work?

That sounds like “I'm cautiously interested.  Mock up a PR so I can
get a better feel for this alternative future” ;).  I'll try and file
one in the next few days.

wking:
On Thu, Nov 10, 2016 at 01:28:19PM -0800, W. Trevor King wrote:
> That sounds like “I'm cautiously interested.  Mock up a PR so I can
> get a better feel for this alternative future” ;).  I'll try and
[...]

PR #626, opened 2016-11-14
Cleanup the spec a bit to remove WG/git text that's not really part of the spec

duglin:
First pass - and a little bit of extra clean-up.  Per the f2f at kubecon.

Signed-off-by: Doug Davis <dug@us.ibm.com>

PR #628, opened 2016-11-15
Transition from Markdown to AsciiDoc

wking:
As discussed in #615.  Rendered HTML and PDF forms are [here][1] and [here][2].  There's a fair bit of detail in most commit messages if you're interested in that sort of thing.

The base of this PR is currently a few commits behind master, but keeping this branch up do date will be a pain.  I think we should collect feedback on the approach here, and if/when the approach collects enough maintainer approval I'll push through a rebase and we can land this quickly (before it goes stale again, which will happen if pretty much anything touching the Markdown lands ;).

[1]: http://wking.github.io/opencontainer-runtime-spec/
[2]: http://wking.github.io/opencontainer-runtime-spec/runtime-spec.pdf


wking:
CC @vbatts, @RobDolinMS.

mrunalp:
The rendered pdf/html look pretty good!

mrunalp:
@wking Sorry I missed this PR earlier. I agree that if everyone is on board, we should land this first.

duglin:
can we do this after https://github.com/opencontainers/runtime-spec/pull/626  ?

wking:
On Tue, Nov 15, 2016 at 12:20:20PM -0800, Doug Davis wrote:
> can we do this after
> https://github.com/opencontainers/runtime-spec/pull/626 ?

Sure.  I'm happy to rebase around anything that lands before this PR.
I'd just rather not have to rebase this one multiple times ;).

duglin:
that's how I feel too :-)

dqminh:
The rendered version looks good ( like really good compared to markdown-rendered :smile: )

mrunalp:
could be rebased now that #626 is merged.

wking:
On Wed, Nov 16, 2016 at 12:19:06PM -0800, Mrunal Patel wrote:
> could be rebased now that #626 is merged.

There are still a lot of open PRs that would conflict with this one
(e.g. #620, #611, #607, #606, #577, …).  Before we rebase this to
worry about content, I'd rather hash out any content-agnostic
technical issues.  Once we have some maintainers approving the
technical approach, I'll rebase to bring the content up to speed and
we can collect official LGTMs on the whole package.

hqhq:
I'm +1 for this change, rendered version looks better than markdown.

RobDolinMS:
Per @crosbymichael, not sure we want to do this at all; but agree we want to hold-off on this conversation until after v1.0.0

crosbymichael:
We will looking into consolidating the spec into a single markdown file to make publishing easier.

vbatts:
there will be pains in either staying with markdown or switching to asciidoc

PR #632, opened 2016-11-26
Exclude hook state definition explicitly

WeiZhang555:
Hook State data is passed from runtime to hook, the data structure isn't
defined in spec, and we should clearly clarify this in case reader keeps
trying to find the definition from here.

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>

### DIFF
--- a/config.md
+++ b/config.md
@@ -324,6 +324,7 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
 Hooks allow one to run programs before/after various lifecycle events of the container.
 Hooks MUST be called in the listed order.
 The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+The container state data format is defined by various runtime implementations, and is not included in this specification.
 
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -324,6 +324,7 @@ Presently there are `Prestart`, `Poststart` and `Poststop`.
 Hooks allow one to run programs before/after various lifecycle events of the container.
 Hooks MUST be called in the listed order.
 The state of the container is passed to the hooks over stdin, so the hooks could get the information they need to do their work.
+The container state data format is defined by various runtime implementations, and is not included in this specification.
```
WeiZhang555: Ouch, my mistake. I searched the https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go and didn't find the hook state definition, so I thought it would be a runtime specific field and we never defined this in specs, it turns out to be here: https://github.com/opencontainers/runtime-spec/blob/master/specs-go/state.go

I'll revoke this and send an opposite PR to add definition of Hook State in this spec later.

Sorry for my misread~

PR #633, opened 2016-11-28
Rename `version` to `ociVersion` according to runtime.md

WeiZhang555:
<del>
* `Root` is path to the container's rootfs which can be useful information
to hook, and the official sample runtime implementation--runc--has
already added this field as payload to hook, we should add this to the
spec.
    
* `Status` seems like unnecessary field, because `State` is payload to
hook and hook is executed in special time point, e.g.
1.  `Prestart` is a list of hooks to be run before the container process is
executed, indicating a status of `created`.
2. `Poststart` is a list of hooks to be run immediately after the container
process is started, indicating a status of `running`.
3. `Poststop` is a list of hooks to be run after the container process exits,
indicating a status of `stopped`.
So the `Status` field is redundant. 
</del>

According to definition of [state](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state), we should rename
`version` field in `state.go` to `ociVersion`

Signed-off-by: Zhang Wei <zhangwei555@huawei.com>

### DIFF
--- a/specs-go/state.go
+++ b/specs-go/state.go
@@ -3,7 +3,7 @@ package specs
 // State holds information about the runtime state of the container.
 type State struct {
 	// Version is the version of the specification that is supported.
-	Version string `json:"version"`
+	Version string `json:"ociVersion"`
 	// ID is the container ID
 	ID string `json:"id"`
 	// Status is the runtime state of the container.


crosbymichael:
I think status is included because this state object can be used as the output for `runtime state <id>`

WeiZhang555:
Oh, my... I'm too confused now, it seems that we have many conflicts between specs and runc.

runc containerState: https://github.com/opencontainers/runc/blob/master/list.go#L24-L41
runc hookState: https://github.com/opencontainers/runc/blob/master/libcontainer/configs/config.go#L246-L252

specs description about container state:
https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state
specs structure about state:
https://github.com/opencontainers/runtime-spec/blob/master/specs-go/state.go

In my understanding, they should be the same, but currently they conflict with each other.

conflicting fields include: 
* "version" vs "ociVersion"
* missing "root" field in specs, and also "root" vs "rootfs"
* missing "status" in hooks
* extra "created" field in runc `state`

I'll try to make them synchronous, but the breaking of backward compatibility seems inevitable :cry:

WeiZhang555:
Update:

only update `ociVersion` field because `state.go` didn't follow `runtime.md`

Others:
* missing "root" field in specs, and also "root" vs "rootfs": `root` is redundant, as we already have `bundle`, rootfs can be retrieved from [bundle/config.md](https://github.com/opencontainers/runtime-spec/blob/master/config.md#root-configuration)
* missing "status" in hooks: modify runc
* extra "created" field in runc state: keep it only in runc, not add it in spec. Unless all persons agree that it's truly necessary, but this can be discussed seperately.
* "bundlePath" vs "bundle": runc should follow the spec

Please take a look. @opencontainers/runc-maintainers @opencontainers/runtime-spec-maintainers

wking:
c8f5a569 looks good to me, and seems like a fairly straightforward
bugfix.

hqhq:
LGTM

vbatts:
LGTM

PR #645, opened 2016-12-28
config.md: add empty limit for key of annotations

Mashimiao:
empty key does not make any senses, we should limit it.

### DIFF
--- a/config.md
+++ b/config.md
@@ -388,7 +388,7 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 This information MAY be structured or unstructured.
 Annotations MUST be a key-value map where both the key and value MUST be strings.
 While the value MUST be present, it MAY be an empty string.
-Keys MUST be unique within this map, and best practice is to namespace the keys.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -388,7 +388,7 @@ The semantics are the same as `Path`, `Args` and `Env` in [golang Cmd](https://g
 This information MAY be structured or unstructured.
 Annotations MUST be a key-value map where both the key and value MUST be strings.
 While the value MUST be present, it MAY be an empty string.
-Keys MUST be unique within this map, and best practice is to namespace the keys.
+Keys MUST be unique and MUST NOT be an empty string within this map, and best practice is to namespace the keys.
```
wking: I'd rather not split “unique within this map”.  How about splitting each idea into its own sentence:

> Keys MUST be unique within this map.
> Keys MUST NOT be an empty string.

We can drop the “best practice” portion, because the following line is “Keys SHOULD be named using a reverse domain notation…”, which covers that idea more explicitly.  If we want to give explicit namespacing advice to annotation-spec devs without access to a domain name, I'd rather address that directly (in a separate PR?).

crosbymichael:
LGTM

mrunalp:
LGTM

PR #646, opened 2017-01-05
Perfect json content

q384566678:

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -89,7 +89,10 @@
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/pids/limit",
                                 "$ref": "defs.json#/definitions/int64"
                             }
-                        }
+                        },
+                        "required": [
+                            "limit"
+                        ]
                     },
                     "blockIO": {
                         "id": "https://opencontainers.org/schema/bundle/linux/resources/blockIO",
@@ -222,7 +225,11 @@
                                         "limit": {
                                             "$ref": "defs.json#/definitions/uint64"
                                         }
-                                    }
+                                    },
+                                    "required": [
+                                        "pageSize",
+                                        "limit"
+                                    ]
                                 }
                             },
                             {
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -59,6 +59,9 @@
             "description": "Configures the container's root filesystem.",
             "id": "https://opencontainers.org/schema/bundle/root",
             "type": "object",
+            "required": [
+                "path"
+            ],
             "properties": {
                 "path": {
                     "id": "https://opencontainers.org/schema/bundle/root/path",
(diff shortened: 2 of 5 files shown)


specs-go/config.go
```diff
@@ -17,7 +17,7 @@ type Spec struct {
 	// Mounts configures additional mounts (on top of Root).
 	Mounts []Mount `json:"mounts,omitempty"`
 	// Hooks configures callbacks for container lifecycle events.
-	Hooks Hooks `json:"hooks"`
+	Hooks Hooks `json:"hooks,omitempty"`
```
wking: This is covered in #427.

specs-go/state.go
```diff
@@ -13,5 +13,5 @@ type State struct {
 	// BundlePath is the path to the container's bundle directory.
 	BundlePath string `json:"bundlePath"`
 	// Annotations are the annotations associated with the container.
-	Annotations map[string]string `json:"annotations"`
+	Annotations map[string]string `json:"annotations,omitempty"`
```
wking: The in-flight #507 is adding the associated REQUIRED and OPTIONAL wording to `runtime.md`.

schema/config-linux.json
```diff
@@ -89,7 +89,10 @@
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/pids/limit",
                                 "$ref": "defs.json#/definitions/int64"
                             }
-                        }
+                        },
+			"required": [
```
wking: Inconsistent indent.  Try running `make fmt` from the `schema` directory.

schema/config-schema.json
```diff
@@ -97,6 +118,10 @@
                 "user": {
                     "id": "https://opencontainers.org/schema/bundle/process/user",
                     "type": "object",
+                    "required": [
+			"uid",
+			"gid"
```
wking: These are [only required on Linux/Solaris][1], and there's no way to enforce that constraint in JSON Schema without some serious `oneOf` hoop-jumping.

[1]: https://github.com/opencontainers/runtime-spec/blob/ced3365b4d5856371a61cb9b0f9c2e69c3163643/config.md#linux-and-solaris-user

schema/config-schema.json
```diff
@@ -170,8 +195,6 @@
         "ociVersion",
         "platform",
         "process",
-        "root",
-        "mounts",
-        "hooks"
```
wking: Removing `hooks` is covered by the in-flight #427.

specs-go/state.go
q384566678: Since it is optional, it should not be so modified?

wking: Adding `omitempty` (as you do here and as #507 does not do) makes sense.  Adding REQUIRED/OPTIONAL to the Markdown spec (as #507 does and you do not do here) also makes sense.  Once both PRs land, we'll be in the right place ;).

crosbymichael:
LGTM

mrunalp:
LGTM

PR #647, opened 2017-01-05
config-linux: Add restriction for duplicated device path

Mashimiao:
I think runtime should generate an error, if devices has duplicated device path.
Because we don't know which one is really needed.

Signed-off-by: Ma Shimiao <njmiaoge@126.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -115,11 +115,14 @@ Each entry has the following structure:
 * **`type`** *(string, REQUIRED)* - type of device: `c`, `b`, `u` or `p`.
   More info in [mknod(1)][mknod.1].
 * **`path`** *(string, REQUIRED)* - full path to device inside container.
+  If a [file][file.1] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
 * **`major, minor`** *(int64, REQUIRED unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
 * **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
   You can also control access to devices [with cgroups](#device-whitelist).
 * **`uid`** *(uint32, OPTIONAL)* - id of device owner.
 * **`gid`** *(uint32, OPTIONAL)* - id of device group.
+
+The same `type`, `major` and `minor` SHOULD NOT be used for multiple devices.
 
 ###### Example
 
@@ -608,6 +611,7 @@ The values MUST be absolute paths in the [container namespace][container-namespa
 [cgroup-v2]: https://www.kernel.org/doc/Documentation/cgroup-v2.txt
 [devices]: https://www.kernel.org/doc/Documentation/devices.txt
 [devpts]: https://www.kernel.org/doc/Documentation/filesystems/devpts.txt
+[file.1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_164
 
 [mknod.1]: http://man7.org/linux/man-pages/man1/mknod.1.html
 [mknod.2]: http://man7.org/linux/man-pages/man2/mknod.2.html


config-linux.md
```diff
@@ -121,6 +121,9 @@ Each entry has the following structure:
 * **`uid`** *(uint32, OPTIONAL)* - id of device owner.
 * **`gid`** *(uint32, OPTIONAL)* - id of device group.
 
+If a `devices` contains duplicated devices with same `path`, the runtime MUST generate an error.
```
wking: Duplication in the `devices` array is a subset of this issue, which is that a path in `devices` already exists in the container filesystem being constructed.  Besides an earlier `devices` entry, this could also happen if a file (in the "everything is a file" sense) exists in the `root.path` directory at the target location, although you could also have conflicts sure to `mounts` entries.  I think this line should be under the `path` entry above with text like:

> If a [file][1] already exists at `path` that does not match the requested file, the runtime MUST generate an error instead of creating the requested file.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_164

Mashimiao: How about this?
`If a file already exists at `path` that does not match the requested device, the runtime MUST generate an error.`
We don't care what runtime already did now, we need it to generate an error and return.

config-linux.md
```diff
@@ -115,7 +115,9 @@ Each entry has the following structure:
 * **`type`** *(string, REQUIRED)* - type of device: `c`, `b`, `u` or `p`.
   More info in [mknod(1)][mknod.1].
 * **`path`** *(string, REQUIRED)* - full path to device inside container.
+  If a file already exists at `path` that does not match the requested device, the runtime MUST generate an error.
 * **`major, minor`** *(int64, REQUIRED unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
+  Same `major, minor` for different devices is acceptable, but it is not recommended.
```
wking: Maybe move this back outside the properties list and say something like:

> Using the same `type`, `major`, and `minor` for multiple `devices` entries is allowed but not recommend.

Mashimiao:
On 01/05/2017 10:24 PM, W. Trevor King wrote:
> Maybe move this back outside the properties list and say something like:
> 
>     Using the same |type|, |major|, and |minor| for multiple |devices| entries is allowed but not recommend.

It doesn't matter about duplicated `type`, we can add same `type` devices.
Noting same `type` is not recommended is not correct.

wking:
> It doesn't matter about duplicated `type`, we can add same `type` devices.

You can use the same `major` on multiple devices too.  What we're recommending is that you don't use the same `(type, major, minor)` tuple.

Mashimiao:
On 01/06/2017 09:43 AM, W. Trevor King wrote:
> You can use the same |major| on multiple devices too. What we're recommending is that you don't use the same |(type, major, minor)| tuple.

That's what I don't agree with.
The tuple (type, major, minor) will tell users only when all of the three items are same is not recommended.
Actually, what we don't recommend is same (major,minor) for multiple devices.

wking:
On Thu, Jan 05, 2017 at 05:57:23PM -0800, Ma Shimiao wrote:
> 01/06/2017 09:43 AM, W. Trevor King:
> > You can use the same |major| on multiple devices too. What we're
> > recommending is that you don't use the same |(type, major, minor)|
> > tuple.
> 
> That's what I don't agree with.  The tuple (type, major, minor) will
> tell users only when all of the three items are same is not
> recommended.  Actually, what we don't recommend is same
> (major,minor) for multiple devices.

I don't see any problem with a user creating a c15/0 (/dev/js0,
joystick) and a b15/0 (Sony CD-ROM) in the same ‘devices’ array [1].
Do you?

[1]: https://www.kernel.org/doc/Documentation/admin-guide/devices.txt

Mashimiao:
On 01/06/2017 10:35 AM, W. Trevor King wrote:
> I don't see any problem with a user creating a c15/0 (/dev/js0,
> joystick) and a b15/0 (Sony CD-ROM) in the same ‘devices’ array [1].
> Do you?

Of course not.
I have never said it is a problem.
I just think major,minor represents a device ID, it's better to be unique.
And I just want to recommend users to set them to be unique.
We have no need to care about whether type is same or not.
I'm not sure what do you mean.

wking:
On Thu, Jan 05, 2017 at 06:47:08PM -0800, Ma Shimiao wrote:
> I just think major,minor represents a device ID, it's better to be
> unique.  And I just want to recommend users to set them to be
> unique.  We have no need to care about whether type is same or not.

I'm fine with the uniqueness recommendation (although I'm also fine
without it [1]).  But (major, minor) does not represent a device ID.
In my earlier example [2], both c15/0 and b15/0 have the same majors
and minors, but one is a joystick and the other is a CD-ROM [2].  To
represent a device ID you need the whole (type, major, minor) tuple.

[1]: https://github.com/opencontainers/runtime-tools/pull/297#issuecomment-270441079
[2]: https://github.com/opencontainers/runtime-spec/pull/647#issuecomment-270816836

Mashimiao:
In [major.3](http://man7.org/linux/man-pages/man3/major.3.html), the description says a device ID consist of two parts: major, minor.
In my opinion, As ID, it should be unique. I'm not sure why it is not in Kernel.
I still think let major,minor to be unique is better for devices.
But if @opencontainers/runtime-spec-maintainers object, I can remove it in this PR.

wking:
On Thu, Jan 05, 2017 at 09:24:41PM -0800, Ma Shimiao wrote:
> In major.3, the description says a device ID consist of two parts:
> major, minor.  In my opinion, As ID, it should be unique.

(major, minor) is unique within a given file type.  That type just
gets passed into mknod(2) via a separate argument.

> … I can remove it in this PR.

I'm fine with removing the uniqueness suggestion, but I'm also fine
keeping it as long as the unique object is the full (type, major,
minor) tuple.

Mashimiao:
On 01/06/2017 01:44 PM, W. Trevor King wrote:
> (major, minor) is unique within a given file type. That type just
> gets passed into mknod(2) via a separate argument.

I think it's not.

$ sudo mknod test1 b 15 0
$ sudo mknod test2 b 15 0
$ ll
brw-r--r--  1 root      root       15,   0 Jan  6 13:57 test
brw-r--r--  1 root      root       15,   0 Jan  6 13:57 test2

I can create two node(15, 0) with file type b.

config-linux.md
```diff
@@ -608,6 +610,7 @@ The values MUST be absolute paths in the [container namespace][container-namespa
 [cgroup-v2]: https://www.kernel.org/doc/Documentation/cgroup-v2.txt
 [devices]: https://www.kernel.org/doc/Documentation/devices.txt
 [devpts]: https://www.kernel.org/doc/Documentation/filesystems/devpts.txt
+[file.1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_164
```
wking: [Trailing whitespace][1].

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/189381782#L293

Mashimiao: fixed, thanks

hqhq:
I agree the recommend against should to the whole (type, major, minor) tuple.

Mashimiao:
I'm not sure how Kernel manage devices currently.
From (here)[http://unix.stackexchange.com/questions/124225/are-the-major-minor-number-unique], it seems identify by type and major,minor is reasonable. But Kernel seems does not work like that.

Mashimiao:

hqhq:
LGTM

Mashimiao:
@mrunalp  @vbatts  @dqminh  PTAL

config-linux.md
```diff
@@ -115,12 +115,15 @@ Each entry has the following structure:
...
   More info in [mknod(1)][mknod.1].
 * **`path`** *(string, REQUIRED)* - full path to device inside container.
+  If a [file][file.1] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
 * **`major, minor`** *(int64, REQUIRED unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
 * **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
   You can also control access to devices [with cgroups](#device-whitelist).
 * **`uid`** *(uint32, OPTIONAL)* - id of device owner.
 * **`gid`** *(uint32, OPTIONAL)* - id of device group.
 
+Using same `type` and `major, minor` for mulitple devices is allowed, but not recommended.
```
wking: Maybe replace:

    `type` and `major, minor`

with:

    `type`, `major`, and `minor`

`major` and `minor` are separate properties, so I don't think we want them in the same monospace span.

wking: And "allowed but not recommended" sound like "SHOULD NOT".

dqminh: also typo at `mulitple`

Mashimiao:
@wking @RobDolinMS  @mrunalp  @hqhq   PTAL

config-linux.md
```diff
@@ -115,12 +115,15 @@ Each entry has the following structure:
...
   More info in [mknod(1)][mknod.1].
 * **`path`** *(string, REQUIRED)* - full path to device inside container.
+  If a [file][file.1] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
 * **`major, minor`** *(int64, REQUIRED unless **`type`** is `p`)* - [major, minor numbers][devices] for the device.
 * **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
   You can also control access to devices [with cgroups](#device-whitelist).
 * **`uid`** *(uint32, OPTIONAL)* - id of device owner.
 * **`gid`** *(uint32, OPTIONAL)* - id of device group.
 
+Same `type`, `major` and `minor` SHOULD NOT be used for multiple devices.
```
wking: nit: “Same” → “The same”.  Otherwise a78a4c1 looks good to me.

Mashimiao: fixed, thanks

hqhq:
LGTM

mrunalp:
LGTM

PR #648, opened 2017-01-05
Allow negative value for some resource fields

hqhq:
Carry #499

For these values, cgroup kernal APIs accept -1 to set
them as unlimited, as docker and runc all support
update resources, we should not set drawbacks in spec.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -283,15 +283,15 @@ For more information, see [the memory cgroup man page][cgroup-v1-memory].
 
 The following parameters can be specified to setup the controller:
 
-* **`limit`** *(uint64, OPTIONAL)* - sets limit of memory usage in bytes
+* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage in bytes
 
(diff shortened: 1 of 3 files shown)


config-linux.md
```diff
@@ -283,15 +283,15 @@ For more information, see [the memory cgroup man page][cgroup-v1-memory].
 
 The following parameters can be specified to setup the controller:
 
-* **`limit`** *(uint64, OPTIONAL)* - sets limit of memory usage in bytes
+* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage in bytes
```
wking: These properties are not tied directly to their backing cgroup property.  I think we should either:

a. Be explicit about the cgroup property which MUST be used to satisfy the configuration (in which case the kernel docs will explain the 0 and -1 cases) or
b. Explicitly define the 0 and -1 cases here.

tianon:
LGTM

IMO the additional clarity as to why this is an `int64` rather than a `uint64` that @wking points out would be excellent as well, but I don't personally think it's a blocker for this PR.

crosbymichael:
LGTM

PR #649, opened 2017-01-11
config-linux: Lift no-tweaking namespace restriction

### LINKED ISSUE #17: Allow Updates

### LINKED ISSUE #305: [Tracker] Live Container Updates

wking:
This restriction originally landed via #158.  The hostname case landed via #214, citing the namespace restriction.  The restriciton extended to runtime namespaces in #538.  There was also a proposal in-flight to get config-wide consistency around the no-tweaking concept (#540).

In today's meeting, the maintainer consensus was to [strike the no-tweaking restriction][2], which is what I've done here.  I've removed the ROADMAP entry because this gives folks a way to adjust existing containers (launch a new container which joins and tweaks the original).

The hostname entry still mentions the UTS namespace to provide a guard against accidental foot-gunning.  There was no no-tweaking language for properties related to other namespaces (e.g. 'mounts').  Maybe the other namespaces have more obvious names.

Fixes #17.
Fixes #305.

[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-01-11-22.04.log.html#l-117

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -18,14 +18,6 @@ Could be solved by lifecycle/ops and create/start split discussions
 
 *Owner:* vishh & duglin
 
-### Live Container Updates
-
-Should we allow dynamic container updates to runtime options?
-
-Proposal: make it an optional feature
-
-*Owner:* hqhq (was vishh) robdolinms, bcorrie
-
 ### Version Schema
 
 Decide on a robust versioning schema for the spec as it evolves.
--- a/config-linux.md
+++ b/config-linux.md
@@ -39,7 +39,6 @@ The following parameters can be specified to setup namespaces:
 
 If a path is specified, that particular file is used to join that type of namespace.
 If a namespace type is not specified in the `namespaces` array, the container MUST inherit the [runtime namespace](glossary.md#runtime-namespace) of that type.
-If a new namespace is not created (because the namespace type is not listed, or because it is listed with a `path`), runtimes MUST assume that the setup for that namespace has already been done and error out if the config specifies anything else related to that namespace.
 If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST error out.
 
 ###### Example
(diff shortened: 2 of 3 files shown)


mrunalp:
LGTM

crosbymichael:
LGTM

hqhq:
Why would this fix update issues? Update is more about tweaking cgroup resources.

wking:
On Wed, Jan 11, 2017 at 07:46:50PM -0800, Qiang Huang wrote:
> Why would this fix update issues? Update is more about tweaking
> cgroup resources.

You create a new config like:

  {
    …
    "linux": {
      "cgroupsPath": "/my/existing/container"
      "resources": {
        "pids": {
          "limit": 4096
        }
      }
    }
  }

and create/delete a container with that config.  That will join your
existing container's ‘pids’ controller, adjust the limit, and die
(leaving the adjusted value in place, although this is made clearer by
#651).

There's still no way to use the runtime join and tweak rlimits,
capabilities, etc. but you can call prlimit on the container process
without entering any container namespace, so I don't see a need to
involve the runtime in that.  However, I'm not sure if there are
prlimit analogs for capabilities, etc., and folks without a container
process ID (#459) might need some special tooling around this.  So
there's still work to be done, but *cgroup resources* seem like
they're sufficiently handled by this PR and #651.

PR #651, opened 2017-01-12
runtime: Clarify UTS and mount cleanup on 'delete'

wking:
Now that #649 allows us to get into this sort of situation.  This sort of ownership may also apply to other resources (cgroups?), but we can handle them in follow-up PRs (or here, I don't mind).

Also drop “Configuration” from the root header.  Everything in that file is a configuration.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -209,6 +209,43 @@ A container joins an existing [UTS namespace](config-linux.md#namespaces) on Lin
 
 When the example container is deleted, neither removing the preexisting UTS namespace nor undoing the hostname change are allowed.
 
+##### Joining cgroups (Linux)
+
+A container joins an existing [cgroup](config-linux.md#control-groups) on Linux and changes a [memory limit](config.md#memory) with the following [configuration](config.md):
+
+```json
+{
+    "ociVersion": "1.0.0-rc3",
+    "platform": {
+        "os": "linux",
+        "arch": "amd64"
+    },
+    "process": {
+        "cwd": "/",
+        "args": [
+            "sh"
+        ],
+        "user": {
+            "uid": 1,
+            "gid": 1
+        }
+    },
+    "root": {
+        "path": "rootfs"
+    },
+    "linux": {
+        "cgroupsPath": "/some/existing/container",
+        "resources": {
+            "memory": {
+                "limit": 100000
+            }
+        }
+    }
+}
+```
+
+When the example container is deleted, neither removing the preexisting cgroup not undoing the memory limit change are allowed.
+
 ## Hooks
 Many of the operations specified in this specification have "hooks" that allow for additional actions to be taken before or after each operation.
 See [runtime configuration for hooks](./config.md#hooks) for more information.
--- a/runtime.md
+++ b/runtime.md
@@ -121,11 +121,96 @@ When the process in the container is stopped, irrespective of it being as a resu
 This operation MUST generate an error if it is not provided the container ID.
 Attempting to delete a container that does not exist MUST generate an error.
 Attempting to delete a container whose process is still running MUST generate an error.
-Deleting a container MUST delete the resources that were created during the `create` step.
-Note that resources associated with the container, but not created by this container, MUST NOT be deleted.
 Once a container is deleted its ID MAY be used by a subsequent container.
 
+Deleting a container MUST delete the resources that were created during the `create` step.
+Resources associated with the container, but not created by this container, MUST NOT be altered.
+
+#### Examples of resource ownership
+
+This section contains informative elaborations of the above “resources created by the container” requirements.
+It is not intended to be exhaustive.
+
+##### Joining mount namespaces (Linux)
+
+A container joins an existing [mount namespace](config-linux.md#namespaces) on Linux and pivots [root](config.md#root) into the `rootfs` directory with the following [configuration](config.md):
+
+```json
+{
+    "ociVersion": "1.0.0-rc3",
+    "platform": {
+        "os": "linux",
+        "arch": "amd64"
+    },
+    "process": {
+        "cwd": "/",
+        "args": [
+            "sh"
+        ],
+        "user": {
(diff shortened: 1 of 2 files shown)


wking:
I've added a cgroup-ownership line and polished the earlier bits up slightly with abf8855 → 083f23e.

mrunalp:
LGTM

RobDolinMS:
@dqminh Want to take a look?  This needs another maintainer.

runtime.md
```diff
@@ -121,11 +121,18 @@ When the process in the container is stopped, irrespective of it being as a resu
...
-Note that resources associated with the container, but not created by this container, MUST NOT be deleted.
 Once a container is deleted its ID MAY be used by a subsequent container.
 
+Deleting a container MUST delete the resources that were created during the `create` step.
+Resources associated with the container, but not created by this container, MUST NOT be deleted.
+
+On Linux, [`mounts`](config.md#mounts), [`root`](config.md#root), [`linux.devices`](config-linux.md#devices), and other filesystem changes belong to the container which created the [container mount namespace][container-namespace3].
+If a container joins an existing mount namespace and applies those settings, deleting the container MUST NOT reverted the settings.
+Similarly [`hostname`](config.md#hostname) changes belong to the container which created the container UTS namespace.
+[`linux.resources`](config-linux.md#control-groups) settings belong to the container which created the associated cgroup.
```
dqminh: I dont know, this seems overly specified. I think the concept of "deleting resources owned only by the container" makes more sense, i.e. you dont delete changes to shared resources (i.e. when container make changes to a shared namespace ).

wking: > I think the concept of "deleting resources owned only by the container" makes more sense…

I tried to cover that generically in the [preceding paragraph][1].  This paragraph gives some explicit examples to illustrate the concept and avoid ambiguity about who owns a given resource.  Do you feel that the preceding paragraph is sufficiently clear without these more-detailed examples?

[1]: https://github.com/opencontainers/runtime-spec/pull/651/commits/2e7155815d47dc415ae0b76a69a47584d7784ca7#diff-b84a8d65d8ed53f4794cd2db7e8ea731R126

dqminh:
```
Deleting a container MUST delete the resources that were created during the `create` step.
Resources associated with the container, but not created by this container, MUST NOT be deleted.
```

For ownership concept, it is reasonably enough for me.

I think its not harmful to have more examples to illustrate the point, but they need to be clearly spelled out as example so they wont be mistook as the only cases that we will support.

WDYT about something like:

```
In the following example, when the container joins an existing mount namespace on Linux and changes root with the following config.json

{
root: /new/root,
linux: { namespaces: [ {type: mount, path: /proc/1234/ns/mnt} ] },
}

deleting the container MUST not remove the existing mount namespace, and MUST not undo the changes to root

In the following example, when the container joins an existing UTS namespace on Linux and changes hostname with the following config.json

{
hostname: runtimespec,
linux: { namespaes: [{type: uts, path: /proc/1234/ns/uts} ] }
}

deleting the container MUST not remove the existing UTS namespace, and MUST not undo the changes to hostname.
```

wking: > WDYT about something like…

I'm fine with a bit more hand-holding like that.  To be absolutely clear that these examples were informative (and not adding new normative restrictions), I've reworded them to avoid MUST and split them off into new subsections with 083f23e → 52a5199.  I've also added a similar cgroup example so we don't focus exclusively on namespaces.  And I've also filled out the configs because I don't like fragments (although we have an existing fragment example [here][1]).  I personally prefer the compactness of 083f23e, but maybe maintainers prefer 52a5199 or a sweet spot somewhere in between?  If you want something in the middle, please give me a list of corners to cut (e.g. “drop the subsections” or “use JSON fragments for the configs”).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config-linux.md#example-3

wking: And I've just pushed 52a5199 → eeacccd adding a new mount namespace to the UTS example, otherwise the (required) `root.path` entry would pivot the host mount namespace.

runtime.md
```diff
@@ -121,11 +121,133 @@ When the process in the container is stopped, irrespective of it being as a resu
...
+            {
+                "type": "mount",
+                "path": "/proc/1234/ns/mnt"
+            }
+        ]
+    }
+}
+```
+
+When the example container is deleted, neither removing the preexisting mount namespace nor undoing the pivot into `rootfs` are allowed.
```
hqhq: I think deleting namespace is a bit overly specified, it probably is not true since on Linux, namespace is destroyed when the last process that is a member of the namespace terminates.

So if we join a shared namespace and the original container exits, our container's terminates will delete this namespace, and that's something we should not interfere.

runtime.md
```diff
@@ -121,11 +121,133 @@ When the process in the container is stopped, irrespective of it being as a resu
...
 Attempting to delete a container that does not exist MUST generate an error.
 Attempting to delete a container whose process is still running MUST generate an error.
-Deleting a container MUST delete the resources that were created during the `create` step.
-Note that resources associated with the container, but not created by this container, MUST NOT be deleted.
 Once a container is deleted its ID MAY be used by a subsequent container.
 
+Deleting a container MUST delete the resources that were created during the `create` step.
+Resources associated with the container, but not created by this container, MUST NOT be altered.
+
+#### Examples of resource ownership
```
crosbymichael: I don't see why we should have examples like this in the spec at all.  Its too much overspecifying things and your descriptions are wrong anyways on how namespaces are "destroyed".

crosbymichael: My example I know we have examples in the repo showing off how to populate fields but examples like this describing functionality/implementation don't belong in my opinion.

crosbymichael:
Closing because we should not have implementation specific examples detailing functionality like this in the spec.  The original two sentences are enough and provide the context required for runtime authors on their platforms.

PR #652, opened 2017-01-12
defs.json: add misssing int64Pointer

Mashimiao:
int64Pointer is used in config-linux.json, but hasn't been defined.

### DIFF
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -56,6 +56,16 @@
                 }
             ]
         },
+        "int64Pointer": {
+            "oneOf": [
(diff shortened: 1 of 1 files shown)


wking:
On Thu, Jan 12, 2017 at 12:57:44AM -0800, Ma Shimiao wrote:
> int64Pointer is used in config-linux.json, but hasn't been defined.

Those references came in recently with #648, and fa74102 looks like a
good fix to me.

mrunalp:
LGTM

PR #655, opened 2017-01-13
schema: fix items based on latest spec

Mashimiao:

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -245,6 +245,10 @@
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/kernel",
                                 "$ref": "defs.json#/definitions/int64Pointer"
                             },
+                            "kernelTCP": {
+                                "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/kernelTCP",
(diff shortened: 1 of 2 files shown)


dqminh:
LGTM

schema/config-linux.json
```diff
@@ -245,6 +245,10 @@
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/kernel",
                                 "$ref": "defs.json#/definitions/int64Pointer"
                             },
+                            "kernelTCP": {
+                                "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/kernelTCP",
+                                "$ref": "defs.json#/definitions/int64Pointer"
```
wking: We definitely need a `kernelTCP` entry (catching up with #235), but after #555 I'm not sure we want to use pointers here.  The optional-ness is represented in JSON Schema by not listing these properties in a `required` array; we don't have to (or want to?) allow people to explicitly set a `null` value.

mrunalp:
LGTM

RobDolinMS:
@wking Can your comment be addressed in a separate PR?

vbatts:
LGTM

PR #657, opened 2017-01-14
config: Improve seccomp format to be more expressive

GrantSeltzer:
I'm opening this to continue [this discussion](https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/BkTgy0bZrXQ) about improving the seccomp specification. @mrunalp @runcom @justincormack

While of course runtime-spec shouldn't make changes with specific projects in mind, it's worth noting that this format is currently being used in [cri-o]() and [docker]().

The change proposes that Syscalls be grouped together in the case that their listings be exactly the same. Conditionals can be expressed on architectures and capabilities also. Comments are a nice feature for dev-ops teams as well.

There's inconsistencies ([example](https://github.com/opencontainers/runc/issues/1273)) in the use of seccomp specifications across oci projects, I think heading in the direction of this improvement can help standardize adoption of the spec. 

See the following example of this format: 
```
{
    "archMap": [{
        "architecture": "SCMP_ARCH_X86_64",
        "subArchitectures": [
            "SCMP_ARCH_X86",
            "SCMP_ARCH_X32"
        ]
    }],
    "syscalls": [{
        "names": [
            "accept",
            "accept4",
            "access",
            "alarm",
            "alarm",
            "bind",
            "brk",
            "capget",
            "capset",
            "chdir",
            "chmod"
        ],
        "action": "SCMP_ACT_ALLOW",
        "args": [],
        "comment": "this is a comment!",
        "includes": {},
        "excludes": {}
    }, {
        "names": [
            "iopl",
            "ioperm"
        ],
        "action": "SCMP_ACT_ALLOW",
        "args": [],
        "comment": "",
        "includes": {
            "caps": [
                "CAP_SYS_RAWIO"
            ]
        },
        "excludes": {}
    }, {
        "names": [
            "clone"
        ],
        "action": "SCMP_ACT_ALLOW",
        "args": [{
            "index": 1,
            "value": 2080505856,
            "valueTwo": 0,
            "op": "SCMP_CMP_MASKED_EQ"
        }],
        "comment": "",
        "includes": {
            "arches": [
                "s390",
                "s390x"
            ]
        },
        "excludes": {
            "caps": [
                "CAP_SYS_ADMIN"
            ]
        }
    }]
}
```

Signed-off-by: grantseltzer <grantseltzer@gmail.com>

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -365,13 +365,6 @@ type LinuxDeviceCgroup struct {
 	Access string `json:"access,omitempty"`
 }
 
-// LinuxSeccomp represents syscall restrictions
-type LinuxSeccomp struct {
-	DefaultAction LinuxSeccompAction `json:"defaultAction"`
-	Architectures []Arch             `json:"architectures"`
-	Syscalls      []LinuxSyscall     `json:"syscalls,omitempty"`
-}
-
 // Solaris contains platform specific configuration for Solaris application containers.
 type Solaris struct {
 	// SMF FMRI which should go "online" before we start the container process.
@@ -469,6 +462,13 @@ type WindowsNetworkResources struct {
 	EgressBandwidth *uint64 `json:"egressBandwidth,omitempty"`
 }
 
+// LinuxSeccomp represents syscall restrictions
+type LinuxSeccomp struct {
+	DefaultAction LinuxSeccompAction `json:"defaultAction"`
+	Architectures []Arch             `json:"architectures,omitempty"`
+	Syscalls      []LinuxSyscall     `json:"syscalls"`
+}
+
 // Arch used for additional architectures
 type Arch string
 
@@ -529,7 +529,8 @@ type LinuxSeccompArg struct {
 
 // LinuxSyscall is used to match a syscall in Seccomp
 type LinuxSyscall struct {
-	Name   string             `json:"name"`
-	Action LinuxSeccompAction `json:"action"`
-	Args   []LinuxSeccompArg  `json:"args,omitempty"`
+	Names   []string           `json:"names"`
+	Action  LinuxSeccompAction `json:"action"`
+	Args    []LinuxSeccompArg  `json:"args"`
+	Comment string             `json:"comment"`
 }
--- a/config-linux.md
+++ b/config-linux.md
@@ -538,12 +538,17 @@ Operator Constants:
    "seccomp": {
        "defaultAction": "SCMP_ACT_ALLOW",
        "architectures": [
-           "SCMP_ARCH_X86"
+           "SCMP_ARCH_X86",
+           "SCMP_ARCH_X32"
        ],
        "syscalls": [
            {
-               "name": "getcwd",
-               "action": "SCMP_ACT_ERRNO"
+               "names": [
+                   "getcwd",
+                   "chmod"
+               ],
+               "action": "SCMP_ACT_ERRNO",
+               "comment": "stop exploit x"
            }
        ]
    }
--- a/config.md
+++ b/config.md
@@ -710,12 +710,17 @@ Here is a full example `config.json` for reference.
         "seccomp": {
             "defaultAction": "SCMP_ACT_ALLOW",
             "architectures": [
-                "SCMP_ARCH_X86"
+                "SCMP_ARCH_X86",
+                "SCMP_ARCH_X32"
             ],
             "syscalls": [
                 {
-                    "name": "getcwd",
-                    "action": "SCMP_ACT_ERRNO"
+                    "names": [
+                        "getcwd",
+                        "chmod"
+                    ],
+                    "action": "SCMP_ACT_ERRNO",
+                    "comment": "stop exploit x"
                 }
             ]
         },
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -63,8 +63,10 @@
         "Syscall": {
             "type": "object",
             "properties": {
-                "name": {
-                    "type": "string"
+                "names": {
+                    "type": [
+                        "string"
+                    ]
                 },
                 "action": {
                     "$ref": "#/definitions/SeccompAction"


justincormack:
Sorry for not making a proposal yet...

This format is definitely better, and was deliberately designed to be fully backward compatible.

However, it is still impossible to express many valid seccomp programs using this format.

See https://github.com/seccomp/libseccomp/issues/44 for a discussion of some of the issues.

One option would be to allow `{"program":"base64 encoded bytes of bpf code"}` but it is not terribly human readable.

GrantSeltzer:
> One option would be to allow `{"program":"base64 encoded bytes of bpf code"}` but it is not terribly human readable.

Can you elaborate on where that bit would go in the specification? 

I read through the thread you linked, is the issue with seccomp itself or the programs that are generating/reading the seccomp profiles? Not

justincormack:
That would be an option to replace the entire seccomp part of the specification. The problem is with the JSON specification of the profiles which is defined by libseccomp, not with seccomp itself. In particular you cannot write rules that say "these values for this syscall are ok, but these ones are not".

GrantSeltzer:
So you're saying something like this doesn't work?

```
 "syscalls": [{
        "names": [
            "syscallA"
        ],
        "action": "SCMP_ACT_ALLOW",
        "args": [],
        "comment": "syscallA with no arguments, allowed",
        "includes": {},
        "excludes": {}
    }, {
        "names": [
            "syscallA"
        ],
        "action": "SCMP_ACT_ERRNO",
        "args": ["fakeArg","otherFakeArg"],
        "comment": "syscallA with some specific args, not allowed",
        "includes": {},
        "excludes": {}
    }
]}
```

mheon:
There are certain types of filters that can't be expressed via libseccomp's current syntax - the most common one would be embedding a blacklist within a whitelist (or vice versa), as is mentioned in the libseccomp issue @justincormack linked.

There may be other ways around this aside from directly inputting user-specified BPF (layered Seccomp filters, for example). I'd say this is an issue for another time, though. This is a definite improvement over the existing filter syntax, and is already being used by Docker in the wild. We should update the spec to match what's actually being used before we consider changing it again.

specs-go/config.go
```diff
@@ -469,6 +462,23 @@ type WindowsNetworkResources struct {
...
 }
 
+// LinuxSeccomp represents syscall restrictions
+type LinuxSeccomp struct {
+	DefaultAction LinuxSeccompAction `json:"defaultAction"`
+	// Architectures is kept to maintain backward compatibility for projects
+	// that already use the old seccomp profile.
+	Architectures []Arch          `json:"architectures,omitempty"`
+	ArchMap       []Architecture  `json:"archMap,omitempty"`
+	Syscalls      []*LinuxSyscall `json:"syscalls"`
```
wking: Why are the values becoming pointers here?  Is there a reason you'd want an explicit `null` in a `syscalls` array?

specs-go/config.go
```diff
@@ -469,6 +462,23 @@ type WindowsNetworkResources struct {
 	EgressBandwidth *uint64 `json:"egressBandwidth,omitempty"`
 }
 
+// LinuxSeccomp represents syscall restrictions
+type LinuxSeccomp struct {
+	DefaultAction LinuxSeccompAction `json:"defaultAction"`
+	// Architectures is kept to maintain backward compatibility for projects
+	// that already use the old seccomp profile.
+	Architectures []Arch          `json:"architectures,omitempty"`
+	ArchMap       []Architecture  `json:"archMap,omitempty"`
```
wking: `archMap` is a strange name for a property that is not an object/map.  Maybe call the new setting `architectures2`?  Or take advantage of the pre-1.0 lack of backwards compat (SemVer wording for pre-releases in [§9][1]) and just use `architectures` for the new field.  I'm not a maintainer, but my personal preference is for the latter, since then we don't have to define how to handle configs that set both properties simultaneously.

[1]: http://semver.org/spec/v2.0.0.html#spec-item-9

justincormack: yeah, it maps one architecture to others, I guess it might be confusing.

specs-go/config.go
justincormack: For backwards compatibility with the old format.

wking: The old format [does not use pointer values there][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/specs-go/config.go#L372

justincormack: oh yes. Hmm, maybe @runcom remembers

specs-go/config.go
GrantSeltzer: ArchMap threw me off at first as well. `Architectures2` doesn't really give it any meaning, and I'd prefer not to change the old field. `ArchAndSubArches`? `ArchHierarchy`?

wking: > Architectures2 doesn't really give it any meaning...

Like `dup2` (and 3 and 4) the meaning is "This is what we should have done for `architectures{n-1}`, but we aren't changing that because we're preserving backwards compatible".  But if `architectures` was poorly named to begin with (I don't know), picking a better name now makes sense.

wking: > If we implement this change, would there be precedence to eventually remove the old Architectures? As in give a few releases warning that it'll be deprecated?

Removing or changing semantics for currently-valid values for existing property requires a major version bump.  Of course, it doesn't have to be the *next* major version bump, but given the pace of the spec I don't see a reason to give a two+ major window.

And yeah, if we pick a different name it will look less silly once we drop the original.  My preference is still to drop the original now (before we cut 1.0), but if we intend to drop the original when we cut 2.0 that would be a reason to not choose `architectures2`.

specs-go/config.go
runcom: No reason behind those being pointers. When I refactored the seccomp profile in docker I just kept the pointers from their previous version (https://github.com/docker/docker/pull/26200/files#diff-05cb14d3dc2005d727f9d28cd0daece7L7)

GrantSeltzer: I can't think of an advantage to having them as pointers. As @wking pointed out, there's no time we'd want an explicit null. Going to go ahead and change that over.

GrantSeltzer:
@mrunalp PTAL

crosbymichael:
@justincormack do you have any feedback on this?  We think a change like this would be good to have in 1.0 as it greatly reduces the size of the spec and its a much nicer output.  

I think BPF support can come at a later date as its more additive than a schema change

specs-go/config.go
```diff
@@ -529,7 +539,28 @@ type LinuxSeccompArg struct {
 
 // LinuxSyscall is used to match a syscall in Seccomp
 type LinuxSyscall struct {
-	Name   string             `json:"name"`
-	Action LinuxSeccompAction `json:"action"`
-	Args   []LinuxSeccompArg  `json:"args,omitempty"`
+	Name     string            `json:"name,omitempty"`
```
crosbymichael: Why do we have a `Name` field here?

specs-go/config.go
```diff
@@ -529,7 +539,28 @@ type LinuxSeccompArg struct {
...
+	Names    []string          `json:"names,omitempty"`
+	Action   Action            `json:"action"`
+	Args     []LinuxSyscallArg `json:"args"`
+	Comment  string            `json:"comment"`
+	Includes Filter            `json:"includes"`
```
crosbymichael: How are the includes and excludes meant to be used?

specs-go/config.go
```diff
@@ -469,6 +462,23 @@ type WindowsNetworkResources struct {
...
+	Architectures []Arch         `json:"architectures,omitempty"`
+	ArchMap       []Architecture `json:"archMap,omitempty"`
+	Syscalls      []LinuxSyscall `json:"syscalls"`
+}
+
+// Architecture is used to represent an specific architecture
+// and its sub-architectures
+type Architecture struct {
+	Arch      Arch   `json:"architecture"`
+	SubArches []Arch `json:"subArchitectures"`
```
crosbymichael: How is subarches supposed to be used?

specs-go/config.go
GrantSeltzer: Conditionals are for arches and capabilities. Something like this:
```
{
        "names": [
            "clone"
        ],
        "action": "SCMP_ACT_ALLOW",
        "args": [{
            "index": 1,
            "value": 2080505856,
            "valueTwo": 0,
            "op": "SCMP_CMP_MASKED_EQ"
        }],
        "comment": "",
        "includes": {
            "arches": [
                "s390",
                "s390x"
            ]
        },
        "excludes": {
            "caps": [
                "CAP_SYS_ADMIN"
            ]
        }
    }
```

specs-go/config.go
GrantSeltzer: For backwards compatibility. This is another example of the conversation above, we have to decide at what point should we break previous versions, if at all.

specs-go/config.go
GrantSeltzer: Something like this:

```
    "archMap": [{
        "architecture": "SCMP_ARCH_X86_64",
        "subArchitectures": [
            "SCMP_ARCH_X86",
            "SCMP_ARCH_X32"
        ]
    }],
```

crosbymichael: But what is the usecase for it?

specs-go/config.go
crosbymichael: I think its already broken with the change so we can remove it

specs-go/config.go
justincormack: The use case is that people often want to configure amd64 machines to allow i386 images to run and if you want to do this you must specify, or it will be disallowed. This is mainly here so you can specify all the combinations, so that you can ship a single seccomp profile for multiple architectures, rather than having to package a different one for each architecture.

justincormack:
@crosbymichael yes, I think it makes the spec more compact and portable so we should have it for 1.0

specs-go/config.go
justincormack: @crosbymichael we managed to make the format fully backward compatible when we changed it in Docker. We didnt deprecate anything, but I guess we can fix it up in Docker.

specs-go/config.go
justincormack: I dont know that the conditionals should be in the spec, I think they should be dealt with upstream.

specs-go/config.go
justincormack: This should probably not be in the spec either - the fixed form is better, as it is an exact spec.

justincormack:
So basically the aggregation parts (`names`) should be there, but not the conditional parts...

specs-go/config.go
crosbymichael: Then should it just be a single field or an array?

specs-go/config.go
crosbymichael: Ya, i'm not sure if we should specifies these conditional filters.  You can handle this in the caller and not add the syscall if you don't want it.

mrunalp: @justincormack Not sure what you mean by upstream. Wouldn't be want filters that disable syscalls only for certain arguments?

crosbymichael: @mrunalp there are no such things in libseccomp to support these conditionals (includes, excludes), it has to be implemented by the runtime author.  Instead of having every runtime handle this, the runtime author should just not add the rule if they don't want it based on setting a cap.

specs-go/config.go
runcom: @justincormack @crosbymichael so "archMap" should just be an array right?

specs-go/config.go
runcom: +1 to deprecate this is docker to get rid of this field

specs-go/config.go
justincormack: I dont see we need `ArchMap`. The list in `Architectures` is sufficient.

specs-go/config.go
GrantSeltzer: Ok, removing the `Name` field

specs-go/config.go
GrantSeltzer: Ok, Removing conditionals

specs-go/config.go
GrantSeltzer: Agreed, removing  `ArchMap`

GrantSeltzer:
@crosbymichael @mrunalp @runcom 

Made the changes from feedback, ready for further review/discussion

specs-go/config.go
```diff
@@ -529,7 +529,19 @@ type LinuxSeccompArg struct {
 
 // LinuxSyscall is used to match a syscall in Seccomp
 type LinuxSyscall struct {
-	Name   string             `json:"name"`
-	Action LinuxSeccompAction `json:"action"`
-	Args   []LinuxSeccompArg  `json:"args,omitempty"`
```
wking: You seem to be replacing `LinuxSeccompArg` with `LinuxSyscallArg`, but that leaves a dangling `LinuxSeccompArg` type (at least as of c36c819):

```
$ git grep LinuxSeccompArg origin/pr/657
origin/pr/657:specs-go/config.go:// LinuxSeccompArg used for matching specific syscall arguments in Seccomp
origin/pr/657:specs-go/config.go:type LinuxSeccompArg struct {
```

Also, we'll need to update the Markdown spec to cover this change, since there is definitely a schema change going on and the Go types are not normative.  The JSON Schema will need updating as well.

```
$ git show --stat --oneline origin/pr/657
c36c819 improve seccomp format to be more expressive
 specs-go/config.go | 32 ++++++++++++++++++++++----------
 1 file changed, 22 insertions(+), 10 deletions(-)
```

specs-go/config.go
```diff
@@ -529,7 +529,8 @@ type LinuxSeccompArg struct {
...
+	Names   []string           `json:"names,omitempty"`
```
crosbymichael: You probably no longer need an `omitempty` here now that `Name` has been removed.

GrantSeltzer: Good callout

specs-go/config.go
GrantSeltzer: Fixed that, did not mean for that replacement.

Updated markdown as well.

crosbymichael:
LGTM

mrunalp:
LGTM

GrantSeltzer:
Thanks for the merge!

PR #658, opened 2017-01-17
[Linux Config] Clarify which filesystem is being referenced

RobDolinMS:
Replaces PR #577


RobDolinMS:
I think the problem is in my sign-off name.  I'll try another PR.

PR #659, opened 2017-01-18
change the type of hugepageLimits.limit to uint64

Kxuan:
The limit should not be lower than 0.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -414,7 +414,7 @@ Each entry has the following structure:
 
 * **`pageSize`** *(string, REQUIRED)* - hugepage size
 
-* **`limit`** *(int64, REQUIRED)* - limit in bytes of *hugepagesize* HugeTLB usage
+* **`limit`** *(uint64, REQUIRED)* - limit in bytes of *hugepagesize* HugeTLB usage
 
 ###### Example
 


wking:
On Tue, Jan 17, 2017 at 11:21:54PM -0800, Xuan wrote:
> The limit should not be lower than 0.
> …
>   * change the type of hugepageLimits.limit to uint64

This was just changed from a uint64 to an int64 in #648, although
we're still missing docs explaining what -1 is for [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/648#discussion_r94778793

Kxuan:
Ok. I get it.

I'm trying to add hugepageLimit to runtime-tools, but the type of HugepageLimit.Limit is different between [runtime-tools](https://github.com/opencontainers/runtime-tools/blob/master/Godeps/_workspace/src/github.com/opencontainers/runtime-spec/specs-go/config.go#L201) and [runtime-spec](https://github.com/opencontainers/runtime-tools/blob/master/Godeps/_workspace/src/github.com/opencontainers/runtime-spec/specs-go/config.go#L201).
I think we need to find some method to sync the runtime-spec and runtime-tools.

Thanks

wking:
On Tue, Jan 17, 2017 at 11:45:43PM -0800, Xuan wrote:
> I think we need to find some method to sync the runtime-spec and
> runtime-tools.

@crosbymichael has called for a spec-schema feature freeze in the near
future [1], although as far as I know the runtime-spec maintainers
have not officially put a date on that yet.  If/when that happens, it
will give runC and other Go consumers time to catch up anything they
don't vendor directly from runtime-spec (or to bump their vendored
version if they do vendor runtime-spec).

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-01-11-22.04.log.html#l-33

PR #661, opened 2017-01-18
[Runtime] Allow for excluding pid from state

RobDolinMS:
Replaces #459

Signed-off-by: Rob Dolin (MSFT) <robdolin@microsoft.com>

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -21,6 +21,7 @@ The value MAY be one of:
 
     Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the container process, as seen by the host.
+The pid MAY be excluded or set to a value indicating empty or null if it is not seen by the host. 
 * **`bundlePath`**: (string) is the absolute path to the container's bundle directory.
 This is provided so that consumers can find the container's configuration and root filesystem on the host.
 * **`annotations`**: (map) contains the list of annotations associated with the container.


runtime.md
```diff
@@ -21,6 +21,7 @@ The value MAY be one of:
 
     Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`**: (int) is the ID of the container process, as seen by the host.
+The pid MAY be excluded or set to a value indicating empty or null if it is not seen by the host.
```
dqminh: i wonder if its useful to further clarified what it means by "pid is not seen by the host". I only understand this phrase when reading background of https://github.com/opencontainers/runtime-spec/pull/459/files.

Perhaps something like:

```
pid MAY be excluded when the container's kernel is isolated from the host / or the host does not manage the container pids
```

hqhq:
@RobDolinMS Any specific reasons that you can't do rebase in https://github.com/opencontainers/runtime-spec/pull/459 ? I'll be better to do base there so we can keep good tracking.

runtime.md
wking: nit: there's some [trailing whitespace][1] on the current line.

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/193148351#L293

crosbymichael:
Please rebase your original PRs and not create new ones.

Thanks! ;)

PR #545, opened 2016-08-29
fix blkio related validation

Mashimiao:
Signed-off-by: Ma Shimiao mashimiao.fnst@cn.fujitsu.com

PR #662, opened 2017-01-19
schema: Drop pointers and nulls

wking:
Maintainers feel (and I agree) that there's [no point in explicitly allowing a null value when callers can simply leave the property unset][1].  This commit removes all references to “pointer” and “null” from the JSON Schema to support that decision.  While optional properties [may sometimes be represented as pointer types in Go][2], optional properties should be represented in JSON Schema by not including the properties in the `required` array.

Spun off from [this comment][3] as [requested][4] by @RobDolinMS.

[2]: https://github.com/opencontainers/runtime-spec/blob/c368be6a74d7383f9b6c9a4b7a3e73d65537b08a/style.md#optional-settings-should-not-have-pointer-go-types
[3]: https://github.com/opencontainers/runtime-spec/pull/655#discussion_r96018318
[4]: https://github.com/opencontainers/runtime-spec/pull/655#issuecomment-273622407

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -72,11 +72,11 @@
                         "properties": {
                             "blkioWeight": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/blockIO/blkioWeight",
-                                "$ref": "defs-linux.json#/definitions/blkioWeightPointer"
+                                "$ref": "defs-linux.json#/definitions/blkioWeight"
                             },
                             "blkioLeafWeight": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/blockIO/blkioLeafWeight",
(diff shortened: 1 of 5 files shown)


Mashimiao:
This is part duplicated with #545

wking:
> This is part duplicated with #545

Ah, yup. Happy to rebase if that lands first.

wking:
On Sun, Jan 22, 2017 at 08:23:12PM -0800, W. Trevor King wrote:
> > This is part duplicated with #545
> 
> Ah, yup. Happy to rebase if that lands first.

And #545 has landed, so I've rebased this around it with b554d2e →
0927437.

mikebrow:
Reviewed looks good.

crosbymichael:
LGTM

hqhq:
LGTM

PR #664, opened 2017-01-22
runtime: Only require 'pid' in the state for created/running statuses

wking:
Because during creation (before `created`) we may not have a container process yet (e.g. if we're still reading the configuration or setting up cgroups), and in the `stopped` phase the PID is no longer meaningful.

Also add OPTIONAL/REQUIRED and remove colons for consistency with the `config.md`.

This documents (as best I can tell) the current semantics and gives #459 / #661 a well-defined REQUIREment to relax.

It's also the tip commit from #507 (currently at 08d52a0) cherry-picked onto master.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -7,24 +7,26 @@ Whether other entities using the same, or other, instance of the runtime can see
 
 ## State
 
-The state of a container MUST include, at least, the following properties:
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

crosbymichael:
LGTM

Issue #663, opened 2017-01-20
Should Terminal be in the process spec?


crosbymichael:
We have had this for a while and I'm starting to question if this is the best place for this.  Because of the console work and much of this has to be coordinated outside of the runtime and container I don't know if the spec is the right place to have this.  

As an example, the shim with containerd, we have to create the socket and wait to get the pty master back from the runtime, the only way the shim would know if it is supposed to create this socket and wait is if:

a. the higher level pass down the information that you should get a console from the container
b. we read the spec 

Reading the spec id kinda weird when the higher layers make the spec and give it to you, you open and unmarshal it to figureout this one field, then hand it off to the runtime.  

It is more ideal if the higher levels just say, "hey i want a terminal with this container" and the shim can just launch runc with `--tty` or `--console-socket` and if this is specified then runc creates one, if not, it does not.   This information is not split between the spec and higher levels.  

What do you all think?  Have you ran into this same issue implementing the spec or high level systems with runc or other runtimes? 

@mlaventure wdyt?

mlaventure:
I personally think that the `Terminal` is not a property of a process, it's more the "environment" in which it runs.

All binaries that cares about having a terminal check at startup what's their current IOs points to.

So :+1: from me to move outside the spec and let the users/tools provide a console if so desired.

wking:
I'm fine dropping it from the config, but think the command-line API should cover how to set up a terminal (since we still want the runtime doing the terminal-creation inside the container namespace).  And this issue may be a dup of #494, where @crosbymichael [floated dropping the `terminal` property in a comment][1].

[1]: https://github.com/opencontainers/runtime-spec/issues/494#issuecomment-225251219

tianon:
@RobDolinMS can you comment regarding Windows implications? (especially since "console size" is something that was added to the spec explicitly for Windows initially, IIRC)

IMO, it is kind of nice to be able to have the bundle author note that the process they've configured as the "container process" requires a TTY, but can definitely see the value from the `runc` perspective on punting that to the higher layers (since it needs to be able to coordinate, as you described); definitely think it'd be good to get comments from other runtime authors actually implementing the spec (mostly just wanted to ping Rob for explicit Windows comments but figured I might as well throw in my own thoughts while I was at it).

hqhq:
> So 👍 from me to move outside the spec and let the users/tools provide a console if so desired.

[...]

Issue #24, opened 2015-06-25
linux-config: clean-up namespaces docs


philips:
The docs don't reflect things we discussed in-person:

- Adding the path field
- Making the namespaces OS specific
- Plumbing through the hooks to make changes to the namespace

@crosbymichael volunteered for this one I think.

mrunalp:
@philips I can take on the hooks portion of it (talked with @crosbymichael today and he told me look into it). I have assigned myself to #20

philips:
@mrunalp Can you also document the restrictions on joining namespaces from a new user namespace that was discussed? I believe @LK4D4 had details and maybe even linux code/doc references we can use.

mrunalp:
@philips Yes, sure. @LK4D4 and I discussed how a hook would have to be provided to allow customization of other namespaces from a user namespace. @LK4D4 could help review that :)

philips:
@mrunalp Can you get a draft of your current thinking up sometime tomorrow?

mrunalp:
There are PRs open for the first two items for Linux. I will put up a draft for the namespaces tomorrow.

vbatts:
discussed during review.

PR #666, opened 2017-01-23
linux: relax filesystem requirements for container

dqminh:
change MUST to SHOULD so containers are not required to have all these filesystems mounted.

Related to discussions in https://github.com/opencontainers/runc/pull/1176

cc @cyphar @crosbymichael @philips @justincormack

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -8,7 +8,7 @@ The Linux container specification uses various kernel features like namespaces, 
 The Linux ABI includes both syscalls and several special file paths.
 Applications expecting a Linux environment will very likely expect these file paths to be setup correctly.
 
-The following filesystems MUST be made available in each container's filesystem:
+The following filesystems SHOULD be made available in each container's filesystem:
 
 |   Path   |  Type  |
 | -------- | ------ |


justincormack:
I believe `/proc` MUST still be supplied, and `/dev/pts` in the case where you specify a console, and `/sys` if you specify `sysctl` config. Also `/dev` seems curiously missing...

wking:
On Mon, Jan 23, 2017 at 08:00:44AM -0800, Justin Cormack wrote:
> Also `/dev` seems curiously missing...

It does not have to be a separate filesystem.  I'd floated devtmpfs in
the past [1,2], but folks can use other mechanisms to provide the
required devices [3].

If we're making these mounts optional for security reasons [4], we'd
be overturning an earlier decision on adding these (without a way to
opt-out [5]).  For example, @philips has “considered making it
optional in appc but I don't think it hurts anything to have it” for
/dev/shm [6].

And if we do lift the requirement for these mounts, I'd rather just
drop the “Default Filesystems” section entirely.  Users who need these
mounts would have to explicitly set them up in their configuration (as
runtime-tools continues to do, see opencontainers/runtime-tools#24) if
they wanted to guarantee their presence.  Users who don't need them
can omit them from their configuration.  I don't see how keeping a
SHOULD-strength section (vs. dropping it completely) helps either of
those workflows.

[1]: https://github.com/opencontainers/runtime-spec/issues/95#issuecomment-131265209
[2]: https://github.com/opencontainers/runtime-spec/pull/164#issuecomment-138978263
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config-linux.md#default-devices
[4]: https://github.com/opencontainers/runc/pull/1176#issuecomment-263366010
[5]: https://github.com/opencontainers/runtime-spec/issues/95#issuecomment-128175117
[6]: https://github.com/opencontainers/runtime-spec/issues/95#issuecomment-128189315

dqminh:
> I believe /proc MUST still be supplied, and /dev/pts in the case where you specify a console, and /sys if you specify sysctl config.

Makes sense. WDYT about:

- /proc must be mounted
- /sys must be mounted if sysfs config is set
- /dev/pts must be available if terminal= true ( dont we want to take it out of process, i dont quite remember ?? )
- leave out /dev/shm and the rest. I dont actually had any special requirements to not have /dev/shm, but i also feel like `you use shm do you need /dev/shm. runc is a low level tool and should allow for weird use cases. In particular if you want to totally lock down ability to write to files by having no writeable mounts you want to remove /dev/shm` makes sense

justincormack:
@wking `/dev/shm` is writeable (if it is not writeable it cannot be used for shm), and therefore can be used to fill a chunk of memory, possibly enough to cause your actual processes to crash if an attacker found a way to write to it. Without this you can have a totally read only container. So I think it is worthwhile to be able to remove it.

wking:
On Mon, Jan 23, 2017 at 09:38:46AM -0800, Daniel, Dao Quang Minh wrote:
> - /dev/pts must be available if terminal= true ( dont we want to
>   take it out of process, i dont quite remember ?? )

You may be thinking about #663.

> - leave out /dev/shm and the rest. I dont actually had any special
>   requirements to not have /dev/shm…

If we're rolling back the earlier “you won't need to opt out” position
in one case, I feel like we should roll it back for all of them.
Explicitly declaring these mounts is not particularly complicated
(opencontainers/runtime-spec#24) and we've already been wrong about
the need for opting-out once.

Kxuan:
> /sys must be mounted if sysfs config is set

sysctl writes value to /proc/sys, not /sys. so it is not needed to mount sysfs to /sys.

gregkh:
@Kxuan   sysctl is not sysfs, please don't get those confused.   sysfs should be mounted at /sys, if the container needs it.  Depends on what is running in it.

Kxuan:
@gregkh yes, I know it. but they are discussing about "/sys if you specify sysctl config."

cyphar:
@Kxuan `sysctl`s are set through the `/proc/sys` virtual filesystem -- not `/sys`. The two are _very_ different (and @gregkh was one of the people who authored `/sys`'s `sysfs` so he definitely knows what he's talking about :wink:).

crosbymichael:
LGTM

The spec does not specify what you MUST populate in your own spec.  If I'm writing a spec I can add whatever I want in it.  This section is meant as a hint to spec authors that these filesystems should be made available.  

We need to stick to specing the schema for the fields in the spec and not spec what users put in it.

wking:
>  This section is meant as a hint to spec authors that these filesystems should be made available.

With the phrasing in 279c3c0 it looks like we're going from “the runtime MUST supply these even if the config doesn't call for them in `mounts`” to “the runtime SHOULD supply these even if the config doesn't call for them in `mounts`” (and I [don't think we want that][1]).  If the intention is actually to go to “config authors probably want to include `mounts` entries for these”, then I think we need to make that more explicit.  Something like:

> On Linux, configuration authors are likely to want `mounts` to contain some of the following:
>
> [entries from opencontainers/runtime-tools#24]

[1]: https://github.com/opencontainers/runtime-spec/pull/666#issuecomment-274550344

mrunalp:
LGTM

PR #670, opened 2017-01-25
schema: allow Device.fileMode set to null.

Kxuan:
The `fileMode` is marked optional in the [description](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices)
In [spec-go](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L347), fileMode can be set to null.
However, in the schema, a null type is not allowed for fileMode.

This patch allows fileMode set to null, and consistent with other optional fields.

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -117,7 +117,14 @@
                     "$ref": "defs.json#/definitions/FilePath"
                 },
                 "fileMode": {
-                    "$ref": "#/definitions/FileMode"
+                    "oneOf": [
+                        {
+                            "$ref": "#/definitions/FileMode"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ]
                 },
                 "major": {
                     "$ref": "#/definitions/Major"


wking:
On Tue, Jan 24, 2017 at 06:04:26PM -0800, Xuan wrote:
> In spec-go, fileMode can be set to null.  However, in the schema, a
> null type is not allowed for fileMode.

spec-go has it as an omitempty pointer, so instead of setting it to
null in the JSON, you should not set the property at all.  More on
this in #662 and in [1].

Kxuan:
> specs-go has it as an omitempty pointer, so instead of setting it to
null in the JSON, you should not set the property at all.  More on
this in #662 and in [1].

@wking ok, but I think we should make the optional fields unified. All of them could accept null or all of them could not accept null.

wking:
> I think we should make the optional fields unified. All of them could accept null or all of them could not accept null.

I agree.  #662 removes all instances of `null` from the JSON Schema, after which it will be consistent.

mikebrow:
So we're closing this PR correct?

Issue #303, opened 2016-01-13
[Tracker] Base Config Compatibility 


RobDolinMS:
(copied from roadmap.md)
Ensure that the base configuration format is viable for various platforms
Systems:
- Solaris
- Windows
- Linux

vbatts:
We'll need some folks to contribute the technical requirements for these platforms, as Linux is what is currently getting our attention.

RobDolinMS:
Currently comparing the Spec structs from:
* https://github.com/docker/containerd/blob/master/vendor/src/github.com/opencontainers/specs/specs-go/config.go
* https://github.com/docker/docker/blob/master/libcontainerd/windowsoci/oci_windows.go

RobDolinMS:
Have been in-touch with Jesse from Oracle and expecting more info soon.

hqhq:
Looks like this needs to be addressed before 1.0, I'm adding it to 1.0 milestone.

jlbutler:
@RobDolinMS as you're heading this effort up, can you make a call on whether or not the final round of cleanup in PR #673 is enough to close the issue out? If not, please let me know what else I can do to wrap this up.

crosbymichael:
Since #673 was merged I'm going to close this as resolved.  If this is not the case please open actionable issues that can be addressed.

PR #673, opened 2017-01-30
narrative cleanup in support of Base Config Compatibility #303

jlbutler:
Some minor narrative nits cleaned up in support of resolving issue #303. Most changes attempt to make it clear that multiple platforms are supported but keep the Linux examples and references in place. In the Process section, Linux-only options were left under the original header, anything which is supported by at least one other platform was pulled out.

### DIFF
--- a/config.md
+++ b/config.md
@@ -2,11 +2,15 @@
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
 The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).
+Platform-specific configuration schema are defined in the [platform-specific documents](#platform-specific-configuration) linked below.
 
 The configuration file contains metadata necessary to implement standard operations against the container.
 This includes the process to run, environment variables to inject, sandboxing features to use, etc.
 
-Below is a detailed description of each field defined in the configuration format.
+Below is a detailed description of each field defined in the configuration format and valid values are specified.
+Platform-specific fields are identified as such.
+For all platform-specific configuration values, the scope defined below in the [Platform-specific configuration](#platform-specific-configuration) section applies.
+
 
 ## Specification version
 
@@ -22,11 +26,11 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 
 ## Root Configuration
 
-**`root`** (object, REQUIRED) configures the container's root filesystem.
+**`root`** (object, REQUIRED) specifies the container's root filesystem.
 
 * **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
-  The path can be an absolute path (starting with /) or a relative path (not starting with /), which is relative to the bundle.
-  For example (Linux), with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
+  The path is either an absolute path or a relative path to the bundle.
+  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
   A directory MUST exist at the path declared by the field.
 * **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false.
 
@@ -41,25 +45,27 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 
 ## Mounts
 
-**`mounts`** (array, OPTIONAL) configures additional mounts (on top of [`root`](#root-configuration)).
+**`mounts`** (array, OPTIONAL) specifies additional mounts beyond [`root`](#root-configuration).
 The runtime MUST mount entries in the listed order.
-The parameters are similar to the ones in [the Linux mount system call](http://man7.org/linux/man-pages/man2/mount.2.html).
-For Solaris, the mounts corresponds to fs resource in zonecfg(8).
+For Linux, the parameters are as documented in [the mount system call](http://man7.org/linux/man-pages/man2/mount.2.html).
+For Solaris, the mount entry corresponds to the 'fs' resource in zonecfg(8).
+For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [SetVolumeMountPoint](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365561(v=vs.85).aspx).
+
 
 * **`destination`** (string, REQUIRED) Destination of mount point: path inside container.
   This value MUST be an absolute path.
-  For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
-  For the Solaris operating system, this corresponds to "dir" of the fs resource in zonecfg(8).
+  * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
+  * Solaris: corresponds to "dir" of the fs resource in zonecfg(8).
 * **`type`** (string, REQUIRED) The filesystem type of the filesystem to be mounted.
-  Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
-  Windows: ntfs.
-  Solaris: corresponds to "type" of the fs resource in zonecfg(8).
+  * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  * Windows: the type of file system on the volume, e.g. "ntfs".
+  * Solaris: corresponds to "type" of the fs resource in zonecfg(8).
 * **`source`** (string, REQUIRED) A device name, but can also be a directory name or a dummy.
-  Windows: the volume name that is the target of the mount point, \\?\Volume\{GUID}\ (on Windows source is called target).
-  Solaris: corresponds to "special" of the fs resource in zonecfg(8).
+  * Windows: the volume name that is the target of the mount point, \\?\Volume\{GUID}\ (on Windows source is called target).
+  * Solaris: corresponds to "special" of the fs resource in zonecfg(8).
 * **`options`** (list of strings, OPTIONAL) Mount options of the filesystem to be used.
-  Linux: [supported][mount.8-filesystem-independent] [options][mount.8-filesystem-specific] are listed in [mount(8)][mount.8].
-  Solaris: corresponds to "options" of the fs resource in zonecfg(8).
+  * Linux: [supported][mount.8-filesystem-independent] [options][mount.8-filesystem-specific] are listed in [mount(8)][mount.8].
+  * Solaris: corresponds to "options" of the fs resource in zonecfg(8).
 
 ### Example (Linux)
 
@@ -93,8 +99,6 @@ For Solaris, the mounts corresponds to fs resource in zonecfg(8).
 ]
 ```
 
-See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [SetVolumeMountPoint](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365561(v=vs.85).aspx) in Windows.
-
 ### Example (Solaris)
 
 ```json
@@ -113,13 +117,12 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 ]
 ```
 
-
 ## Process
 
-**`process`** (object, REQUIRED) configures the container process.
+**`process`** (object, REQUIRED) specifies the container process.
 
 * **`terminal`** (bool, OPTIONAL) specifies whether a terminal is attached to that process, defaults to false.
-  On Linux, a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
+  As an example, if set to true on Linux a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
 * **`consoleSize`** (object, OPTIONAL) specifies the console size of the terminal if attached, containing the following properties:
   * **`height`** (uint, REQUIRED)
   * **`width`** (uint, REQUIRED)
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -1,7 +1,7 @@
 # Container Configuration file
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
-The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).
+The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go). [**`Platform-specific`**](#platform) configuration schema are defined in platform-specific documents.
```
wking: There is already some documentation for this idea [here][1].  If you think it needs a mention in the top of this file, I'd consider just adding a link to the local `#platform-specific-configuration` (which actually links to the platform-specific documentation).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config.md#platform-specific-configuration

config.md
```diff
@@ -119,7 +118,6 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 **`process`** (object, REQUIRED) configures the container process.
 
 * **`terminal`** (bool, OPTIONAL) specifies whether a terminal is attached to that process, defaults to false.
-  On Linux, a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
```
wking: This language is new in #518; we probably don't want to remove it in this PR without addressing why directly.  There's also the recent #663 which might end up removing `process.terminal` completely.

config.md
```diff
@@ -52,7 +51,7 @@ For Solaris, the mounts corresponds to fs resource in zonecfg(8).
   For the Solaris operating system, this corresponds to "dir" of the fs resource in zonecfg(8).
 * **`type`** (string, REQUIRED) The filesystem type of the filesystem to be mounted.
   Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
-  Windows: ntfs.
+  Windows: corresponds to the type of file system on the volume, e.g. 'ntfs'.
```
wking: I would be much happier if we could link to some Windows docs listing allowed types here.  “e.g. `ntfs`” doesn't help with compliance testing (which types should get checked before a runtime is deemed compliant?).

config.md
```diff
@@ -25,8 +25,7 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 **`root`** (object, REQUIRED) configures the container's root filesystem.
 
 * **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
-  The path can be an absolute path (starting with /) or a relative path (not starting with /), which is relative to the bundle.
-  For example (Linux), with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
+  The path can is either an absolute path or a relative path to the bundle. On Linux for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
```
wking: You're dropping the [one sentence per line][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/style.md#one-sentence-per-line

config.md
```diff
@@ -128,27 +126,25 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-
-For Linux-based systems the process structure supports the following process specific fields:
-
-* **`capabilities`** (array of strings, OPTIONAL) capabilities is an array that specifies Linux capabilities that can be provided to the process inside the container.
-  Valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html).
+* **`capabilities`** (array of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html).
```
wking: I think you need to say something like “OPTIONAL on Linux and Solaris, undefined on Windows” or something to get that idea across.  Also, if other platforms besides Linux support `process.capabilities` we need to link to their documentation for allowed values, otherwise compliance testing is difficult.

config.md
```diff
@@ -128,27 +126,25 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
...
 * **`rlimits`** (array of objects, OPTIONAL) allows setting resource limits for a process inside the container.
   Each entry has the following structure:
 
-    * **`type`** (string, REQUIRED) - the 'type' field are the resources defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html).
-    * **`soft`** (uint64, REQUIRED) - the value that the kernel enforces for the corresponding resource.
-    * **`hard`** (uint64, REQUIRED) - the ceiling for the soft limit that could be set by an unprivileged process.
-      Only privileged process (under Linux: one with the CAP_SYS_RESOURCE capability) can raise a hard limit.
+    * **`type`** (string, REQUIRED) - the platform resource being limited, for example as defined in [the man page](http://man7.org/linux/man-pages/man2/setrlimit.2.html).
```
wking: This is not “for example” on Linux.  The values defined in `setrlimit(2)` MUST be supported by compliant runtimes.  For other platforms, we need new links to their supported values, not a weakening of the Linux values.

jlbutler:
@RobDolinMS please review in the context of #303. It seems a lot of other changes have come in to establish config compatibility, this just cleans up a few nits.

config.md
jlbutler: That's what I intended to do, typo. Will fix, thanks.

jlbutler: I also broke the 'one sentence per line' rule here, will fix.

config.md
jlbutler: Will fix, thanks.

config.md
jlbutler: I thought using an example rather than implying only 'ntfs' would be better. I pinged irc for supported file systems but no reply yet. It's possible that only 'ntfs' is supported, but I thought we might see 'refs' eventually.

@RobDolinMS could you advise what should go here?

config.md
jlbutler: Ok thanks for pointing that out. It seemed like an implementation note more than anything, so I just dropped it. I'll put it back for now; I'm aware of #663.

config.md
jlbutler: On the wording, is there a reason to label this as explicitly undefined? If it's OPTIONAL, implementations may leave it out of their scope.

As far as documentation, on reading this spec I really take the man page references to be just example references for clarity. We have various places where documentation is not explicitly referred to. If that should be part of this (or another PR), I think it would extend to more than this field. Thoughts?

config.md
jlbutler: Again, I took this to be an example. I don't see how this infers that the field is not required nor how it weakens the spec. It's just that the specific man page reference is specific only to Linux, it serves as an example as to what can be valid for a particular platform.

This loops back to previous comment, if this spec should have references to all documentation for all configuration options on all supported platforms, it seems more work than these specific fields.

config.md
wking: > If it's OPTIONAL, implementations may leave it out of their scope.

Right, but when a Linux config sets `capabilities`, the runtime has to apply them in a particular way.  And when a Windows config sets `capabilities`, the runtime doesn't have to do anything.

> As far as documentation, on reading this spec I really take the man page references to be just example references for clarity. We have various places where documentation is not explicitly referred to. If that should be part of this (or another PR), I think it would extend to more than this field.

I agree that this is a bigger issue than we want to handle in this PR.  I just think we need to have clearly-specified values (otherwise how do the compliance tests or folks writing a config know what values their generic OCI-compliant runtime must support?).  And while I don't think we need to strengthen all of that in this PR, I'd rather not weaken anything here ;).

jlbutler: For this PR, how would you feel about a generic "valid values for other platforms are as documented for that specific platform" or similar type of statement? Without some abstraction, we should probably refer to specific documentation for all platforms.

The abstraction proofs the spec against platform-specific documentation changes (content or location), which seems like an additional benefit.

jlbutler:
@opencontainers/runtime-spec-maintainers @RobDolinMS @jhowardmsft PTAL, thank you.

Based on this week's discussion on the call and in IRC, I've made changes to clarify what constitutes a valid value for a platform-specific property. Note the reference at the top of the file to the platform-specific scope below. I am hoping that this abstraction is enough to allow for implementors to feel free to support a platform as they see fit, while maintaining the spec's strength.

There are various nits which improve consistency and readability, I'm hoping it's simpler to review taken as a whole, rather than breaking out nits into separate PRs.

config.md
```diff
@@ -128,27 +131,25 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-
-For Linux-based systems the process structure supports the following process specific fields:
-
-* **`capabilities`** (array of strings, OPTIONAL) capabilities is an array that specifies Linux capabilities that can be provided to the process inside the container.
-  Valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html).
+* **`capabilities`** (array of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
```
wking: I think this is still too wishy.  MUST runtimes support all those valid values?  E.g. [is a runtime that errors on `CAP_AUDIT_READ` still compliant?][1]. What are valid values on Solaris?  On Windows (now that you've removed the earlier "For Linux-based systems the process structure supports the following process specific fields:" guard)?

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-02-01.log.html#t2017-02-01T22:48:39

jlbutler: I don't see that it's any different from before, only that Linux isn't the only thing referenced, and the man page isn't an implied extension of the spec. I have attempted to clarify with statements in the 'Platform-specific configuration' section. Applying those rules for all fields, not just 'capabilities', the answers to your questions are yes, valid values for Solaris are documented with the Solaris doc set, same for Windows. A runtime may support any combination of valid values for that platform, and there are rules for what it must do if it chooses to not do so.

wking: > A runtime may support any combination of valid values for that platform, and there are rules for what it must do if it chooses to not do so.

But then you could have a “compliant” no-op runtime, which always errors out with “sorry, I don't support that”.  I think it makes sense to set a minimum bar for what you have to implement before you are compliant (e.g. “Linux runtimes MUST implement at least these capabilities: …”.  I'm ok with that minimum bar being bounded by the local platform (e.g. a compliant runtime can error on `CAP_X` if and only if the kernel doesn't support `CAP_X`), but I don't have a good way to word that requirement generically.

jlbutler: I think we might be getting a bit contrived with an runtime that implements nothing... for example, no file system types for mounts are support. That said, this isn't meant to be a design document. I would say that as long as a clear error message is returned for any valid value for a field, yes it's spec-compliant.

wking: >  I would say that as long as a clear error message is returned for any valid value for a field, yes it's spec-compliant.

That's an internally consistent approach, although it makes “OCI runtime-spec compliant” a pretty cheap badge, and means the only way to know that a runtime will be able to handle a given config will be “try to use the runtime to launch that config”.  If the maintainers want to go in this direction, they may want to re-evaluate the effort currently being put into runtime-tools' compliance testing.

jlbutler: I am having difficulty drawing that conclusion. As far as I know, there is no requirement for a minimum set of platform-specific configuration fields to be populated. This change doesn't seem to undo anything, and in fact corrects an error. For this specific field, I've changed the statement that valid values for all platforms are documented in a Linux man page, to a statement which says that valid values are platform-specific, and valid values must be used for this field. It even uses the same man page as an example. I am fine with claiming that a minimum amount of functionality must be implemented, but I don't see how that has bearing on this set of changes. I do not see how this introduces an argument that we avoid or weaken runtime compliance testing or certification.

wking: > As far as I know, there is no requirement for a minimum set of platform-specific configuration fields to be supported.

I agree that the previous language made this unclear, and that we want to add clarity moving forward.

> For this specific field, I've changed the statement that valid values for all platforms are documented in a Linux man page…

The old wording makes sense in master, where we [restrict `process.capabilities` to Linux][1].  A narrow expansion (to support Solaris as well) would be:

> On Linux, valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html).
> On Solaris, valid values are the strings for capabilities defined in [wherever Solaris documents these].

By using “For example”, it goes from being ambiguous about runtime support to reading more like “look in capabilities(7) for inspiration, but the runtime may not support all the capabilities documented there”, which is less ambiguous (in the direction of allowing a compliant no-op runtime).

And by extending the field to Windows (by lifting the Linux-only restriction), we open up a new can of worms (having to figure out what the Windows analog to capabilities is, assuming there is one with [capabilities not even being in POSIX][2]).

>  I would be fine with a discussion around whether or not a minimum amount of functionality must be implemented, but I don't see how that has bearing on this set of changes

I think the “For example” wording moves us further from where I'd like to be, with a (possibly platform-dependent) defined set of valid values which runtimes MUST support.  But the wording is ambiguous in master, and still ambiguous in ab7cca36 (is “For example” saying “these values may not work on compliant non-Linux runtimes but MUST work on compliant Linux runtimes” or  “these values may not work on compliant runtimes, even if they are Linux runtimes”?).  So I'm ok punting the discussion on what runtimes MUST support for now, landing whatever folks want in the meantime, and coming back around to “what MUST runtimes support?” in a different issue/PR (hopefully to be resolved before 1.0).

[1]: https://github.com/opencontainers/runtime-spec/blame/7278567eb2fbf95e96cc5ad8b3c22dc8fc81ecba/config.md#L132
[2]: http://wt.tuxomania.net/publications/posix.1e/

vbatts:
Good stuff, so far. I'm still reviewing it tho.

config.md
```diff
@@ -41,25 +45,27 @@ For example, if a configuration is compliant with version 1.1 of this specificat
...
-  For the Windows operating system, one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
-  For the Solaris operating system, this corresponds to "dir" of the fs resource in zonecfg(8).
+  * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
+  * Solaris: corresponds to "dir" of the fs resource in zonecfg(8).
 * **`type`** (string, REQUIRED) The filesystem type of the filesystem to be mounted.
-  Linux: *filesystemtype* argument supported by the kernel are listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
-  Windows: ntfs.
-  Solaris: corresponds to "type" of the fs resource in zonecfg(8).
+  * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  * Windows: the type of file system on the volume, e.g. "ntfs".
```
vbatts: yea. This allows for better future-proofing.

config.md
```diff
@@ -2,11 +2,15 @@
 
 The container's top-level directory MUST contain a configuration file called `config.json`.
 The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).
+Platform-specific configuration schema are defined in [**`platform-specific documents`**](#platform-specific-configuration) linked below.
```
wking: We probably don't need backticks or bold around “platform-specific documents”.  The usual link styling makes the link-ness obvious enough.

config.md
vbatts: I think this is fine. Again, if further linting is required by an application to ensure a host has the expected capabilities, that is not going to be defined here.

config.md
```diff
@@ -289,14 +290,16 @@ For Windows based systems the user structure has the following fields:
 
 ## Platform-specific configuration
 
-[**`platform.os`**](#platform) is used to lookup further platform-specific configuration.
+[**`platform.os`**](#platform) is used to specify platform-specific configuration.
+Runtime implementations MAY support any valid values for platform-specific fields as part of this configuration.
+Implementations MUST generate a clear error message for valid values it chooses to not support and MUST generate an error when invalid values are encountered.
```
vbatts: does "a clear error message" mean that the implementation would not continue running and would exit non-zero? Just curious.

wking: > does "a clear error message" mean that the implementation would not continue running and would exit non-zero?

Exiting non-zero on `create` errors is [covered][1] by the in-flight #513.  The currently landed error language has “[as if the operation were never attempted][2]” wording and “[MUST generate an error and a new container MUST NOT be created][3]” wording, which seems to cover everything but the exit-code (which only makes sense in the context of the #513 API).

[1]: https://github.com/opencontainers/runtime-spec/pull/513/files#diff-0c1e5a5baa5649945697a530e09acbf5R66
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc4/runtime.md#L69
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc4/runtime.md#L93

jlbutler:
Rebased, fixed up link nits and clarified the error and exit requirements in the platform-specific scope. Thanks @wking and @vbatts .

config.md
crosbymichael: What @vbatts said.  You cannot base compliance on system values @wking and you cannot just make up some minimum.  It can support as many values as the underlying host supports for fields like this.  Just because you use a runtime on an older kernel that doesn't have a CAP does not mean its no longer a compliant runtime.

LGTM @jlbutler

config.md
```diff
@@ -289,14 +290,16 @@ For Windows based systems the user structure has the following fields:
 
 ## Platform-specific configuration
 
-[**`platform.os`**](#platform) is used to lookup further platform-specific configuration.
+[**`platform.os`**](#platform) is used to specify platform-specific configuration.
+Runtime implementations MAY support any valid values for platform-specific fields as part of this configuration.
+Implementations MUST error out when invalid values are encountered and MUST generate a clear error message and exit non-zero when encountering valid values it chooses to not support.
```
wking: I think we want to drop “exit non-zero”, because `config.md` should be independent of the API used to invoke the runtime.  “generate a clear error message” is generic, and the relationship between error generation and exit-code should be [setup in the API docs for APIs which use exit-codes][1].

And “clear” is hard to test.  Maybe make it “MUST [generate an error][2]”?  Then we'd trust on runtime authors to make it clear.  If we don't trust runtime authors, we should say something enforceable, but that's hard to do when [we don't even require the runtime to expose error to the user][2].

[1]: https://github.com/opencontainers/runtime-spec/pull/513/files#diff-0c1e5a5baa5649945697a530e09acbf5R66
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/runtime.md#errors

config.md
wking: > Just because you use a runtime on an older kernel that doesn't have a CAP does not mean its no longer a compliant runtime.

That's not the situation I'm concerned about.  I'm concerned about the situation where the host *does* support a particular `CAP_*` (or whatever) but the runtime does not.  Instead of “it can support…”, I'd rather have “it MUST support values from {some values} for values supported by the host platform”.  That bases compliance on system values (which you don't like), but means that you can actually test runtimes for meaningful compliance (e.g. the no-op runtime would be non-compliant).

config.md
jlbutler: I believe I've covered these concerns in the latest commit. Thanks!

config.md
vbatts: @wking then the runtime errors out. end-users investigate. complain to the runtime developers or write a fix. That kind-of linting and validation needs to be secondary to this spec.

vbatts:
LGTM

crosbymichael:
LGTM

config.md
wking: > then the runtime errors out. end-users investigate. complain to the runtime developers or write a fix. That kind-of linting and validation needs to be secondary to this spec.

So what kind of `CAP_*` values should the integration test suite use?  Currently it [sets][1] [these][2], and fails runtimes which [die in `create`][3] (currently `start`, because runtime-tools hasn't been ported around #384, opencontainers/runtime-tools#285) or [run without a matching set][4].  Will we update that to say that any runtime is compliant as long as it:

* Dies in `create` or
* Succeeds in `create` and has a container environment matching the configuration

And then test all combinations (or some sampling) of `CAP_*` values regardless of platform support?

[1]: https://github.com/opencontainers/runtime-tools/blob/2d92f6557e64d4f9a0e799a75fdf153cec13dffa/test_runtime.sh#L80
[2]: https://github.com/opencontainers/runtime-tools/blob/2d92f6557e64d4f9a0e799a75fdf153cec13dffa/generate/generate.go#L57-L70
[3]: https://github.com/opencontainers/runtime-tools/blob/2d92f6557e64d4f9a0e799a75fdf153cec13dffa/test_runtime.sh#L82-L85
[4]: https://github.com/opencontainers/runtime-tools/blob/2d92f6557e64d4f9a0e799a75fdf153cec13dffa/cmd/runtimetest/main.go#L175-L194

PR #675, opened 2017-02-02
Add ambient and bounding capability support

crosbymichael:
Closes #668

### DIFF
--- a/config.md
+++ b/config.md
@@ -132,7 +132,13 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-* **`capabilities`** (array of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
+* **`capabilities`** (object, OPTIONAL) is an object containing arrays that specifies the sets of capabilities for the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
+  capabilities contains the following properties:
+    * **`effective`** (array of strings, OPTIONAL) - the `effective` field is an array of effective capabilities that are kept for the process.
+    * **`bounding`** (array of strings, OPTIONAL) - the `bounding` field is an array of bounding capabilities that are kept for the process.
+    * **`inheritable`** (array of strings, OPTIONAL) - the `inheritable` field is an array of inheritable capabilities that are kept for the process.
+    * **`permitted`** (array of strings, OPTIONAL) - the `permitted` field is an array of permitted capabilities that are kept for the process.
+    * **`ambient`** (array of strings, OPTIONAL) - the `ambient` field is an array of ambient capabilities that are kept for the process.
 * **`rlimits`** (array of objects, OPTIONAL) allows setting resource limits for a process inside the container.
   Each entry has the following structure:
 
@@ -191,11 +197,30 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
     "apparmorProfile": "acme_secure_profile",
     "selinuxLabel": "system_u:system_r:svirt_lxc_net_t:s0:c124,c675",
     "noNewPrivileges": true,
-    "capabilities": [
-        "CAP_AUDIT_WRITE",
-        "CAP_KILL",
-        "CAP_NET_BIND_SERVICE"
-    ],
+    "capabilities": {
+        "bounding": [
+            "CAP_AUDIT_WRITE",
+            "CAP_KILL",
+            "CAP_NET_BIND_SERVICE"
+        ],
+       "permitted": [
+            "CAP_AUDIT_WRITE",
+            "CAP_KILL",
+            "CAP_NET_BIND_SERVICE"
+        ],
+       "inheritable": [
+            "CAP_AUDIT_WRITE",
+            "CAP_KILL",
+            "CAP_NET_BIND_SERVICE"
+        ],
+        "effective": [
+            "CAP_AUDIT_WRITE",
+            "CAP_KILL",
+        ],
+        "ambient": [
+            "CAP_NET_BIND_SERVICE"
+        ]
+    },
     "rlimits": [
         {
             "type": "RLIMIT_NOFILE",
@@ -446,11 +471,30 @@ Here is a full example `config.json` for reference.
             "TERM=xterm"
         ],
         "cwd": "/",
-        "capabilities": [
-            "CAP_AUDIT_WRITE",
-            "CAP_KILL",
-            "CAP_NET_BIND_SERVICE"
-        ],
+        "capabilities": {
+            "bounding": [
+                "CAP_AUDIT_WRITE",
+                "CAP_KILL",
+                "CAP_NET_BIND_SERVICE"
+            ],
+            "permitted": [
+                "CAP_AUDIT_WRITE",
+                "CAP_KILL",
+                "CAP_NET_BIND_SERVICE"
+            ],
+            "inheritable": [
+                "CAP_AUDIT_WRITE",
+                "CAP_KILL",
+                "CAP_NET_BIND_SERVICE"
+            ],
+            "effective": [
+                "CAP_AUDIT_WRITE",
+                "CAP_KILL",
+            ],
+            "ambient": [
+                "CAP_NET_BIND_SERVICE"
+            ]
+        },
         "rlimits": [
             {
                 "type": "RLIMIT_CORE",
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -135,9 +135,43 @@
                 },
                 "capabilities": {
                     "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities",
-                    "type": "array",
-                    "items": {
-                        "$ref": "defs-linux.json#/definitions/Capability"
+                    "type": "object",
+                    "properties": {
+                        "bounding": {
+                            "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/bounding",
+                            "type": "array",
+                            "items": {
+                                "$ref": "defs-linux.json#/definitions/Capability"
+                            }
+                        },
+                        "permitted": {
+                            "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/permitted",
+                            "type": "array",
+                            "items": {
+                                "$ref": "defs-linux.json#/definitions/Capability"
+                            }
+                        },
+                        "effective": {
+                            "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/effective",
+                            "type": "array",
+                            "items": {
+                                "$ref": "defs-linux.json#/definitions/Capability"
+                            }
+                        },
+                        "inheritable": {
+                            "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/inheritable",
+                            "type": "array",
+                            "items": {
+                                "$ref": "defs-linux.json#/definitions/Capability"
+                            }
+                        },
+                        "ambient": {
+                            "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/ambient",
+                            "type": "array",
+                            "items": {
+                                "$ref": "defs-linux.json#/definitions/Capability"
+                            }
+                        }
                     }
                 },
                 "apparmorProfile": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -44,8 +44,8 @@ type Process struct {
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
 	Cwd string `json:"cwd"`
-	// Capabilities are Linux capabilities that are kept for the container.
-	Capabilities []string `json:"capabilities,omitempty" platform:"linux"`
+	// Capabilities are Linux capabilities that are kept for the process.
+	Capabilities *LinuxCapabilities `json:"capabilities,omitempty" platform:"linux"`
 	// Rlimits specifies rlimit options to apply to the process.
 	Rlimits []LinuxRlimit `json:"rlimits,omitempty" platform:"linux"`
 	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container.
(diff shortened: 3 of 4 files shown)


schema/config-schema.json
```diff
@@ -135,10 +135,17 @@
                 },
                 "capabilities": {
                     "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities",
-                    "type": "array",
-                    "items": {
-                        "$ref": "defs-linux.json#/definitions/Capability"
-                    }
+                    "type": "object",
+					"properties": {
```
wking: Probably want to reformat this to get consistent indents (`make -C schema fmt`).

schema/config-schema.json
```diff
@@ -135,10 +135,17 @@
...
+						"bounding": {
+							"id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/bounding",
+							"$ref": "defs-linux.json#/definitions/Capability"
```
wking: This (and `ambient`) should still be arrays.  So:

    "type": "array",
    "items": {
      "$ref": "defs-linux.json#/definitions/Capability"
    }

instead of the line I'm commenting on.

config.md
```diff
@@ -131,8 +131,11 @@ See links for details about [mountvol](http://ss64.com/nt/mountvol.html) and [Se
 
 For Linux-based systems the process structure supports the following process specific fields:
 
-* **`capabilities`** (array of strings, OPTIONAL) capabilities is an array that specifies Linux capabilities that can be provided to the process inside the container.
+* **`capabilities`** (object, OPTIONAL) capabilities is a whitelist of capabilities for the bounding and ambient sets for Linux processes.
   Valid values are the strings for capabilities defined in [the man page](http://man7.org/linux/man-pages/man7/capabilities.7.html).
+  capabilities contains the following properties:
+    * **`bounding`** (array of strings, OPTIONAL) - the 'bounding' field is the whitelist of bounding capabilities that are kept for the process.
+	* **`ambient`** (array of strings, OPTIONAL) - the 'ambient' field is the whitelist of ambient capabilities that are kept for the process.
```
wking: Inconsistent indent?  And you can probably drop “the '…' field is the” and just start in:

> * **`ambient`** (array of strings, OPTIONAL) A whitelist of …

config.md
```diff
@@ -189,11 +192,15 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are
     "apparmorProfile": "acme_secure_profile",
     "selinuxLabel": "system_u:system_r:svirt_lxc_net_t:s0:c124,c675",
     "noNewPrivileges": true,
-    "capabilities": [
-        "CAP_AUDIT_WRITE",
-        "CAP_KILL",
-        "CAP_NET_BIND_SERVICE"
-    ],
+    "capabilities": {
+		"bounding": [
```
wking: Inconsistent indent (all the rest of the examples use four spaces).

justincormack:
LGTM

mrunalp:
While this looks good, we do manipulate the inherited, permitted and effective sets as well in runc. Should we make a statement about that or rename bounding to something else?

crosbymichael:
@mrunalp I guess if we want to be uber explicit with the settings we can add all the fields but I don't think its totally useful for most containers.  Everyone will set the additional fields the same anyways so keeping a bounding set makes things much more straightforward.

justincormack:
I think I can imagine a situation where you want permitted but not effective or inherited, eg you want to allow some suid executables to work, but the capabilities to be only executed by those, not by the initial process. So I think its probably better to have all four sets, even if most applications just set e=i=p.

crosbymichael:
@mrunalp what do you think?  Should we add the rest and remove bounding? 

Anyone else have any input on this?

mrunalp:
@crosbymichael We need ambient and bounding for sure. If we add the other 3 sets as well then we will have to make a statement that the runtime will apply these to the init process before it execs the user specified process. The capabilities could change on the basis of being root or not, executed binary being setuid or not and having any filesystem capabilities.

crosbymichael:
@mrunalp not much of a difference from today right?

mrunalp:
@crosbymichael Yes, that's what we do today. I think we can just add all the sets and leave it up to the user to set them correctly. For e.g. whatever is set in ambient has to be in both permitted and inherited. We can just point the user to the man page for capabilities.

crosbymichael:
Ok, I updated this PR with the changes

config.md
```diff
@@ -131,7 +131,12 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
...
+* **`capabilities`** (object of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
+  capabilities contains the following properties:
+    * **`effective`** (array of strings, OPTIONAL) - the 'bounding' field is the whitelist of bounding capabilities that are kept for the process.
+    * **`inheritable`** (array of strings, OPTIONAL) - the 'bounding' field is the whitelist of bounding capabilities that are kept for the process.
+    * **`permitted`** (array of strings, OPTIONAL) - the 'bounding' field is the whitelist of bounding capabilities that are kept for the process.
+    * **`ambient`** (array of strings, OPTIONAL) - the 'ambient' field is the whitelist of ambient capabilities that are kept for the process.
```
mrunalp: We need bounding as well.

config.md
```diff
@@ -445,11 +455,15 @@ Here is a full example `config.json` for reference.
...
-            "CAP_AUDIT_WRITE",
-            "CAP_KILL",
-            "CAP_NET_BIND_SERVICE"
-        ],
+        "capabilities": {
+            "bounding": [
+                "CAP_AUDIT_WRITE",
+                "CAP_KILL",
+            ],
+            "ambient": [
```
mrunalp: Just for correctness, we will have set permitted/inheritable as well to atleast the same caps as ambient. We could set p/i/e to same as bounding in this example.

mrunalp: Actually ambient has a different cap so would need to include that as well in p/i and optionally in e as well.

config.md
```diff
@@ -190,11 +196,27 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are
...
     "noNewPrivileges": true,
-    "capabilities": [
-        "CAP_AUDIT_WRITE",
-        "CAP_KILL",
-        "CAP_NET_BIND_SERVICE"
-    ],
+    "capabilities": {
+        "bounding": [
+            "CAP_AUDIT_WRITE",
+            "CAP_KILL",
```
justincormack: the example should have `CAP_NET_BIND_SERVICE` in the bounding set or it cannot be applied as an ambient capability.

crosbymichael:
@justincormack updated

dqminh:
LGTM

config.md
mrunalp: `CAP_NET_BIND_SERVICE` also needs to be added to permitted/inherited.
From capabilities(7)
```
The ambient capability set obeys the invariant that no capability can ever be ambient if it is not both permitted and inheritable.
```

crosbymichael: @mrunalp does that mean we should or should not have it in bounding?

lucab:
According to this new specification, what is the semantics of `effective`? I'm asking because the effective set is adjusted by the kernel on exec(3) starting from [file-cap,permitted-set,ambient-set], so I'm not sure what is the behavior expected from a runtime receiving an explicit `effective` field.

config.md
mrunalp: @crosbymichael We should have it in bounding, permitted and inheritable.

mrunalp:
@lucab yes. What we mean here is that the runtime sets effective before exec'ing and then it gets calculated per the rules in the kernel on exec. Maybe we can clarify that.

mrunalp:
We still need the examples updated for this.

crosbymichael:
@mrunalp fixed.

mrunalp:
LGTM

config.md
```diff
@@ -132,7 +132,13 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-* **`capabilities`** (array of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
+* **`capabilities`** (object of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
```
wking: I don't see any precedence for “object of …”, and would recommend just using “object” (which is [what we use for `annotations`][1]).

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc4/config.md#L392

config.md
```diff
@@ -132,7 +132,13 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
...
+  capabilities contains the following properties:
+    * **`effective`** (array of strings, OPTIONAL) - the 'effective' field is the whitelist of effective capabilities that are kept for the process.
```
wking: We usually use backticks around property names, so “'effective'” → “`effective`” (and similarly in the following lines).  Grepping for `'` in our `*.md`, I see no examples of single-quoted property names in master:

```
$ git grep "'[^' ]*'" origin/master -- *.md
origin/master:config.md:For Solaris, the mount entry corresponds to the 'fs' resource in zonecfg(8).
origin/master:config.md:  As an example, the ['no_new_privs' kernel doc](https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt) has more information on how this is achieved using a prctl system call on Linux.
```

Both of those are references to external stuff (not OCI properties), and I think they should have been backticked as well ;).

wking: Also, the new “whitelist” language would require the removal of [this runtime-tools check][1] and give configuration authors no way to clearly drop a given capability.  I'd rather we leave these arrays as “this will be the set of process capabilities after `create` finishes” and just focus on adding the additional sets in this PR.  We can address the change from “this set” to “at least this set” in a separate PR.

[1]: https://github.com/opencontainers/runtime-tools/blob/a652af160a151258b9f0d697a7ab064b4aa9c1cd/cmd/runtimetest/main.go#L192

specs-go/config.go
```diff
@@ -44,8 +44,8 @@ type Process struct {
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
 	Cwd string `json:"cwd"`
-	// Capabilities are Linux capabilities that are kept for the container.
-	Capabilities []string `json:"capabilities,omitempty" platform:"linux"`
+	// Capabilities are Linux capabilities that are kept for the process.
+	Capabilities *LinuxCapabilities `json:"capabilities,omitempty" platform:"linux"`
```
wking: @jlbutler just removed the Linux-specific-ness of `process.capabilities` in #673 (at least [in the Markdown spec][1]).  With this change, it becomes even less clear to me how other OSes are supposed to configure their capabilities (although I [wasn't clear on that since #673][2], so it's not a lot worse ;).

[2]: https://github.com/opencontainers/runtime-spec/pull/673/files#r99353136

justincormack: No other systems have capabilities; there was a withdrawn Posix draft that the Linux implementation was based on, and then customised for Linux. They should have remained Linux specific.

wking: > No other systems have capabilities…

capabilies(7) references [this][1] for the withdrawn POSIX.1e draft, and that has:

> Why Posix.1e was abandoned is difficult to understand from today's (July 2014) point of view. Solaris, Irix, Linux, and probably other Unices seemed to recognize the standard. On the other hand the FreeBSD project found counter arguments and didn't integrate capabilities ('fine grained privileges') by default.

I don't have access to Solaris or Irix for testing, or enough interest in them to find relevant docs, but I'd caution against making this Linux-specific (again) without consulting someone with Solaris access.  On the other hand, maybe the policy is “unless you document your valid values, we may drop support for this property on your system”, and I'd be +1 for that, giving the Solaris folks some reasonable period to provide the valid-values docs, and then dropping Solaris capability support if they don't.

[1]: http://wt.tuxomania.net/publications/posix.1e/

justincormack: But here we are adding the Linux specific types such as ambient. Solaris capabilities are a bit similar by the look of it (they have E, I, P but different capabilities) but that doesn't mean they should be at the top level.

justincormack: Solaris docs are http://docs.oracle.com/cd/E23824_01/html/821-1456/prbac-2.html

wking: > Solaris capabilities are a bit similar by the look of it (they have E, I, P but different capabilities) but that doesn't mean they should be at the top level.

Is that “I don't think there is enough overlap to be worth sharing the same namespace”?  If the Solaris folks are ok using `bounding` instead of their `limited`, we could just make `ambient` Linux-specific.  In other situations where only the values are different (e.g. [`mounts` entries][1]), the maintainer preference seems to be “share the same property, but define the different per-platform values”.

On the other hand, if Windows has nothing equivalent, the [pro-namespacing policy here][2] suggests namespacing the whole `capabilities` structure to something POSIX-ish-specific.  I tried to raise the “[how will this be handled on Windows?][3]” issue in #673, but am still not clear on if/whether Windows supports something capability-like.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/config.md#mounts
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-18-17.01.log.html#l-101
[3]: https://github.com/opencontainers/runtime-spec/pull/673#discussion_r99353136

crosbymichael:
@mrunalp @hqhq pushed another update to fix last bits of feedback

config.md
```diff
@@ -132,7 +132,13 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-* **`capabilities`** (array of strings, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
+* **`capabilities`** (object, OPTIONAL) is an array that specifies the set of capabilities of the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
```
mrunalp: Can we reword  `is an array` to `is an object containing arrays of capability types` or something similar?

mrunalp:
LGTM

config.md
```diff
@@ -132,7 +132,13 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
...
+* **`capabilities`** (object, OPTIONAL) is an object containing arrays that specifies the sets of capabilities for the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [CAPABILITIES(7)](http://man7.org/linux/man-pages/man7/capabilities.7.html) man page.
+  capabilities contains the following properties:
+    * **`effective`** (array of strings, OPTIONAL) - the `effective` field is a array of effective capabilities that are kept for the process.
+    * **`bounding`** (array of strings, OPTIONAL) - the `bounding` field is a array of bounding capabilities that are kept for the process.
+    * **`inheritable`** (array of strings, OPTIONAL) - the `inheritable` field is a array of inheritable capabilities that are kept for the process.
+    * **`permitted`** (array of strings, OPTIONAL) - the `permitted` field is a array of permitted capabilities that are kept for the process.
+    * **`ambient`** (array of strings, OPTIONAL) - the `ambient` field is a array of ambient capabilities that are kept for the process.
```
hqhq: an array?

dqminh: LGTM when typo is fixed

crosbymichael:
@hqhq @dqminh fixed the typos

hqhq:
LGTM

mrunalp:
LGTM

dqminh:
LGTM

PR #677, opened 2017-02-03
runtime: Add 'exit' to state for collecting the container exit code

wking:
This gives us a more portable way to discover the container exit code (vs. [requiring callers to use subreapers][1] or other platform-specific approaches which require knowledge of the runtime implementation).

There would still be no portable way to discover *when* the container had exited (short of polling `state`), but we can handle that separately.

[1]: https://github.com/opencontainers/runc/pull/827#r63598816


crosbymichael:
You cannot have this as required because that forces the runtime to always be running to update the state files.

This change just forces long running runtimes and will impact performance and usage down the road when you run at scale. 

If you leave the runtime running it returns the container's exit status as its exit status anyway if you make that tradeoff.  This does not belong in the spec.

PR #678, opened 2017-02-03
config-linux: Drop the default-filesystem section

wking:
Users who need these mounts would have to explicitly set them up in their configuration (as runtime-tools [continues to do][1]) if they wanted to guarantee their presence.  Users who don't need them can omit them from their configuration.  I don't see how keeping a SHOULD-strength runtime requirement helps either of those workflows.

This is one of the possible approaches I'd [floated][2] in #666.

[1]: https://github.com/opencontainers/runtime-tools/pull/24
[2]: https://github.com/opencontainers/runtime-spec/pull/666#issuecomment-274550344

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -3,20 +3,6 @@
 This document describes the schema for the [Linux-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
 The Linux container specification uses various kernel features like namespaces, cgroups, capabilities, LSM, and filesystem jails to fulfill the spec.
 
-## Default Filesystems
-
-The Linux ABI includes both syscalls and several special file paths.
-Applications expecting a Linux environment will very likely expect these file paths to be setup correctly.
-
-The following filesystems SHOULD be made available in each container's filesystem:
-
-|   Path   |  Type  |
-| -------- | ------ |
-| /proc    | [procfs](https://www.kernel.org/doc/Documentation/filesystems/proc.txt)   |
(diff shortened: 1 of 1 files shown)


hqhq:
This section still looks useful to me.

wking:
> This section still looks useful to me.

So do you prefer #679?  Or do you like master how it is?

If you like master, do you read the current master wording [as a suggestion to runtime authors][1] or [as a suggestion to config authors][2]?

[1]: https://github.com/opencontainers/runtime-spec/pull/679#issue-205252535
[2]: https://github.com/opencontainers/runtime-spec/pull/666#issuecomment-277067251

crosbymichael:
I prefer the master version

wking:
> I prefer the master version

And I'm citing [your earlier comment][1] for the “this is advice to config authors” interpretation.  Is that actually your interpretation?  Do you think the master wording is clearly not advice to runtime authors (it's not clear to me)?  It might be easier to talk about this if we talk over how SHOULD validation will work.  Will this SHOULD be checked by runtime-tools' `validate` package (aimed at validating configuration JSON) or runtime-tools' `runtimetest` command (aimed at validating runtime implementations)?

[1]: https://github.com/opencontainers/runtime-spec/pull/666#issuecomment-277067251

crosbymichael:
Closing as this section is marked as `SHOULD` and is found useful by multiple maintainers.

PR #679, opened 2017-02-03
config: Make default Linux filesystems an example

wking:
The MUST default-filesystem wording altered in #666 had read (to me, anyway) as:

> The runtime MUST supply these even if the config doesn't call for them in mounts.

with #666 weaking it to:

> The runtime SHOULD supply these even if the config doesn't call for them in mounts.

But that's not very useful (callers that *need* a given mount will still have to configure it explicitly).  However, one interpretation of the #666 wording seems to be [something like][1]:

> Config authors probably want to include mounts entries for these.

That's fine, and this commit tries to make that interpretation more obvious by shifting the config recommendation over to the Linux `mounts` example.

This is one of the possible approaches I'd [floated][2] in #666.  Another approach is just dropping the section (#678), so this PR is a parallel alternative to #678.

[1]: https://github.com/opencontainers/runtime-spec/pull/666#issuecomment-277067251
[2]: https://github.com/opencontainers/runtime-spec/pull/666#issuecomment-277086447

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -3,20 +3,6 @@
 This document describes the schema for the [Linux-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
 The Linux container specification uses various kernel features like namespaces, cgroups, capabilities, LSM, and filesystem jails to fulfill the spec.
 
-## Default Filesystems
-
-The Linux ABI includes both syscalls and several special file paths.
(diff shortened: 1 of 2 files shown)


crosbymichael:
REJECT 

Please don't open two conflicting PRs for the same thing, at the same time.  Its confusing and just splits discussions.  

Closing this in favor of the first one opened where there is discussion happening.

PR #681, opened 2017-02-07
config: Move valid-value rules to their own section

wking:
This wording just landed via #673, but the rule is generic and not unique to platform-specific properties.

Also adjust the wording somewhat to match the more established wording from the “Extensibility” section.


crosbymichael:
I think the original wording is much more straightforward and easier to understand than the changes in this PR.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #683, opened 2017-02-08
schema/validate: get correct canonical of schemaPath

Mashimiao:

Before this fix, we can't get correct schemaPath, because it's overridden.

### DIFF
--- a/schema/validate.go
+++ b/schema/validate.go
@@ -19,12 +19,12 @@ func main() {
 
 	schemaPath := os.Args[1]
 	if !strings.Contains(schemaPath, "://") {
-		schemaPath, err := filepath.Abs(schemaPath)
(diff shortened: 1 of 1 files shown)


schema/validate.go
```diff
@@ -19,12 +19,12 @@ func main() {
 
 	schemaPath := os.Args[1]
 	if !strings.Contains(schemaPath, "://") {
-		schemaPath, err := filepath.Abs(schemaPath)
+		absPath, err := filepath.Abs(schemaPath)
```
wking: Obsolete now that #552 has landed [with this change][1]?

[1]: https://github.com/opencontainers/runtime-spec/pull/552/files#diff-8ce800f4e5ad55c2ec4bdcd49613f4c9R44

Mashimiao:
OK, I will colse this PR

PR #685, opened 2017-02-09
config.md: args of Hooks should be optional

Mashimiao:

### DIFF
--- a/config.md
+++ b/config.md
@@ -328,7 +328,7 @@ Hooks allow for the configuration of custom actions related to the [lifecycle](r
     Entries in the array contain the following properties:
     * **`path`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execv`'s *path*][ieee-1003.1-2001-xsh-exec].
       This specification extends the IEEE standard in that **`path`** MUST be absolute.
-    * **`args`** (array of strings, REQUIRED) with the same semantics as [IEEE Std 1003.1-2001 `execv`'s *argv*][ieee-1003.1-2001-xsh-exec].
+    * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001 `execv`'s *argv*][ieee-1003.1-2001-xsh-exec].
     * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
     * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
   * **`poststart`** (array, OPTIONAL) is an array of [post-start hooks](#poststart).
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -95,7 +95,6 @@
                 }
             },
             "required": [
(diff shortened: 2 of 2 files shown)


config.md
```diff
@@ -328,7 +328,7 @@ Hooks allow for the configuration of custom actions related to the [lifecycle](r
     Entries in the array contain the following properties:
     * **`path`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execv`'s *path*][ieee-1003.1-2001-xsh-exec].
       This specification extends the IEEE standard in that **`path`** MUST be absolute.
-    * **`args`** (array of strings, REQUIRED) with the same semantics as [IEEE Std 1003.1-2001 `execv`'s *argv*][ieee-1003.1-2001-xsh-exec].
+    * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001 `execv`'s *argv*][ieee-1003.1-2001-xsh-exec].
```
wking: If we're thinking about adjusting this, I [still][1] prefer Python's approach which makes `path` optional.  I'd also like to [unify the approach to process invocation between `hooks` and `process`][2] (having different semantics for `args` depending on the context is just confusing).  But if those are not on the table, this change looks ok to me, and gets us back closer to the [old][3] [Go semantics with their optional Args][4].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/luaaorsya10/IdjzUYl9GQAJ
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Cn5L6g1prgA
[3]: https://github.com/opencontainers/runtime-spec/pull/427/files#diff-c9c91c29b41257aea3a3403cc606ad99L383
[4]: https://golang.org/pkg/os/exec/#Cmd

hqhq: +1 for keeping consistency for `args`'s semantics between `process` and `hooks`, either current version for process or go semantics works for me.

crosbymichael:
LGTM

mrunalp:
LGTM

wking:
On Wed, Feb 15, 2017 at 02:12:48PM -0800, Mrunal Patel wrote:
> Merged #685.

My reading of the language this landed is that:

  {
    …
    "hooks": {
      "prestart": [
        {
          "path": "/foo/bar"
        }
      ]
    }
  }

would invoke /foo/bar with an empty/unset argument list.  Something
like:

  $ head *.c
  ==> x.c <==
  #include <stdio.h>

  int main(int argc, char **argv) {
    printf("invoked with argc = %d\n", argc);
    return 0;
  }

  ==> y.c <==
  #include <unistd.h>
  
  int main(int argc, char **argv) {
    execv("./x", NULL);
    return 1;
  }

  $ gcc -o x x.c
  $ gcc -o y y.c
  $ ./y
  invoked with argc = 0

Was that the intention?

Or was the intention to restore the old Go semantics where ‘args’
falls back to ‘[path]’ if ‘args’ was unset [1]?  If so, we probably
need a follow-up PR that updates the ‘args’ documentation to specify
the fallback.

[1]: https://golang.org/pkg/os/exec/#Cmd

PR #688, opened 2017-02-14
schema: remove major,minor from device required items

Mashimiao:
When type is p, major and minor is not required.
So, we should not leave them in required.


wking:
0e808fcad looks good to me.

hqhq:
LGTM

mrunalp:
LGTM

PR #690, opened 2017-02-14
schema/config-linux: add pattern limit for deviceCgroup

Mashimiao:

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -194,7 +194,8 @@
                     "type": "boolean"
                 },
                 "type": {
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^[acb]$"
                 },
                 "major": {
                     "$ref": "#/definitions/Major"


schema/defs-linux.json
```diff
@@ -194,7 +194,8 @@
                     "type": "boolean"
                 },
                 "type": {
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^[acb]?$"
```
wking: We should drop the `?`, because the empty string is not a legal value.  The field is optional, so it may be *unset*.  But it cannot be an empty string (for reasons similar to our not supporting explicit `null`s, #662).

schema/defs-linux.json
```diff
@@ -203,7 +204,8 @@
                     "$ref": "#/definitions/Minor"
                 },
                 "access": {
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^[rwm]{1,3}$"
```
wking: This regexp would allow `rrr`, which should be invalid.  But it's probably close enough, since a long chain of `|`ed options would be tedious to write, and I can't think of a better way to test for `r?w?m?` which is order-agnostic and includes at least one character.

Mashimiao: In my opinion ,`^(?!.*(.).*\1)[rwm]{1,3}$` will work well, but it seems json schema does not support `?!`.

wking: > … but it seems json schema does not support ?!.

JSON Schema [supports the ECMA 262 regular expression dialect][1], which [includes `(?!…)`][2], and your proposed regexp looks good to me.  Maybe just a limitation of gojsonschema?  It looks like it chokes on that pattern with:

    invalid character '1' in string escape code

probably because it's [using Go's stock regexp package][3], which [does not support `(?!…)`][4].  Some related discussion in xeipuuv/gojsonschema#93.

So I'm fine with your current, loose regexp.  If/when gojsonschema gains support for `(?!…)`, we can move to your more-specific regexp.

[1]: https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.8
[2]: http://www.ecma-international.org/ecma-262/5.1/#sec-15.10.1
[3]: https://github.com/xeipuuv/gojsonschema/blob/6b67b3fab74d992bd07f72550006ab2c6907c416/validation.go#L660
[4]: https://github.com/google/re2/wiki/Syntax

schema/defs-linux.json
Mashimiao: updated.

schema/defs-linux.json
Mashimiao: Yes, by testing, `?!` and `\1` have not been supported by go regexp.

wking:
f621c2f3 looks good to me.

mrunalp:
We have pointers to the kernel documentation for device cgroups which is good enough. I am closing this in line with the other PRs that we closed that try to specify exact kernel limits on values.

wking:
> I am closing this in line with the other PRs that we closed that try to specify exact kernel limits on values.

This position is now backed by #780 and related recent changes.  But I don't think “we're punting this to the kernel” is a valid position as long as the spec [continues to list the valid values][1] which this PR was validating.  Can we revisit this PR as one of the special cases where we are inlining explicit values?  Or should I file a pull request removing the current valid-value listing for this parameter?

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config-linux.md#L223

Mashimiao:
@opencontainers/runtime-spec-maintainers  as @wking  said, should we reopen this PR?

wking:
>  Or should I file a pull request removing the current valid-value listing for this parameter?

I've gone this way with #841.

wking:
> > Or should I file a pull request removing the current valid-value listing for this parameter?
>
> I've gone this way with #841.

With #841 rejected, I'm back to thinking we should re-open this pull request.  Although I'm fine waiting to see what @crosbymichael covers in his PR to [remove the device defaults][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/841#issuecomment-305573623

hqhq:
As it's not introducing new kernel limitations but reflection of spec, I think we can reconsider it.

hqhq:
@wking Regarding kernel limitations in OCI, I think it's not a simple "we'll have them" or "we don't have them" option, we should see these limitations in different ways. For limitations such as for some magic numbers which are given specific meanings by kernel and look like they're possible to be changed in the future, we should not introduce such limitations in OCI, and for other limitations which seem to be more common sense and stable, like linux or unix level limitation, I think we can have them.

hqhq:
@Mashimiao As the second pattern is not actually accurate, is it really helpful for validation tools? In my opinion if it's not accurate, it'll be a wrong reflection and could cause confusion.

Mashimiao:
The second pattern is not exceedingly needed. As gojsonschema don't support accurate validation, we can handle it  in validation tools.

Mashimiao:
If moved it, the PR has nothing different with #918

hqhq:
@Mashimiao It has the original discussion :)

wking:
> ...it'll be a wrong reflection and could cause confusion.

If there's any chance of this, we need to reiterate that the JSON Schema are going to be too permissive.  That's unavoidable with JSON Schema.  Once we establish that, I think a dozen or so characters to get pretty-close validation are worth adding here.

Mashimiao:
I removed pattern limit for access, now just added pattern for device type.
Can we merge this now?

hqhq:
LGTM

schema/defs-linux.json
```diff
@@ -194,7 +194,8 @@
                     "type": "boolean"
                 },
                 "type": {
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^[acb]$"
```
cyphar: This is not correct, you can have empty `type` fields. It should be `^[acb]?$`.

wking: > It should be `^[acb]?$`.

No, it's right without the `?`.  Empty-string values are not valid in the JSON.  This property can be *unset* in JSON, which we cover inthe JSON Schema by [not listing `type` in `required`][1].  We test the unset-`type` case with [this][2], via [this][3] and [this][4].

In the non-pointer Go property, "unset" is not an option, but the `omitempty` means empty-string Go values are treated as unset for JSON Schema validation.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/schema/defs-linux.json#L224
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/schema/test/config/good/spec-example.json#L257
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/schema/Makefile#L21
[4]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/.travis.yml#L25

vbatts:
_house keeping:_ with us currently considering the reconciliation of cgroup v2, I'm not sure this is something we want to bring in at this point

PR #691, opened 2017-02-16
Add OCI Runtime Abstract

### LINKED ISSUE #506: Suggestion: Add an Abstract section to spec
It would be useful to have a brief summary describing the OCI Runtime spec.  
This may just be a single paragraph of a few sentences.
This would be useful for people coming new to the spec or wanting to have a consistent brief description of the spec.

crosbymichael:
Closes #506

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -2,6 +2,13 @@
 
 The [Open Container Initiative](http://www.opencontainers.org/) develops specifications for standards on Operating System process and application containers.
 
+# Abstract
+
+The OCI Runtime Specification aims to specify the configuration, execution environment, and lifecycle a container.
+
+A container's configuration is specified as the `config.json` for the supported platforms and details the fields that enable the creation of a container.
+The execution environment is specified to ensure that applications running inside a container have a consistent environment between runtimes along with common actions defined for the container's lifecycle.
+
 # Platforms
 
 Platforms defined by this specification are:


mrunalp:
LGTM

spec.md
```diff
@@ -2,6 +2,15 @@
 
 The [Open Container Initiative](http://www.opencontainers.org/) develops specifications for standards on Operating System process and application containers.
 
+# Abstract
+
+The OCI Runtime Specification aims to specify the inputs and outputs for running containers on various platforms.
+
+A container's input or configuration is specified as the `config.json` for the supported platforms and details fields that enable containers to run on that platform.
+It contains host specific information for the creation of a container.
```
wking: I think you want “host specific” → “host-specific”, although `master` has inconsistent hyphenation for this.

spec.md
```diff
@@ -2,6 +2,15 @@
 
 The [Open Container Initiative](http://www.opencontainers.org/) develops specifications for standards on Operating System process and application containers.
 
+# Abstract
+
+The OCI Runtime Specification aims to specify the inputs and outputs for running containers on various platforms.
+
+A container's input or configuration is specified as the `config.json` for the supported platforms and details fields that enable containers to run on that platform.
```
wking: You have “on various platforms” in the previous paragraph, so I don't think you need “for the supported platforms” here.  Maybe just echo the [current configuration description][1]?

> A container's input is a configuration which contains metadata necessary to implement standard operations against the container. This includes the process to run, environment variables to inject, sandboxing features to use, etc.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/config.md#container-configuration-file

spec.md
```diff
@@ -2,6 +2,15 @@
...
+It contains host specific information for the creation of a container.
+
+The outputs of a container are defined as both actions that can be performed on a container after it has been created and the environment details that the container provides to applications running inside of it.
```
wking: These don't sound like “the outputs of a container” to me.  Actions are more like additional inputs.  The container environment is more of a runtime output, and that's probably what you're trying to focus on here.  I'd consider container output to be things like filesystem alterations which persist after container exit (e.g. simulation results written to a bind-mounted volume).  And containers which host services don't fit cleanly into the the idea of “container output” at all.  Maybe de-emphasize input/output and instead sketch the configuration, actions, and lifecycle in a few sentences each?  Something like:

> Containers are [configured](config.md) with metadata necessary to implement standard operations against the container. This includes the process to run, environment variables to inject, sandboxing features to use, etc.  Runtimes [create](runtime.md#create) containers from that configuration, and several [operations](runtime.md#operations) are defined to manage the container over its [lifecycle](runtime.md#lifecycle).

cyphar:
I don't really like the "input and outputs" description either. I'd prefer if we instead emphasised that the runtime specification defines the execution environment configuration and lifecycle management of containers. There are inputs and outputs involved, but the key point is that we're standardising the *lifecycle* and the *configuration* of container runtimes.

crosbymichael:
@cyphar updated based on feedback

cyphar:
lgtm

dqminh:
LGTM

mrunalp:
LGTM

PR #696, opened 2017-02-24
Fix config-schema.json

q384566678:
Signed-off-by: zhouhao zhouhao@cn.fujitsu.com

### DIFF
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -230,7 +230,6 @@
         "ociVersion",
         "platform",
         "process",
-        "root",
-        "mounts"
+        "root"
     ]
 }
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -192,6 +192,11 @@
                     "items": {
                         "id": "https://opencontainers.org/schema/bundle/linux/rlimits/0",
                         "type": "object",
+                        "required": [
+                            "type",
+                            "soft",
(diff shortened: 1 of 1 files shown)


schema/config-schema.json
```diff
@@ -118,6 +118,10 @@
                 "user": {
                     "id": "https://opencontainers.org/schema/bundle/process/user",
                     "type": "object",
+                    "required": [
+                        "uid",
+                        "gid"
```
wking: We can't add this to the JSON Schema, because they're only required on Linux.  [Windows entries will not have them][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/config.md#windows-user

q384566678: That being the case, should the `user` field should be removed from the `config-schema`, into different types（config-linux.json/config-windows.json etc）.

wking: > should the user field should be removed from the `config-schema`, into different types (`config-linux.json`/`config-windows.json` etc).

You could do that if you like, but I don't think it's worth the trouble.  I'd rather leave the user fields optional in the JSON Schema and check for any platform-specific requirements in runtime-tools' Go validation.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #699, opened 2017-02-27
Make mount type and source optional

### LINKED ISSUE #470: [ Config ] Mounts | type is underspecified
The type parameter is a required string and this is defined for Linux however there is no information for Solaris and there is a suggestion of just using "ntfs" for Windows.  

I don't know how important this parameter is for Linux, but if this is not defined for Solaris and there is only one value for Windows, maybe this should be optional; or required for Linux but otherwise optional?

crosbymichael:
These are optional on multiple platforms and should be left up to the
runtime/host system for validation.

Closes #470

### DIFF
--- a/config.md
+++ b/config.md
@@ -57,11 +57,11 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
   This value MUST be an absolute path.
   * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
   * Solaris: corresponds to "dir" of the fs resource in zonecfg(8).
-* **`type`** (string, REQUIRED) The filesystem type of the filesystem to be mounted.
+* **`type`** (string, OPTIONAL) The filesystem type of the filesystem to be mounted.
   * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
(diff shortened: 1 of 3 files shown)


mrunalp:
LGTM

config.md
```diff
@@ -57,11 +57,11 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
   This value MUST be an absolute path.
   * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
   * Solaris: corresponds to "dir" of the fs resource in zonecfg(8).
-* **`type`** (string, REQUIRED) The filesystem type of the filesystem to be mounted.
+* **`type`** (string, OPTIONAL) The filesystem type of the filesystem to be mounted.
```
wking: This probably needs an `omitempty` [here][1] and dropping the `required` entry [here][2].

[1]: https://github.com/opencontainers/runtime-spec/blob/2d491b03fa3c09680c4b557b74825e0db6b04a7b/specs-go/config.go#L116
[2]: https://github.com/opencontainers/runtime-spec/blob/2d491b03fa3c09680c4b557b74825e0db6b04a7b/schema/defs.json#L145

crosbymichael:
@mrunalp @hqhq @dqminh PTAL

mrunalp:
LGTM

config.md
wking: > This probably needs an `omitempty`…

These were addressed in 7c5796d → 5246151, and 5246151 looks good to me.

hqhq:
LGTM

PR #701, opened 2017-02-27
config: Make process optional

wking:
Since #384, it's possible for a container process to never execute user-specified code (e.g. you can call `create`, `kill`, `delete` without calling `start`).  For folks who expect to do that, there's no reason to define `process.args`.

The narrow solution to this would be to make `process.args` optional, but that approach was [rejected in favor of this all-or-nothing approach to `process` handling][1].  What an unset `process` means for the implementation-defined container process between `create` and `start` is unclear, but clarifying how that is handled is a separate issue (#700) independent of whether `process` is optional or not.

[1]: https://github.com/opencontainers/runtime-spec/pull/620#issuecomment-282820279

### DIFF
--- a/config.md
+++ b/config.md
@@ -120,7 +120,8 @@ For Windows, see links for details about [mountvol](http://ss64.com/nt/mountvol.
 
 ## Process
 
-**`process`** (object, REQUIRED) specifies the container process.
+**`process`** (object, OPTIONAL) specifies the container process.
+  This property is REQUIRED when [`start`](runtime.md#start) is called.
 
 * **`terminal`** (bool, OPTIONAL) specifies whether a terminal is attached to that process, defaults to false.
   As an example, if set to true on Linux a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
--- a/runtime.md
+++ b/runtime.md
@@ -107,6 +107,7 @@ This operation MUST generate an error if it is not provided the container ID.
 Attempting to start a container that does not exist MUST generate an error.
 Attempting to start an already started container MUST have no effect on the container and MUST generate an error.
 This operation MUST run the user-specified program as specified by [`process`](config.md#process).
+This operation MUST generate an error if `process` was not set.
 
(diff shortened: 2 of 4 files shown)


cyphar:
I would like it if we had some language in `create` that states that `process.args` should only be "checked" when you hit `start`. `create` should allow for the process to block, waiting to execute a program that doesn't exist.

runC doesn't do this now, but I would like some language to make it clear that this is meant to be a supported workflow.

wking:
On Mon, Feb 27, 2017 at 02:00:08PM -0800, Aleksa Sarai wrote:
> I would like it if we had some language in `create` that states that
> `process.args` should only be "checked" when you hit `start`.

Runtime validation at create-time is currently completely optional
[1].  And it is clear that subsequent changes to the configuration do
not effect the container [2].

I'm guessing that you're interested in runC's (still current?)
create-time lookup for the initial config arg [3], but as far as I can
tell “an executable at process.args[0] MUST exist at create-time is
not a spec requirement.  We do require an executable at
process.args[0] by start-time [4] (which may never happen), but that
shouldn't require a create-time check.

> runC doesn't do this now, but I would like some language to make it
> clear that this is meant to be a supported workflow.

#489, #620, and now this PR have been my attempts at
clarifying/streamlining our support for the “I'm never going to call
‘start’” workflow.  Are you missing some language for that workflow in
this PR?  If so, please tell me what you want added.

Or are you looking for language to cover the “there's nothing at
process.args[0] now, but I'll have it in place by the time I call
‘start’”?  If so, that seems like a separate issue.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc4/runtime.md#L97 
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc4/runtime.md#L100
[3]: https://github.com/opencontainers/runc/pull/827#r63464201
[4]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc4/config.md#L130

mrunalp:
LGTM

crosbymichael:
LGTM

PR #702, opened 2017-02-28
runtime: Remove status redefinitions from operations

wking:
These states are already defined in the "State" section.  There's no need to redefine them in the operation sections.

Operation-level redefinitions are dicy anyway, because they imply something testable about the immediately-after-this-operation time, and it's not possible to run race-free tests of that time (e.g. the process could die for other reasons between the successful `create` call and the `state` call you made to look for a `created` status).


crosbymichael:
plz rebase

wking:
> plz rebase

Rebased onto master with bb429f4 → 2d8d50c, resolving trivial conflicts with #707.

crosbymichael:
LGTM

tianon:
LGTM

Issue #692, opened 2017-02-16
How to set rlimit value to unlimited


sak0:
[opencontainers/runc#1283](https://github.com/opencontainers/runc/issues/1283)

runtime should be support type for "RLIMI_INFINITY" ?

wking:
On Wed, Feb 15, 2017 at 10:17:31PM -0800, cuihaozhi wrote:
> runtime should be support type for "RLIMI_INFINITY" ?

Yeah, I think the runtime-spec would need to be patched to add this
support, if maintainers feel like it should be supported.  Previous
discussion in the thread starting with [1].  Now that we've landed
#648, I expect maintainers would prefer -1 in the JSON to represent
RLIM_INFINITY, which would mean making hard/soft signed integer
properties.

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/nfuv8NnOsUg
     Subject: rlimits
     Date: Thu, 20 Oct 2016 12:29:17 -0700
     Message-ID: <CAJvN1BOB9esrq29_2=rzSquRsRR6Rc10huM6WQoNKNfyFVdufw@mail.gmail.com>

mrunalp:
`RLIM_INFINITY` is defined as unsigned long in the kernel. See https://github.com/torvalds/linux/search?utf8=%E2%9C%93&q=RLIM_INFINITY so one should be able to set it today.

hqhq:
Yes kernel defines `RLIM_INFINITY` as unsigned long, we should use the same value otherwise people won't be able to set values larger than MaxInt64 but smaller than MaxUint64. Regarding set to unlimited, since kernel defines `RLIM_INFINITY = ~0UL` (some architectures might override it), there is no real unlimited value but setting it as a very large one, so users can always set it by themselves, with the fact that:
```
uint64(-1) = 18446744073709551615
int64(18446744073709551615) = -1
```
people can still use `-1` to set it as unlimited when update as long as we do this forced conversion in runtime before generating the config. In this case, upper level tools need to define this value as `int64`, but this limitation should be introduced by upper level tools when they see fit, not by spec.

I'm thinking maybe we should revert https://github.com/opencontainers/runtime-spec/pull/648 , types should always be the same as kernel's, we can only change `cfs_quota_us` as https://github.com/opencontainers/runtime-spec/pull/499 suggested since the kernel type is `signed long long` and takes `-1` as a real unlimited setting.

[...]

PR #704, opened 2017-03-01
Set specs value the same as kernel API input

hqhq:
This partially revert #648 , after a second thought, I think we
should use specs value the same as kernel API input, see:
https://github.com/opencontainers/runtime-spec/issues/692#issuecomment-281889852

For memory and hugetlb limits *.limit_in_bytes, cgroup APIs take the values
as string, but the parsed values are unsigned long, see:
https://github.com/torvalds/linux/blob/v4.10/mm/page_counter.c#L175-L193

For `cpu.cfs_quota_us` and `cpu.rt_runtime_us`, cgroup APIs take the input
value as signed long long, while `cpu.cfs_period_us` and `cpu.rt_periof_us`
take the input value as unsigned long long.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -285,15 +285,15 @@ For more information, see [the memory cgroup man page][cgroup-v1-memory].
 
 The following parameters can be specified to setup the controller:
 
-* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage in bytes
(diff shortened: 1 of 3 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #706, opened 2017-03-02
config-linux.md: fix seccomp

q384566678:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -496,43 +496,70 @@ Seccomp provides application sandboxing mechanism in the Linux kernel.
 Seccomp configuration allows one to configure actions to take for matched syscalls and furthermore also allows matching on values passed as arguments to syscalls.
 For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
 The actions, architectures, and operators are strings that match the definitions in seccomp.h from [libseccomp][] and are translated to corresponding values.
-A valid list of constants as of libseccomp v2.3.2 is shown below.
 
-Architecture Constants
-* `SCMP_ARCH_X86`
-* `SCMP_ARCH_X86_64`
-* `SCMP_ARCH_X32`
-* `SCMP_ARCH_ARM`
-* `SCMP_ARCH_AARCH64`
-* `SCMP_ARCH_MIPS`
-* `SCMP_ARCH_MIPS64`
-* `SCMP_ARCH_MIPS64N32`
-* `SCMP_ARCH_MIPSEL`
-* `SCMP_ARCH_MIPSEL64`
-* `SCMP_ARCH_MIPSEL64N32`
-* `SCMP_ARCH_PPC`
-* `SCMP_ARCH_PPC64`
-* `SCMP_ARCH_PPC64LE`
-* `SCMP_ARCH_S390`
-* `SCMP_ARCH_S390X`
-* `SCMP_ARCH_PARISC`
-* `SCMP_ARCH_PARISC64`
+**`seccomp`** (object, OPTIONAL)
 
-Action Constants:
-* `SCMP_ACT_KILL`
-* `SCMP_ACT_TRAP`
-* `SCMP_ACT_ERRNO`
-* `SCMP_ACT_TRACE`
-* `SCMP_ACT_ALLOW`
+The following parameters can be specified to setup seccomp:
 
-Operator Constants:
-* `SCMP_CMP_NE`
-* `SCMP_CMP_LT`
-* `SCMP_CMP_LE`
-* `SCMP_CMP_EQ`
-* `SCMP_CMP_GE`
-* `SCMP_CMP_GT`
-* `SCMP_CMP_MASKED_EQ`
+* **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as `syscalls[].action`.
+
+* **`architectures`** *(array of strings, OPTIONAL)* - the architecture used for system calls.
+    A valid list of constants as of libseccomp v2.3.2 is shown below.
+
+    * `SCMP_ARCH_X86`
+    * `SCMP_ARCH_X86_64`
+    * `SCMP_ARCH_X32`
+    * `SCMP_ARCH_ARM`
+    * `SCMP_ARCH_AARCH64`
+    * `SCMP_ARCH_MIPS`
+    * `SCMP_ARCH_MIPS64`
+    * `SCMP_ARCH_MIPS64N32`
+    * `SCMP_ARCH_MIPSEL`
+    * `SCMP_ARCH_MIPSEL64`
+    * `SCMP_ARCH_MIPSEL64N32`
+    * `SCMP_ARCH_PPC`
+    * `SCMP_ARCH_PPC64`
+    * `SCMP_ARCH_PPC64LE`
+    * `SCMP_ARCH_S390`
+    * `SCMP_ARCH_S390X`
+    * `SCMP_ARCH_PARISC`
+    * `SCMP_ARCH_PARISC64`
+
+* **`syscalls`** *(array of objects, REQUIRED)* - match a syscall in seccomp.
+
+    Each entry has the following structure:
+
+    * **`names`** *(array of strings, REQUIRED)* - the names of the syscalls.
+
+    * **`action`** *(string, REQUIRED)* - the action for seccomp rules.
+        A valid list of constants as of libseccomp v2.3.2 is shown below.
+
+        * `SCMP_ACT_KILL`
+        * `SCMP_ACT_TRAP`
+        * `SCMP_ACT_ERRNO`
+        * `SCMP_ACT_TRACE`
+        * `SCMP_ACT_ALLOW`
+
+    * **`args`** *(array of objects, OPTIONAL)* - the specific syscall in seccomp.
+
+        Each entry has the following structure:
+
+        * **`index`** *(uint, REQUIRED)* - the index for syscall arguments in seccomp.
+
+        * **`value`** *(uint64, REQUIRED)* - the value for syscall arguments in seccomp.
+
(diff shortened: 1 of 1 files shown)


config-linux.md
```diff
@@ -490,9 +490,9 @@ For more information, see [the man page](http://man7.org/linux/man-pages/man8/sy
    }
 ```
 
-## seccomp
+## Seccomp
 
-Seccomp provides application sandboxing mechanism in the Linux kernel.
+**`seccomp`** (object, OPTIONAL) provides application sandboxing mechanism in the Linux kernel.
```
cyphar: I don't feel this is correct, mainly because "Seccomp" here is referring to the kernel feature not the object in the spec. One of the other "Seccomp" references should be changed.

wking: > One of the other "Seccomp" references should be changed.

+1 to not changing this instance, but I don't think there is another reference which should be changed instead.  I think the seccomp section is just missing a Markdown spec, so we need to add completely new language along the lines of:

```
* **`seccomp`** (object, OPTIONAL)

    The following parameters can be specified to setup seccomp:

    * **`defaultAction`** (string, REQUIRED) …something about what this means…

        Runtimes MUST support at least the following values:

        * `SCMP_ACT_KILL`
        * `SCMP_ACT_TRAP`
        …
    …
```

config-linux.md
```diff
@@ -498,41 +498,56 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
+  * **`syscalls`** *(object, REQUIRED)* - match a syscall in seccomp.
+
+    * **`names`** *(array of strings, REQUIRED)* - the name of the syscall.
+
+    * **`action`** *(string, REQUIRED)* - the action for syscall.
+
+    * **`args`** *(object, OPTIONAL)* - the specific syscall in seccomp.
+      
+	* **`op`** *(string, REQUIRED)* - the operator for syscall arguments in seccomp. Implementations MUST support at least the following values: 
+
```
q384566678: Do not support further indentation here, what good way? In addition there are some fields, whether it is necessary to write all?

wking: > Do not support further indentation here, what good way?

Markdown lists should be nested with [four-space indents][1] (see also #495).  So you want something like:

```
* **`seccomp`** (object, OPTIONAL)
    The following parameters can be specified to setup seccomp:

    * **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp.

    * **`architectures`** *(array, OPTIONAL)* - the architecture used for system calls.
        Implementations MUST support at least the following values:

        * `SCMP_ARCH_X86`
        …
```

[1]: http://daringfireball.net/projects/markdown/syntax#list

q384566678: updated

config-linux.md
```diff
@@ -498,41 +498,59 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
-* `SCMP_CMP_LE`
-* `SCMP_CMP_EQ`
-* `SCMP_CMP_GE`
-* `SCMP_CMP_GT`
-* `SCMP_CMP_MASKED_EQ`
+**`seccomp`** (object, OPTIONAL)
+
+The following parameters can be specified to setup seccomp:
+
+* **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp.
```
wking: I think this entry needs another line like:

> Allowed values are the same as `sycalls[].action`.

to give some guidance on the expected string values.

config-linux.md
```diff
@@ -498,41 +498,59 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
+
+* **`architectures`** *(array, OPTIONAL)* - the architecture used for system calls.
```
wking: “array” → “array of strings”.

config-linux.md
```diff
@@ -498,41 +498,59 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
+    * `SCMP_ARCH_MIPSEL64N32`
+    * `SCMP_ARCH_PPC`
+    * `SCMP_ARCH_PPC64`
+    * `SCMP_ARCH_PPC64LE`
+    * `SCMP_ARCH_S390`
+    * `SCMP_ARCH_S390X`
+    * `SCMP_ARCH_PARISC`
+    * `SCMP_ARCH_PARISC64`
+
+* **`syscalls`** *(object, REQUIRED)* - match a syscall in seccomp.
```
wking: “object” → “array of objects”.  And then a new line like:

> Each entry has the following structure:

to match the precedent set [here][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config-linux.md#L79

config-linux.md
```diff
@@ -498,41 +498,59 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
+    * **`action`** *(string, REQUIRED)* - the action for seccomp rules.
+        Implementations MUST support at least the following values:
+
+        * `SCMP_ACT_KILL`
+        * `SCMP_ACT_TRAP`
+        * `SCMP_ACT_ERRNO`
+        * `SCMP_ACT_TRACE`
+        * `SCMP_ACT_ALLOW`
+
+    * **`args`** *(object, OPTIONAL)* - the specific syscall in seccomp.
```
wking: “object” → “array of objects”. And then a new line like:

> Each entry has the following structure:

Also, you list `op`, but it looks like you also need to document the [`index`, `value`, and `valueTwo` properties][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/specs-go/config.go#L541-L543

crosbymichael:
LGTM

config-linux.md
```diff
@@ -498,41 +498,69 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
-* `SCMP_CMP_LE`
-* `SCMP_CMP_EQ`
-* `SCMP_CMP_GE`
-* `SCMP_CMP_GT`
-* `SCMP_CMP_MASKED_EQ`
+**`seccomp`** (object, OPTIONAL)
+
+The following parameters can be specified to setup seccomp:
+
+* **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as sycalls[].action.
```
wking: Backticks around `syscalls[].action` (and fix the `sycalls` → `syscalls` typo).  Also maybe split a new line at the sentence break.

q384566678: updated, thanks!

config-linux.md
```diff
@@ -498,41 +498,69 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
+    * `SCMP_ARCH_S390`
+    * `SCMP_ARCH_S390X`
+    * `SCMP_ARCH_PARISC`
+    * `SCMP_ARCH_PARISC64`
+
+* **`syscalls`** *(array of objects, REQUIRED)* - match a syscall in seccomp.
+
+    Each entry has the following structure:
+
+    * **`names`** *(array of strings, REQUIRED)* - the name of the syscall.
```
hqhq: the names of the syscalls.

config-linux.md
```diff
@@ -498,41 +498,69 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
+
+    * **`args`** *(array of objects, OPTIONAL)* - the specific syscall in seccomp.
+
+        Each entry has the following structure:
+
+        * **`index`** *(uint, REQUIRED)* - the index for syscall arguments in seccomp.
+
+        * **`value`** *(uint64, REQUIRED)* - the value for syscall arguments in seccomp.
+
+        * **`valueTow`** *(uint, REQUIRED)* - the value for syscall arguments in seccomp.
```
hqhq: Why is this `uint` not `uint64`?

config-linux.md
```diff
@@ -498,41 +498,69 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
 The actions, architectures, and operators are strings that match the definitions in seccomp.h from [libseccomp][] and are translated to corresponding values.
 A valid list of constants as of libseccomp v2.3.2 is shown below.
```
hqhq: These two sentences should be moved to below sections and adapted accordingly.

config-linux.md
```diff
@@ -498,41 +498,69 @@ For more information about Seccomp, see [Seccomp][seccomp] kernel documentation.
...
-* `SCMP_CMP_GT`
-* `SCMP_CMP_MASKED_EQ`
+**`seccomp`** (object, OPTIONAL)
+
+The following parameters can be specified to setup seccomp:
+
+* **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as `syscalls[].action`.
+
+* **`architectures`** *(array of strings, OPTIONAL)* - the architecture used for system calls.
+    Implementations MUST support at least the following values:
```
hqhq: As said above, these values are copied from certain version of libseccomp, we should not use MUST here to force implementations use certain version of libseccomp.

q384566678: updated, PTAL

hqhq:
LGTM

config-linux.md
```diff
@@ -496,43 +496,70 @@ Seccomp provides application sandboxing mechanism in the Linux kernel.
...
+
+    * **`args`** *(array of objects, OPTIONAL)* - the specific syscall in seccomp.
+
+        Each entry has the following structure:
+
+        * **`index`** *(uint, REQUIRED)* - the index for syscall arguments in seccomp.
+
+        * **`value`** *(uint64, REQUIRED)* - the value for syscall arguments in seccomp.
+
+        * **`valueTow`** *(uint64, REQUIRED)* - the value for syscall arguments in seccomp.
```
wking: `valueTow` -> `valueTwo`

q384566678:
@crosbymichael @hqhq PTAL

hqhq:
LGTM

crosbymichael:
LGTM

PR #709, opened 2017-03-03
style: Backtick `DOC_FILES`, `_N`, and `N`.

wking:
These are literal quotes, and we don't want Markdown to confuse the underscores with [emphasis delimiters][1].

[1]: http://daringfireball.net/projects/markdown/syntax#em


hqhq:
LGTM

dqminh:
LGTM

PR #712, opened 2017-03-06
schema/defs-linux: Fix type for seccomp names

wking:
The:

    "type": [
      "string"
    ]

syntax added in #657 is not valid:

    $ ./validate ./config-schema.json <../config.json
    - linux.seccomp.syscalls.0.names: Invalid type. Expected: string, given: array


hqhq:
LGTM

mrunalp:
LGTM

PR #687, opened 2017-02-09
make link usage consistent across the specification

jlbutler:
Some hopefully useful link cleanup throughout the specification. Closes #676

Signed-off-by: Jesse Butler <jesse.butler@oracle.com>

PR #713, opened 2017-03-07
config-linux: make interface name clear

Mashimiao:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -438,7 +438,7 @@ The following parameters can be specified to setup the controller:
 
 * **`priorities`** *(array, OPTIONAL)* - specifies a list of objects of the priorities assigned to traffic originating from processes in the group and egressing the system on various interfaces.
   The following parameters can be specified per-priority:
-    * **`name`** *(string, REQUIRED)* - interface name
+    * **`name`** *(string, REQUIRED)* - interface name in [runtime network namespace](glossary.md#runtime-namespace)
     * **`priority`** *(uint32, REQUIRED)* - priority applied to the interface
 
 ###### Example


crosbymichael:
Are you sure this is correct?  There are an awful lot of `eth0` when you run containers.

Mashimiao:
cgroups works on host, you can't run a container with limiting an interface which not exists on host., cgroups will not accept that. For example by runc:
$ cat ~/testdir/config.json  |grep -A 5 -B 5 eth0 
    		},
			"network": {
        		"classID": 1048577,
			    "priorities": [
           		{
               		"name": "eth0",
               		"priority": 500
           		}
        		]		
			}
		},

$ sudo runc run --bundle ~/testdir test
container_linux.go:247: starting container process caused "process_linux.go:386: container init caused \"process_linux.go:327: setting cgroup config for ready process caused \\\"failed to write eth0 500 to net_prio.ifpriomap: write /sys/fs/cgroup/net_cls,net_prio/test/net_prio.ifpriomap: no such device\\\"\""

cyphar:
While I'm sure this is a correct statement, this is IMO another instance where we shouldn't be adding wording around Linux kernel implementation details. If someone wants to put a random value in a field they'd better understand what they're actually doing (in terms of the kernel features they're using) otherwise things would break no matter how much guidance we give.

From my PoV I see no reason why this restriction will always be true in the kernel -- how is the case handled where you have a file descriptor set up in a different network namespace (for instance)? if the kernel changes some behaviour (or they extend it) then we'd have to update the spec.

Mashimiao:
I think if we don't make it clear, people may be confused interface name should be which in container network namespace or in runtime network namespace.
At least, it seems @crosbymichael  is thinking about interface name is which in container network namespace.

crosbymichael:
I think it mis-read it the first time sorry.  It does seem correct.  I'm indifferent about adding this line, I think its fine leaving it out like @cyphar said but its also fine adding it as the additional is pretty straight forward.

Mashimiao:

hqhq:
LGTM

config-linux.md
```diff
@@ -438,7 +438,7 @@ The following parameters can be specified to setup the controller:
 
 * **`priorities`** *(array, OPTIONAL)* - specifies a list of objects of the priorities assigned to traffic originating from processes in the group and egressing the system on various interfaces.
   The following parameters can be specified per-priority:
-    * **`name`** *(string, REQUIRED)* - interface name
+    * **`name`** *(string, REQUIRED)* - interface name in [runtime network namespace](glossary.md#runtime-namespace)
```
tianon: It appears the rest of this file is using footer links (`[link text][link ID]` with then a `[link ID]: URL` at the bottom of the file such as `[container-namespace2]: glossary.md#container_namespace`); should we stay consistent on that?

wking: > It appears the rest of this file is using footer links…

I think those just slipped through #687 and have filed #799 to catch them up.  @Mashimiao's current inline links match the current style for inline links.

tianon:
LGTM

PR #714, opened 2017-03-07
remove comment

q384566678:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -549,8 +549,7 @@ Operator Constants:
                    "getcwd",
                    "chmod"
                ],
-               "action": "SCMP_ACT_ERRNO",
-               "comment": "stop exploit x"
+               "action": "SCMP_ACT_ERRNO"
            }
        ]
    }
--- a/config.md
+++ b/config.md
@@ -766,8 +766,7 @@ Here is a full example `config.json` for reference.
                         "getcwd",
                         "chmod"
                     ],
-                    "action": "SCMP_ACT_ERRNO",
-                    "comment": "stop exploit x"
+                    "action": "SCMP_ACT_ERRNO"
                 }
             ]
         },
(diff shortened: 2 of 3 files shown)


cyphar:
NACK. There's a very good reason that JSON doesn't support comments. While you might not agree with it, it is a valid reason. Not to mention that the runtime-spec **specifically** tells runtimes to ignore fields they don't recognise (so using `comment` should already work).

crosbymichael:
What does this PR actually do?  I don't understand.

wking:
On Tue, Mar 07, 2017 at 06:06:15AM -0800, Aleksa Sarai wrote:
> NACK. There's a very good reason that JSON doesn't support comments.

This PR is catching the JSON Schema up with [1].  If you don't want
comments in the JSON, you probably want to file a PR removing the Go
field.

[1]: https://github.com/opencontainers/runtime-spec/pull/657/files#diff-7f24d60f0cbb9c433e165467e3d34838R535

cyphar:
-_-

crosbymichael:
I think that just slipped by.  Its not necessary for the spec so we should probably remove it.  The spec should be short and concise, only containing information required for the creation/execution of a container.  

I can see that a comment is a nice to have but not required and probably shouldn't be in the spec.

q384566678:
So the result of your discussion is to delete the `comment` field? Is there a clear rule to delete `comment`?

crosbymichael:
@q384566678 ya, I don't think it should have been added in the first place and just slipped by in review.  We don't have any other  comment fields in the spec.

q384566678:
@crosbymichael As you said, I think  should delete the comment in the spec, as I did in https://github.com/opencontainers/runtime-spec/pull/714/commits/b7a172ddfca13a6657c1f8da506006642ea484dd.

cyphar:
:+1:

config-linux.md
```diff
@@ -549,8 +549,7 @@ Operator Constants:
                    "getcwd",
                    "chmod"
                ],
-               "action": "SCMP_ACT_ERRNO",
-               "comment": "stop exploit x"
+               "action": "SCMP_ACT_ERRN"
```
hqhq: typo?

q384566678: updated, PTAL

hqhq:
LGTM

crosbymichael:
LGTM

PR #719, opened 2017-03-09
README: Document the alternate-week meeting times

wking:
On Thu, Mar 09, 2017 at 08:37:55AM -0700, Chris Aniszczyk [wrote][1]:
> summarizing the discussion, how about we just alternative the time like Jonathan discussed?
>
> I believe that's the only fair thing to do and it's reassuring to hear from people like Phil who would be able to make the time along with others.
>
> I'm fine with the current time (1000AEST/1500PST/0000CET) and then 0400AEST/0900PST/1800CET

Folks with a [POSIX `date` command][2] can find the week number with:

    $ date +%V
    10

That means that next week will be an odd week with the 2pm PDT meeting (because the US goes into daylight savings time this weekend), which is 2016-03-15 21:00 UTC.

There may be some doubling up around the end of the year, but we're usually canceling meetings around then anyway.

I'd be happier with meeting times anchored to UTC to make life easier for folks outside of the US, but one change at a time ;).  And I think UTC-anchored times have been rejected before, but I haven't dug up a reference for that.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/p0mTOspVgd0/mh7FYse2BAAJ
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,11 @@ When in doubt, start on the [mailing-list](#mailing-list).
 
 ### Weekly Call
 
-The contributors and maintainers of all OCI projects have a weekly meeting Wednesdays at 2:00 PM (USA Pacific).
+The contributors and maintainers of all OCI projects have a weekly meeting on Wednesdays at:
+
+* 9:00 AM (USA Pacific), during [even weeks][iso-week].
+* 2:00 PM (USA Pacific), during [odd weeks][iso-week].
+
 Everyone is welcome to participate via [UberConference web][uberconference] or audio-only: 415-968-0849 (no PIN needed.)
 An initial agenda will be posted to the [mailing list](#mailing-list) earlier in the week, and everyone is welcome to propose additional topics or suggest other agenda alterations there.
 Minutes are posted to the [mailing list](#mailing-list) and minutes from past calls are archived to the [wiki][runtime-wiki].
@@ -148,6 +152,7 @@ Read more on [How to Write a Git Commit Message][how-to-git-commit] or the Discu
 [dev-list]: https://groups.google.com/a/opencontainers.org/forum/#!forum/dev
 [how-to-git-commit]: http://chris.beams.io/posts/git-commit
 [irc-logs]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/
+[iso-week]: https://en.wikipedia.org/wiki/ISO_week_date#Calculating_the_week_number_of_a_given_date
 [oci]: https://www.opencontainers.org
 [runtime-wiki]: https://github.com/opencontainers/runtime-spec/wiki
 [uberconference]: https://www.uberconference.com/opencontainers


crosbymichael:
Please wait until discussions are finished before making changes to any schedule.

jonboulle:
What denotes the discussion as "finished"? This seems like a hasty closure.

> I'd be happier with meeting times anchored to UTC to make life easier for folks outside of the US,

+1, I think the conversation today showed there are plenty of interested participants outside PST.

wking:
On Thu, Mar 09, 2017 at 09:41:02AM -0800, Jonathan Boulle wrote:
> What denotes the discussion as "finished"? This seems like a hasty
> closure.

There seemed to be general agreement on the split (if not the times)
in the list thread, so I figured it was time to put a hashed proposal
forward to collect more formal consensus.  But I'm happy to continue
to update this PR to track whatever the consensus seems to be if the
thread continues to evolve.

crosbymichael:
There were still replies coming in and much of the discussion happened over the evening where many in the US timezone have not responded or voiced their comments.

README.md
```diff
@@ -60,7 +60,11 @@ When in doubt, start on the [mailing-list](#mailing-list).
 
 ### Weekly Call
 
-The contributors and maintainers of all OCI projects have a weekly meeting Wednesdays at 2:00 PM (USA Pacific).
+The contributors and maintainers of all OCI projects have a weekly meeting on Wednesdays at:
+
+* 9:00 AM (USA Pacific), during [even weeks][iso-week].
```
sameo: 8:00 AM PST, i.e. 5:00PM CET would work better for several of us.

jonboulle: thanks for chiming in, cross-reference: https://groups.google.com/a/opencontainers.org/d/msg/dev/p0mTOspVgd0/ULXnARy9BAAJ

wking:
On Thu, Mar 09, 2017 at 10:42:09AM -0800, Samuel Ortiz wrote:
> +* 9:00 AM (USA Pacific), during [even weeks][iso-week].
> 
> 8:00 AM PST, i.e. 5:00PM CET would work better for several of us.

I haven't heard anyone arguing for 9am over 8am, and you two are both
happy with 8am, so I've updated to use 8am Pacific in 765c091 → 39adc01.

wking:
I've pushed 39adc01 → afbe47b pulling in the Doodle poll results to motivate 8am Pacific and 5pm Pacific times, although how relevant those results are as folks move in and out of various daylight savings times is not clear to me.

wking:
I've pushed afbe47b → 8f99038 swapping the odd/even order to get next
week's meeting at 8am Pacific to be Europe-friendly for KubeCon [1].
I think we should re-open this PR at some point so GitHub will notice
the branch changes; until then, I guess folks can look at my branch
[2].

I've also added a meeting.ics file [3], which folks can import into
their calendar app to see how it plays.  There are two events there,
because I was unable to find the right RRULE invocation to cover both
times in one event, and RFC 5545 does not allow multiple RRULEs in one
VEVENT [4] (maybe @tianon can figure out some RRULE magic?).

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-03-22-21.00.log.html
[2]: https://github.com/wking/opencontainer-runtime-spec/tree/alternating-meetings
[3]: https://raw.githubusercontent.com/wking/opencontainer-runtime-spec/alternating-meetings/meeting.ics
[4]: https://tools.ietf.org/html/rfc5545#appendix-A.1

tianon:
Naw, two `VEVENT` objects is exactly what I was planning.  I don't think we can re-open this PR now; GitHub says "The alternating-meetings branch was force-pushed or recreated."  (not sure if re-pushing the commit it was at when it closed would do it or not, but probably not worth the trouble to try -- a new PR seems much simpler IMO)

PR #729, opened 2017-03-15
[Config Linux] Replace "required" (lower case)

RobDolinMS:
Replaces "required" (lower case) with "necessary" since this instance does not seem to be intended as "REQUIRED"

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -186,7 +186,7 @@ You can configure a container's cgroups via the `resources` field of the Linux c
 Do not specify `resources` unless limits have to be updated.
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
-A runtime MUST at least use the minimum set of cgroup controllers required to fulfill the `resources` settings.
+A runtime MUST at least use the minimum set of cgroup controllers necessary to fulfill the `resources` settings.
 However, a runtime MAY attach the container process to additional cgroup controllers supported by the system.
 
 ###### Example


config-linux.md
```diff
@@ -186,7 +186,7 @@ You can configure a container's cgroups via the `resources` field of the Linux c
 Do not specify `resources` unless limits have to be updated.
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
-A runtime MUST at least use the minimum set of cgroup controllers required to fulfill the `resources` settings.
+A runtime MUST at least use the minimum set of cgroup controllers necessary to fulfill the `resources` settings.
 However, a runtime MAY attach the container process to additional cgroup controllers supported by the system.
```
wking: Do we even want to keep the line you're bumping?  It sounds like any runtime which violated that constraint would necessarily violate some more specific constraint on a `resources` setting.  Maybe replace this paragraph with:

> Runtimes MAY attach the container process to additional cgroup controllers beyond those required to fulfill the `resources` settings.

tianon:
LGTM

Suffers from a Travis failure: (might need a rebase and/or DCO munging?)
```console
$ TRAVIS_COMMIT_RANGE="${TRAVIS_COMMIT_RANGE/.../..}" make .gitvalidation
git-validation -v -run DCO,short-subject,dangling-whitespace -range 78e6667ae2d67aad100b28ee9580b41b7a24e667..HEAD
2017/03/15 22:55:26 exit status 128
make: *** [.gitvalidation] Error 1
The command "TRAVIS_COMMIT_RANGE="${TRAVIS_COMMIT_RANGE/.../..}" make .gitvalidation" exited with 2.
```

Seems to me that regardless of whether we keep the line, avoiding the word `REQUIRED` where it isn't intended to be a strict validation-related usage is a sane thing to do.

wking:
On Wed, May 10, 2017 at 01:17:47PM -0700, Tianon Gravi wrote:
> Seems to me that regardless of whether we keep the line, avoiding
> the word `REQUIRED` where it isn't intended to be a strict
> validation-related usage is a sane thing to do.

Agreed, although I don't see the point of making the narrow fix if we
can come to an agreement on dropping the redundant MUST.  I've filed
#800 with a version of my earlier proposal that avoids “required” in
case maintainers agree that the current MUST is redundant.

tianon:
Given https://github.com/opencontainers/runtime-spec/pull/800, my preference is to close this and merge that one instead (since IMO it makes the intent of this bit more clear).

wking:
Obsolete with #800 landed?

tianon:
Yeah, thanks @wking (and @RobDolinMS)

PR #730, opened 2017-03-16
runtime: Drop "Barring access control concerns"

wking:
This wording landed without comment as part of #225.  However, I'm not entirely clear on the exception it's making.  It may be trying to say something like:

> Just because you were authorized to manage that container when you created it doesn't mean you're still authorized to perform operation X on it now.  Maybe you've lost privileges in the meantime.

But as far as compliance testing is concerned, the same test harness will be calling `create` and the subsequent operations.  That harness will be reporting MUST violations if the runtime refuses a subsequent operation, and removing the access-control loophole makes it more obvious that the runtime's refusal is non-compliant.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -2,7 +2,7 @@
 
 ## <a name="runtimeScopeContainer" />Scope of a Container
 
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

hqhq:
LGTM

PR #733, opened 2017-03-16
runtime: Drop "not supported by the base OS" loophole

wking:
All of these operations should be supported on all of our compliance-tested platforms.  If there are cases where a given OS cannot support one of these operations, it should be discussed in that operation's section, and the "Unless otherwise stated" qualifier makes space for that.  But leaving the reader to decide on whether a host OS supports a given operation seems too unstable for a specification.

I've also dropped the "OCI compliant" qualifier, because the (not) compliant language in `spec.md` already ties OCI compliance to the RFC 2119 language.  There's no point in mentioning "OCI compliant" outside of that section, because the "MUST" already brings in all of the compliance associations we need.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -82,7 +82,7 @@ Unless otherwise stated, logging a warning does not change the flow of the opera
 
 ## <a name="runtimeOperations" />Operations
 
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #734, opened 2017-03-16
runtime: Container-scope-wide uniqueness for container IDs

wking:
As written:

>  MUST be unique across all containers on this host.

seems unenforceable.  How will runC know what IDs are in use with other runtimes on that host?  And how will a runtime in one namespace know which IDs are used in another namespace?

The create-time language is much closer to being enforcable, by invoking the container-scope idea.  To firm this up, I've adjusted the wording to aligh with the name of the scope section, and I've linked to that scope section.

With the create-time check in place, I don't think there's any purpose to documenting the scope again in the state JSON section, so I've dropped those two lines.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -11,8 +11,6 @@ The state of a container includes the following properties:
 
 * **`ociVersion`** (string, REQUIRED) is the OCI specification version used when creating the container.
 * **`id`** (string, REQUIRED) is the container's ID.
-This MUST be unique across all containers on this host.
-There is no requirement that it be unique across hosts.
 * **`status`** (string, REQUIRED) is the runtime state of the container.
 The value MAY be one of:
 
@@ -99,7 +97,7 @@ This operation MUST return the state of a container as specified in the [State](
 `create <container-id> <path-to-bundle>`
 
 This operation MUST [generate an error](#errors) if it is not provided a path to the bundle and the container ID to associate with the container.
-If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST [generate an error](#errors) and a new container MUST NOT be created.
+If the ID provided is not unique across all containers within the [scope of the container](#scope-of-a-container), or is not valid in any other way, the implementation MUST [generate an error](#errors) and a new container MUST NOT be created.
 Using the data in [`config.json`](config.md), this operation MUST create a new container.
 This means that all of the resources associated with the container MUST be created, however, the user-specified program MUST NOT be run at this time.
 If the runtime cannot create the container as specified in [`config.json`](config.md), it MUST [generate an error](#errors) and a new container MUST NOT be created.


jonboulle:
Can't you just change it to "managed by this runtime on this host"?

wking:
On Thu, Mar 16, 2017 at 04:31:28AM -0700, Jonathan Boulle wrote:
> Can't you just change it to "managed by this runtime on this host"?

I don't think so, because not all ‘create’ invocations will be able to
share the set of currently allocated IDs.  For example, runC uses
/run/runc/{container-ID} to track allocated IDs.  But if you launch
runC from a separate mount namespace which masks that directory:

  $ CID=374957eb2656a3418b0723ced9d5d177b8630a117c3f43591507d71f91b6ec36
  $ runc state ${CID} | jq .pid
  7784
  $ runc exec ${CID} runc state ${CID}
  $ runc exec ${CID} ls /run/runc
  ls: cannot access /run/runc: No such file or directory

Your wording is fine if by “host” you mean “whichever portions of the
host environment that the runtime uses to communicate currently
allocated IDs”.  But that's what the “Scope of a Container” is
establishing and what I'm trying to do in this PR.

Do you feel like repeating the ‘create’ restriction where we define
the state's ‘id’ property helps?  I'd rather avoid double-MUSTing the
same restriction.  If we want some docs around the state ‘id’
property, I'd rather have something like:

  See the [`create`](#create) section for restrictions on `id`.

runtime.md
```diff
@@ -99,7 +97,7 @@ This operation MUST return the state of a container as specified in the [State](
 `create <container-id> <path-to-bundle>`
 
 This operation MUST [generate an error](#errors) if it is not provided a path to the bundle and the container ID to associate with the container.
-If the ID provided is not unique across all containers within the scope of the runtime, or is not valid in any other way, the implementation MUST [generate an error](#errors) and a new container MUST NOT be created.
+If the ID provided is not unique across all containers within the [scope of the container](#scope-of-a-container), or is not valid in any other way, the implementation MUST [generate an error](#errors) and a new container MUST NOT be created.
```
crosbymichael: This does not make any sense to me.

wking: > This does not make any sense to me.

Which portion?  Does the previous “scope of the runtime” make sense to you?  What does it mean to you?  Something other than the [scope of the container][1]?  Does the scope of the container (as [we currently define it][1]) make sense to you?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/runtime.md#scope-of-a-container

hqhq: @wking In my understanding, this statement means container ID should be unique within all containers created by one runtime, but two containers created by different runtimes on the same host can be the same.

Your change to `unique id across all containers within the scope of the container` doesn't make sense to me. I think it's not `id within the scope of xxx`, but `all containers within the scope of xxx`.

wking: @hqhq, I'm not sure what you're asking for.  Maybe:

> If the ID provided is not unique within the [scope of the container](#scope-of-a-container), or is not valid in any other way, the implementation MUST [generate an error](#errors) and a new container MUST NOT be created.

?

hqhq: The [scope of the container](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/runtime.md#scope-of-a-container) we link here, doesn't quite fit this statement, it's scopt of **A** container, if an ID is unique or not, would be compared with other containers', which is out of the scope of **A** container.

wking: > ... if an ID is unique or not, would be compared with other containers', which is out of the scope of A container.

I expect

    $ funC create $ID ...

to be checking for an exiting $ID container using the same funC-specific state-sharing mechanism that

   $ fun state $ID

and other operations on that container are using.  I think the "Scope of a Container" section is really about the scope of this state-sharing (and I'm happy to rename that section if it helps clarify this).

Do you see a point in distinguishing between "state-sharing for ID deduping" and "state-sharing for everything else"?

hqhq: > I think the "Scope of a Container" section is really about the scope of this state-sharing 

I don't quite see that, maybe the `scope of a container` section needs to be clarified.

wking: > > I think the "Scope of a Container" section is really about the scope of this state-sharing
>
> I don't quite see that, maybe the scope of a container section needs to be clarified.

My previous attempt at clarifying the connection is [here][1].  I'm happy to cherry-pick that into this PR and make any necessary adjustments if that helps.  But I don't see how the runtime is expected to enforce ID uniqueness without using something like its state sharing mechanism.

[1]: https://github.com/duglin/runtime-spec/pull/1/commits/2fff5454fb7f20beec6110ff8725fc8507f4cce8

mrunalp:
I'll create another PR per discussion today to clarify runtime instance.

PR #736, opened 2017-03-16
runtime-linux: Condition /proc/self/fd symlinks on source existence

wking:
Since #666 it's no longer guaranteed that `/proc` will exist.  And there doesn't seem to be much point in requiring symlinks which will be known broken.

This commit also tightens the timing.  Before it was just “after the container has `/proc` mounted”, which could have happened during the `delete` operation (if the container authors wanted to be especially ornery).  With this commit, I've put the creation in [step 2 of the lifecycle][1].  And within step 2, it happens after `mounts` has been processed.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/runtime.md#L54-L57


tianon:
LGTM

PR #737, opened 2017-03-16
config: Lead off with the purpose of the config

wking:
Instead of leading off with links to a bunch of other places, notes on the Go tags, etc., make things more inviting by leading off with a big-picture summary of what the configuration is about.

Also drop the `config.json` existence MUST because:

1. This section defines the configuration format, and doesn't need to be tied to a particular filename.
2. The bundle spec (in `bundle.md`) already [has][1]:

    > This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.

The `config.md` line may have been useful when it was added (77d44b10).  But since the `bundle.md` line landed in #210, I think it's been redundant.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/bundle.md#L14

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,11 +1,11 @@
 # <a name="containerConfigurationFile" />Container Configuration file
 
+This configuration file contains metadata necessary to implement standard operations against the container.
+This includes the process to run, environment variables to inject, sandboxing features to use, etc.
+
 The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).
 [Platform](spec.md#platforms)-specific configuration schema are defined in the [platform-specific documents](#platform-specific-configuration) linked below.
 For properties that are only defined for some [platforms](spec.md#platforms), the Go property has a `platform` tag listing those protocols (e.g. `platform:"linux,solaris"`).
-
-The configuration file contains metadata necessary to implement standard operations against the container.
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

config.md
```diff
@@ -1,13 +1,12 @@
 # <a name="containerConfigurationFile" />Container Configuration file
 
-The container's top-level directory MUST contain a configuration file called `config.json`.
```
cyphar: Did you mean to drop this line? Maybe you should include it below the moved paragraph?

wking: > Did you mean to drop this line?

Yes, see “Also drop the `config.json` existence MUST because” and later.  If that is contentious, I can add it back in this PR and file a follow-up PR for its removal.

vbatts: this seems like a bizzare flyby. :-1:

wking: > this seems like a bizzare flyby

It didn't seem contentious to me, and it's all in the same section of the spec.  But it is an orthogonal change, so I dropped it from this PR with c223790 → 0952dd8 and filed #756 for just the redundant filename removal.

crosbymichael:
rebase plz

wking:
Rebased with 0952dd8 → ff071cc, working around the just-landed #756 which [I'd spun off this PR earlier][1] :p.

[1]: https://github.com/opencontainers/runtime-spec/pull/737#discussion_r109963257

mrunalp:
LGTM

crosbymichael:
LGTM

PR #738, opened 2017-03-16
config: Fix broken JSON in Windows mount example

wking:
The `key: value` syntax is for objects, but `mounts` is an array.


mrunalp:
LGTM

hqhq:
LGTM

PR #747, opened 2017-03-29
spec: Define the property-specification syntax

wking:
And add platform annotations.

By formally defining our syntax, we do a better job of making that syntax enforcable.  We can also make platform-scoping very obvious, and no longer need to rely on “For Linux-based systems the process supports…” guards and similar.

The only intentional semantic change is that we now allow, platform-named properties that don't match platform.os.  For example:

    {
      "platform": {
        "os": "linux",
        …
      },
      "windows": {
        …
      },
      …
    }

and even:

    {
      "platform": {
        "os": "linux",
        …
      },
      "windows": "foo",
      …
    }

are both legal now.  This rolls back the “MUST NOT be set otherwise” requirements which had landed in #673 to make those properties consistent with our [general][1]:

> Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown property.

[and][2]:

> The state MAY include additional properties.

extensibility policies.  The alternative would be cross-platform specification of those properties which then forbid the properties on non-matching platforms, but then the cross-platform definition would not match the platform-specific Go type.  For example:

    Linux *Linux `json:"linux,omitempty" platform:"linux"`

Blocking the non-matching-platform properties didn't seem to be worth the potential confusion of the Go `platform` tag not matching the Markdown `platforms` tag.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#extensibility
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/runtime.md#L31

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -26,7 +26,7 @@ For more information, see the [namespaces(7)][namespaces.7_2] man page.
 Namespaces are specified as an array of entries inside the `namespaces` root field.
 The following parameters can be specified to setup namespaces:
 
-* **`type`** *(string, REQUIRED)* - namespace type. The following namespace types are supported:
+* **`type`** *(string, REQUIRED, linux)* - namespace type. The following namespace types are supported:
     * **`pid`** processes inside the container will only be able to see other processes inside the same container.
     * **`network`** the container will have its own network stack.
     * **`mount`** the container will have an isolated mount table.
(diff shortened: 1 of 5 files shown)


wking:
On Wed, May 10, 2017 at 02:14:28PM -0700, Mrunal Patel wrote:
> Closed #747.

Without the ABNF and associated docs [1], there's nothing connecting
(for example) to unpack:

  foo (object, REQUIRED) Something about foo…

into:

  foo values MUST be objects.  Configurations MUST include the foo
  property.  Something about foo…

I think we want *something* to help with that unpacking, even if it
doesn't use an ABNF (although I think ABNF is a nice, compact way to
clarify the syntax).  Do we have any alternative plans for clarifying
the syntax?

PR #748, opened 2017-03-29
README: E.123 number for meeting call

wking:
And fix a period/paren swap.  The period/paren swap and contry-code addition both bring the local version more in line with [the image-spec version][1] and seem like good changes.

I've diverged from the image-spec version by using space separation in the phone number instead of hyphens.  From [E.123][3]:

> 9.1  Grouping of digits in a telephone number⁵ should be accomplished by means or spaces⁶ unless an agreed upon explicit symbol (e.g. hyphen) is necessary for procedural purposes. Only spaces should be used in an international number.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/README.md#weekly-call
[3]: https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-E.123-200102-I!!PDF-E&type=items

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ The contributors and maintainers of all OCI projects have a weekly meeting on We
 
 There is an [iCalendar][rfc5545] format for the meetings [here](meeting.ics).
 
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

vbatts:
LGTM

PR #749, opened 2017-03-29
*: Replace "array" type with "array of objects"

wking:
We have a few different element types in our arrays, so it's useful to clarify the element type for the property being specified.  Before this commit:

    $ sed -n 's|.*\*\*`\([^`]*\)`\*\*[^(]*(\([^,]*\),.*|\2|p' *.md | sort | uniq -c | grep array
          7 array
          1 array of ints
          8 array of objects
         13 array of strings

All of the bare `array` instances turned out to be arrays of objects.


hqhq:
LGTM

mrunalp:
LGTM

PR #750, opened 2017-03-30
schema: modify Makefile

q384566678:
Signed-off-by: zhouhao zhouhao@cn.fujitsu.com

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -1,10 +1,11 @@
 
-default: help
+default: validate
 
 help:
-	@echo "Usage: make <target>"
+	@echo "Usage: make [target]"
 	@echo
 	@echo " * 'fmt' - format the json with indentation"
+	@echo " * 'help' - show this help information"
 	@echo " * 'validate' - build the validation tool"
 
 fmt:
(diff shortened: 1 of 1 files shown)


hqhq:
I don't think it's necessary and it'll make the help information non-accurate.

Rejected.

vbatts:
I don't see that it makes the help info incorrect. I'm two-ways about the default make target being `help`. Honestly, most of the time I expect the default target to be sensible, like building the tool or running the tests.
This PR LGTM in that sense

hqhq:
@vbatts Not that incorrect, just not completely accurate.
```
$ make
Usage: make <target>

 * 'fmt' - format the json with indentation
 * 'validate' - build the validation tool
```
To me it means "make needs an target, which can be fmt or validate", if we make validate default and target is optional, I would like to see something like:
```
$ make
Usage: make [target]

 * 'fmt' - format the json with indentation
 * 'validate' - build the validation tool (default)
 * 'help' - show this help information
```

I'm also fine with changing the default target to be running the tests, just needs to fix help information as well.

vbatts:
@hqhq oh oh, good catch. so yes, using square brackets there would be more accurate :smile_cat:

q384566678:
@hqhq Thanks for your advice, update, PTAL.

hqhq:
LGTM

q384566678:
@vbatts @mrunalp @titanous PTAL

vbatts:
LGTM

PR #752, opened 2017-03-30
GOVERNANCE.md: fix typo

q384566678:


dqminh:
LGTM.

Since we copied the template from https://github.com/opencontainers/project-template/blob/master/GOVERNANCE.md, the fix also needs to be ported there.

q384566678:
OK, I will do it.

mrunalp:
LGTM

PR #753, opened 2017-04-04
config: Replace '$GOOS' with 'GOOS', etc.

### LINKED ISSUE #477: Environment variables use platform-specific semantics
The spec refers to environment variables using the Linux semantics $FOO, $BAR, etc.  

Suggest using phrasing "for the environment variable FOO", "environment variable $BAR", etc.

wking:
Windows' [`cmd.exe` prefers wrapping in `%`][1] (e.g. `%GOOS%`), and @RobDolinMS has asked for the POSIX `$` to be removed (#477).  In this case, I think the context is sufficient that we can just use the variable name instead of Rob's recommended “for the environment variable FOO”.

Fixes #477.

[1]: https://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/ntcmds_shelloverview.mspx?mfr=true

### DIFF
--- a/config.md
+++ b/config.md
@@ -298,12 +298,12 @@ For Windows based systems the user structure has the following fields:
 
 * **`os`** (string, REQUIRED) specifies the operating system family of the container configuration's specified [`root`](#root) file system bundle.
   The runtime MUST generate an error if it does not support the specified **`os`**.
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

vbatts:
LGTM

PR #756, opened 2017-04-05
config: Drop the redundant config.json filename requirement

wking:
Because:

1. This section defines the configuration format, and doesn't need to be tied to a particular filename.
2. The bundle spec (in `bundle.md`) already has:

    > This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.

The config.md line may have been useful when it was added (77d44b10).  But since the bundle.md line landed in 106ec2da (#210), I think it's been redundant.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,6 +1,5 @@
 # <a name="containerConfigurationFile" />Container Configuration file
 
-The container's top-level directory MUST contain a configuration file called `config.json`.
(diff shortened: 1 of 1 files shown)


mrunalp:
LGTM

crosbymichael:
LGTM

PR #703, opened 2017-02-28
schema: add username for user field

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #760, opened 2017-04-07
config: process.user.username is implementation-defined on Windows

### LINKED ISSUE #618: Windows process.username optional but POSIX process.uid required?
On POSIX (currently Linux and Solaris), [`uid` and `gid` are required][1]. My preferred approach here is to [make those optional][2] and [use platform defaults][3]:

    If unset, the runtime will not attempt to manipulate the user ID (e.g. not calling setuid(2) or similar).

But the maintainer consensus is that they want those to be [explicitly][4] [required][5] [properties][6].

The Windows `username`, on the other hand, [*is* optional][7], although the default behavior is unclear.  I see no discussion in #565 to suggest whether this was intentionally approved or not.

So the current spec seems to have different policies for whether user information is required, and it would be nice to make that consistent.  I'd rather make it consistent by making the POSIX fields optional.  Is the implicit support for an optional Windows `username` sufficient grounds to make the POSIX fields optional?  Or was the optional `username` an accident?  Or do maintainers approve of the current inconsistency?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#linux-and-solaris-user
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/DWdystx5X3A
[3]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216076069
[4]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-04-17.00.log.html#l-44
[5]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937010
[6]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937090
[7]: https://github.com/opencontainers/runtime-spec/blob/cc983bb6f4ca85194dc6d64ee300cca00e495070/config.md#windows-user

wking:
On POSIX (currently Linux and Solaris), [`uid` and `gid` are required][1]. My preferred approach here is to [make those optional][2] and [use platform defaults][3]:

    If unset, the runtime will not attempt to manipulate the user ID (e.g. not calling setuid(2) or similar).

But the maintainer consensus is that they want those to be [explicitly][4] [required][5] [properties][6].

The Windows `username`, on the other hand, [*is* optional][7], although the default behavior is unclear.  I see no discussion in #565 to suggest whether this was intentionally approved or not.  And in #618 (where I asked about the inconsistency), @crosbymichael [said][8]:

> No, both should be made explicit unless there is something on windows that prohibits this.

So this commit is making that happen.

Pinging some Windows folks for review of the “unless there is something on windows that prohibits this” condition: @jhowardmsft, @RobDolinMS, @jstarks.

Note that there is also a JSON Schema PR for `username` in flight with #703, but that is orthogonal to this change (the JSON Schema cannot require `username` even if the field is REQUIRED on Windows, because it is not required, or even specified, on other OSes).

Fixes #618.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#linux-and-solaris-user
[2]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/DWdystx5X3A
[3]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216076069
[4]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-05-04-17.00.log.html#l-44
[5]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937010
[6]: https://github.com/opencontainers/runtime-spec/pull/417#issuecomment-216937090
[7]: https://github.com/opencontainers/runtime-spec/blob/cc983bb6f4ca85194dc6d64ee300cca00e495070/config.md#windows-user
[8]: https://github.com/opencontainers/runtime-spec/issues/618#issuecomment-277105273

### DIFF
--- a/config.md
+++ b/config.md
@@ -260,6 +260,8 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
 For Windows based systems the user structure has the following fields:
 
 * **`username`** (string, OPTIONAL) specifies the user name for the process.
+    The value MUST NOT be an empty string.
+    The default `username` is implementation-defined.
 
 ### Example (Windows)
 
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -130,6 +130,11 @@
                         "additionalGids": {
                             "id": "https://opencontainers.org/schema/bundle/process/user/additionalGids",
                             "$ref": "defs.json#/definitions/ArrayOfGIDs"
+                        },
+                        "username": {
+                            "id": "https://opencontainers.org/schema/bundle/process/user/username",
+                            "type": "string",
+                            "minLength": 1
                         }
                     }
                 },


jhowardmsft:
Does NOT LGTM. This absolutely IS an optional field on Windows.

wking:
On Thu, Apr 06, 2017 at 10:43:58PM -0700, John Howard wrote:
> This absolutely IS an optional field on Windows.

uid/gid are technically optional on Linux too (just don't call setuid,
and you inherit the runtime caller uid/gid, modulo namespace mapping
[1]).  The spec has a policy (which I don't agree with, but I'm not a
maintainer) that the uid and gid values are required anyway (on
Linux).  See links in my original post for maintainers on this point.

If ‘username’ stays optional, then I think the spec should be updated
to clarify what happens when it is not specified.  For example, is
there a default username (e.g. is it inherited from the runtime
caller)?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#user-namespace-mappings

jhowardmsft:
image

And with an explicitly added user
image

Feel free to clarify. But this PR does not LGTM.

wking:
> And no, it's not inherited at all.

So which of the following would make more sense in the spec.

> If unset, `username` defaults to `ContainerAdministrator`.

or

> The default `username` is implementation-defined.

?

jhowardmsft:
I would prefer `The default username is implementation-defined.`

wking:
On Mon, Apr 24, 2017 at 09:31:32AM -0700, John Howard wrote:
> I would prefer `The default username is implementation-defined.`

Done with 4e3945f → a3ef036.  I've also added JSON Schema for this,
and forbidden the empty-string value.  If anyone with Windows
experience feels that empty-string usernames should be allowed, please
let me know.

crosbymichael:
Closing based on feedback from the windows team

wking:
> Closing based on feedback from the windows team

What feedback from the Windows team?  The current commit (a3ef036) has been [endorsed][1] by @jhowardmsft, who's the only Windows member involved in this PR.  Without this PR, the property remains:

* Undefined (as in [nasal demons][2]) when someone takes advantage of the OPTIONAL-ness and leaves the property unset.  With this PR that use-case would be implementation-defined.
* Unchecked in the JSON Schema.  With this PR, we'd validate that the value was a string.

And [as I said earlier][3], I'm fine dropping the new “but not the empty string” requirement if anyone has issues with it.

[1]: https://github.com/opencontainers/runtime-spec/pull/760#issuecomment-296731380
[2]: https://en.wikipedia.org/wiki/Undefined_behavior
[3]: https://github.com/opencontainers/runtime-spec/pull/760#issuecomment-296743531

wking:
Since the [pivot from REQUIRED to OPTIONAL][1], the JSON Schema change in this PR mostly matches that which just landed in #703.  So the only remaining features of this PR (which I think we still want) are:

* Making the default `username` implementation-defined (it is currently undefined).
* Forbidding empty-string usernames (although I'm fine dropping this if Windows somehow supports empty-string usernames).

[1]: https://github.com/opencontainers/runtime-spec/pull/760#issuecomment-296743531

PR #761, opened 2017-04-07
config: Specify height/width units (characters) for consoleSize

wking:
This is pretty clear from the examples like [this][1]:

    "consoleSize": {
      "height": 25,
      "width": 80
    }

But examples are not normative, so this commit makes the character units normative.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L180


mrunalp:
LGTM

crosbymichael:
LGTM

Issue #762, opened 2017-04-07
config-linux: Is `syscalls` really required?


wking:
While reviewing our required/optional settings, I noticed that `linux.seccomp.sycalls` is required, but that we don't require an entry in the array.  That means `"syscalls": []` would be technically valid, and I'm pretty sure that's not what we want.

If it makes sense to have a `seccomp` property that does not need `syscalls` entries, then `syscalls` should be optional.

If it does not makes sense to have an empty/unset `syscalls` then it should be required *and* have a minimum length of one.

Before #657, `syscalls` was `omitempty` (and therefore more optional-feeling, although there was no real Markdown spec for seccomp before #706, so it's hard to know).  My current impression is that the property should be optional (and get its old `omitempty` back), but I'm fine either way as long as we have a consistent position.

[1]: https://github.com/opencontainers/runtime-spec/pull/657#discussion_r102613331

wking:
`syscalls`' `names` property is in the same boat.  If [`names` is required][1], I think we need to also require it to have a length ≥ 1.

[1]: https://github.com/opencontainers/runtime-spec/blame/0946333a3604d653ddbe8af1674fc6cf00d9f6d5/config-linux.md#L617

PR #763, opened 2017-04-07
specs-go/config: Add omitempty to LinuxSyscall.Args

wking:
It used to have this, but the `omitempty` [was][1] [dropped][2] in #657.  However, the docs that landed in #706 [list the property as optional][3], and if it is optional, we can leave it unset instead of serializing an empty array.

If it should be required, then we should also require it to have a minimum length of 1 (#762).

Ping @GrantSeltzer, @q384566678.

[1]: https://github.com/opencontainers/runtime-spec/pull/657/files#diff-7f24d60f0cbb9c433e165467e3d34838L534
[2]: https://github.com/opencontainers/runtime-spec/pull/657/files#diff-7f24d60f0cbb9c433e165467e3d34838R534
[3]: https://github.com/opencontainers/runtime-spec/pull/706/files#diff-428eec4013a655816cdefafd5d3505f1R543

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -551,7 +551,7 @@ type LinuxSeccompArg struct {
 type LinuxSyscall struct {
 	Names  []string           `json:"names"`
 	Action LinuxSeccompAction `json:"action"`
-	Args   []LinuxSeccompArg  `json:"args"`
+	Args   []LinuxSeccompArg  `json:"args,omitempty"`
 }
 
 // LinuxIntelRdt has container runtime resource constraints


q384566678:
If we do not add `omitempty` to `Args`, the following error occurs when the validate command is executed:

> 'LinuxSyscall.Args' should not be empty.

In the assignment to the `syscalls`, there is the case of [Args value](https://github.com/opencontainers/runtime-tools/blob/master/generate/seccomp/seccomp_default.go#L422) is empty, so in order to avoid this conflict, I think we should add `omitempty`.

cyphar:
:+1: I've hit this with `umoci` very recently because of the `DefaultSeccomp` value in runtime-tools.

wking:
If maintainers visit this PR in the mood to clean up the Go config
types, there are some more fixups in #759 (which I'm happy to submit
as stand-alone PRs if you want them without the round-trip testing).

hqhq:
LGTM

mrunalp:
LGTM

PR #764, opened 2017-04-07
config: Require strictly-positive timeout values

wking:
If the timeout value was zero, the hook would always error, and there doesn't seem to be much point to that.  And I'm not sure what negative timeouts would mean.  By adding this restriction, we do not limit useful hook entries, and we give the validation code grounds to warn users attempting to validate configs which are poorly defined or have useless hook entries.

Removing the pointer is not strictly required, but keeps up with our [anti-pointer zero-value style][1] now that Go's zero-value is clearly invalid.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/style.md#optional-settings-should-not-have-pointer-go-types

### DIFF
--- a/config.md
+++ b/config.md
@@ -357,6 +357,7 @@ Hooks allow for the configuration of custom actions related to the [lifecycle](r
     * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001 `execv`'s *argv*][ieee-1003.1-2001-xsh-exec].
     * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
     * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
+      If set, `timeout` MUST be greater than zero.
   * **`poststart`** (array of objects, OPTIONAL) is an array of [post-start hooks](#poststart).
     Entries in the array have the same schema as pre-start entries.
   * **`poststop`** (array of objects, OPTIONAL) is an array of [post-stop hooks](#poststop).
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -91,7 +91,8 @@
                     "$ref": "#/definitions/Env"
                 },
                 "timeout": {
-                    "type": "integer"
+                    "type": "integer",
+                    "minimum": 1
                 }
             },
             "required": [


mrunalp:
LGTM

crosbymichael:
I don't think the pointer should be removed here

wking:
> I don't think the pointer should be removed here

Because?  We removed similar in unnecessary pointers in #653.

erikh:
This was done for similar reasons (which I presume are nulls being supported) in image-spec recently, it seems like a valid place here: https://gist.github.com/eef405c950a123c1b178aa706e935139

Just seagulling, hope it's ok.

wking:
On Thu, Apr 13, 2017 at 08:26:40AM -0700, Erik Hollensbe wrote:
> This was done for similar reasons (which I presume are nulls being
> supported) in image-spec recently, it seems like a valid place here:
> opencontainers/image-spec#633

In that case (and also in opencontainers/image-spec#625 and
opencontainers/image-spec#626), you need a pointer for the optional
property, because of golang/go#11939.  However, in this case ‘Timeout’
is an int, and (unlike structs) those *do* have zero values that
omitempty supports.  So the non-pointer int property works, and is in
line with our current policy [1] (since #653 / #656).  Maybe the
policy is closer to:

  New properties should not have pointer Go types when the zero value
  works with omitempty and is a no-op or invalid.  For existing
  properties, maintainers will decide on a case-by-case basis with no
  recognizable pattern, so don't worry too much about trying to guess
  what they want.

;)

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/style.md#optional-settings-should-not-have-pointer-go-types

crosbymichael:
You cannot apply just blanket rules to everything, its not all 1 or 0s.  If you remove the pointer, and a `0` is invalid, you make every config invalid because `0` is the default value.  So how are we supposed to validate this at all without a pointer on the `int` when it will always be 0?

specs-go/config.go
```diff
@@ -126,7 +126,7 @@ type Hook struct {
 	Path    string   `json:"path"`
 	Args    []string `json:"args,omitempty"`
 	Env     []string `json:"env,omitempty"`
-	Timeout *int     `json:"timeout,omitempty"`
```
crosbymichael: Add back the pointer to this field so we can validate a actual user provided 0 value.

wking: > Add back the pointer to this field so we can validate a actual user provided 0 value.

This seems like it would be an issue with the string properties de-pointerized by #653.  I think that validation tools should be using the JSON Schema or `interface{}` parsing to check this sort of thing (pull request to get that to happen in opencontainers/runtime-tools#197).  If the goal is to be able to distinguish “unset” from “set to an invalid/no-op value”, then we do want a pointer here, but probably also want to get the pointer back for [`CgroupsPath`][1], etc.  Or is there a reason why we want to be able to easily identify “`timeout` set to 0” in Go but don't need that for “`cgroupsPath` set to an empty string”?

[1]: https://github.com/opencontainers/runtime-spec/pull/653/files#diff-7f24d60f0cbb9c433e165467e3d34838L142

wking: In the interest of landing the normative spec change, I've dropped the de-pointerization here with 8510055 → ecf7314 and opened #772 around whether we need invalid-zero-value detection for optional parameters.

mrunalp:
LGTM

crosbymichael:
LGTM

Issue #755, opened 2017-04-05
Linux capabilities not listed


justincormack:
The Linux capabilities supported are not listed in the spec, but as they are strings and not numbers they probably ought to be, with the spec changed as more are added. If they were numbers it would seem ok, as you could just pass a new number for a new cap, but with strings, each implementation needs to add support to convert the string to a number so it seems sane to have them explicitly listed and a minor version spec bump should another be added (a rare event).

vbatts:
fair. this is one of those rabbit-holes of effectively specifying the linux kernel API. Thoughts on how to make this age well?

cyphar:
@justincormack But the `kill` operation requires specifying a `signal` (which is a string) but we don't really provide any strong requirements what the strings mean. Should we also specify signal names as well (I'm just wondering which side of the consistency fence we want to land on)?

wking:
On Wed, Apr 05, 2017 at 03:01:41AM -0700, Justin Cormack wrote:
> The Linux capabilities supported are not listed in the spec, but as
> they are strings and not numbers they probably ought to be, with the
> spec changed as more are added.

My preferred way to do this would be to reference a tagged release of
the man-pages listing (e.g. [1]).  But inlining them all would be
acceptable too.  However, in the past there has been resistance to
requiring support for particular values (see [2] and later).

[1]: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/tree/man7/capabilities.7?h=man-pages-4.10
[2]: https://github.com/opencontainers/runtime-spec/pull/673#discussion_r99353136

justincormack:
Yeah, not entirely sure, I just noticed this when I was thinking where is the list of all caps I can get from the spec. Its all because of C's use of macros deep down. I am fine with us deciding to just put some wording instead of the list, but thought I would bring it up for discussion. It is a question of whether there is an explicit spec version that supports `CAP_SYS_FOO` or some implementations do and some dont, and you cant work it out from the spec version that you specify.

Currently we do specify the list of namespace kinds (which seems reasonable as they need special handling), and the list of seccomp architectures (kind of weird), and seccomp actions (less weird).

justincormack:
@wking the man pages are a poor source, they are often outdated. Better to link to the kernel source.

justincormack:
hmm, the comment in #673 covers much of the same ground, with no real conclusion. So going to close this on the grounds that its already discussed ;)

vbatts:
[...]

PR #766, opened 2017-04-11
schema/defs-linux: Drop 'Capability' type

wking:
The current “[For example, valid values for Linux…][1]” wording in config.md does not seem strong enough to support this condition, especially since the spec makes no claims about what valid capabilities are for non-Linux OSes.  And `process.capabilities` has been nominally legal for non-Linux OSes since #673.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L135-L136

### DIFF
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -141,35 +141,35 @@
                             "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/bounding",
                             "type": "array",
                             "items": {
-                                "$ref": "defs-linux.json#/definitions/Capability"
+                                "type": "string"
                             }
                         },
                         "permitted": {
                             "id": "https://opencontainers.org/schema/bundle/process/linux/capabilities/permitted",
                             "type": "array",
                             "items": {
-                                "$ref": "defs-linux.json#/definitions/Capability"
+                                "type": "string"
                             }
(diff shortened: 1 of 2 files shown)


wking:
Related to #755 and the discussion starting [here][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/673#discussion_r99353136

crosbymichael:
Lets keep as is, the string's schema is what we set here.

tianon:
LGTM

As I've mentioned over on https://github.com/opencontainers/runtime-spec/pull/790, we're punting to the kernel, and it just happens to include `CAP_` in all current valid values, so we should simply specify "string" here IMO.

wking:
On Tue, May 09, 2017 at 01:54:04PM -0700, Tianon Gravi wrote:
> LGTM
> 
> As I've mentioned over on
> https://github.com/opencontainers/runtime-spec/pull/790, we're
> punting to the kernel, and it just happens to include `CAP_` in all
> current valid values, so we should simply specify "string" here IMO.

So now that #790 has landed, are we ready to reopen this one?

crosbymichael:
LGTM

mrunalp:
LGTM

Issue #746, opened 2017-03-28
config-linux: Many config properties without RFC 2119 language


wking:
I've filed #745 to add RFC 2119 language to `disableOOMKiller`, but there are many other settings in `config-linux.md` without RFC 2119 wording to connect them with [compliance testing][1] (e.g. [`oomScoreAdj`][2], [`memory`][3], …).  Without RFC 2119 language, it is not very clear what config-author and runtime responsibilities are intended to be.  And it is clear that, whatever the intentions were:

* A config would be technically valid with ridiculous values in the setting (e.g. `"oomScoreAdj" = 999999999999999999999999999999999999999999999999999999999999999999`).
* A runtime would be technically valid if it silently ignored the setting.

I think we should address this before cutting 1.0.  #745 is a (small) step in that direction, but I wanted to open this issue to get feedback on the issue in general (without getting bogged down on a specific setting).

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#notational-conventions
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#set-oom_score_adj
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#memory

wking:
#728 is another PR that is partially addressing this.

wking:
In the meeting today, folks were asking for a way to close this issue, so here's a checklist of all the properties in config-linux for whether or not they have sufficient runtime-facing RFC 2119 language for compliance testing (as of 138ad89):

* [ ] `namespaces` (#767)
* [ ] `uidMappings`
* [ ] `gidMappings`
* [x] `devices`
    * [ ] `type`
    * [ ] `path`
    * [ ] `major`
    * [ ] `minor`
    * [ ] `fileMode`
    * [ ] `uid`
    * [ ] `gid`
* [x] default devices
    * [x] `/dev/null`
    * [x] `/dev/zero`
    * [x] `/dev/random`
    * [x] `/dev/urandom`
    * [x] `/dev/tty`
    * [ ] `/dev/console`
    * [ ] `/dev/ptmx`
* [x] `cgroupsPath`
* [ ] `resources`
    * [ ] `devices`
        * [ ] `allow`
        * [ ] `type`
        * [ ] `major`
        * [ ] `minor`
        * [ ] `access`
    * [ ] `disableOOMKiller`
    * [ ] `oomScoreAdj`
[...]

PR #767, opened 2017-04-12
config-linux: RFC 2119 tightening for namespaces

wking:
Previously we had no MUST-level runtime requirements for namespace entries in valid configs.  This commit attempts to pin those down.  For more background on hierarchical namespaces, see [here][1].  For more background on the owning user namespace idea, see [here][2], [here][3], and [here][4].

The “`path` not associated with a namespace of type `type`” condition ensures that runtimes don't blindly call `setns(2)` on the path without setting `nstype` nonzero.

Part of #746.

Ping @cyphar, especially on the owning-user-namespace condition.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a7306ed8d94af729ecef8b6e37506a1c6fc14788
[2]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6786741dbf99e44fb0c0ed85a37582b8a26f1c3b
[3]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=e5ff5ce6e20ee22511398bb31fb912466cf82a36
[4]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d95fa3c76a66b6d76b1e109ea505c55e66360f3c

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -35,11 +35,14 @@ The following parameters can be specified to setup namespaces:
     * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container.
     * **`cgroup`** the container will have an isolated view of the cgroup hierarchy.
 
-* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace)
+* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace).
+    The runtime MUST place the container process in the namespace associated with that `path`.
+    The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
 
-If a path is specified, that particular file is used to join that type of namespace.
+    If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
+
 If a namespace type is not specified in the `namespaces` array, the container MUST inherit the [runtime namespace](glossary.md#runtime-namespace) of that type.
-If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST error out.
+If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST [generate an error](runtime.md#errors).
 
 ###### Example
 


config-linux.md
```diff
@@ -35,11 +35,17 @@ The following parameters can be specified to setup namespaces:
...
     * **`cgroup`** the container will have an isolated view of the cgroup hierarchy.
 
-* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace)
+* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace).
+    The runtime MUST place the container process in the namespace associated with that `path`.
+    The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
+
+    If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
+    For hierarchical namespaces (e.g. `pid`, `user`), the new container namespace MUST be a child of the [runtime namespace](glossary.md#runtime-namespace) of that type.
+    For seeded namespaces (e.g. `mount`, `uts`), the new container namespace MUST be seeded by the runtime namespace of that type.
```
cyphar: It's not clear to me why these two restrictions are necessary (who the parent / what the seed should be). On the one hand they are useful for some circumstances but I'm having trouble convincing myself that these requirements are not going make weird restrictions for a non-cli interface.

wking: > On the one hand they are useful for some circumstances…

For example, if you ask for a new `uts` namespace but do not set the optional [`hostname`][1], having the seed defined means that the hostname in the container UTS namespace is well-defined (it will be whatever the hostname was in the runtime UTS namespace).

This is less of an issue for the `mount` namespace, because with [`root.path`] REQUIRED, there's no way to avoid clobbering whatever mounts you got from your seed (which makes *not* asking for a new `mount` namespace exciting ;).

> … but I'm having trouble convincing myself that these requirements are not going make weird restrictions for a non-cli interface…

We already have some of “runtime namespace” conditions (e.g. for [`linux.namespaces[].path`][3] which have this “caller may not have a good idea what the runtime namespace actually is” situation.  And this is clearly an issue already for namespaces which are *not* set in [`linux.namespaces[]`][4].

I expect runtimes, CLI and otherwise, will document what they consider the runtime namespaces to be to avoid any ambiguity.  For a CLI, it would probably, but not necessarily, be the caller's namespaces.  For a web UI, it might be:

> The runtime mount namespace will only contain a root directory owned by 0:0 with 755 permissions.  The runtime UTS namespace will have a `test` hostname.  …

But I don't see how this has an impact unique to the “defining the seed namespace” use case.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#hostname
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#root
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config-linux.md#L38
[4]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config-linux.md#L41

crosbymichael: You are specifying implementation specific functionality vs just telling people what the field in the spec is used for.

wking: > You are specifying implementation specific functionality vs just telling people what the field in the spec is used for.

If we want to explicitly make the parent / seed implementation-defined and not tied to the implementation-defined runtime namespaces, I guess we could do that.  But I think “I this container to run in a new user/pid namespace that is a child of the runtime user/pid namespace” should be something that has a portable config expression.  Otherwise it becomes very unclear what to put in the `hostID` field for `(u|g)idMappings`, because you don't know what namespace will be used to interpret the `hostID`s.

And while I expect the child/seeded namespaces will be created via a `clone(2)` call or an `unshare(2)` call, the wording here does not require that.  If the kernel grows new ways to create new namespaces, runtimes will be free to use them, as long as they preserve the required relationship between namespaces.

config-linux.md
```diff
@@ -35,11 +35,17 @@ The following parameters can be specified to setup namespaces:
     * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container.
     * **`cgroup`** the container will have an isolated view of the cgroup hierarchy.
 
-* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace)
+* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace).
+    The runtime MUST place the container process in the namespace associated with that `path`.
+    The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
```
mrunalp: We can drop the text below this line.

config-linux.md
```diff
@@ -35,11 +35,17 @@ The following parameters can be specified to setup namespaces:
     * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container.
     * **`cgroup`** the container will have an isolated view of the cgroup hierarchy.
 
-* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace)
+* **`path`** *(string, OPTIONAL)* - an absolute path to namespace file in the [runtime mount namespace](glossary.md#runtime-namespace).
+    The runtime MUST place the container process in the namespace associated with that `path`.
```
crosbymichael: 39, 40 and 42 all good to keep

wking: > 39, 40 and 42 all good to keep

Done (punting the other lines to #795).

mrunalp:
LGTM

crosbymichael:
LGTM

PR #768, opened 2017-04-12
config-linux: Make linux.seccomp.syscalls OPTIONAL

### LINKED ISSUE #762: config-linux: Is `syscalls` really required?

wking:
Fixes #762, since the issue there has been quiet, and [1.0 may be almost upon us][1], so I'm nervous about leaving the issue to cook for too long ;).

This commit has gone with OPTIONAL, because a seccomp config which only sets `defaultAction` seems potentially valid.  But I'd be ok with the “REQUIRED with length > 0” approach to fixing #762 too.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-04-12-14.59.log.html#l-13

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -610,7 +610,10 @@ The following parameters can be specified to setup seccomp:
     * `SCMP_ARCH_PARISC`
     * `SCMP_ARCH_PARISC64`
 
-* **`syscalls`** *(array of objects, REQUIRED)* - match a syscall in seccomp.
+* **`syscalls`** *(array of objects, OPTIONAL)* - match a syscall in seccomp.
+
+    While this property is OPTIONAL, some values of `defaultAction` are not useful without `syscalls` entries.
+    For example, if `defaultAction` is `SCMP_ACT_KILL` and `syscalls` is empty or unset, the kernel will kill the container process on its first syscall.
 
     Each entry has the following structure:
 
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -251,7 +251,10 @@
                             "$ref": "defs-linux.json#/definitions/Syscall"
                         }
                     }
-                }
+                },
+                "required": [
+                    "defaultAction"
+                ]
             },
             "sysctl": {
                 "id": "https://opencontainers.org/schema/bundle/linux/sysctl",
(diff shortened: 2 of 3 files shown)


dklyle:
Technically, OPTIONAL is the right value, but unless you specify the default action for seccomp to be SCMP_ACT_ALLOW the result will be an error at run time.

I would suggest an additional clarification to this fact in config-linux.md would be very helpful if marking syscall as OPTIONAL.

wking:
On Tue, Apr 25, 2017 at 01:32:26PM -0700, David Lyle wrote:
> I would suggest an additional clarification to this fact…

I've added a narrower example in fa423ce → bb4519c.  How does that
look to you?

dklyle:
I think that's much better. I think a more general statement including your text might be more informative, something along the lines of:

"Note that depending on the value of `defaultAction` specified, syscalls should be non-empty. For example, if `defaultAction` is `SCMP_ACT_KILL` and `syscalls` is empty or unset, the kernel will kill the container process on its first syscall."

But what you have provides that inference. LGTM

wking:
On Tue, Apr 25, 2017 at 02:47:24PM -0700, David Lyle wrote:
> "Note that depending on the value of `defaultAction` specified,
> syscalls should be non-empty.

I've shifted closer to this with bb4519c → 42984e8 while still trying
to avoid an underspecified should (even lowercase).  How does that
look?

dklyle:
Even better. Looks great.

mrunalp:
LGTM

tianon:
LGTM

hqhq:
LGTM

PR #769, opened 2017-04-12
config-linux: Require at least one entry in linux.seccomp.sycalls[].names

wking:
I expect the (undocumented) intention here is to iterate through `names` and call `seccomp_rule_add(3)` or similar for each name.  In that case, an empty `names` makes the whole syscall entry a no-op, and with this commit we can warn users who are validating such configs.

If, on the other hand, we were comfortable with no-op syscall entries, we'd want to make `names` OPTIONAL.

Warning folks who accidentally empty (or don't set) `names` seems more useful to me, and doesn't restrict the useful config space, so that's what I've gone with in this commit.

`minItems` is documented [here][1], and there is an example of its use [here][2]:

```
  "options": {
    "type": "array",
    "minItems": 1,
    "items": { "type": "string" },
    "uniqueItems": true
  },
```

[Part][3] of #762.

[1]: https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.11
[2]: http://json-schema.org/example2.html
[3]: https://github.com/opencontainers/runtime-spec/issues/762#issuecomment-292450654

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -615,6 +615,7 @@ The following parameters can be specified to setup seccomp:
     Each entry has the following structure:
 
     * **`names`** *(array of strings, REQUIRED)* - the names of the syscalls.
+        `names` MUST contain at least one entry.
 
     * **`action`** *(string, REQUIRED)* - the action for seccomp rules.
         A valid list of constants as of libseccomp v2.3.2 is shown below.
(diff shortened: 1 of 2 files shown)


wking:
This is a breaking change (adding a new MUST-level restriction), so it would be good to get it into 1.0.  Otherwise we'd have to delay this until 2.0.

mrunalp:
LGTM

tianon:
LGTM

vbatts:
LGTM

PR #770, opened 2017-04-13
config-linux.md: Increase the valid value of rootfsPropagation

q384566678:
According to the  [shared subtree](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt).

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -670,7 +670,7 @@ The following parameters can be specified to setup seccomp:
 ## <a name="configLinuxRootfsMountPropagation" />Rootfs Mount Propagation
 
 **`rootfsPropagation`** (string, OPTIONAL) sets the rootfs's mount propagation.
-Its value is either slave, private, or shared.
+Its value is either slave, private, shared or unbindable.
 The [Shared Subtrees][sharedsubtree] article in the kernel documentation has more information about mount propagation.
 
 ###### Example


wking:
On Wed, Apr 12, 2017 at 06:38:11PM -0700, Zhou Hao wrote:
> According to the change of [shared subtree](https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt).

Is that recent?  It looks like MS_UNBINDABLE is circa 2005 [1], and
part of v2.6.15-rc1.  But +1 on the spec change.

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9676f0c6389b62bd6b24d77d4b3abdbcfa32d0f2
     [PATCH] unbindable mounts, 2005-11-07

wking:
Looks like the JSON Schema could also use an enum.

q384566678:
@wking But why not use `unbindable`?

wking:
> But why not use `unbindable`?

`unbindable` is good with me.  I just though "change" was an odd choice of words for a feature from 2005 ;).  The enum entry should update the entry in `config-linux.json`.

mrunalp:
LGTM

q384566678:
@hqhq @vbatts PTAL

q384566678:
@crosbymichael PTAL

crosbymichael:
LGTM

PR #771, opened 2017-04-14
config: Document 'rbind' and 'bind' mount options extensions

wking:
My last attempt to get (recursive) bind mounts documented was in #530, which was rejected based on “[`MS_REC` makes no sense what-so-ever outside of a bind mount][1]”.  That [doesn't make sense to me][2], with `mount(8)` exposing `MS_REC | MS_SHARED` as `--make-rshared`, `MS_REC | MS_SLAVE` as `--make-rslave`, etc., but here's a reroll that uses `rbind` and `bind` instead, with a goal of specifying the (recursive) bind syntax so runtime-tools can rely on that instead of [blindly assuming runtimes support an unspecified bind syntax][3].

[1]: https://github.com/opencontainers/runtime-spec/pull/530#issuecomment-257028720
[2]: https://github.com/opencontainers/runtime-spec/pull/530#issuecomment-258664079
[3]: https://github.com/opencontainers/runtime-tools/blob/18a122b45a71765b09c6a451008a63687040b74a/generate/generate.go#L799-L817

### DIFF
--- a/config.md
+++ b/config.md
@@ -55,18 +55,73 @@ For Windows, see [mountvol][mountvol] and [SetVolumeMountPoint][set-volume-mount
 
 * **`destination`** (string, REQUIRED) Destination of mount point: path inside container.
   This value MUST be an absolute path.
+  * Linux: runtimes MUST pass this value to [`mount(2)`][mount.2] as the `target` argument.
   * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
   * Solaris: corresponds to "dir" of the fs resource in [zonecfg(1M)][zonecfg.1m].
-* **`type`** (string, OPTIONAL) The filesystem type of the filesystem to be mounted.
-  * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
-  * Windows: the type of file system on the volume, e.g. "ntfs".
-  * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+* **`type`** (string, OPTIONAL) The type of filesystem to mount.
+    If `type` is unset, the runtime MAY ask the kernel to guess the desired type.
+    Depending on the mount `options`, the `type` field MAY be ignored.
+    * Linux: when `type` is set, runtimes MUST pass the value to [`mount(2)`][mount.2] as the `filesystemtype` argument.
+        `type` is ignored when `options` contains `bind` or `rbind`; see the `MS_BIND` description in [mount(2)][mount.2].
+    * Windows: the type of file system on the volume, e.g. "ntfs".
+    * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`source`** (string, OPTIONAL) A device name, but can also be a directory name or a dummy.
-  * Windows: the volume name that is the target of the mount point, \\?\Volume\{GUID}\ (on Windows source is called target).
-  * Solaris: corresponds to "special" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+    * Linux: when `type` is set, runtimes MUST pass the value to [`mount(2)`][mount.2] as the `source` argument.
+        When `type` is not set, the value used for the `source` argument is unspecified.
+    * Windows: the volume name that is the target of the mount point, \\?\Volume\{GUID}\ (on Windows source is called target).
+    * Solaris: corresponds to "special" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`options`** (list of strings, OPTIONAL) Mount options of the filesystem to be used.
-  * Linux: supported options are listed in the [mount(8)][mount.8] man page. Note both [filesystem-independent][mount.8-filesystem-independent] and [filesystem-specific][mount.8-filesystem-specific] options are listed.
-  * Solaris: corresponds to "options" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+    * Linux: runtimes MUST call [`mount(2)`][mount.2] with an initially-zero `mountflags` argument altered by applying the following `options` in the listed order:
+
+        | Option          | Set bits                 | Clear bits                                                                            |
+        | --------------- | ------------------------ | ------------------------------------------------------------------------------------- |
+        | `acl`           | `MS_POSIXACL`            |                                                                                       |
+        | `async`         |                          | `MS_SYNCHRONOUS`                                                                      |
+        | `atime`         |                          | `MS_NOATIME`                                                                          |
+        | `bind`          | `MS_BIND`                |                                                                                       |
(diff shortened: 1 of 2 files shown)


mrunalp:
I think we should link to the kernel documents instead.

dqminh:
I dont think there's a kernel doc for this. Since we are converting the list of known strings values (i.e. rbind,noatime,slave etc to a list of mount flags), maybe having a table of required mappings to refer to here is useful.

For example taken from https://github.com/opencontainers/runc/blob/653207bc29a6d2d62b5d4f55b596467cb715a128/libcontainer/specconv/spec_linux.go#L630

flags | syscall
------- | ---------
bind |  syscall.MS_BIND
rbind | syscall.MS_BIND &#124; syscall.MS_REC

wking:
On Wed, Apr 26, 2017 at 08:28:36AM -0700, Daniel, Dao Quang Minh wrote:
> rbind | syscall.MS_BIND &#124; syscall.MS_REC

Now that we're becoming very clear about the mapping between ‘options’
and MS_* constants, do maintainers want to revisit a separate string
for MS_REC (like the ‘recursive’ I floated in #530)?  Or are we
forging ahead with r* options for common MS_REC & MS_* pairings?

We could just link to mount(2) or [1] to define the MS_* values
instead of including a table.  ‘rlimit’ would no longer be supported,
which has a backwards-compat cost, but we'd be able to get out of the
business of defining a wrapper around the kernel API (and based on
#780 that's a business we want to get out of).

[1]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/include/uapi/linux/fs.h?id=refs/tags/v4.10.12#n104

crosbymichael:
@wking NO

When I said rec makes no sense outside of binds, that was because you used binds in your examples and I try to simplify things for you because you take everything so literally.  

rec is never used by itself and the format and options are spec'd based on `fstab` style that are common between `/etc/fstab` and `mount(8)`.

I like @dqminh suggestion on the table if you feel like you must write more docs on this. Just follow the list that is already in the code and don't make up anything, document as is.

wking:
> I like @dqminh suggestion on the table…

That's pretty much what I have now, with the list [here][1].  It currently only covers `bind` and `rbind`.  Do I need to convert that list to a table?  Do I need to fill it out with [the other entries from runC][2]?

I expect we do need to fill it out somewhat, but don't think that has to happen in this PR.  This PR can establish the pattern (table or list, wording pattern for each entry, etc.) and then once we get something landed I'll file a followup adding the other entries that aren't covered by the mount(8) references.

[1]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R74
[2]: https://github.com/opencontainers/runc/blob/653207bc29a6d2d62b5d4f55b596467cb715a128/libcontainer/specconv/spec_linux.go#L630

crosbymichael:
I'm not a fan of the current PR contents and would rather not specify rbind and bind without the others.  They should be no different than any of the others.

wking:
> I'm not a fan of the current PR contents…

Is there anything other than which values are explicitly listed in the spec you would like to see changed?

> … and would rather not specify rbind and bind without the others. They should be no different than any of the others.

Of the [options listed in runC's `flags`][1], `bind` and `rbind` are the only ones not covered by [mount(8) filesystem-independent options ][2]'s (source [here][3]).  RunC differs from the mount page in defining [`defaults` as a no-op][4], while the [man page defines it as `rw` (clear `MS_RDONLY`), `suid` (clear `MS_NOSUID`), `dev` (clear `MS_NODEV`), `exec` (clear `MS_NOEXEC`), `auto` (clear `MS_NOAUTO`), `nouser` (clear `MS_USER`), and `async` (clear `MS_SYNCHRONOUS`)][5].  The man page also defines [`silent` and `loud`][6] (set and clear `MS_SILENT`), which are not supported in runC (at least as of 1.0.0-rc3).  The other filesystem-indepentent options from the man page look like they're backed by `data` and not `mountflags` in the `mount(2)` call.

So if this PR lands as it stands in bd28d9bd, I expect runC will need to fix `default` and add `silent` and `loud`.  If we want runC to be compliant as it stands, we'll have to drop the filesystem-independent `mount(8)` reference and inline a list with all of the `MS_*`-backed entries and then say something like “any unrecognized options are passed to `mount(2)` as `data`”.  Or is there a third approach I'm missing that still allows compliance testing?  And which approach sounds best to you?

[1]: https://github.com/opencontainers/runc/blob/v1.0.0-rc3/libcontainer/specconv/spec_linux.go#L626-L650
[2]: http://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT%20OPTIONS
[3]: https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/sys-utils/mount.8?h=v2.29.2#n835
[4]: https://github.com/opencontainers/runc/blob/v1.0.0-rc3/libcontainer/specconv/spec_linux.go#L629
[5]: https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/sys-utils/mount.8?h=v2.29.2#n935
[6]: https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/tree/sys-utils/mount.8?h=v2.29.2#n1068

wking:
Rerolled in bd28d9b→9073ddf to target `mount(2)` and replace the list with a table [as discussed in yesterday's meeting][1].  Since I was clarifying the `mountflags` and `data` values, I went ahead and clearly tied the other properties to `mount(2)` as well.  This should be a no-op as far as spec-consumers are concerned unless they were using the unspecified `bind` `type` without putting `(r)bind` in `options`.

There are still a number of differences vs. runC where runC diverges from `mount(8)`'s filesystem-independent options:

* The effect of `defaults`, [as described earlier][2].  We could follow runC and make `defaults` a no-op, but I don't see a point to that (who would put `defaults` in their `options` *knowing* it is a no-op?).
* Missing support: `silent`, `loud` ([as described earlier][2]), `(no)iversion`, `(no)lazytime`.  And the mount(8) filesystem-specific options which are represented by `MS_*` flags: `(no)acl`.  And the mount(8) double-dash options which are represented by `MS_*` flags: `move`.  We could follow runC and not define options for the missing `MS_*` settings, but that just breaks consumers who try to use them (since runtimes would have to pass `lazytime`, etc. through in `data` instead of setting `MS_LAZYTIME`).

I'm happy to submit patches to runC to correct those issues.  But if runC patches are not acceptable and maintainers won't accept this PR unless I strictly match the current runC implementation, please say so again and I'll spin them off into follow-up PRs.

More details in the commit message for those who are interested.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-05-09-20.07.log.html#l-24
[2]: https://github.com/opencontainers/runtime-spec/pull/771#issuecomment-297838316

wking:
I've filed opencontainers/runC#1460, opencontainers/runC#1461, opencontainers/runC#1462, and opencontainers/runC#1463 addressing the [missing options from mount(8)][1].

I have not filed a runC PR for `MS_MOVE` yet, because the current spec does not claim support for it.  We'll see how the missing-option PRs play out, and I may file a `MS_MOVE` PR if they go over well.

[1]: https://github.com/opencontainers/runtime-spec/pull/771#issuecomment-300559556

wking:
Can we re-open this now that opencontainers/runc#1460 has landed?

PR #774, opened 2017-04-18
schema: add `clean` to Makefile

q384566678:

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -16,3 +16,6 @@ validate: validate.go
 	go get -d ./...
 	go build ./validate.go
 
+clean:
+	rm -f validate
+


mrunalp:
LGTM

schema/Makefile
```diff
@@ -16,3 +16,6 @@ validate: validate.go
 	go get -d ./...
 	go build ./validate.go
 
+clean:
```
wking: This is also `.PHONY`.

q384566678: I don't think that is necessary. `make clean` can be executed correctly.

wking: > I don't think that is necessary. `make clean1 can be executed correctly.

So can `make validate`, with or without the `.PHONY` declaration.  `.PHONY` *keeps* the target working even if someone creates a `clean` file.  And while that is unlikely, I don't see why we shouldn't protect ourselves from it.

q384566678: If you do not add `.PHONY`, `make validate` can not be repeated, the following error occurs：
`make: 'validate' is up to date.`
so add the `.PHONY`.

wking: > If you do not add `.PHONY`, `make validate` can not be repeated…

Fair.  An example closer to your `clean` would be the [the `clean` entry in the main `Makefile`][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/Makefile#L97-L98

crosbymichael:
LGTM

PR #775, opened 2017-04-21
schema: Add enumeration to rootfsPropagation values

q384566678:

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -1,5 +1,14 @@
 {
     "definitions": {
+        "RootfsPropagation": {
+            "type": "string",
+            "enum": [
+                "private",
+                "shared",
+                "slave",
+                "unbindable"
+            ]
+        },
         "SeccompArch": {
             "type": "string",
             "enum": [
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -227,7 +227,7 @@
             },
             "rootfsPropagation": {
                 "id": "https://opencontainers.org/schema/bundle/linux/rootfsPropagation",
-                "type": "string"
+                "$ref": "defs-linux.json#/definitions/RootfsPropagation"
             },
             "seccomp": {
                 "id": "https://opencontainers.org/schema/bundle/linux/seccomp",


q384566678:
@wking I think it should be in the `config.go` file to increase the corresponding value, but I do not know how to name it, is it should be made such a style:
`const (`
`	ArchX86         Arch = "SCMP_ARCH_X86"`

wking:
On Thu, Apr 20, 2017 at 10:34:58PM -0700, Zhou Hao wrote:
> @wking I think it should be in the `config.go` file to increase the
> corresponding value, but I do not know how to name it, is it should
> be made such a style:
> `const (`
> `	ArchX86         Arch = "SCMP_ARCH_X86"`

I'm not sure what you mean.  I'd expect the current [1]:

  Its value is either slave, private, or shared.

should be adjusted to:

  Runtimes MUST support `slave`, `private`, [whatever other entries
  are in the enum].

[1]: https://github.com/opencontainers/runtime-spec/blob/cfc95a56767ecc7249bb43ac29bf8e9ee06ad415/config-linux.md#rootfs-mount-propagation

schema/defs-linux.json
```diff
@@ -1,5 +1,14 @@
 {
     "definitions": {
+        "RootfsPropagation": {
+            "type": "string",
+            "enum": [
+                "private",
+                "shared",
+                "slave",
+                "unbindable"
```
wking: You probably want to drop `unbindable` until #770 lands.

wking: #770 has landed, so there's no need to drop `unbindable` anymore.

tianon:
I'm +1 on this (as discussed on our maintainer call just now) -- for values like "CPU Quota", the value in our spec is the value as defined in the kernel, so it makes a lot of sense to punt on "valid values" to the kernel docs/implementation.

For values like `rootfsPropagation`, the value in our spec is a custom string which we've created for the purposes of this spec that maps to some other value (defined by a kernel enum), so in my mind we _must_ enumerate the expected values (and possibly even the kernel mappings unless they're already pretty obvious, which IMO they are in this case).

tianon:
LGTM

(in other words)

mrunalp:
LGTM

PR #777, opened 2017-04-21
Change Windows CPU Percent to Maximum

darrenstahlmsft:
In order to increase the granularity of CPU resource control, change the CPU Percent (0-100) resource setting to CPU Maximum (0-10000)

The HCS API accepts a ProcessorMaximum, which is a percentage multiplied by 100. This PR allows finer control over the CPU resources of Windows containers by extending the ProcessorMaximum into the OCI spec, rather than truncating to a 0-100 percent.

The discussion context is https://github.com/moby/moby/issues/32014 and https://github.com/moby/moby/pull/32608. The implementation of NanoCPUs in Moby should be able to scale to precision smaller than a single percent of total CPU resources.

/cc @jhowardmsft 

Signed-off-by: Darren Stahl <darst@microsoft.com>

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -40,7 +40,7 @@ The following parameters can be specified:
 
 * **`shares`** *(uint16, OPTIONAL)* - specifies the relative weight to other containers with CPU shares.
 
-* **`percent`** *(uint, OPTIONAL)* - specifies the percentage of available CPUs usable by the container.
+* **`maximum`** *(uint, OPTIONAL)* - specifies the portion of processor cycles that this container can use as a percentage times 100.
 
 #### Example
 
@@ -48,7 +48,7 @@ The following parameters can be specified:
     "windows": {
         "resources": {
             "cpu": {
-                "percent": 50
+                "maximum": 5000
             }
         }
     }
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -34,9 +34,9 @@
                                 "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/shares",
                                 "$ref": "defs-windows.json#/definitions/cpuShares"
                             },
-                            "percent": {
-                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/percent",
-                                "$ref": "defs.json#/definitions/percent"
+                            "maximum": {
+                                "id": "https://opencontainers.org/schema/bundle/windows/resources/cpu/maximum",
+                                "$ref": "defs.json#/definitions/uint16"
                             }
                         }
                     },
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -460,8 +460,8 @@ type WindowsCPUResources struct {
 	Count *uint64 `json:"count,omitempty"`
 	// CPU shares (relative weight to other containers with cpu shares). Range is from 1 to 10000.
 	Shares *uint16 `json:"shares,omitempty"`
-	// Percent of available CPUs usable by the container.
-	Percent *uint8 `json:"percent,omitempty"`
+	// Specifies the portion of processor cycles that this container can use as a percentage times 100.
+	Maximum *uint16 `json:"maximum,omitempty"`
 }
 
 // WindowsStorageResources contains storage resource management settings.


schema/defs.json
```diff
@@ -41,10 +41,10 @@
             "minimum": 0,
             "maximum": 18446744073709552000
         },
-        "percent": {
+        "maximum": {
             "type": "integer",
             "minimum": 0,
```
wking: You probably want to bump this from 0 to 1 to match [your Markdown docs][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/777/files#diff-36e882271af2f3ad47923159cb746598R43

darrenstahlmsft: Looking at this again. I think this is actually a generic percent definition, rather than specific to CPU Percent (Oops), so I should not touch it. Am I reading that right?

wking: > Looking at this again. I think this is actually a generic percent definition, rather than specific to CPU Percent…

Ah, right.  But it looks like the Windows field is the sole consumer.  I'd recommend dropping `percent` from `defs.json` and inlining the new limits in the `maximum` entry in `config-windows.json`.  But I'm not a maintainer, so you may want to wait for one of them to chime in.

darrenstahlmsft: I think I agree with that, and I've made the change locally, but I'll wait for more input before I push those changes.

wking: Another possibility, now that this shares the same 1–10k range as `cpuShares`, would be to rename the `cpuShares` entry in `defs-windows.json` to [`basisPoint`][1] or some such (`permyriad`?  `perTenThousand`?).  That's generic enough that it could live in `defs.json`

[1]: https://en.wikipedia.org/wiki/Basis_point

config-windows.md
```diff
@@ -40,15 +40,15 @@ The following parameters can be specified:
 
 * **`shares`** *(uint16, OPTIONAL)* - specifies the relative weight to other containers with CPU shares. The range is from 1 to 10000.
 
-* **`percent`** *(uint, OPTIONAL)* - specifies the percentage of available CPUs usable by the container.
+* **`maximum`** *(uint, OPTIONAL)* - specifies the portion of processor cycles that this container can use as a percentage times 100 . Range is from 1 to 10000.
```
dqminh: I think we want the change from percent to maximum, but we dont want to encode the limit into the specs. Also https://github.com/opencontainers/runtime-spec/pull/780

wking: > I think we want the change from percent to maximum, but we dont want to encode the limit into the specs.

The important exception to this, discussed in [today's meeting][1], is when the value is unique to this spec and there are no matching kernel docs.  I expect this property, and likely `shares` ([adjusted][2] in #780) both fall into that category, and need to retain their runtime-spec-defined limits.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-04-26-15.00.html
[2]: https://github.com/opencontainers/runtime-spec/pull/780/files#diff-36e882271af2f3ad47923159cb746598L41

darrenstahlmsft: FWIW, I agree that for the Windows shares and maximum values, the ranges should be left in the spec. While the values could be followed through to the implementation [in Microsoft/hcsshim](https://github.com/Microsoft/hcsshim/blob/master/interface.go), the ranges are not to change between kernel versions, and not easily mapped back to matching docs without looking at implementation.

jhowardmsft: (Not a maintainer) but I concur with @darrenstahlmsft - on Windows, the ranges probably should remain in the OCI spec.

darrenstahlmsft:
Should I remove the limits from this PR and add a new PR to add them back in? I think this change is important, and that the limits should remain in the spec for Windows, but if that's the only thing holding this PR then I can split them up to get this one merged.

crosbymichael:
@darrenstahlmsft would probably be better.

darrenstahlmsft:
Updated to remove limits. Once this is merged I'll open another PR to continue that discussion.

crosbymichael:
LGTM

dqminh:
LGTM

PR #778, opened 2017-04-24
schema/Makefile: Add .PHONY declarations for phony targets

wking:
The only non-phony target (where the target name matches the output file) is `validate`, but we need `.PHONY` there because the Go dependencies are not represented in the `Makefile`.  This commit adds the missing `.PHONY` declarations to the other targets, which truly are phony.

Spun off from [this discussion][1] and following the example set by [the `clean` entry in the main `Makefile`][2].

[1]: https://github.com/opencontainers/runtime-spec/pull/774#discussion_r112252004
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/Makefile#L97-L98

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -1,6 +1,8 @@
 
+.PHONY: default
 default: validate
 
+.PHONY: help
 help:
 	@echo "Usage: make [target]"
 	@echo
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM

hqhq:
LGTM

I would prefer them in one line though.

dqminh:
LGTM

> I would prefer them in one line though.

I have the exact opposite preference :stuck_out_tongue_closed_eyes: I prefer PHONY to be attached to their targets.

mrunalp:
LGTM

PR #779, opened 2017-04-25
schema: Add a range limit to Swappiness

q384566678:
https://github.com/opencontainers/runtime-spec/blame/master/specs-go/config.go#L296

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -198,7 +198,7 @@
                             },
                             "swappiness": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/swappiness",
-                                "$ref": "defs.json#/definitions/uint64"
+                                "$ref": "defs-linux.json#/definitions/memorySwappiness"
                             }
                         }
                     },
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -213,6 +213,11 @@
                 "allow"
             ]
         },
+        "memorySwappiness": {
+            "type": "integer",
+            "minimum": 0,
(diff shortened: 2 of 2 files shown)


wking:
On Mon, Apr 24, 2017 at 10:48:20PM -0700, Zhou Hao wrote:
> https://github.com/opencontainers/runtime-spec/blame/master/specs-go/config.go#L296

The Go comments are not normative [1], so you cannot use them as
grounds for validation.  We could bring that range limit into the
normative Markdown of config-linux.md, but the current maintainer
position seems to be against carrying kernel limits (e.g. see #755).
On the other hand, the spec isn't particularly consistent about
whether it carries kernel limits (e.g. it has them in [2,3,4,…]), so
maybe maintainers will want to carry this one.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L4
[2]: https://github.com/opencontainers/runtime-spec/blame/4754b55de94f6559f32589858d4027ee7bf61a5a/config-linux.md#L114
[3]: https://github.com/opencontainers/runtime-spec/blame/4754b55de94f6559f32589858d4027ee7bf61a5a/config-linux.md#L218
[4]: https://github.com/opencontainers/runtime-spec/blame/4754b55de94f6559f32589858d4027ee7bf61a5a/config-linux.md#L355

hqhq:
I don't think they should be in spec, I've opened https://github.com/opencontainers/runtime-spec/pull/780 to remove the remain ones.

mrunalp:
Closing since we are getting rid of tracking kernel limits.

PR #780, opened 2017-04-26
Remove range limit which depend on kernel

hqhq:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -352,14 +352,14 @@ For more information, see [the kernel cgroups documentation about blkio][cgroup-
 
 The following parameters can be specified to setup the controller:
 
-* **`blkioWeight`** *(uint16, OPTIONAL)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules. The range is from 10 to 1000.
+* **`blkioWeight`** *(uint16, OPTIONAL)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules.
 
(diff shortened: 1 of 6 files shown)


dqminh:
LGTM

mrunalp:
LGTM

config-linux.md
```diff
@@ -352,14 +352,14 @@ For more information, see [the kernel cgroups documentation about blkio][cgroup-
 
 The following parameters can be specified to setup the controller:
 
-* **`blkioWeight`** *(uint16, OPTIONAL)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules. The range is from 10 to 1000.
+* **`blkioWeight`** *(uint16, OPTIONAL)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules.
```
wking: Punting to the kernel is fine when it's clear which kernel parameter you're punting to.  In this case, the spec-reader has to figure out that `blkioWeight` is (MUST be?) implemented with [`blkio.weight` in the `blkio` controller][1], which has:

> Currently allowed range of weights is from 10 to 1000

The current docs [make it clear that `blockIO` represents the `blkio` controller][2], but mapping between the individual properties and the controller API is left up to the reader's imagination (with no RFC 2119 language for compliance testing, #746).  That leaves the semantics of these values pretty wiggly, and outside of validation and compliance testing.

Also the kernel's “Currently allowed” wording is troublesome.  If we punt to the kernel for semantics, it's not clear to me how we handle evolving kernel semantics unless we pin the property to a particular kernel version or trust the kernel to maintain backwards compatibility for as long as this version of the spec is useful.

[1]: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config-linux.md#L350

PR #783, opened 2017-04-27
Remove range limit

q384566678:
Fixes #780 

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -293,7 +293,7 @@ type LinuxMemory struct {
 	Kernel *uint64 `json:"kernel,omitempty"`
 	// Kernel memory limit for tcp (in bytes)
 	KernelTCP *uint64 `json:"kernelTCP,omitempty"`
-	// How aggressive the kernel will swap memory pages. Range from 0 to 100.
+	// How aggressive the kernel will swap memory pages.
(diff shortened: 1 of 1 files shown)


specs-go/config.go
```diff
@@ -458,7 +458,7 @@ type WindowsMemoryResources struct {
 type WindowsCPUResources struct {
 	// Number of CPUs available to the container.
 	Count *uint64 `json:"count,omitempty"`
-	// CPU shares (relative weight to other containers with cpu shares). Range is from 1 to 10000.
+	// CPU shares (relative weight to other containers with cpu shares).
```
wking: You may want to wait on this until #777 lands, because the Windows folks are sounding like they'd like to return the explicit ranges to their settings ([here][1] and [here][2]).  On the other hand, there's no need to list the range in the Go comments even if the Markdown spec gives a valid range.  If we want to remove range comments from the Go types regardless of whether the spec gives a valid range, we can go ahead with this removal without waiting on #777.

[1]: https://github.com/opencontainers/runtime-spec/pull/777#discussion_r113536160
[2]: https://github.com/opencontainers/runtime-spec/pull/777#discussion_r113543747

hqhq:
LGTM

crosbymichael:
LGTM

PR #784, opened 2017-04-27
schema: bug fixes

q384566678:
Signed-off-by: zhouhao <zhouhao@cn.fujitsu.com>

PR #785, opened 2017-04-27
schema: Add 'test' target to the Makefile

wking:
And fill in some known-good and known-bad examples.  We can make this as detailed as we want, but this commit just adds enough to know that:

* The full-file spec examples are valid.
* The JSON Schema can distinguish valid examples from invalid JSON.

This will help catch JSON Schema typos like those being addressed by #784 (and until that lands the Travis test for this PR will fail, which is what it's designed check).


wking:
Rebased onto master and fixed expected-failure validation with e688390 → 0b7efd2.  And now that #784 has landed, the tests pass.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #787, opened 2017-05-01
config-linux: RFC 2119 wording for intelRdt

wking:
So we can compliance-test runtimes for these settings.

Also remove the tutorial, since the kernel docs should provide sufficient documentation on that front.  The kernel can be patched if they do not, and we do not include tutorials for other config-linux settings in this spec.

See also opencontainers/runC#433 and opencontainers/runC#1279.

CC @xiaochenshen.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -478,86 +478,29 @@ The following parameters can be specified to setup the controller:
 
 ## <a name="configLinuxIntelRdt" />IntelRdt
 
-Intel platforms with new Xeon CPU support Intel Resource Director Technology
-(RDT). Cache Allocation Technology (CAT) is a sub-feature of RDT, which
-currently supports L3 cache resource allocation.
+**`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
+    If `intelRdt` is set, the runtime MUST write the container process ID to the `<container-id>/tasks` file in a mounted `resctrl` pseudo-filesystem, using the container ID from [`start`](runtime.md#start) and creating the `<container-id>` directory if necessary.
+    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
 
-This feature provides a way for the software to restrict cache allocation to a
-defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
-The different subsets are identified by class of service (CLOS) and each CLOS
-has a capacity bitmask (CBM).
-
-In Linux kernel, it is exposed via "resource control" filesystem, which is a
-"cgroup-like" interface.
-
-Comparing with cgroups, it has similar process management lifecycle and
-interfaces in a container. But unlike cgroups' hierarchy, it has single level
-filesystem layout.
-
-Intel RDT "resource control" filesystem hierarchy:
-```
-mount -t resctrl resctrl /sys/fs/resctrl
-tree /sys/fs/resctrl
-/sys/fs/resctrl/
-|-- info
-|   |-- L3
-|       |-- cbm_mask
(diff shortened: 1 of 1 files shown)


config-linux.md
```diff
@@ -478,82 +478,25 @@ The following parameters can be specified to setup the controller:
...
-currently supports L3 cache resource allocation.
+**`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
+    If `intelRdt` is set, the runtime MUST write the container process ID to the `<container-id>/tasks` file in a mounted `resctrl` pseudo-filesystem, using the container ID from [`start`](runtime.md#start) and creating the `<container-id>` directory if necessary.
+    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
 
-This feature provides a way for the software to restrict cache allocation to a
-defined 'subset' of L3 cache which may be overlapping with other 'subsets'.
-The different subsets are identified by class of service (CLOS) and each CLOS
-has a capacity bitmask (CBM).
+    If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` psuedo-filesystems.
```
xiaochenshen: > ```+    If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` psuedo-filesystems.```

How about using these words? There should not be more than one `resctrl` filesystems.
```+    If `intelRdt` is not set, the runtime MUST NOT manipulate the `resctrl` psuedo-filesystem.```

File: config-linux.md [hunk 2]
```diff
@@ -478,82 +478,25 @@ The following parameters can be specified to setup the controller:
...
 
-**`intelRdt`** (object, OPTIONAL) represents the L3 cache resource constraints in Intel Xeon platforms.
+    If `l3CacheSchema` is not set, runtimes MUST not write to `schemata` files in any `resctrl` psuedo-filesystems.
 
-For more information, see [Intel RDT/CAT kernel interface][intel-rdt-cat-kernel-interface].
+### Example
 
-The following parameters can be specified for the container:
+Consider a two-socket machine with two L3 caches where the default CBM is 0xfffff and the max CBM length is 20 bits.
+This configuration assigns 4/5 of L3 cache id 0 and the whole L3 cache id 1 for the container:
```
xiaochenshen: Since the inline tutorial will be removed, I would like to improve the original wording to prevent confusion to new users of Intel RDT, and provide a more sophisticated example.
```
+Tasks inside the container only have access to the "upper" 80% of L3 cache id 0 and the "lower" 50% L3 cache id 1:

"linux": {
	"intelRdt": {
		"l3CacheSchema": "L3:0=ffff0;1=3ff"
	}
}
```

wking: > … provide a more sophisticated example…

I've pulled in your suggested example with edef71f → cbd2c02.  How does that look?

config-linux.md
wking: > There should not be more than one resctrl filesystems.

But there might be more than one (just like proc).  I expect you'd only actually *have* multiple `resctrl` filesystems mounted when you were setting up nested containers (e.g. docker-in-docker), but the “any” wording isn't much more complicated and avoids confusion in the multiple-mount corner case.

config-linux.md
xiaochenshen: @wking Thanks. It looks good to me.

config-linux.md
xiaochenshen: @wking I got it. It makes sense in nested containers case.

hqhq:
LGTM

wking:
I fixed a “MUST not” → “MUST NOT” and rebased onto master with cbd2c02 → b11ade4.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #790, opened 2017-05-09
config: strengthen punt to kernel for valid "capabilities" strings

tianon:
Opening this as we discussed on our review call today.  The intent here is to _not_ have an explicit list of "string to kernel constant" mappings in the specification, and to instead expect implementations to make a reasonable effort to have up-to-date mappings (especially for kernel features that might be newer than the particular specification release), and to error out for anything that they can't map.

As I mentioned on the call, I do not feel that this is a "1.0 blocker" -- the existing language is sufficient in my eyes to get the point across to implementors what they're expected to do.

### DIFF
--- a/config.md
+++ b/config.md
@@ -132,7 +132,7 @@ For Windows, see [mountvol][mountvol] and [SetVolumeMountPoint][set-volume-mount
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-* **`capabilities`** (object, OPTIONAL) is an object containing arrays that specifies the sets of capabilities for the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [capabilities(7)][capabilities.7] man page.
+* **`capabilities`** (object, OPTIONAL) is an object containing arrays that specifies the sets of capabilities for the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [capabilities(7)][capabilities.7] man page, such as `CAP_CHOWN`. Any value which cannot be mapped to a relevant kernel interface MUST cause an error.
   capabilities contains the following properties:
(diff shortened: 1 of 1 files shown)


tianon:
This is related to https://github.com/opencontainers/runtime-spec/issues/755 and https://github.com/opencontainers/runtime-spec/pull/766 (which should probably be re-opened, since `CAP_` isn't actually prescribed by the specification -- we're punting to the kernel and the runtime implementation to do best-effort and the Linux kernel just happens to include `CAP_` for all current valid values).

config.md
```diff
@@ -132,7 +132,7 @@ For Windows, see [mountvol][mountvol] and [SetVolumeMountPoint][set-volume-mount
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
   This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-* **`capabilities`** (object, OPTIONAL) is an object containing arrays that specifies the sets of capabilities for the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [capabilities(7)][capabilities.7] man page.
+* **`capabilities`** (object, OPTIONAL) is an object containing arrays that specifies the sets of capabilities for the process(es) inside the container. Valid values are platform-specific. For example, valid values for Linux are defined in the [capabilities(7)][capabilities.7] man page, such as `CAP_CHOWN`. Any value which cannot be mapped to a relevant kernel interface MUST cause an error.
```
wking: If we're looking for a clear way to specify the “relevant kernel interface”, [`cap_from_text(3)` is part of the withdrawn POSIX.1e draft specification][1] (and therefore has some chance of being cross-platform) and is supported by [libcap][2].  Without clarification, I'd expect the kernel intefaces to be [`capset(2)`][3] and/or [`prctl(2)`][4], but both of those happen after the capabilities have been converted to integers.

[1]: http://www.man7.org/linux/man-pages/man3/cap_from_text.3.html#CONFORMING_TO
[2]: https://git.kernel.org/pub/scm/linux/kernel/git/morgan/libcap.git/
[3]: http://www.man7.org/linux/man-pages/man2/capset.2.html
[4]: http://www.man7.org/linux/man-pages/man2/prctl.2.html

crosbymichael:
LGTM

mrunalp:
LGTM

PR #791, opened 2017-05-09
Makefile: Add .install.* to .PHONY

wking:
The `$(var:pattern=replacement)` syntax is documented [here][1].

Spun off from #759.

[1]: https://www.gnu.org/software/make/manual/html_node/Text-Functions.html#index-patsubst-1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,9 +80,10 @@ else
 	git-validation -v -run DCO,short-subject,dangling-whitespace -range $(EPOCH_TEST_COMMIT)..HEAD
 endif
 
+TOOLS := golint gitvalidation
 
-.PHONY: install.tools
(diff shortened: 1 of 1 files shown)


dqminh:
why do we need this ?

wking:
On Tue, May 09, 2017 at 09:23:14PM -0700, Daniel, Dao Quang Minh wrote:
> why do we need this ?

We don't *need* it, but putting targets in .PHONY when they don't
correspond to on-disk files is good practice.  See previous discussion
around [1], and a similar addition in #778.

Assuming that we agree that the .install.* targets should be .PHONY,
the TOOLS variable I'm adding just gives us a DRY way to accomplish
that.

crosbymichael:
Closing as we are going to remove PHONY

PR #792, opened 2017-05-09
specs-go/config: Use a pointer for Process.ConsoleSize

wking:
Avoid injecting `"consoleSize":{"height":0,"width":0}` when serializing with Go's stock JSON serializer.  Using a pointer for this optional struct property works around golang/go#11939.

Spun off from #759.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #786, opened 2017-04-28
config-solaris.md: fix info

q384566678:
Signed-off-by: zhouhao <zhouhao@cn.fujitsu.com>

PR #794, opened 2017-05-09
config: Update zonecfg(1M) link to match config-solaris.md

wking:
With the same link update that config-solaris.md got in #786, since the old link 404s.


crosbymichael:
LGTM

hqhq:
LGTM

Issue #731, opened 2017-03-16
runtime: Ambiguous namespace for state `pid` and `bundle`


wking:
The spec [has][1]:

* **`pid`** (int, REQUIRED when `status` is `created` or `running`) is the ID of the container process, as seen by the host.
* **`bundle`** (string, REQUIRED) is the absolute path to the container's bundle directory.

I'm assuming those are both “in the runtime namespace”.  However the “[runtime namespace][2]” wording needs to be tightened up in this case, because `create` and `state` may have been called from different namespaces.  Is the `pid` value always in the `create` runtime namespace?  Or when `state` is called from another PID namespace, is the container process ID translated into that PID namespace (if it is even visible)?  The same ambiguity applies to `bundle` and mount namespaces.

I'd raised this issue [on dev@ a while back][3], but with 1.0 approaching (#726), and dev@-based discussions being pretty quiet, I thought I'd cross-post here in case that helps with triage/prioritization.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/runtime.md#state
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/glossary.md#runtime-namespace
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/ujtABQoCmgk

cyphar:
I had some discussion about possibilities of how a runtime could implement namespace-agnostic ways. opencontainers/runc#1224. I'm not sure if it makes sense to define this in the spec, but it's an interesting idea IMO.

Maybe if we make `state` return a path to a `procfs` directory which users can then read `stat` and other pseudo-files to parse. Of course, this will massively break VM-based runtimes so it's probably not the best idea in the world.

wking:
On Thu, Mar 16, 2017 at 07:00:45AM -0700, Aleksa Sarai wrote:
> Maybe if we make `state` return a path to a `procfs` directory…

That assumes that when the ‘state’ call succeeds, the ‘state’ caller
will share a mount namespace (or enough of the mount contents) to be
able to get to the right proc by following that path.  This is
basically the same issue we currently have with the ‘bundle’ value.
For runC (which uses the filesystem to share state), this condition is
likely satisfied.  But other runtimes may have different approaches to
sharing state (maybe they communicate with a state-registry daemon
over TCP), so I don't think we can rely on this condition being
satisfied in general.

> Of course, this will massively break VM-based runtimes so it's
[...]

PR #795, opened 2017-05-09
config-linux: Specify relationships for new namespaces

wking:
Spun off from #767, since these [are][1] [contentious][2].  I still think we want to say something about these relationships.

We already have some of “runtime namespace” conditions (e.g. when a type is not listed in `linux.namespaces[]`), so runtimes should already have implementation-specific wording around what the runtime namespaces are (we don't explicitly make them implementation-defined, although we probably should).  Anyhow, that's not a new concept added by this commit.

### Seeded namespaces

For example, if you ask for a new `uts` namespace but do not set the optional [`hostname`][3], having the seed defined means that the hostname in the container UTS namespace is well-defined (it will be whatever the hostname was in the runtime UTS namespace).

This is less of an issue for the mount namespace, because with [`root.path` REQUIRED][4], there's no way to avoid clobbering whatever mounts you got from your seed (which makes not asking for a new mount namespace exciting ;).

### Hierarchical namespaces

I think “I want this container to run in a new user/pid namespace that is a child of the runtime user/pid namespace” should be something that has a portable config expression.  Otherwise it becomes very unclear what to put in the `hostID` field for `(u|g)idMappings`, because you don't know what namespace will be used to interpret the hostIDs.

### Namespace ownership

This is another case where I think specified clarity is essential.  A new network namespace will not be very useful if you don't know who owns it.

[1]: https://github.com/opencontainers/runtime-spec/pull/767#discussion_r115591844
[2]: https://github.com/opencontainers/runtime-spec/pull/767#discussion_r115592437
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#hostname
[4]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#root

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -40,6 +40,9 @@ The following parameters can be specified to setup namespaces:
     The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
 
     If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
+    For hierarchical namespaces (e.g. `pid`, `user`), the new container namespace MUST be a child of the [runtime namespace](glossary.md#runtime-namespace) of that type.
+    For seeded namespaces (e.g. `mount`, `uts`), the new container namespace MUST be seeded by the runtime namespace of that type.
+    When `type` is not `user`, new namespaces MUST be owned by the container `user` namespace.
 
 If a namespace type is not specified in the `namespaces` array, the container MUST inherit the [runtime namespace](glossary.md#runtime-namespace) of that type.
 If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST [generate an error](runtime.md#errors).
--- a/glossary.md
+++ b/glossary.md
@@ -15,7 +15,7 @@ For example, namespaces, resource limits, and mounts are all part of the contain
 
 ## <a name="glossaryContainerNamespace" />Container namespace
 
-On Linux, a leaf in the [namespace][namespaces.7] hierarchy in which the [configured process](config.md#process) executes.
+On Linux, the [namespaces][namespaces.7] in which the [configured process](config.md#process) executes.
 
 ## <a name="glossaryJson" />JSON
 
@@ -30,9 +30,7 @@ It reads the [configuration files](#configuration) from a [bundle](#bundle), use
 
 ## <a name="glossaryRuntimeNamespace" />Runtime namespace
 
-On Linux, a leaf in the [namespace][namespaces.7] hierarchy from which the [runtime](#runtime) process is executed.
-New container namespaces will be created as children of the runtime namespaces.
-
+On Linux, the namespaces from which new [container namespaces](config-linux.md#namespaces) are created.
 
 [JSON]: https://tools.ietf.org/html/rfc7159
 [UTF-8]: http://www.unicode.org/versions/Unicode8.0.0/ch03.pdf


wking:
This [got kicked around some in today's meeting][1] (although my notes are thin).  My impression was that there is concern that we are documenting the kernel (which I'm trying to avoid) and that these requirements are somewhat tautological with the source of the `unshare(2)` or `clone(2)` being the [“runtime namespace” by definition][2] (which I agree with).  It's been a while since I looked at the glossary entry, and now that I do I think:

* The current glossary entry makes all namespaces sound hierarchical, when they aren't.
* The current glossary entry has no RFC 2119 language for compliance testing.

I think we want to land something like the wording in this PR (as of c9b61187) with RFC 2119 teeth, and then update the glossary entry to be something like:

> On Linux, a the namespace from which the container namespaces [link to `config-linux.md#namespaces`] are created.

Which embraces the overlap by trimming the incorrect glossary entry and referencing the RFC 2119 teeth here.

I don't think we should leave these relationships alone with an argument like “namespaces can be confusing, so let's not say obvious things like this”.  If namespaces are confusing, I think that's a good reason to say obvious things, because “obvious” is in the eye of the beholder.  And it's especially good if we can phrase them specifically enough for compliance testing, since that removes the eye-of-the-beholder ambiguity.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-05-10-21.03.log.html#l-115
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/glossary.md#runtime-namespace

tianon:
I feel like this is already implied (given that it's how the kernel itself works), and that there isn't a lot of value added by explicitly mentioning it or attempting to validate runtime behavior around it.

I'd vote to either close or defer discussion to post-1.0.

wking:
On Wed, May 10, 2017 at 09:39:10PM -0700, Tianon Gravi wrote:
> I feel like this is already implied (given that it's how the kernel
> itself works)…

That's how the kernel works if you're calling unshare(2) or clone(2)
from the runtime namespaces to create new container namespaces.  But
there's nothing in the spec that says you have to do it that way.  You
could have a namespace-generating process that creates new namespaces
and pipes file descriptors for them into the container process on a
Unix socket which the container process setns(2)s.  That's how the
kernel works too, but I doubt it's what most callers will expect.

> … and that there isn't a lot of value added by explicitly mentioning
> it or attempting to validate runtime behavior around it.

If my feeling is that this is a useful addition and yours is “it
doesn't add much”, I'd recommend erring on the side of adding three
lines to config-linux.md and trimming a line from the glossary (I've
included my proposal from [1] in a reroll from c9b6118 → cf8306e).
I'm also happy to PR runtime-tools tests if that helps.  If there is
anything else I can do to reduce the maintenance cost of this change
(which I expect to be small), please let me know.

[1]: https://github.com/opencontainers/runtime-spec/pull/795#issuecomment-300636762

glossary.md
```diff
@@ -30,9 +30,7 @@ It reads the [configuration files](#configuration) from a [bundle](#bundle), use
 
 ## <a name="glossaryRuntimeNamespace" />Runtime namespace
 
-On Linux, a leaf in the [namespace][namespaces.7] hierarchy from which the [runtime](#runtime) process is executed.
-New container namespaces will be created as children of the runtime namespaces.
-
+On Linux, the namespaces from which new [container namespaces](config-linux.md#namespaces) are created.
```
hqhq: Why not link to the glossary container namespace above? Looks more suitable to me.

hqhq:
> I feel like this is already implied (given that it's how the kernel itself works), and that there isn't a lot of value added by explicitly mentioning it or attempting to validate runtime behavior around it.
I'd vote to either close or defer discussion to post-1.0.

Agreed.

> I'd recommend erring on the side of adding three
lines to config-linux.md and trimming a line from the glossary

I'm +1 on this.

config-linux.md
```diff
@@ -40,6 +40,9 @@ The following parameters can be specified to setup namespaces:
     The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
 
     If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
+    For hierarchical namespaces (e.g. `pid`, `user`), the new container namespace MUST be a child of the [runtime namespace](glossary.md#runtime-namespace) of that type.
+    For seeded namespaces (e.g. `mount`, `uts`), the new container namespace MUST be seeded by the runtime namespace of that type.
```
vbatts: NACK on this language.

vbatts:
A couple of the clarifying sentences are okay, but the seeding and hierarchical namespace language is too prescriptive and not okay to add. closing. add prs for the simpler sentences if still desired

wking:
On Wed, May 24, 2017 at 08:55:44AM -0700, v1.0.0.batts wrote:
> A couple of the clarifying sentences are okay, but the seeding and
> hierarchical namespace language is too prescriptive and not okay to
> add. closing.

Without these statements, a config including [1] could lead to a new
container PID namespace that is *not* a child of the runtime
namespace.  With a PID namespace hierarchy like:

  A
  |-- B
  `-- C

where B is the runtime PID namespace, the container PID namespace D
could be created in any of the following positions:

  A
  |-- B
  |-- C
  `-- D'

  A
  |-- B
  |   `-- D''
  `-- C

  A
  |-- B
  `-- C
      `-- D'''

  A
  |-- B
  `-- C
      `-- E
          `-- D'''

  etc.

But I'd like to restrict runtimes to only D''.  Do you feel that is
too perscriptive?  Do you want to allow runtimes to use the D', D''',
D''', etc., positions?  If so, can you explain why?  If not, why allow
them?

> add prs for the simpler sentences if still desired

I've spun the glossary changes out into #852 and #853.

cyphar:
> Do you want to allow runtimes to use the D', D''',
D''', etc., positions?  If so, can you explain why?  If not, why allow
them?

If the runtime can implement all of the relevant lifecycle stuff with those semantics, they can go for it. I don't understand why it is a necessity that `ps aux` in the caller's namespace must container the container PIDs. That's not required anywhere and would actually break any daemon-based container runtimes (at least it would make their jobs harder for no obviously beneficial reason).

Not to mention there might be some very odd usecases where you might actually want the functionality of joining namespaces you couldn't have created.

Without having something as awful as `cgroupsPath` for namespaces, mandating any rules on this that aren't extensible on this topic will break implementations and make people very unhappy. Please don't make a PR for `nsPath`, it will be a waste of time to review (such a path doesn't have any kernel parallel and not every hierarchy is a VFS).

wking:
On Wed, May 24, 2017 at 11:02:24AM -0700, Aleksa Sarai wrote:
> If the runtime can implement all of the relevant lifecycle stuff
> with those semantics, they can go for it.

Lifecycle semantics are not the problem, especially if we address
#731.  If we flip this over to user namespaces (also hierarchical),
one problem is “what do I put in hostID?” [1].  If the goal is for the
new container namespace to map B's UID 1000 (which happens to be to
UID 2000 in A and unmapped in C) to D's PID 0, do I put 1000 in
linux.uidMappings.hostID?  Do I put 2000 in there?  The correct value
depends on whether the runtime will create D', D'', D''', etc.

> I don't understand why it is a necessity that `ps aux` in the
> caller's namespace must container the container PIDs. That's not
> required anywhere and would actually break any daemon-based
> container runtimes (at least it would make their jobs harder for no
> obviously beneficial reason).

Perhaps you want the container killed when PID namespace B's init
exits.  The way to support daemon-based container runtimes is to have
those runtimes clearly document their runtime namespace.  Calling a
‘create’ action from a shell in namespace B doesn't mean that B is the
runtime namespace.  The runtime namespace doesn't even have to be on
the same host.  But whatever the runtime namespace is (which should be
implementation-defined), you need the container namespaces to be
children/seeded from the runtime namespaces in order to know what to
put in linux.uidMappings.hostID and similar or to how to have the
runtime create nested PID namespaces.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#user-namespace-mappings

PR #797, opened 2017-05-10
spec.md: add MUST NOT and SHALL NOT for judging compliance

Mashimiao:
MUST NOT and SHALL NOT mean an absolute prohibition of the spec.
They should be one of the requirements to judge compliance.


tianon:
Is it necessary to spell this out so explicitly?  `MUST` and `SHALL` already have special meaning, so `MUST xxx` would imply that a runtime has to `xxx` to be compliant, which mentally seems fine to me for values of `xxx` starting with `NOT`, ala runtime has to `NOT x, y, or z`

tianon:
We discussed this in a review session, and consensus is that this is already implied by the language we have. :heart:

PR #799, opened 2017-05-10
*: Use inline links for remaining internal references

wking:
Since #687, the official style is to [only use reference-style links for external links][1].  I expect the remaining three entries just slipped through.  This commit adjusts everything found with:

    $ git grep ']: [a-z]' | grep -v http

It also fixes the underscore → hyphen in the `glossary.md#container-namespace` target and updates the `capabilities` location to catch up with #329.

[1]: https://github.com/opencontainers/runtime-spec/blob/f56fe8ba78f9d70c63fbd301cf1cd680218742fa/style.md#links

### DIFF
--- a/style.md
+++ b/style.md
@@ -8,12 +8,12 @@ For example, this paragraph will span three lines in the Markdown source.
 
 ## <a name="styleHex" />Traditionally hex settings should use JSON integers, not JSON strings
 
-For example, [`"classID": 1048577`][class-id] instead of `"classID": "0x100001"`.
+For example, [`"classID": 1048577`](config-linux.md#network) instead of `"classID": "0x100001"`.
 The config JSON isn't enough of a UI to be worth jumping through string <-> integer hoops to support an 0x… form ([source][integer-over-hex]).
 
 ## <a name="styleConstantNames" />Constant names should keep redundant prefixes
 
-For example, `CAP_KILL` instead of `KILL` in [**`linux.capabilities`**][capabilities].
+For example, `CAP_KILL` instead of `KILL` in [**`process.capabilities`**](config.md#process).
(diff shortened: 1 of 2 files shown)


style.md
```diff
@@ -8,12 +8,12 @@ For example, this paragraph will span three lines in the Markdown source.
...
 ## <a name="styleHex" />Traditionally hex settings should use JSON integers, not JSON strings
 
-For example, [`"classID": 1048577`][class-id] instead of `"classID": "0x100001"`.
+For example, [`"classID": 1048577`](config-linux.md#network) instead of `"classID": "0x100001"`.
 The config JSON isn't enough of a UI to be worth jumping through string <-> integer hoops to support an 0x… form ([source][integer-over-hex]).
 
 ## <a name="styleConstantNames" />Constant names should keep redundant prefixes
 
-For example, `CAP_KILL` instead of `KILL` in [**`linux.capabilities`**][capabilities].
+For example, `CAP_KILL` instead of `KILL` in [**`process.capabilities`**](config.md#process).
```
tianon: Shouldn't this be `config.md#configProcess`?

(`## <a name="configProcess" />Process` is the section header of `config.md` which mentions `capabilities`)

wking: > Shouldn't this be `config.md#configProcess`?

That would work too.  I'm currently using the auto-generated fragments to match @mrunalp's [preference][1].  I don't care either way, although I'd prefer the spec consistently use whichever way maintainers want.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-03-03.log.html#t2017-03-03T18:02:12

tianon:
LGTM

Thanks for clarifying @wking. :+1:

mrunalp:
LGTM

PR #800, opened 2017-05-10
config-linux: Remove redundant MUST for minimum cgroup controllers

wking:
This is my suggested alternative to #729, which is currently removing the lowercase “required” but keeping the redundant MUST.  Any runtime which violated the MUST would necessarily violate some more specific constraint on a `resources` setting, so this PR removes the MUST entirely.


tianon:
LGTM

This reads more clearly to me. :+1:

mrunalp:
LGTM

PR #802, opened 2017-05-11
Makefile: Remove unnecessary .PHONY entries

wking:
The only `.PHONY` entry we *need* is for `schema/validate`, since that's a real file but we haven't told Make about its real dependencies (which involve complicated Go lookups).  I'm personally in favor of using `.PHONY` for all targets that aren't on-disk files, because it hints to readers that the rule is not generating a file at the target.  But there has been resistance to adding .PHONY entries to all such cases (e.g. #791), so this commit brings us around to a internally-consistent “only use `.PHONY` when you always need it” position.

That means that, for example, users who create files named `clean` will turn `clean` the target into a no-op, but runtime-spec maintainers are ok with that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,6 @@ DOC_FILES := \
 
 default: docs
 
-.PHONY: docs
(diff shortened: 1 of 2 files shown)


tianon:
LGTM

crosbymichael:
LGTM

PR #803, opened 2017-05-11
config.md: specify config usage

Mashimiao:
config.json is used for creation of container not for
operations of container, we should specify this.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,6 +1,6 @@
 # <a name="containerConfigurationFile" />Container Configuration file
 
-This configuration file contains metadata necessary to implement standard operations against the container.
+This configuration file contains metadata necessary to implement [standard operations](runtime.md#operations) against the container.
 This includes the process to run, environment variables to inject, sandboxing features to use, etc.
 
 The canonical schema is defined in this document, but there is a JSON Schema in [`schema/config-schema.json`](schema/config-schema.json) and Go bindings in [`specs-go/config.go`](specs-go/config.go).


config.md
```diff
@@ -1,6 +1,6 @@
 # <a name="containerConfigurationFile" />Container Configuration file
 
-This configuration file contains metadata necessary to implement standard operations against the container.
+This configuration file contains metadata necessary to implement standard operations against the creation of a container.
```
wking: `process.args` is used by `start` and some hooks are used by `delete`.  But `kill` doesn't need anything from then config.  So we can draw a distinction here, but your current wording isn't quite it.

I'd rather link "standard operations" to the `runtime.md` section and maybe weaken it to "some standard operations".  But I'd avoid calling out creation without also mentioning the other impacted operations.

tianon:
I think I tend to agree with @wking here -- in my mind, the `config.json` (and I suppose the rest of the bundle) is the specification of the desired state of the container, and set of information for how to find/reference/interact with it.

Regardless, I think this is something we could discuss further clarifying post-1.0?

Mashimiao:
@tianon  @wking  updated as @wking  suggested

tianon:
LGTM

mrunalp:
Without the config.json there is no container and no operations to run against a container so I am in favor of closing this PR.

config.md
```diff
@@ -1,6 +1,6 @@
 # <a name="containerConfigurationFile" />Container Configuration file
 
-This configuration file contains metadata necessary to implement standard operations against the container.
+This configuration file contains metadata necessary to implement some [standard operations](runtime.md#operations) against the container.
```
crosbymichael: Lets remove the "some" part but keep the link

Mashimiao:
OK, updated.

mrunalp:
LGTM

tianon:
LGTM

PR #804, opened 2017-05-11
config-linux: Remove explicit 'null' from device cgroup values

wking:
Catch the Markdown spec up with the JSON Schema change in #662.  The Markdown is canonical, so we could restore the explicit-null handling to the JSON Schema instead, but the [maintainers feel][1] (and I agree) that there's no point in explicitly allowing a null value when callers can simply leave the property unset.


mrunalp:
LGTM

dqminh:
LGTM

PR #805, opened 2017-05-11
schema/test/config/good/minimal: Drop 'process'

wking:
It's optional since #701 which landed yesterday.

### DIFF
--- a/schema/test/config/good/minimal-for-start.json
+++ b/schema/test/config/good/minimal-for-start.json
@@ -1,0 +1,20 @@
+{
+    "ociVersion": "1.0.0",
+    "platform": {
+        "os": "linux",
+        "arch": "amd64"
+    },
+    "root": {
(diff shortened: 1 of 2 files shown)


mrunalp:
I think minimal should reflect a normal full flow with start and should have a process section.

wking:
> I think minimal should reflect a normal full flow with start and should have a process section.

Can I just add another for that?  `minimal-for-start.json`?

mrunalp:
Yeah, that works for me.

wking:
On Thu, May 11, 2017 at 10:00:31AM -0700, Mrunal Patel wrote:
> Yeah, that works for me. 

Done with 9b2daa5 → 90423a0.

mrunalp:
LGTM

dqminh:
LGTM

PR #807, opened 2017-05-11
config.md: fix typo of context

Mashimiao:
`list of strings` should be replaced by `array of strings`

### DIFF
--- a/config.md
+++ b/config.md
@@ -55,14 +55,14 @@ For Solaris, the mount entry corresponds to the 'fs' resource in the [zonecfg(1M
   This value MUST be an absolute path.
   * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
   * Solaris: corresponds to "dir" of the fs resource in [zonecfg(1M)][zonecfg.1m].
-* **`type`** (string, OPTIONAL) The filesystem type of the filesystem to be mounted.
-  * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+* **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
+  * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
   * Windows: this field MUST NOT be supplied.
   * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`source`** (string, OPTIONAL) A device name, but can also be a directory name or a dummy.
   * Windows: a local directory on the filesystem of the container host. UNC paths and mapped drives are not supported.
   * Solaris: corresponds to "special" of the fs resource in [zonecfg(1M)][zonecfg.1m].
-* **`options`** (list of strings, OPTIONAL) Mount options of the filesystem to be used.
+* **`options`** (array of strings, OPTIONAL) Mount options of the filesystem to be used.
   * Linux: supported options are listed in the [mount(8)][mount.8] man page. Note both [filesystem-independent][mount.8-filesystem-independent] and [filesystem-specific][mount.8-filesystem-specific] options are listed.
   * Solaris: corresponds to "options" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 


config.md
```diff
@@ -57,13 +57,13 @@ For Windows, see [mountvol][mountvol] and [SetVolumeMountPoint][set-volume-mount
   * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
   * Solaris: corresponds to "dir" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`type`** (string, OPTIONAL) The filesystem type of the filesystem to be mounted.
-  * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  * Linux: valid *filesystem type* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
```
wking: I don't think we want this change, since the current spelling matches `mount(2)`.

Mashimiao: Last sentence is "The filesystem type of the filesystem to be mounted" and we don't have filesystemtype in context, here looks weird.
Keeping consistent in context is easy for readers to understand. And I don't think it's so necessary to use parameter `filesystemtype` of mount for explanation in SPEC

wking: > And I don't think it's so necessary to use parameter filesystemtype of mount for explanation…

I was thinking that the current entry (from the ancient 77d44b10) was really indented to be `filesystemtype`, since we don't have a precedence for italicizing English words.  If we're moving away from mount(2)'s `filesystemtype` token, I think we should completely switch to English and use:

> Linux: valid filesystem types supported by the kernel are listed…

Although I'm not clear on how “valid” and “supported by the kernel” interact.  Maybe that will get sorted in #813.

Mashimiao: OK, let's leave filesystemtype alone, just keep list->array fix.
waiting any comments in #813

wking: > waiting any comments in #813.

Briefly covered in the [meeting][1] with @mrunalp and @crosbymichael, this is not a statement about config validation.  The spec has no opinions on what strings are valid values here.  Once some RFC 2119 language lands (e.g. see [here][2] in the rejected #771, which I hope to revive after opencontainers/runc#1442 lands and I can catch runC up with current kernels), it will be clear that all the runtime has to do is hand this value off to the kernel.  Whether the mount work or not is between the config author and the kernel.  So we want to drop `valid` here; this line is just a hint to config authors about where to figure out what the kernel will accept, and has nothing to do with config validation or runtime compliance testing.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-05-12-21.01.log.html#l-40
[2]: https://github.com/opencontainers/runtime-spec/pull/771/files#diff-c9c91c29b41257aea3a3403cc606ad99R64

Mashimiao: So, the conclusion is to keep filesystemtype as it is?

wking: > So, the conclusion is to keep filesystemtype as it is?

@mrunalp and @crosbymichael were fine splitting that up.  I think we want:

>. Linux: filesystem types supported by the kernel are listed…

Mashimiao: modified to `the types of filesystem`

config.md
```diff
@@ -56,14 +56,14 @@ For Windows, see [mountvol][mountvol] and [SetVolumeMountPoint][set-volume-mount
   This value MUST be an absolute path.
   * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
   * Solaris: corresponds to "dir" of the fs resource in [zonecfg(1M)][zonecfg.1m].
-* **`type`** (string, OPTIONAL) The filesystem type of the filesystem to be mounted.
-  * Linux: valid *filesystemtype* supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+* **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
+  * Linux: the types of filesystem supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
```
wking: This still reads awkwardly to me, since there are multiple filesystems listed in `/proc/filesystems`.  Maybe we can drop the “type” echo:

> * Linux: filesystems supported by the kernel are listed in `/proc/filesystems` (e.g., `minix`, `ext2`, `ext3`, …).

Mashimiao: I used `types` not `type`.

wking: > I used `types` not `type`.

But why use “type(s)” at all?  I still prefer the more compact form from [here][1], but if you want to use “type(s)” I'd recommend either:

> Linux: filesystem types supported by the kernel are listed in `/proc/filesystems` (e.g., `minix`, `ext2`, `ext3`, …).

or:

> Linux: types of filesystems supported by the kernel are listed in `/proc/filesystems` (e.g., `minix`, `ext2`, `ext3`, …).

[1]: https://github.com/opencontainers/runtime-spec/pull/807#discussion_r116557329

Mashimiao:

hqhq:
LGTM

crosbymichael:
LGTM

PR #808, opened 2017-05-11
config.md: specify mount source

Mashimiao:


wking:
Some related discussion in #735.  This PR takes the "block relative paths entirely" position, which is hard to enforce withough an exhaustive list of pseudo-filesystems.

crosbymichael:
We want to keep this wording as is.  We don't want to force absolute paths for the spec, that can be left to the user.

PR #810, opened 2017-05-12
config.go: platform-specific properties of process fix

Mashimiao:
capabilities, rlimits and nonewprivilege is not Linux-specific any more.
oomscoreadj becomes Linux-specific.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -44,33 +44,33 @@ type Process struct {
 	// Cwd is the current working directory for the process and must be
 	// relative to the container's root.
 	Cwd string `json:"cwd"`
-	// Capabilities are Linux capabilities that are kept for the process.
-	Capabilities *LinuxCapabilities `json:"capabilities,omitempty" platform:"linux"`
+	// Capabilities are platform-specific capabilities that are kept for the process.
+	Capabilities *Capabilities `json:"capabilities,omitempty"`
 	// Rlimits specifies rlimit options to apply to the process.
-	Rlimits []LinuxRlimit `json:"rlimits,omitempty" platform:"linux"`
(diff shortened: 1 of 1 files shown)


wking:
On Thu, May 11, 2017 at 10:34:15PM -0700, Ma Shimiao wrote:
> capabilities, rlimits and nonewprivilege is not Linux-specific any
> more.

This is catching us up with [1], and I'm still not sure that was a
good idea.  Does Solaris support ambient caps?  Does Windows support
any of this?  I think we want more feedback on that sort of thing
before we double down on it with breaking Go type changes.

> oomscoreadj becomes Linux-specific.

This catches us up with #222 and matches master [2], so +1 to getting
it in.  I think we should spin it off from the capabilities changes so
it doesn't get tied up with them.

[2]: https://github.com/opencontainers/runtime-spec/blame/45c3fd4b5002bc54786b85a19b1a2105502e29a6/config.md#L154

mrunalp:
@RobDolinMS @jlbutler Could you take a look at the changes for capabilities. Will these capabilities ever be supported on Solaris and Windows?

jhowardmsft:
It is highly unlikely these capabilities will ever be relevant on Windows.

vbatts:
i'm closing this, as it's too much to expect these linux features translate across platforms

wking:
On Wed, May 17, 2017 at 02:36:01PM -0700, v1.0.0.batts wrote:
> i'm closing this, as it's too much to expect these linux features
> translate across platforms

So we want a new PR rolling back [1] to make `capabilities` Linux-only
again.  I'll file this when I get time, unless someone beats me to it.

PR #814, opened 2017-05-12
Windows: Add CredentialSpec

jhowardmsft:

The Windows implementation of the docker daemon still passes a number of fields out of band from the OCI runtime spec. This PR addresses the `CredentialSpec` field by adding it to the runtime spec. (https://github.com/moby/moby/blob/master/libcontainerd/client_windows.go#L177-L178)

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -94,3 +94,14 @@ The following parameters can be specified:
         }
    }
 ```
+
+## <a name="configWindowsCredentialSpec" />Credential Spec
+
+You can configure a container's group Managed Service Account (gMSA) via the OPTIONAL `credentialspec` field of the Windows configuration.
+The `credentialspec` is a JSON object whose properties are implementation-defined.
+For more information about gMSAs, see [Active Directory Service Accounts for Windows Containers][gMSAOverview].
+For more information about tooling to generate a gMSA, see [Deployment Overview][gMSATooling].
+
+
+[gMSAOverview]: https://aka.ms/windowscontainers/manage-serviceaccounts
+[gMSATooling]: https://aka.ms/windowscontainers/credentialspec-tools
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -65,6 +65,10 @@
                         }
                     }
                 }
+            },
+            "credentialspec": {
+                "id": "https://opencontainers.org/schema/bundle/windows/credentialspec",
+                "type": "object"
             }
         }
     }
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -432,6 +432,8 @@ type SolarisAnet struct {
 type Windows struct {
 	// Resources contains information for handling resource constraints for the container.
 	Resources *WindowsResources `json:"resources,omitempty"`
+	// CredentialSpec contains a JSON object describing a group Managed Service Account (gMSA) specification.
+	CredentialSpec interface{} `json:"credentialspec,omitempty"`
 }
 
 // WindowsResources has container runtime resource constraints for containers running on Windows.


config-windows.md
```diff
@@ -94,3 +94,42 @@ The following parameters can be specified:
         }
    }
 ```
+
+## <a name="configWindowsCredentialSpec" />Credential Spec
+
+You can configure a container's group Managed Service Accounts (gMSAs) via the OPTIONAL `credentialspec` field of the Windows configuration. For more information about gMSAs, see [Active Directory Service Accounts for Windows Containers][gMSAOverview]. For more information about tooling to generate a gMSA, see [Deployment Overview][gMSATooling].
```
wking: Travis [doesn't like the trailing whitespace here][1] (or after `"GroupManagedServiceAccounts": [` below).

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/231727783#L355

config-windows.md
```diff
@@ -94,3 +94,42 @@ The following parameters can be specified:
...
+## <a name="configWindowsCredentialSpec" />Credential Spec
+
+You can configure a container's group Managed Service Accounts (gMSAs) via the OPTIONAL `credentialspec` field of the Windows configuration. For more information about gMSAs, see [Active Directory Service Accounts for Windows Containers][gMSAOverview]. For more information about tooling to generate a gMSA, see [Deployment Overview][gMSATooling]. 
+
+
+### Example
+
+```json
+    "windows": {
+        "credentialspec": {
```
wking: This value is a JSON object.  Your current docs don't have much to say on the value type (using the `credentialspec` (object, OPTIONAL) …` pattern would help with this), but your current Go type has it listed as a string (which will break if you try to read in your example here).

config-windows.md
jhowardmsft: Ah, yes, fixing.

config-windows.md
jhowardmsft: It's an opaque JSON blob passed as a string. I should have escaped the example and will shortly. The actual contents are very independent from the spec and implementation specific which is why I referred to the tooling to create such an example. I really don't believe the contents should be described in this spec.

jhowardmsft:
Added matching change to schema\config-windows.json

config-windows.md
```diff
@@ -94,3 +94,42 @@ The following parameters can be specified:
         }
    }
 ```
+
+## <a name="configWindowsCredentialSpec" />Credential Spec
+
+You can configure a container's group Managed Service Accounts (gMSAs) via the OPTIONAL `credentialspec` field of the Windows configuration. For more information about gMSAs, see [Active Directory Service Accounts for Windows Containers][gMSAOverview]. For more information about tooling to generate a gMSA, see [Deployment Overview][gMSATooling]. The `credentialspec` MUST be a string containing an escaped JSON object.
```
wking: Instead of making this a string with serialized JSON, can you make it a JSON object, say the object properties are implementation-defined, and use an `interface{}` in the Go type?  That should get you the same flexibility while still letting you (de)serialize in a single pass.

jhowardmsft: Yes, I can change this to be `interface{}`. (Done).  I'm not sure what this should be in `schema\config-windows.json` though if not type `string`? This is unique - no similar example exists in the spec to duplicate.

I have also updated the description in `config-windows.md` to indicate it's implementation-defined.

To be clear though - you are recommending to remove the example (even though technically accurate)? And what about the links? Should they stay (I think they should...)?

wking: > I'm not sure what this should be in schema\config-windows.json though if not type string?

You can just declare it an object and leave it at that (backing RFC drafts [here][1] and [here][2]):

```
"credentialspec": {
    "id": "https://opencontainers.org/schema/bundle/windows/credentialspec",
    "type": "object"
}
```

> To be clear though - you are recommending to remove the example (even though technically accurate)?

Ideally we'd link to the backing Windows docs (which are necessary to satisfy “[implementation-defined][3]”.  Lacking that, I'm in favor of including as much information as we can as long as we're reasonably confident that it won't go stale during the useful lifetime of the next spec release.  If you expect the links and example you have now to be current for the lifetime of the 1.0 runtime spec, I'm fine with them.  If you expect they will go stale before runtime-spec 1.0 is archaic, I'd rather replace them with something generic enough to stay current.

[1]: https://tools.ietf.org/html/draft-wright-json-schema-01#section-4.2
[2]: https://tools.ietf.org/html/draft-wright-json-schema-validation-01#section-6.25
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#notational-conventions

jhowardmsft: > You can just declare it an object and leave it at that

Thanks. Done.

While I don't believe there's any reason that the example will be invalid (as it's already in a released Microsoft product that customers are using), on reflection, it probably is cleaner to remove the example and refer readers back to the source of knowledge, that being the Microsoft documentation. Update done and pushed.

For the links, @patricklang is going to arrange a couple of `aka.ms` redirects, and I'll update the PR with them once I have them.

jhowardmsft:
Updated to `aka.ms` links. Thanks @patricklang 😸

TomSweeneyRedHat:
LGTM

mrunalp:
LGTM

tianon:
LGTM

PR #815, opened 2017-05-12
Windows: Add servicing

jhowardmsft:

Similar to https://github.com/opencontainers/runtime-spec/pull/814, the Windows implementation of the docker daemon still passes a number of fields out of band from the OCI runtime spec. This PR addresses the Servicing field by adding it to the runtime spec. https://github.com/moby/moby/blob/master/libcontainerd/client_windows.go#L152-L154

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -105,3 +105,18 @@ For more information about tooling to generate a gMSA, see [Deployment Overview]
 
 [gMSAOverview]: https://aka.ms/windowscontainers/manage-serviceaccounts
 [gMSATooling]: https://aka.ms/windowscontainers/credentialspec-tools
+
+
+## <a name="configWindowsServicing" />Servicing
+
+When a container terminates, the Host Compute Service indicates if a Windows update servicing operation is pending.
+You can indicate that a container should be started in a mode to apply pending servicing operations via the OPTIONAL `servicing` field of the Windows configuration.
+
+
+### Example
+
+```json
+    "windows": {
+        "servicing": true
+    }
+```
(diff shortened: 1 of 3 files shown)


jhowardmsft:
Adding matching change to config-windows.json

tianon:
Is there any documentation that could be pointed to which explains what this means? (to help users and/or implementers know when it's appropriate to set?)

jhowardmsft:
Rebased.

jhowardmsft:
@tianon Brian from my team may be able to comment further on the above if the following doesn't cover it. 

I'm not aware of any formal documentation. It means that the container has to be started in a special mode to apply a servicing operation (eg Windows update), as during shutdown, it indicated that updates needed to be applied.

I should also mentioned that the change has been validated to be good in the context of moby/moby (https://github.com/Microsoft/docker/tree/jjh/ocispec-moveservicing)

tianon:
Ah, perhaps the description in the markdown could be expanded to just include a few more of those words?  That clears it up well for me (and _I'm_ fully trusting that you've verified the change, was just hoping for a bit more to explain to the layperson what/why :heart:).

jhowardmsft:
@tianon OK, updated `config.md` with the following:

```
When a container terminates, the Host Compute Service indicates if a Windows update servicing operation is pending.
You can indicate that a container should be started in a mode to apply pending servicing operations via the OPTIONAL `servicing` field of the Windows configuration. 
```

tianon:
LGTM

:+1:

jhowardmsft:
Fixed dangling white-space error on the amendment to `config.md` which caused CI to fail

tianon:
LGTM

(redux)

hqhq:
LGTM

PR #818, opened 2017-05-12
Windows: Add Hyper-V isolation fields

jhowardmsft:

Similar to #814, the Windows implementation of the docker daemon still passes a number of fields out of band from the OCI runtime spec. This PR addresses the fields required to support Hyper-V containers by adding them to the runtime spec.

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -130,3 +130,24 @@ You can indicate that a container should be started in an a mode where disk flus
         "ignoreflushesduringboot": true
     }
 ```
+
+## <a name="configWindowsHyperV" />HyperV
+
+`hyperv` is an OPTIONAL field of the Windows configuration. If present, the container MUST be run with Hyper-V isolation. If omitted, the container MUST be run as a Windows Server container.
+
+The following parameters can be specified:
+
+* **`utilityvmpath`** *(string, OPTIONAL)* - specifies the path to the image used for the utility VM. This would be specified if using a base image which does not contain a utility VM image. If not supplied, the runtime will search the container filesystem layers from the bottom-most layer upwards, until it locates "UtilityVM", and default to that path.
+
+* **`sandboxpath`** *(string, REQUIRED)* - specifies the root of the path to the sandbox to be used for the container.
+
+### Example
+
+```json
+    "windows": {
+        "hyperv": {
+            "utilityvmpath": "C:\\\\path\\\\to\\utilityvm",
+            "sandboxpath": "C:\\\\programdata\\\\docker\\\\windowsfilter
+        }
+    }
+```
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -77,6 +77,20 @@
             "ignoreflushesduringboot": {
                 "id": "https://opencontainers.org/schema/bundle/windows/ignoreflushesduringboot",
                 "type": "boolean"
+            },
+            "hyperv": {
+                "id": "https://opencontainers.org/schema/bundle/windows/hyperv",
+                "type": "object",
+                "properties": {
+                    "utilityvmpath": {
+                        "id": "https://opencontainers.org/schema/bundle/windows/hyperv/utilityvmpath",
+                        "type": "string"
+                    },
+                    "sandboxpath": {
+                        "id": "https://opencontainers.org/schema/bundle/windows/hyperv/sandboxpath",
+                        "type": "string"
+                    }
+                }
             }
         }
     }
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -438,6 +438,8 @@ type Windows struct {
 	Servicing bool `json:"servicing,omitempty"`
 	// IgnoreFlushesDuringBoot indicates if the container is being started in a mode where disk writes are not flushed during its boot process.
 	IgnoreFlushesDuringBoot bool `json:"ignoreflushesduringboot,omitempty"`
+	// HyperV contains information for running a container with Hyper-V isolation.
+	HyperV *WindowsHyperV `json:"hyperv,omitempty"`
 }
 
 // WindowsResources has container runtime resource constraints for containers running on Windows.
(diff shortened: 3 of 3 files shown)


config-windows.md
```diff
@@ -94,3 +94,23 @@ The following parameters can be specified:
...
+
+## <a name="configWindowsHyperV" />HyperV
+
+`hyperv` is an OPTIONAL field of the Windows configuration. If present, the container will be run with Hyper-V isolation. If omitted, the container will be run as a Windows Server container.
+
+The following parameters can be specified:
+
+* **`utilityvmpath`** *(string, OPTIONAL)* - specifies the path to the image used for the utility VM. If not supplied, the runtime will search the container filesystem layers from the bottom-most layer upwards, until it locates "UtilityVM", and default to that path.
+
+* **`sandboxpath`** *(string)* - specifies the root of the path to the sandbox to be used for the container.
```
wking: Is this REQUIRED or OPTIONAL?  My very uninformed guess is that it is REQUIRED (if you set `hyperv`, you MUST set `sandboxpath`).  In case it helps (if you like ABNF), I took a stab at explaining the property-definition syntax [here][1] (from #747).

jhowardmsft: @wking Yes, you are correct it is REQUIRED if the Hyper-V structure is supplied. It's not REQUIRED though from a global spec perspective as the Hyper-V structure itself is optional. I took a look at the link, but I'm not entirely sure how I would represent that differently or how you suggest I alter this.

wking: > … but I'm not entirely sure how I would represent that differently or how you suggest I alter this.

I suggest you use:

```
* **`sandboxpath`** *(string, REQUIRED)* - specifies the root of the path to the sandbox to be used for the container.
```

The relevant section from the [link][1] is:

> If an OPTIONAL parent property is an object with a REQUIRED child, the child is only REQUIRED if the parent is set. For example…

and the following example.  For an example of this in the current spec, see [`mounts[].destination`][2], where `mounts` is OPTIONAL, but any entry in that array MUST include `destination`.

[2]: https://github.com/opencontainers/runtime-spec/blame/559acdf05cd4dd0b57d720a3981147597d249280/config.md#L55

jhowardmsft: Got it, thanks. Updated.

config-windows.md
```diff
@@ -94,3 +94,23 @@ The following parameters can be specified:
         }
    }
 ```
+
+## <a name="configWindowsHyperV" />HyperV
+
+`hyperv` is an OPTIONAL field of the Windows configuration. If present, the container will be run with Hyper-V isolation. If omitted, the container will be run as a Windows Server container.
```
tianon: Do you think it would be appropriate to replace `will be` with `MUST be` in these instances?
ala:
```markdown
`hyperv` is an OPTIONAL field of the Windows configuration.
If present, the container MUST be run with Hyper-V isolation.
If omitted, the container MUST be run as a Windows Server container.
```

(Or another of our RFC-requirement words, if `MUST` is too strong)

jhowardmsft: @tianon Sure, updated. I have to be honest it does feel a little strong, but there's nothing closer.

jhowardmsft:
Rebased

config-windows.md
tianon: Ok, if that feels too strong, maybe relaxing the second half to a `SHOULD` would be appropriate?  ie, if Hyper-V is requested via config, the container either MUST run via Hyper-V, or there must be an error, but otherwise, the behavior can be determined by the runtime if necessary?

jhowardmsft: On reflection, MUST is fine.

jhowardmsft:
Rebased again.

jhowardmsft:
Also verified this update is correct in the context of the docker engine.

tianon:
LGTM

config-windows.md
```diff
@@ -112,11 +112,30 @@ For more information about tooling to generate a gMSA, see [Deployment Overview]
...
     }
+```
+
+## <a name="configWindowsHyperV" />HyperV
+
+`hyperv` is an OPTIONAL field of the Windows configuration. If present, the container MUST be run with Hyper-V isolation. If omitted, the container MUST be run as a Windows Server container.
+
+The following parameters can be specified:
+
+* **`utilityvmpath`** *(string, OPTIONAL)* - specifies the path to the image used for the utility VM. If not supplied, the runtime will search the container filesystem layers from the bottom-most layer upwards, until it locates "UtilityVM", and default to that path.
```
TomSweeneyRedHat: It would be nice to have an example of this field in the example below.

dqminh:
Needs rebase now that the other windows PR was just merged 😢

jhowardmsft:
@dqminh Fully expected, no worries at all. Rebased again.

config-windows.md
jhowardmsft: @TomSweeneyRedHat Yes, I thought about that, but it was a deliberate omission. The reason is that to give an example of something meaningful (except `c:\\path\\to\\utilityvm`) is difficult as we don't currently (Windows RS1 and RS2) ship any base image which doesn't include the utility vm image contained within the base image. I expect this to change (in fact will have to) when we support Linux containers on Windows in the RS3 timeframe though, but I still can't at this time provide anything meaningful. Does that make sense?

TomSweeneyRedHat: @jhowardmsft Thanks for the background info.  Your call on this.  If I was personally making the decision, I'd put in the sample that you had in your comment, it at least gets the idea into the users mind.  I would ask though that once a real world sample can be given here that you loop back and either edit or add it.  Cheers!

jhowardmsft: How about I update the comment to

`specified the path to the image used for the utility VM. This would be specified if using a base image which does not contain a utility VM image. If not supplied, the runtime will....` and add the example then? To be fair `\\path\\to\\utilityvm` is probably about as meaningful an example as I could get regardless 😸

tianon:
LGTM

(redux)

crosbymichael:
LGTM

Issue #23, opened 2015-06-25
Clarify readonlyRootfs


taylorb-microsoft:
In confid.md:
We should clarify the readonlyRootfs flag.  My interpretation is that when true, writes to that file system from within the container would result in copy on write and when false they would actually modify the underlying file system.  Is that an accurate understanding of the intent?

philips:
That is great question. My interpretation is that readonlyRootfs means it is mounted RO and nothing is writeable, writes will fail.

philips:
@chanezon wrote the original spec, so what was the thought with readonlyRootfs. Perhaps @crosbymichael knows too.

wking:
On Thu, Jun 25, 2015 at 01:01:42PM -0700, Brandon Philips wrote:
> My interpretation is that readonlyRootfs means it is mounted RO and
> nothing is writeable, writes will fail.

That was my interpretation too.  If you want the CoW semantics, just
copy your container before launching it and don't set readonlyRootfs.
For bonus points, using something like btrfs' snapshots for this
initial copy will mean you get to share unaltered blocks between your
copy and the original.

That said, I agree that disambiguation is good, and “mounted
read-only” is pretty unambiguous ;).

taylorb-microsoft:
Makes since, I'll try to disambiguate.   Is this a feature utilized commonly today?  Windows does not support booting/running the OS from read-only volumes so this would have to be a platform split at least for now.

wking:
On Thu, Jun 25, 2015 at 03:32:56PM -0700, Taylor Brown wrote:
> Is this a feature utilized commonly today?

I don't know how to get a number for that, but the ability to do this
in Docker landed in 1.5 [1], and the associated PR
(docker/docker#10093) references docker/docker#7923 (“Fedora has
supported the concept of a readonly root for a long time…).  The other
referenced issues (docker/docker#2710 and docker/docker#8752) seem to
be more focused on the general ideas of increased security and the
fact that you'll fail early if your application tries to store
information in a volume that you don't intend to save.

[1]: https://blog.docker.com/2015/02/docker-1-5-ipv6-support-read-only-containers-stats-named-dockerfiles-and-more/

taylorb-microsoft:
Interesting, seems like the initial conversation was skeptical on it's usefulness but I do like the idea of it interims of failing deterministically if the process tires to write outside of the intended/defined scope.  I submitted https://github.com/opencontainers/specs/pull/25 to try and cleanup/disambiguate the content.

PR #819, opened 2017-05-13
Windows: Clarify r/o root filesystem

jhowardmsft:

Clarification that Windows does not support a read-only root filesystem.

### DIFF
--- a/config.md
+++ b/config.md
@@ -32,7 +32,7 @@ For example, if a configuration is compliant with version 1.1 of this specificat
   The path is either an absolute path or a relative path to the bundle.
   On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
   A directory MUST exist at the path declared by the field.
-* **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false.
+* **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false. On Windows, this field must be omitted or false.
 
 ### Example
 


config.md
```diff
@@ -32,7 +32,7 @@ For example, if a configuration is compliant with version 1.1 of this specificat
   The path is either an absolute path or a relative path to the bundle.
   On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
   A directory MUST exist at the path declared by the field.
-* **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false.
+* **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false. On Windows, this field must be omitted or false.
```
wking: Previous discussion in #23 and #476.

An alternative to this would be to label the property as Linux/Solaris-specific, so it wouldn't be specified for Windows and Windows runtimes could silently ignore it (using [this][1]).  That gets us to the runtime behavior proposed in #476, but avoids confusing config authors (like the spec wording in #476) because Windows authors would know that the property did not apply to their platform, and non-Windows authors would know that the property did apply (and runtimes would have to implement it) on their platforms.

[1]: https://github.com/opencontainers/runtime-spec/blob/559acdf05cd4dd0b57d720a3981147597d249280/config.md#extensibility

tianon:
LGTM

Forcing this to be a known-supported value like this makes sense IMO -- that allows for it to be supported in a future update if the platform adds support.

This also ends up mirroring other places in the spec where features requested in the configuration that are unsupported by the runtime or the underlying kernel MUST result in an error (which is essentially what this is codifying for Windows).

wking:
> This also ends up mirroring other places in the spec where features requested in the configuration that are unsupported by the runtime or the underlying kernel MUST result in an error ...

To clarify [my alternative proposal][1], what do you think a Linux runtime should do if a Linux config sets the Windows-only `process.user.username`?  I think it can ignore the unknown-to-its-platform property.  Are you suggesting it MUST raise an error?

[1]: https://github.com/opencontainers/runtime-spec/pull/819#discussion_r116351941

tianon:
As a user, anything other than an error would be very surprising behavior.

TomSweeneyRedHat:
LGTM

vbatts:
LGTM

PR #820, opened 2017-05-13
Clarify root path

jhowardmsft:

This changes the root path to OPTIONAL as it is not supported for Hyper-V containers on Windows, where the root filesystem isn't accessible on the host operating system - it only has meaning inside the utility VM running the container.

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -8,17 +8,22 @@ See also [MacOS application bundles][macos_bundle] for a similar use of the term
 The definition of a bundle is only concerned with how a container, and its configuration data, are stored on a local filesystem so that it can be consumed by a compliant runtime.
 
 A Standard Container bundle contains all the information needed to load and run a container.
-This MUST include the following artifacts:
+This includes the following artifacts:
 
 1. <a name="containerFormat01" />`config.json`: contains configuration data.
 This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
 See [`config.json`](config.md) for more details.
 
 2. <a name="containerFormat02" />A directory representing the root filesystem of the container.
-While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
-This directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
+    While the name of this directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
 
-While these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.
+    On Windows, for Windows Server containers, this directory is REQUIRED. For Hyper-V containers, it MUST be omitted.
+
+    On all other platforms, this field is REQUIRED.
+
+    If set, this directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
+
+When supplied, while these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.
 In other words, a tar archive of a *bundle* will have these artifacts at the root of the archive, not nested within a top-level directory.
 
 [macos_bundle]: https://en.wikipedia.org/wiki/Bundle_%28macOS%29
--- a/config.md
+++ b/config.md
@@ -28,10 +28,15 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 
 **`root`** (object, REQUIRED) specifies the container's root filesystem.
 
-* **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
-  The path is either an absolute path or a relative path to the bundle.
-  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
-  A directory MUST exist at the path declared by the field.
+* **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container. The path is either an absolute path or a relative path to the bundle.
+
+    On Windows, for Windows Server Containers, this field is REQUIRED. For Hyper-V Containers, this field MUST be omitted.
+
+    On all other platforms, this field is REQUIRED.
+
+    On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
+
+    If defined, a directory MUST exist at the path declared by the field.
 * **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false. On Windows, this field must be omitted or false.
 
 ### Example
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -59,9 +59,6 @@
             "description": "Configures the container's root filesystem.",
             "id": "https://opencontainers.org/schema/bundle/root",
             "type": "object",
-            "required": [
-                "path"
-            ],
             "properties": {
                 "path": {
                     "id": "https://opencontainers.org/schema/bundle/root/path",
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -96,7 +96,7 @@ type User struct {
 // Root contains information about the container's root filesystem on the host.
 type Root struct {
 	// Path is the absolute path to the container's root filesystem.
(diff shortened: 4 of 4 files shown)


jhowardmsft:
Not sure what travis is complaining about.... can someone help me understand? 😬

config.md
```diff
@@ -28,10 +28,12 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 
 **`root`** (object, REQUIRED) specifies the container's root filesystem.
 
-* **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
+* **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container.
   The path is either an absolute path or a relative path to the bundle.
-  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
-  A directory MUST exist at the path declared by the field.
+  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
```
wking: Travis doesn't like [the trailing whitespace here][1]:

```
$ git show --oneline --check origin/pr/820
23a0b15 Clarify root path
config.md:33: trailing whitespace.
+  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`. 
```

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/231771631#L355

config.md
```diff
@@ -28,10 +28,12 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 
 **`root`** (object, REQUIRED) specifies the container's root filesystem.
 
-* **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
+* **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container.
   The path is either an absolute path or a relative path to the bundle.
-  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
-  A directory MUST exist at the path declared by the field.
+  On Linux, for example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`. 
+  On Windows, for Windows Server Containers, this field is REQUIRED. For Hyper-V Containers, this field MUST NOT be supplied.
```
wking: For writing validation tooling, how should runtime-tools distinguish between a Windows Server config and a Hyper-V config?

This change will also need some knock ons:

* Adjusting [the][1] [bundle][2] [requirements][3] to only require the root filesystem directory when `root.path` is set.  I still don't see a need for that bundle requirement (last discussed in #469) and think that the `config.md` requirement (which you preserve in this PR) for “A directory MUST exist at the path declared by the field” is sufficient for sanity.  But if you can't get consensus around something like that, you'll still need to do something to `bundle.md`, since the current wording there makes it impossible to leave `root.path` unset.
* An `omitempty` [in the Go type][4].
* Dropping the `path` requirement [from the JSON Schema][5].

[1]: https://github.com/opencontainers/runtime-spec/blame/559acdf05cd4dd0b57d720a3981147597d249280/bundle.md#L11
[2]: https://github.com/opencontainers/runtime-spec/blame/559acdf05cd4dd0b57d720a3981147597d249280/bundle.md#L17-L19
[3]: https://github.com/opencontainers/runtime-spec/blame/559acdf05cd4dd0b57d720a3981147597d249280/bundle.md#L21-L21
[4]: https://github.com/opencontainers/runtime-spec/blob/559acdf05cd4dd0b57d720a3981147597d249280/specs-go/config.go#L99
[5]: https://github.com/opencontainers/runtime-spec/blob/559acdf05cd4dd0b57d720a3981147597d249280/schema/config-schema.json#L62-L64

config.md
jhowardmsft: Thanks! Fixed in next push

config.md
jhowardmsft: > For writing validation tooling, how should runtime-tools distinguish between a Windows Server config and a Hyper-V config?

See the associated PR in https://github.com/opencontainers/runtime-spec/pull/818

Have updated the bundle requirements, the go type and the schema

wking: > … how should runtime-tools distinguish between a Windows Server config and a Hyper-V config?

Probably by looking for a `hyperv` property (in flight with #818).  Maybe we should wait until #818 lands and then reroll this PR to mention that property?

bundle.md
```diff
@@ -8,17 +8,22 @@ See also [MacOS application bundles][macos_bundle] for a similar use of the term
...
-While the name of this REQUIRED directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
-This directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
+While the name of this directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
 
-While these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.
+On non-Windows platforms, this directory is REQUIRED.
+
+On Windows, for Windows Server containers, this directory is REQUIRED. For Hyper-V containers, it MUST be omitted.
+
+If set, this directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
```
wking: I think these are all part of the second list entry.  To get multi-paragraph list entries in Pandoc, you need four-space (or tab) indents (more in #495).  Can you use:

```
2. <a name="containerFormat02" />A directory representing the root filesystem of the container.
    While the name of this directory may be arbitrary, users should consider using a conventional name, such as `rootfs`.
 
    On non-Windows platforms, this directory is REQUIRED.

    On Windows, for Windows Server containers, this directory is REQUIRED.
    For Hyper-V containers, it MUST be omitted.

    If set, this directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
```

jhowardmsft: Indeed. Updated.

bundle.md
```diff
@@ -8,17 +8,21 @@ See also [MacOS application bundles][macos_bundle] for a similar use of the term
...
 
-While these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.
-In other words, a tar archive of a *bundle* will have these artifacts at the root of the archive, not nested within a top-level directory.
+    On Windows, for Windows Server containers, this directory is REQUIRED. For Hyper-V containers, it MUST be omitted.
+
+    On all other platforms, this field is REQUIRED.
+
+    If set, this directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
+
+When supplied, while these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle. In other words, a tar archive of a *bundle* will have these artifacts at the root of the archive, not nested within a top-level directory.
```
wking: You seem to be removing the newline before “In other words” while not touching the content of that sentence.  I think you want to [keep that sentence on its own line][1].

Also, “When supplied, while …” is a bit awkward.  I keep circling back to #469 when trying to think up clearer phrasing, so I don't have anything new to propose as an alternative.  But probably something that separates out the same-directory MUST from the definition of “bundle” as the parent of `config.json` (and any other artifacts listed here).

jhowardmsft:
> You seem to be removing the newline before “In other words” while not touching the content of that sentence.  I think you want to keep that sentence on its own line.

Hmmm. Yes, I see from the link. But from a grammatical perspective, it seems more natural to me as it directly relates to the previous sentence. Reverted though.

wking:
On Mon, May 15, 2017 at 01:33:50PM -0700, John Howard wrote:
> But from a grammatical perspective, it seems more natural to me as
> it directly relates to the previous sentence.

The rendered page will include both lines in the same paragraph, since
Markdown needs a blank line (e.g. two consecutive newlines, not just
one) to split paragraphs [1].  The one-sentence-per-line policy is
mostly for convenient diffing (e.g. ‘git blame …’ will continue to
show the second line as last touched by 106ec2da, #446).

[1]: https://daringfireball.net/projects/markdown/syntax#p

tianon:
LGTM

vbatts:
LGTM

PR #821, opened 2017-05-13
Windows - correct mounts

jhowardmsft:

Not sure where the mis-information came from, but the mounts section on Windows has incorrect information.

- The type is not used
- The example was incorrect as the source is not a volume path
- Better not have generic example rather than including a users name ;)
- Clarification of what the source can be (although we hope to be more reaching in future releases here)

### DIFF
--- a/config.md
+++ b/config.md
@@ -49,7 +49,6 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 The runtime MUST mount entries in the listed order.
 For Linux, the parameters are as documented in [mount(2)][mount.2] system call man page.
 For Solaris, the mount entry corresponds to the 'fs' resource in the [zonecfg(1M)][zonecfg.1m] man page.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -91,9 +90,8 @@ For Windows, see [mountvol][mountvol] and [SetVolumeMountPoint][set-volume-mount
 ```json
 "mounts": [
     {
-        "destination": "C:\\Users\\crosbymichael\\My Fancy Mount Point\\",
-        "type": "ntfs",
-        "source": "\\\\?\\Volume\\{2eca078d-5cbc-43d3-aff8-7e8511f60d0e}\\",
+        "destination": "C:\\folder-inside-container",
+        "source": "C:\\folder-on-host}",
```
wking: Is the unbalanced closing brace intentional?

jhowardmsft: That was a typo. Updated.

tianon:
LGTM

hqhq:
LGTM

PR #822, opened 2017-05-14
tfix: replace setup

Mashimiao:
setup is a noun, 'set up' is a verb

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -6,7 +6,7 @@ The Linux container specification uses various kernel features like namespaces, 
 ## <a name="configLinuxDefaultFilesystems" />Default Filesystems
 
 The Linux ABI includes both syscalls and several special file paths.
-Applications expecting a Linux environment will very likely expect these file paths to be setup correctly.
+Applications expecting a Linux environment will very likely expect these file paths to be set up correctly.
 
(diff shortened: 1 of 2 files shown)


config-linux.md
```diff
@@ -24,7 +24,7 @@ Changes to the global resource are visible to other processes that are members o
 For more information, see the [namespaces(7)][namespaces.7_2] man page.
 
 Namespaces are specified as an array of entries inside the `namespaces` root field.
-The following parameters can be specified to setup namespaces:
+The following parameters can be specified to set up namespaces:
```
wking: I'd prefer we use “configure” in the sub-properties template.  Also looking at parameter vs. property vs. field:

```
$ git describe
v1.0.0-rc5-162-g4aed614
$ git grep -i propert *.md | wc -l
23
$ git grep -i parameter *.md | wc -l
19
$ git grep -i field *.md | wc -l
18
```

So I'd like a template like:

> The following properties can be specified to configure {thing}.

In this case that would be:

> The following properties can be specified to configure namespaces:

But this is just me trying to copy-edit for consistency.  The lines are clear as they stand, and slightly more clear with just the “setup” → “set up” change.

tianon:
LGTM

dqminh:
LGTM

PR #811, opened 2017-05-12
consistency and style fix

Mashimiao:
replace file sytem with filesystem for consistent in context
keep one sentence per line

Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #823, opened 2017-05-14
config-linux.md: add cgroupsPath entry and some modifications

Mashimiao:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -171,14 +171,18 @@ Also known as cgroups, they are used to restrict resource usage for a container 
 cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
-The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
-If `cgroupsPath` is:
-* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
-* ... a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
-* ... not specified, the runtime MAY define the default cgroup path.
+### <a name="configLinuxCgroupsPath" />Cgroups Path
+
+**`cgroupsPath`** (string, OPTIONAL) path to the cgroups.
+It can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+
+The value of `cgroupsPath` MUST be either an absolute path or a relative path.
+* In the case of an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroups mount point.
+* In the case of a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroups hierarchy.
+
+If the value is specified, the runtime MUST consistently attach to the same place in the cgroups hierarchy given the same value of `cgroupsPath`.
+If the value is not specified, the runtime MAY define the default cgroups path.
 Runtimes MAY consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
-If a `cgroupsPath` value is specified, the runtime MUST consistently attach to the same place in the cgroup hierarchy given the same value of `cgroupsPath`.
 
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.


config-linux.md
```diff
@@ -171,14 +171,15 @@ Also known as cgroups, they are used to restrict resource usage for a container
 cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
-The path to the cgroups can be specified in the Spec via `cgroupsPath`.
-`cgroupsPath` can be used to either control the cgroup hierarchy for containers or to run a new process in an existing container.
-If `cgroupsPath` is:
-* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
-* ... a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
-* ... not specified, the runtime MAY define the default cgroup path.
+#### <a name="configLinuxCgroupsPath" />Cgroups Path
```
wking: The previous section is `## Control groups`, so you only need `###` here (not `####`).

config-linux.md
```diff
@@ -171,14 +171,15 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
+
+**`cgroupsPath`** (string, OPTIONAL) path to the cgroups. It can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.The value of `cgroupsPath` is either an absolute path or a relative path.
```
wking: I'd recommend [one sentence per line][1], e..g see #811 where you split multi-sentence lines in a few places.  And if you don't split the sentences onto their own lines, you'll want a space before “The value”.

config-linux.md
Mashimiao: I just want keep consistent with blkio, cpu, and so on. Those sections all start with `####`.
Do you mean we should modify them?

config-linux.md
Mashimiao: Fixed, thanks

config-linux.md
wking: > Do you mean we should modify them?

Yes, and I've filed #832 to do that independent of content changes.

config-linux.md
```diff
@@ -171,14 +171,18 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
+#### <a name="configLinuxCgroupsPath" />Cgroups Path
+
+**`cgroupsPath`** (string, OPTIONAL) path to the cgroups.
+It can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+
+The value of `cgroupsPath` is either an absolute path or a relative path.
+* an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroups mount point.
+* a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroups hierarchy.
+
+If the value is specified, the runtime MUST consistently attach to the same place in the cgroup hierarchy given the same value of `cgroupsPath`.
```
wking: I have no preference on “cgroup hierarchy” vs “cgroups hierarchy”, but you have “cgroup hierarchy” here and “cgroups hierarchy” on line 177 (as of f871f49b).

config-linux.md
```diff
@@ -171,14 +171,18 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
-* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
-* ... a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
-* ... not specified, the runtime MAY define the default cgroup path.
+### <a name="configLinuxCgroupsPath" />Cgroups Path
+
+**`cgroupsPath`** (string, OPTIONAL) path to the cgroups.
+It can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+
+The value of `cgroupsPath` is either an absolute path or a relative path.
+* an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroups mount point.
```
TomSweeneyRedHat: Suggest adding "In the case of" in front of this and the next  line.

Mashimiao: Thanks, updated

config-linux.md
```diff
@@ -171,14 +171,18 @@ Also known as cgroups, they are used to restrict resource usage for a container
...
-If `cgroupsPath` is:
-* ... an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroup mount point.
-* ... a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroup hierarchy.
-* ... not specified, the runtime MAY define the default cgroup path.
+### <a name="configLinuxCgroupsPath" />Cgroups Path
+
+**`cgroupsPath`** (string, OPTIONAL) path to the cgroups.
+It can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
+
+The value of `cgroupsPath` is either an absolute path or a relative path.
```
wking: “is” → “MUST be”?  And we can punt to POSIX for formal definitions of [absolute][1] and [relative paths][2].

And do we want to clarify how the empty string is handled (`"cgroupsPath": ""`)?  POSIX doesn't come down firmly on this point, and [allows empty-string paths][3] which [may result in errors or be synonyms for `.`][3].  I'd rather avoid that ambiguity in this spec and forbid explicit empty strings here (although see similar discussion in the rejected #843).

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_02
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_324
[3]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_271

Mashimiao: empty string means the cgroupsPath is not specified, `runtime MAY define the default cgroups path` as the following sentence says

wking: > empty string means the cgroupsPath is not specified…

Does the spec support that?  I think `"cgroupsPath": ""` *is* specifying `cgroupsPath`, because the `cgroupsPath` key exists in the `linux` object.  If we want to conflate `""` with “unset” for this property (or in general?), we'd need to land spec wording to that effect.

Personally, I think life is just as easy for config authors and simpler for runtimes if we forbid empty-string values where they don't make sense.  Runtimes [don't have to validate configs if they don't want][1], so making empty strings illegal would mean that runtimes could handle those configs however they like (erroring out or conflating them with whatever other case they liked).

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/runtime.md#L109

crosbymichael:
We are currently reworking the cgroupsPath description #834.

Mashimiao:
@crosbymichael  yeah, I noticed that, but I don't think we should close this PR. In this PR,  I mainly want to add cgroupsPath entry. As you know, there is not it yet in the spec.

hqhq:
LGTM

crosbymichael:
LGTM

PR #826, opened 2017-05-15
config-linux.md: fix wrong link description about cgroups

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #827, opened 2017-05-15
config-linux: Use the implicit link name shortcut

wking:
#826 added a redundant link name for libseccomp.  This commit restores the old [implicit link name shortcut][1] there and updates the other entries (all of which were for the filesystems table in config-linux) that could also use that shortcut.  This is obviously not a major change, but I think it makes the source easier to read by minimizing the markup so you can focus on the text.

[1]: https://daringfireball.net/projects/markdown/syntax#link

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -12,10 +12,10 @@ The following filesystems SHOULD be made available in each container's filesyste
 
 | Path     | Type   |
 | -------- | ------ |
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM

hqhq:
LGTM

crosbymichael:
LGTM

Issue #817, opened 2017-05-12
Windows targets for 1.0?


wking:
Can somebody from Windows comment on how they see #801, #814, #815, and #816 fitting into 1.0?  Are these 1.0 blockers?  There have been quite a few changes so far since rc5 (and maybe more in #746), but for the most part they've been minor polishing and shifting around well-known properties (e.g. #789).  These feel like larger changes to me, and I'm concerned about pushing them into 1.0 before they've had time to cook.

Do you see these PRs as 1.0 blockers?  As something you're staging here for post-1.0?  Have they already been cooking somewhere else, and therefore maybe needing a shorter cook-time in OCI?

Even if the properties and docs have been cooking for a while somewhere else, I think spec language also needs cooking.  Is there enough here for compliance testing based on RFC 2119 and [our definition of “compliant”][1]?

Ping @jhowardmsft, @darrenstahlmsft, @RobDolinMS.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/spec.md#L40-L41

jhowardmsft:
These fields (and more) are currently passed out of band in docker (see link to specific code) for some time (most since TP3 timeframe of Windows Server 2016). We are pushing to get these in to v1.0 so that Windows can be spec compliant. They are all required fields to run containers on Windows.

wking:
So “we see them as 1.0 blockers” and “the properties are well cooked”.

Do you feel they are sufficient for compliance testing?  It would make
me happier (although I'm not a maintainer, so my happiness is not
relevant ;) if runtime-tools was capabable of doing some Windows
compliance testing.  But like Linux and #746, I expect we only have
time for an eyeball assessment of “we can write tests based on this”
on most properties.

On Fri, May 12, 2017 at 03:41:22PM -0700, John Howard wrote:
> They are all required fields to run containers on Windows.

I don't see REQUIRED in any of them, so you probably mean “necessary
for most useful containers on Windows”.  That makes life a bit easier
though, since you could cut 1.0 without a particular OPTIONAL
property, continue to use it as an extension [1], and then land the
backing spec in 1.1.  That way everything works, and just means:

* 1.0 compliance testing is less meaningful on Windows, and
* 1.0 configs are more vulnerable to interop issues between different
  Windows runtimes (which is probably not a big deal if there is only
  one Windows runtime.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#extensibility

[...]

PR #816, opened 2017-05-12
Windows: Add IgnoreFlushesDuringBoot

jhowardmsft:
Signed-off-by: John Howard <jhoward@microsoft.com>

Similar to #814, the Windows implementation of the docker daemon still passes a number of fields out of band from the OCI runtime spec. This PR addresses the IgnoreFlushesDuringBoot field by adding it to the runtime spec. https://github.com/moby/moby/blob/master/libcontainerd/client_windows.go#L156-L159

PR #801, opened 2017-05-10
Update to Windows network options

darrenstahlmsft:
This moves the Windows network runtime options into the spec. 

NetworkResourceSettings was never used in the runtime spec, and implemented in HNS and libnetwork instead of at runtime on Windows.

/cc @RobDolinMS 

Signed-off-by: Darren Stahl <darst@microsoft.com>

PR #828, opened 2017-05-15
Add LayerFolders to Windows platform config

sunjayBhatia:
LayerFolders is a property that is missing from the runtime spec that is **required** for all useful containers on Windows.

[This issue](https://github.com/opencontainers/runtime-spec/issues/817) mentions extending the runtime spec with "required" platform specific fields.

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -3,6 +3,22 @@
 This document describes the schema for the [Windows-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
 The Windows container specification uses APIs provided by the Windows Host Compute Service (HCS) to fulfill the spec.
 
+## <a name="configWindowsLayerFolders" />LayerFolders
+
+**`layerFolders`** (array of strings, REQUIRED) specifies a list of layer folders the container image relies on. The list is ordered from topmost layer to base layer.
+  `layerFolders` MUST contain at least one entry.
+
+### Example
+
+```json
+    "windows": {
+        "layerFolders": [
+            "C:\\Layers\\layer1",
+            "C:\\Layers\\layer2"
+        ]
+    }
+```
+
 ## <a name="configWindowsResources" />Resources
 
 You can configure a container's resource limits via the OPTIONAL `resources` field of the Windows configuration.
--- a/config.md
+++ b/config.md
@@ -328,7 +328,7 @@ Runtime implementations MAY support any valid values for platform-specific field
 * **`linux`** (object, OPTIONAL) [Linux-specific configuration](config-linux.md).
   This MAY be set if **`platform.os`** is `linux` and MUST NOT be set otherwise.
 * **`windows`** (object, OPTIONAL) [Windows-specific configuration](config-windows.md).
-  This MAY be set if **`platform.os`** is `windows` and MUST NOT be set otherwise.
+  This MUST be set if **`platform.os`** is `windows` and MUST NOT be set otherwise.
 * **`solaris`** (object, OPTIONAL) [Solaris-specific configuration](config-solaris.md).
   This MAY be set if **`platform.os`** is `solaris` and MUST NOT be set otherwise.
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -430,6 +430,8 @@ type SolarisAnet struct {
 
 // Windows defines the runtime configuration for Windows based containers, including Hyper-V containers.
 type Windows struct {
+	// LayerFolders contains a list of absolute paths to directories containing image layers.
+	LayerFolders []string `json:"layerFolders"`
 	// Resources contains information for handling resource constraints for the container.
 	Resources *WindowsResources `json:"resources,omitempty"`
 }
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -4,6 +4,14 @@
         "id": "https://opencontainers.org/schema/bundle/windows",
         "type": "object",
         "properties": {
+            "layerFolders": {
+                "id": "https://opencontainers.org/schema/bundle/windows/layerFolders",
+                "type": "array",
+                "items": {
+                    "$ref": "defs.json#/definitions/FilePath"
+                },
+                "minItems": 1
+            },
             "resources": {
                 "id": "https://opencontainers.org/schema/bundle/windows/resources",
                 "type": "object",
@@ -66,6 +74,9 @@
                     }
                 }
             }
-        }
+        },
+        "required": [
+            "layerFolders"
+        ]
     }
 }


wking:
PullApprove doesn't like your Signed-off-by [1].  I'm guessing that's
because the author is Sunjay Bhatia and the Signed-off-by is Matthew
Horan.  If you're forwarding Matthew's original Signed-off-by, you can
use:

  Signed-off-by: Matthew Horan <mhoran@pivotal.io>
  Signed-off-by: Sunjay Bhatia <sbhatia@pivotal.io>

Using clause (c) in the DCO [2,3].

[1]: https://pullapprove.com/opencontainers/runtime-spec/pull-request/828/
[2]: http://developercertificate.org
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/README.md#sign-your-work

config-windows.md
```diff
@@ -3,6 +3,21 @@
 This document describes the schema for the [Windows-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
 The Windows container specification uses APIs provided by the Windows Host Compute Service (HCS) to fulfill the spec.
 
+## <a name="configWindowsLayerFolders" />LayerFolders
+
+**`layerfolders`** (array of strings, REQUIRED) specifies a list of layer folders the container image relies on. The list is ordered from topmost layer to base layer.
```
wking: Just to be clear, this is something that is not covered by [`mounts`][1] or by [image-spec unpacking][2]?  These layers need to be assembled by the runtime at container-creation time?

[1]: https://github.com/opencontainers/runtime-spec/blob/4aed614c791f4fe291baa2c7cc3dde8734be0876/config.md#example-windows
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/layer.md#applying-changesets

sunjayBhatia: > These layers need to be assembled by the runtime at container-creation time?

Yes, this is true. As mentioned in https://github.com/opencontainers/runtime-spec/issues/817#issuecomment-301203814, the underlying subsystem requires that a list of layer paths be passed to the kernel in order to spin up the container. These are not mounts, and I don't believe any functionality like image-spec unpacking is applicable in this case.

sunjayBhatia:
Thanks for the note about PullApprove. We use [git-duet](https://github.com/git-duet/git-duet) to automatically sign our commits when pairing, and it only adds a single sign-off which corresponds to the other pair. That doesn't work with the requirements of PullApprove.

wking:
On Mon, May 15, 2017 at 12:51:53PM -0700, Sunjay Bhatia wrote:
> Thanks for the note about PullApprove. We use git-duet…

I've filed git-duet/git-duet#51 in case they have any ideas about
making this easier for git-duet users.

jhowardmsft:
@sunjayBhatia Thanks for submitting this - I was just about to add it myself!

(@wking FYI, this is the last of the additions required for Windows - the complete set is https://github.com/opencontainers/runtime-spec/pull/801, https://github.com/opencontainers/runtime-spec/pull/814,  https://github.com/opencontainers/runtime-spec/pull/815, https://github.com/opencontainers/runtime-spec/pull/816, https://github.com/opencontainers/runtime-spec/pull/817 and https://github.com/opencontainers/runtime-spec/pull/818)

@sunjayBhatia There's a few issues here which need addressing. The `LayerPaths` is correct as an array of ordered folders, but it needs to be part of an object at the same level to `Resources` similar to how it is defined in moby/moby libcontainerd\types_windows.go.

```golang
type LayerOption struct {
	// LayerFolderPath is the path to the current layer folder. Empty for Hyper-V containers.
	LayerFolderPath string `json:",omitempty"`
	// Layer paths of the parent layers
	LayerPaths []string
}
```
It also needs an update to `schema\config-windows.json`. Let me know if you need any help with this. And thanks again 😸

sunjayBhatia:
@jhowardmsft I'm not sure I agree with including the `LayerFolderPath` as something that gets included in the runtime spec, mostly because it feels like duplicate info and tying the spec to Docker implementation/`hcsshim` types/fields feels a little odd.

The "current layer" (sandbox layer in `hcsshim` parlance) seems to be equivalent to the bundle directory that an OCI compliant CLI (like `runc`) is given when creating a container. The bundle `config.json` should specify a `Root.Path` to the image that containers created from the bundle are based on and the CLI should activate a sandbox layer in the bundle directory since that is where the container's file system lives. When actually creating a Windows Server container from the current layer with `hcsshim`, we can pass the bundle directory as `LayerFolderPath` in the `hcsshim.ContainerConfig`.

With other config options, we have little choice to do anything other than bubble up the `hcsshim.ContainerConfig` fields up to this spec but I feel like we don't have to do that here.

sunjayBhatia:
Updated `schema/config-windows.json` to reflect the new field

tianon:
LGTM

tianon:
(Taking back my LGTM for now -- @jhowardmsft tells me he's going to review and comment on this one. :heart:)

sunjayBhatia:
@wking any more thoughts on this?

sunjayBhatia:
@jhowardmsft ^

config-windows.md
wking: For required arrays, you probably also want to require at least one entry (otherwise you might as well make the property OPTIONAL).  See #769 for an example of the spec wording and `minItems` JSON Schema I've used for that sort of thing before.

TomSweeneyRedHat: Not my call, but it looks like the variables are all camelCase in here.  Should this instead be "layerFolders"?

sunjayBhatia: thanks @wking @TomSweeneyRedHat, we made these changes and modified the commit for this PR

jhowardmsft:
@sunjayBhatia I'm good with this change. LGTM (not a maintainer).

However, addressing your comment above, it's slightly more complicated. We actually always need to populate a `LayerFolderPath` for HCS which needs to be a top-level item. I'll submit a PR for that shortly. And also at the same time, remove the sandbox path which I recently introduced. Between TP5 and RS1, there was a change in HCS such that if SandboxPath is set, it overwrites LayerFolderPath internally, even though LayerFolderPath is set. IOW, SandboxPath is completely redundant in all RTM version of Windows, and LayerFolderPath is required.

As for your comment about `Root`, there is no `Root` as such in Hyper-V containers, as the root is really only meaningful inside the Utility VM, not the host itself. In Windows Server Containers, there is a root which is a volume-style path (`\\?\Volume{GUID}` style) of the containers root file-system which is mounted on the host, and required.

config-windows.md
```diff
@@ -3,6 +3,22 @@
 This document describes the schema for the [Windows-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
 The Windows container specification uses APIs provided by the Windows Host Compute Service (HCS) to fulfill the spec.
 
+## <a name="configWindowsLayerFolders" />LayerFolders
+
+**`layerFolders`** (array of strings, REQUIRED) specifies a list of layer folders the container image relies on. The list is ordered from topmost layer to base layer.
```
wking: This is the first instance (as far as I know) of a REQUIRED property that is a direct child of a platform property (`windows`, in this case).  Because REQUIRED is only “[if the parent is set][1]”, I think you'll want to adjust [this section][2] to say that `windows` MUST be set when `platform.os` is `windows`.

Discussion in #830 may end up with [`platform` being removed entirely][3].  If that lands first, it will change the phrasing for the `windows` requirement, but requiring the `windows` object for configs aimed at the [`windows` platform][4] will still be something you want.

[2]: https://github.com/opencontainers/runtime-spec/blob/49b0a1fa5f2b6f05195305ebe62d59444ffd73e8/config.md#platform-specific-configuration
[3]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303463569
[4]: https://github.com/opencontainers/runtime-spec/blob/49b0a1fa5f2b6f05195305ebe62d59444ffd73e8/spec.md#platforms

sunjayBhatia: Updated `config.md` to specify that the `windows` configuration section is required if the `windows` platform is set. We can update this again if #830 lands before this PR.

config.md
```diff
@@ -328,7 +328,7 @@ Runtime implementations MAY support any valid values for platform-specific field
 * **`linux`** (object, OPTIONAL) [Linux-specific configuration](config-linux.md).
   This MAY be set if **`platform.os`** is `linux` and MUST NOT be set otherwise.
 * **`windows`** (object, OPTIONAL) [Windows-specific configuration](config-windows.md).
-  This MAY be set if **`platform.os`** is `windows` and MUST NOT be set otherwise.
+  This MUST be set if **`platform.os`** is `windows` and MUST NOT be set otherwise.
```
wking: I'm not entirely clear on the maintainers' preference for this (@mrunalp [was going to document them][1]), but it's possible that they'll want `*Windows` changed to `Windows` [in the Go structure][2] now that this property is required on one platform.  At least, something like that was the reason given for not wanting pointers for [`User.UID` and `User.GID`][3].

[1]: https://github.com/opencontainers/runtime-spec/pull/759#issuecomment-301199323
[2]: https://github.com/opencontainers/runtime-spec/blob/837ee7659b0c115bb32598e4454a1f6d2460b1ea/specs-go/config.go#L29
[3]: https://github.com/opencontainers/runtime-spec/pull/759/files#diff-7f24d60f0cbb9c433e165467e3d34838L87

sunjayBhatia: Thats an interesting point. Will it be a little odd that the other platform specific fields are not pointers as well? I would expect the maintainers would like to keep these fields uniform.

tianon:
LGTM

specs-go/config.go
```diff
@@ -430,6 +430,8 @@ type SolarisAnet struct {
 
 // Windows defines the runtime configuration for Windows based containers, including Hyper-V containers.
 type Windows struct {
+	// LayerFolders contains a list of absolute paths to directories containing image layers.
+	LayerFolders []string `json:"layerFolders"`
```
jhowardmsft: Just spotted this. Should be all lower-case for consistency. And the config-windows.json schema file needs the matching update.

wking: > Should be all lower-case for consistency.

There are currently some camelCase entries in `config-windows.md`.  For example, [the `network` properties][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/2ec18c02474dd49ce20aa30236959a8a80763042/config-windows.md#L84-L87

jhowardmsft: @darrenstahlmsft Could you open a PR to fix ^^

sunjayBhatia: @wking @jhowardmsft i modified the field name as per [this suggestion](https://github.com/opencontainers/runtime-spec/pull/828#discussion_r117719673), should I revert it back?

It looks like [there are linux fields that use camel case as well](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#user-namespace-mappings).

wking: > It looks like there are linux fields that use camel case as well.

The Linux config tries to consistently use camelCase.  I don't know if the maintainers care what convention the Windows-only fields use; probably wait for them to weigh in.

jhowardmsft: OK, lets leave this as is. I'll take an action item to clean up the casing once this is merged. Opened https://github.com/opencontainers/runtime-spec/issues/857 to track.

jhowardmsft:
@wking - Are you OK with this now?

wking:
On Thu, May 25, 2017 at 10:58:34AM -0700, John Howard wrote:
> @wking - Are you OK with this now?

I'm not familiar with the underlying tech, so I can't comment on that.

I've just been pushing for internal consistency, and that looks pretty
good to me with 284ab58e.  My only outstanding concern is the
discrepancy between pointer policy for windows and process.user.uid
[1], but Go types aren't part of the spec so we don't have to resolve
that discrepancy before landing this PR.

[1]: https://github.com/opencontainers/runtime-spec/pull/828#discussion_r118158666

jhowardmsft:
ping @mrunalp @vbatts @crosbymichael Could one of you review please. Thanks. 👍

jhowardmsft:
I should note this is the last of the structural missing pieces in Windows for 1.0. I have proved now that the docker engine using the current spec plus this PR is fully functional.

mrunalp:
LGTM

TomSweeneyRedHat:
LGTM

jhowardmsft:
@mrunalp @tianon Can we merge? Has all the right number of "LGTM"s :smile:

jhowardmsft:
Thanks @Tianon  ❤️

Issue #725, opened 2017-03-14
spec: Unclear scope for platform section


wking:
In opencontainers/runtime-tools#347, @Mashimiao was reading the [platform section][1] as a restriction on [`platform.os`][2].  Something like:

> `platform.os` may have permissive SHOULD language for valid values, but you can ignore that.  Valid configs MUST be using one of these…

I read [the `spec.md` section][1] as a list of certification tracks:

> Runtime-tools will have the following certification tracks: ….  `platform.os` allows many other values (e.g. `android`), and valid configs MAY use those (with questionable usefulness), but there is currently no certification track for them.

The spec should probably be updated to make it very clear which of these interpretations is intended.

With the restrictive approach (where `platform.os == "android"` was illegal), we should be using something like:

```
* **`os`** (string, REQUIRED) specifies the operating system family of the container configuration's specified [`root`](#root) file system bundle.
  The runtime MUST generate an error if it does not support the specified **`os`**.
  Bundles MUST use `linux`, `solaris`, or `windows`.
```

With the permissive approach, we should update the [spec section][1] to make it clear that it does not restrict `platform.os`.  And we probably want some notes about what happens in cases where `platform.os` does not match any of the certification tracks, even if that is “these other configs are possible, but because they have no platform-specific configuration settings, we don't think they are useful enough to be worth a certification track”.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#platforms
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#platform

Mashimiao:
@opencontainers/runtime-spec-maintainers  any comments?

wking:
Any thoughts?  I think clarifying the intention here will help with language if we decide to [“preview” Windows][1] (and Solaris?) in 1.0.

[1]: https://github.com/opencontainers/runtime-spec/issues/817#issuecomment-301228367

PR #830, opened 2017-05-16
Expand platform to be more sophisticated

hqhq:
Currently support is not sufficient for many platforms,
such as ARM, we need more information to identify a
specific platform.

Image-spec already had these entries, so I just picked
them up and applied to runtime-spec.

### DIFF
--- a/config.md
+++ b/config.md
@@ -310,13 +310,18 @@ For Windows based systems the user structure has the following fields:
   The runtime MUST generate an error if it does not support the specified **`arch`**.
   Values for **`arch`** SHOULD use, and runtimes SHOULD understand, **`arch`** entries listed in the Go Language document for [`GOARCH`][go-environment].
   If an architecture is not included in the `GOARCH` documentation, it SHOULD be submitted to this specification for standardization.
+* **`os.version`** (string, OPTIONAL) specifies the operating system version, for example `10.0.10586`.
+* **`os.features`** (array of strings, OPTIONAL) specifies an array of strings, each specifying a mandatory OS feature, for example `win32k`.
+* **`variant`** (string, OPTIONAL) specifies the variant of the CPU, for example `v8` to specify a perticular CPU variant of the ARM CPU.
+* **`features`** (array of strings, OPTIONAL) this property is RESERVED for future versions of the specification.
 
 ### Example
 
 ```json
 "platform": {
     "os": "linux",
-    "arch": "amd64"
+    "arch": "amd64",
+    "variant": "v8"
 }
 ```
 
@@ -338,7 +343,8 @@ Runtime implementations MAY support any valid values for platform-specific field
 {
     "platform": {
         "os": "linux",
-        "arch": "amd64"
+        "arch": "amd64",
+        "variant": "v8"
     },
     "linux": {
         "namespaces": [
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -108,6 +108,14 @@ type Platform struct {
 	OS string `json:"os"`
 	// Arch is the architecture
 	Arch string `json:"arch"`
+	// OSVersion specifies the operating system version, for example `10.0.10586`.
+	OSVersion string `json:"os.version,omitempry"`
+	// OSFeatures specifies an array of strings, each specifying a mandatory
+	// OS feature, for example `win32k`.
+	OSFeatures []string `json:"os.features,omitempry"`
+	// Variant specifies the variant of the CPU, for example `v8` to specify
+	// a perticular CPU variant of the ARM CPU.
+	Variant string `json:"variant,omitempry"`
 }
 
 // Mount specifies a mount for a container.
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -52,7 +52,22 @@
                 "os": {
                     "id": "https://opencontainers.org/schema/bundle/platform/os",
                     "type": "string"
-                }
+                },
+                "os.version": {
+                    "id": "https://opencontainers.org/schema/bundle/platform/os.version",
+                    "type": "string"
+                },
+                "os.features": {
+                    "id": "https://opencontainers.org/schema/bundle/platform/os.features",
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
+                "variant": {
+                    "id": "https://opencontainers.org/schema/bundle/platform/variant",
+                    "type": "string"
+		}
             }
         },
         "root": {


config.md
```diff
@@ -310,13 +310,18 @@ For Windows based systems the user structure has the following fields:
   The runtime MUST generate an error if it does not support the specified **`arch`**.
   Values for **`arch`** SHOULD use, and runtimes SHOULD understand, **`arch`** entries listed in the Go Language document for [`GOARCH`][go-environment].
   If an architecture is not included in the `GOARCH` documentation, it SHOULD be submitted to this specification for standardization.
+* **`os.version`** (string, OPTIONAL) specifies the operating system version, for example `10.0.10586`.
+* **`os.features`** (array of strings, OPTIONAL) specifies an array of strings, each specifying a mandatory OS feature, for example `win32k`.
+* **`variant`** (string, OPTIONAL) specifies the variant of the CPU, for example `v8` to specify a perticular CPU variant of the ARM CPU.
+* **`features`** (array of strings, OPTIONAL) specifies an array of strings, each specifying a mandatory CPU feature, for example `sse4` or `aes`.
```
AkihiroSuda: This is being removed https://github.com/opencontainers/image-spec/pull/672

wking:
On Mon, May 15, 2017 at 11:02:28PM -0700, Qiang Huang wrote:
> Image-spec already had these entries, so I just picked them up and
> applied to runtime-spec.

I'm not excited about having these properties documented in two
different repositories (with an intention to keep them synchronized?).
Do you file a new runtime-spec PR if/when
opencontainers/image-spec#652 lands?  Do we need to synchronize
releases so that there is at least one image-spec release with the
same language as a given runtime-spec release (and vice versa)?

The image and runtime use-cases are a bit different.  For image-spec,
this is registry information, allowing folks to say:

  I want a version of image FOO that can run on PLATFORM.

For runtime-spec, the use case is presumably:

  I'm not sure if my runtime will execute this container on my current
  host or some other platform (e.g. via an internal HTTP API).  If the
  runtime chooses to launch the container on another host, I need it
  to pick one like this.

You probably need the same metadata for both use-cases, but it's not
clear to me if we need to support that runtime-spec use-case out of
the box.  Maybe we should drop ‘platform’ instead and point folks who
need that workflow towards ‘annotations’.

config.md
stevvooe: Yes, please remove this.

hqhq: I see why you remove this field, it's a fair point to me, I'll update. Thanks.

stevvooe:
Should runtime-spec adopt the platform specification from the image-spec? Are these even used in the runtime spec?  What would be the impact of making that change? I think the only addition is to have `architecture` instead of `arch`.

crosbymichael:
Maybe we remove this from runtime spec as we do not consume it anywhere nor intended to when we initially added the fields?  We could leave this as an image-spec concern.

wking:
> Maybe we remove this from runtime spec as we do not consume it anywhere…

We currently consume it [for validation][1], which is one instance of “[the config you're looking at may target a different platform than your current host][2]”.  Without `platform.os`, you don't know if `platform.apparmorProfile` is legal or not.  But as long as there is some way to select your [compliance track][3], I'm fine dropping the rest of `platform`.

[1]: https://github.com/opencontainers/runtime-tools/blob/f49d861b69e359dc9278fcfa6948034a4af58705/validate/validate.go#L279-L282
[2]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-301841186
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#platforms

tianon:
I'd definitely be +1 to punting the `platform` object to the image-spec -- maintaining that list in one place makes a lot of sense (and it's definitely more relevant to the image-spec).

hqhq:
Are we talking about keep `Platform` as is or remove it from runtime-spec?

crosbymichael:
Ya, talking about removing it.  I remember when we added it we were never going to use it in the runtime, it was an informational field.  But this was before the image spec.  If we don't have any real usecases for a runtime to consume it, its worth considering removing it.

cyphar:
At the very least we should make it an annotation.

mattspencer-arm:
I agree with the sentiment of having this defined in one place.  Currently the image-spec references the runtime spec for the definition of platform.architecture and platform.os.  I would suggest as above removing platform from the runtime-spec and having the image spec being the canonical definition.

dqminh:
> Ya, talking about removing it. I remember when we added it we were never going to use it in the runtime, it was an informational field. But this was before the image spec. If we don't have any real usecases for a runtime to consume it, its worth considering removing it.

`platform.os` should at least be useful in config validation though ? Like the current way we are using / wording it

```
[**`platform.os`**](#platform) is used to lookup further platform-specific configuration.
```

wking:
On Tue, May 23, 2017 at 08:01:41AM -0700, Daniel, Dao Quang Minh wrote:
> [**`platform.os`**](#platform) is used to lookup further
>  platform-specific configuration.

It's used for more than lookup though.  For example, both whether
root.path is \ or / delimited and whether ‘proc’ has a chance at being
is a valid mounts[].type depend on the platform.  I suggest we handle
this (and close off #725) with something like:

* `platform` (string, REQUIRED) the platform targetted by this configuration.
    The value MUST be a slug from [the platform list](spec.md#platforms).

crosbymichael:
@dqminh either that or on validation you just pass in the platform you want to validate for.

```bash
runtime-tools validate --platform linux <config.json>
```

It would actually be more robust for validation that you validate what you expect, not what the spec says it is. 


@cyphar they are annotations, you can put whatever you want there, that's way its a `map[string]string`

wking:
> … on validation you just pass in the platform you want to validate for.

I don't see a robustness difference, but I'm fine with this approach (and completely dropping `platform` from the config) too.

specs-go/config.go
```diff
@@ -108,6 +108,14 @@ type Platform struct {
 	OS string `json:"os"`
 	// Arch is the architecture
 	Arch string `json:"arch"`
+	// OSVersion specifies the operating system version, for example `10.0.10586`.
```
jhowardmsft: I'd pick a more up to date build. And it's actually a quad-tuple these days (it was a triplet). eg `10.0.14393.1198` is the current fully-patched Windows Server 2016 RS1 build number.

jhowardmsft: (Historical - 10586 was a pre-release, IIRC Technical Preview 4 of Windows Server 2016)

wking: > I'd pick a more up to date build.

My take on the discussion is that [runtime-spec will likely drop `platform`][1] or [keep only `platform.os`][2].  Either way, this line is unlikely to land in runtime-spec, so we probably don't need to worry about it.  And the image-spec equivalent is getting , so I doubt we need to worry about this line.

And the image-spec analog got a quad-touble in the recently-landed opencontainers/image-spec#651.

[1]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303463569
[2]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303426610

hqhq:
Thanks for all your comments, I think we have conclusion not to expand platform and sync with image-spec, I opened a new PR to simply it in #850 , I'll close this PR.

stevvooe:
> I think we have conclusion not to expand platform and sync with image-spec, 

How did we arrive at this conclusion? #850 is pretty insufficient.

I would favor just removing this altogether rather than make this just a string.

tianon:
@stevvooe and then use the per-platform objects to differentiate for validation? (and update them to all be mutually-exclusive directly, rather than mutually-exclusive based on the value of `platform.os`)

ie, if the `windows` property is set, the `linux` and `solaris` properties MUST NOT be, etc (rather than the current language which hinges around `platform.os`)

tianon:
If I don't need any of the platform-specific properties for a given bundle, isn't that like saying that the bundle doesn't care which platform it runs on?  It seems safe enough to me, especially since we already don't make any guarantees about the "runnability" of the binaries in the rootfs (since that's really more of an image-spec / distribution concern).  I don't see any reason to need to _require_ one of them, but making them mutually exclusive is probably a good idea.

wking:
On Wed, May 24, 2017 at 03:00:27PM -0700, Tianon Gravi wrote:
> If I don't need any of the platform-specific properties for a given
> bundle, isn't that like saying that the bundle doesn't care which
> platform it runs on?

You might still need platform-specific properties
(e.g. process.user.uid and soon mounts[].type, #854) without needing
any of the platform-specific properties under ‘linux’.  If ‘linux’ is
not set (with your approach, or ‘"platform": "linux" is not set with
the #850 approach, or ‘--platform linux’ is not set with
@crosbymichael's approach [1]), then the validation tooling has no way
to validate:

  {
    "ociVersion": "1.0.0",
    "root": {
      "path": "rootfs"
    },
    "process": {
      "cwd": "/",
      "args": [
        "sh"
      ],
      "user": {
        "uid": 0,
        "gid": 0
      }
    }
  }

Is that a valid Windows config (which happens to set the
undefined-for-Windows process.user.uid and process.user.gid)?  Is that
a valid Linux/Solaris config?  I think it's better to require some
explicit declaration of the intended compliance track [2] to avoid the
complication of guessing based on which properties are set and what
they're set to.

[1]: https://github.com/opencontainers/runtime-spec/pull/830/#issuecomment-303463569
[2]: https://github.com/opencontainers/runtime-spec/blob/2ec18c02474dd49ce20aa30236959a8a80763042/spec.md#platforms

wking:
On Wed, May 24, 2017 at 03:11:33PM -0700, W. Trevor King wrote:
> Is that a valid Windows config (which happens to set the
> undefined-for-Windows process.user.uid and process.user.gid)?

This will no longer be an option once #828 lands and starts requiring
‘windows’ for all Windows configs [1]:

  This MUST be set if **`platform.os`** is `windows` and MUST NOT be
  set otherwise.

In that case, the issue will be distinguishing between Solaris and
Linux compliance tracks, and there's not a lot of separation there at
the moment outside of the ‘linux’ and ‘solaris’ prorperties.  However,
that seems like a brittle condition to hang sane compliance testing
on.  For example, it would be impossible to violate the new-in-#828
‘windows’ condition, even if you *intended* to target ‘windows’ but
forgot to set that property.  An explicit compliance track declaration
is harder to muff.

[1]: https://github.com/opencontainers/runtime-spec/pull/828/files#diff-c9c91c29b41257aea3a3403cc606ad99L331

Siyler1 (approved):

Issue #806, opened 2017-05-11
config-linux: Move 'disableOOMKiller' under 'memory'?


wking:
It's backed by `memory.oom_control` (see the in-flight #745), so I expect it should go with the rest of the [memory-controller config][0].  I expect consistency in these groupings will be useful as we work through #746.

Looking at the history, the initial request landing a setting for this in the Docker/OCI ecosystem seems to be docker/libcontainer#417, which added `Cgroup.OomKillDisable`.  That commit was carried from libcontainer into runC (opencontainers/runc@295c7086) where it is now [`Resources.OomKillDisable`][1].  From runC it was carried into this repo (with some renaming) in #51.  Subsequent early doc updates landed in #137 and #199.  In none of those can I find discussion about why the setting is not already under `memory`.  I expect the reason is that the runC structures are flat, so “under `memory`” is not a thing there.  But in this spec, `resources` has per-controller sub-properties.  The fact that `disableOOMKiller` belonged to the `memory` controller may have been overlooked in #51 and never revisited until this issue.

I'm happy to submit a PR if the move sounds appropriate.

[0]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#memory
[1]: https://github.com/opencontainers/runc/blob/2daa11574b12d2dd7f6dad3ffaa302871f21bd12/libcontainer/configs/cgroup_unix.go#L114

wking:
If we want this before 2.0, it needs to go in the [1.0.0 milestone][1].  If we don't want this before 2.0, and are ok splitting the memory controller across multiple `linux.resources` entries, then maintainers should close this issue with a note to that effect.

[1]: https://github.com/opencontainers/runtime-spec/milestone/7

PR #834, opened 2017-05-17
Clarify cgroups path handling behavior

mrunalp:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -183,13 +183,15 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
 The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
-The cgroups will be created if they don't exist.
 
-You can configure a container's cgroups via the `resources` field of the Linux configuration.
-Do not specify `resources` unless limits have to be updated.
+The runtime MUST create the cgroups specified by the `cgroupsPath` if they don't exist.
+If `cgroupsPath` is empty, then the behavior is runtime implementation specific.
+
+The runtime MUST ensure that the container process is attached to the cgroups specified by `cgroupsPath`.
+If any property is set under `resources` then the runtime MUST set it for the container.
+
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
-Runtimes MAY attach the container process to additional cgroup controllers beyond those necessary to fulfill the `resources` settings.
 
 ### Example
 


config-linux.md
```diff
@@ -183,13 +183,16 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
 Implementations of the Spec can choose to name cgroups in any manner.
 The Spec does not include naming schema for cgroups.
 The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
-The cgroups will be created if they don't exist.
 
-You can configure a container's cgroups via the `resources` field of the Linux configuration.
-Do not specify `resources` unless limits have to be updated.
+The runtime MUST create the cgroups specified by the `cgroupsPath` if they don't exist.
```
wking: Which cgroups?  Only those needed to fulfil `resources`, no?  Or MUST they create a new `devices` cgroup even if there is nothing in `linux.resources.devices`?

config-linux.md
```diff
@@ -183,13 +183,16 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
...
+If `cgroupsPath` is empty, then the behavior is runtime implementation specific.
```
wking: I think you mean “[implementation-defined][1]”.

And I liked the old language (where [the runtime could pick a default `cgroupsPath`][2] much more than the huge hole you poke with an unqualified “implementation-defined”.  The runtime could drive all sorts of crazy things unrelated to cgroups through this big loophole.

[2]: https://github.com/opencontainers/runtime-spec/blame/30362737548149b000128dbc059d7e2b13d1032e/config-linux.md#L179

config-linux.md
```diff
@@ -183,13 +183,16 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
...
+
+The runtime MUST ensure that the container process is attached to the cgroups specified by `cgroupsPath`.
+If any property is set under `resources` then the runtime MUST set it for the container.
+Check individual properties for any specific handling.
```
wking: Do we need this line?  It sounds like “and don't forget to read the rest of the spec!” :p.

config-linux.md
```diff
@@ -183,13 +183,16 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
...
+
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
-Runtimes MAY attach the container process to additional cgroup controllers beyond those necessary to fulfill the `resources` settings.
```
wking: I don't think we want to drop this line.  I think the additional attaching is surprising, so if we don't forbid runtimes from doing it (which I don't think we can with runC), I'd like to keep mentioning this as an runtime prerogative.

tianon:
LGTM

config-linux.md
wking: To tie this back into its original motivation, the line you're removing here is descended from @cyphar's 4291fd1d5, which was part of #493.  There is a bit of related discussion [here][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/493#r66303914

config-linux.md
vbatts: i think this new sentence is fine and not any more of a "loophole" than the old language

config-linux.md
vbatts: fair. this sentence is a bit ambiguous.

config-linux.md
vbatts: i think this sentence can stay

config-linux.md
wking: > i think this new sentence is fine and not any more of a "loophole" than the old language

With the new language, config authors should not leave `cgroupsPath` empty unless they have read the implementation-specific docs for any runtime that might consume their config.  With the old language… I'm not sure.  It looks like this PR (as of 81c9f3116) still has the “MAY define the default” line.  So what is this new line about?  Is it just underlining that the default may not be a fixed string (e.g. maybe the runtime defaults to the container ID), so if you leave this unset you might get new cgroup or might join an existing cgroup?  If so, I'd rather clarify that *that* is what is being implementation-defined.  The current language in 81c9f3116 supports runtimes which error out on unset `cgroupsPath` (or ignore `hooks`, etc., etc.), and I don't think we want to give that much leeway here. 

  With the old language, config authors could safely leave `cgroupsPath` empty if they didn't care

config-linux.md
mrunalp: They must create all for the use case of orchestrator preparing the cgroups for you and just be asking the runtime to add the container process to the cgroups.

config-linux.md
mrunalp: Sure, I can take this line out.

config-linux.md
mrunalp: Given my comment above, we would want the runtime to join all the cgroups specified by the cgroupsPath.

config-linux.md
wking: > They must create all for the use case of orchestrator preparing the cgroups for you and just be asking the runtime to add the container process to the cgroups.

How should the runtime distinguish between v1 and v2 cgroups (more on this [here][1])?  I think a safer approach for orchestrators would be to tell the runtime to leave cgroups alone and set them up completely in a hook (both creating/initializing any required cgroups and joining the runtime to them).  On the other hand, with #237 rejected, there's no clear way to warn the runtime that you'll be messing with cgroups in the hook, so the runtime might be surprised with:

1. No `cgroupsPath` or `resources`, so the runtime creates a new devices cgroup and adds the container process.
2. Orchestration hook (or other post-start action) creates the cgroups that the orchestrator wants and moves the container process into them.
3. The runtime looks in the devices cgroup it setup in step 1 and… where did that process go?

[1]: https://github.com/opencontainers/runtime-spec/issues/20#issuecomment-198512331

config-linux.md
wking: Removing this line is clearer in light of [this comment][1].  So I think this line should stay if runtimes remain free to *not* join all cgroups (known to the runtime?  Mentioned in the spec?), but should be dropped if we start requiring them to join all of those cgroups.

[1]: https://github.com/opencontainers/runtime-spec/pull/834#discussion_r117346486

config-linux.md
cyphar: > They must create all for the use case of orchestrator preparing the cgroups for you and just be asking the runtime to add the container process to the cgroups.

But should the runtime join any extra cgroups? The old language handled this case.

config-linux.md
cyphar: I too prefer the old language. What is the reason for this change @mrunalp?

config-linux.md
```diff
@@ -183,13 +183,15 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
...
 The Spec does not include naming schema for cgroups.
 The Spec does not support per-controller paths for the reasons discussed in the [cgroupv2 documentation][cgroup-v2].
-The cgroups will be created if they don't exist.
 
-You can configure a container's cgroups via the `resources` field of the Linux configuration.
-Do not specify `resources` unless limits have to be updated.
+The runtime MUST create the cgroups specified by the `cgroupsPath` if they don't exist.
+If `cgroupsPath` is empty, then the behavior is runtime implementation specific.
+
+The runtime MUST ensure that the container process is attached to the cgroups specified by `cgroupsPath`.
```
cyphar: In cgroupv1 (which we do support but not explicitly) it is not clear what cgroups are specified by a single path. Could you clarify what was wrong with the old language (which had similar semantics but was more explicit about how different controllers are handled).

Also with cgroupv2 it's not clear what `cgroup.controllers` should be set to in `cgroupsPath`. Should it be `+everything` or just `+whatever_is_necessary`? The former makes rootless containers no longer possible in an OCI runtime, while the latter should also allow for runtimes to join extra cgroups.

cyphar:
NACK, I'm not sure I understand the loosening of some of the semantics of `cgroupsPath` (we spent quite a while getting it to this point and I would prefer if we didn't go back :wink:). There's still some more discussion necessary IMO.

mrunalp:
@cyphar @wking Okay, let us consider the 4 cases and then comment on what we would want for each. We can then add/remove/keep the language: 
1. cgroupsPath NOTSET resources NOTSET: We have two options here: don't do anything or make it implementation-defined. Making it implementation defined, gives more flexibility so I am leaning towards that. 
2. cgroupsPath SET resources NOTSET: Considering the orchestrator use case, I was leaning towards joining all the cgroups controllers. We don't know what all resources the orchestrator has set so safest would be to join all. An option would be to additionally allow specifying what controllers should be joined. 
3. cgroupsPath NOTSET resources SET: Here we could have the runtime only join the controllers (at a path of its own choice) required to satisfy the resource requirements or all. I don't have a strong preference as this is least likely to be used in production. 
4. cgroupsPath SET resources SET: Again similar to option 2, we need to figure out whether runtime should join all controllers or only the ones necessary to satisfy the resources. Also, could have an option (like in 2) to specify the controllers explicitly (which would need to be validated as being a superset of controllers required to satisfy resources).

wking:
On Thu, May 18, 2017 at 05:47:26PM -0700, Mrunal Patel wrote:
> 1. cgroupsPath NOTSET resources NOTSET: We have two options here:
>    don't do anything or make it implementation-defined. Making it
>    implementation defined, gives more flexibility so I am leaning
>    towards that.

But it means that you might confuse runtimes that decide to put the
container process in a cgroup and expect it to stay there [1].  And
there may be problems with rootless containers [2] if the runtime
decides to error out if it can't accomplish whatever it would like to
do in this case.  My preference is revive #237 in this case, which was
designed to make it easy for folks to experiment with alternative
cgroup managers.  Lots more discussion on this front in [3].

> 2. cgroupsPath SET resources NOTSET: Considering the orchestrator
>    use case, I was leaning towards joining all the cgroups
>    controllers. We don't know what all resources the orchestrator
>    has set so safest would be to join all. An option would be to
>    additionally allow specifying what controllers should be joined.

“Join all the cgroups” still suffers from “v1 or v2?” and “including
ones the runtime may not be aware of?” (e.g. /sys/fs/cgroup/systemd?)
ambiguities.  It's nice to have a way to join existing cgroups for the
‘exec’ and sub-container use-cases, but as long as v1 cgroups are
available and cases 3 and 4 are ambiguous, I'm not sure how to
accomplish that.

> 3. cgroupsPath NOTSET resources SET: Here we could have the runtime
>    only join the controllers (at a path of its own choice) required
>    to satisfy the resource requirements or all.

I would prefer this (easier to succeed for the rootless case) and this
is what the master wording already encourages.

> 4. cgroupsPath SET resources SET: Again similar to option 2, we need
>    to figure out whether runtime should join all controllers or only
>    the ones necessary to satisfy the resources. Also, could have an
>    option (like in 2) to specify the controllers explicitly (which
>    would need to be validated as being a superset of controllers
>    required to satisfy resources).

I think “has that cgroup property set” is sufficient for explicitly
wanting to join the cgroup.  So “I'd like to join a devices cgroup but
not tweak it” would be:

  "linux": {"resources": {"devices": {}}}

possibly including a cgroupsPath.  As long as we collect properties by
cgroup (we'd want to address #806), that gives a clear way to request
joining cgroups known to the spec.  There would be no way to request
to join cgroups not known to the spec (e.g. a systemd cgroup), but you
can handle that in hooks.  And if the runtime decided to join some
oddball cgroup as well, then it's up to the runtime to not get
surprised by [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/834#discussion_r117349177
[2]: https://github.com/opencontainers/runtime-spec/pull/834#discussion_r117382470
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/qWHoKs8Fsrk

wking:
I feel like “what should happen in cgroup case $X?” for a number of
cases is a discussion that could go on for a while (and has been going
on for about as long as there has been a runtime spec).  If the
current master wording is not clear or ideal, and with rc6 and a
feature freeze imminent, I think we may be in the same boat on this
subject as the Windows spec discussed in #817.  The easy out is to
provide a way to say “please leave the cgroups alone; I'll handle them
myself” in the config (whether that's a revived #237, some other
property combination, or a new property in its own right).  Then users
concerned about portability (if the spec is not clear enough on the
cgroup specification, #746) or with use cases not addressed by the
currently specified cgroup config have a portable way to use their own
cgroup manager.  We can cut rc6 and 1.0 with that, and continue the
discussion about what should happen in any other cases without holding
up the 1.0 release.  If subsequent discussion on whether cgroup
joining is greedy or not settles down close to whatever goes out with
rc6/1.0, it may be able to go out as 1.1 without the breaking-change
2.0 bump.

vbatts:
Closes: #745

mrunalp:
I'll remove line 188 and add more language to clarify what cgroups to join.

hqhq:
@mrunalp What exactly issue does this PR try to fix? Seems that there are still some concerns not clarified, can this be moved to post 1.0?

config-linux.md
```diff
@@ -183,13 +183,15 @@ If a `cgroupsPath` value is specified, the runtime MUST consistently attach to t
...
+If any property is set under `resources` then the runtime MUST set it for the container.
+
 For example, to run a new process in an existing container without updating limits, `resources` need not be specified.
 
-Runtimes MAY attach the container process to additional cgroup controllers beyond those necessary to fulfill the `resources` settings.
```
hqhq: What about controllers not specified in `cgroupsPath` nor `resources`? This line was supposed to cover that case.

mrunalp:
@hqhq I am fine closing this for now.

PR #809, opened 2017-05-12
config.md: minor changes for process

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #835, opened 2017-05-18
config: Make capabilities, noNewPrivileges, and rlimits Linux-only (again)

wking:
This is the alternative to the recently-closed #810.

Roll back the genericization from #673.  Lifting the restriction there seems to have been motivated by “Solaris supports capabilities”, but that was before the split into a capabilities object which happened in #675.  It's not clear if Solaris supports ambient caps, or [what Solaris API `rlimits` or `noNewPrivileges` were punting to][1].  @jhowardmsft has recently [confirmed that Windows does not support capabilities and is unlikely to do so in the future][2].  John's statement didn't directly address `rlimits` or `noNewPrivileges`, but we can always restore any of these properties to the Solaris/Windows platforms if/when we get docs about which API we're punting to on those platforms.

Also add some backticks, remove the hyphens in `OPTIONAL) - the`, standardize lines I touch to [use “the process”][3], and use four-space indents here to keep Pandoc happy (see #495).

This PR will conflict with #809, and I'm happy to rebase if that one lands first.

[1]: https://github.com/opencontainers/runtime-spec/pull/673#discussion_r99353136
[2]: https://github.com/opencontainers/runtime-spec/pull/810#issuecomment-301594590
[3]: https://github.com/opencontainers/runtime-spec/pull/809#discussion_r116297660

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -47,7 +47,7 @@ type Process struct {
 	// Capabilities are Linux capabilities that are kept for the process.
 	Capabilities *LinuxCapabilities `json:"capabilities,omitempty" platform:"linux"`
 	// Rlimits specifies rlimit options to apply to the process.
-	Rlimits []LinuxRlimit `json:"rlimits,omitempty" platform:"linux"`
+	Rlimits []POSIXRlimit `json:"rlimits,omitempty" platform:"linux,solaris"`
 	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container.
 	NoNewPrivileges bool `json:"noNewPrivileges,omitempty" platform:"linux"`
 	// ApparmorProfile specifies the apparmor profile for the container.
@@ -214,8 +214,8 @@ type LinuxIDMapping struct {
 	Size uint32 `json:"size"`
 }
 
-// LinuxRlimit type and restrictions
-type LinuxRlimit struct {
+// POSIXRlimit type and restrictions
+type POSIXRlimit struct {
 	// Type of the rlimit to set
 	Type string `json:"type"`
 	// Hard is the hard limit for the specified type
--- a/config.md
+++ b/config.md
@@ -145,17 +145,33 @@ For all platform-specific configuration values, the scope defined below in the [
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2001's `environ`][ieee-1003.1-2001-xbd-c8.1].
 * **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2001 `execvp`'s *argv*][ieee-1003.1-2001-xsh-exec].
     This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
-* **`rlimits`** (array of objects, OPTIONAL) allows setting resource limits for a process inside the container.
+
+### <a name="configLinuxAndSolarisProcess" />Linux and Solaris Process
+
+For POSIX-based systems (Linux and Solaris), the `process` object supports the following process-specific properties:
+
+* **`rlimits`** (array of objects, OPTIONAL) allows setting resource limits for the process.
     Each entry has the following structure:
 
-    * **`type`** (string, REQUIRED) - the platform resource being limited, for example on Linux as defined in the [setrlimit(2)][setrlimit.2] man page.
-    * **`soft`** (uint64, REQUIRED) - the value of the limit enforced for the corresponding resource.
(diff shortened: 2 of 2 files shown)


dqminh:
Looks conflicted with https://github.com/opencontainers/runtime-spec/pull/845/files ( rlimits is not linux-specific )

The changes to capabilities and nonewprivileges looks sane though.

wking:
On Tue, May 23, 2017 at 07:43:14AM -0700, Daniel, Dao Quang Minh wrote:
> Looks conflicted with
> https://github.com/opencontainers/runtime-spec/pull/845/files (
> rlimits is not linux-specific )

Do you have Solaris and/or Windows docs for what API they'll use to
implement rlimits?  #845 looks like a subset of the rejected #810,
although as I mentioned in the topic post for this PR, I couldn't find
anyone speaking directly to rlimits support in Windows/Solaris.

jhowardmsft:
> Do you have Solaris and/or Windows docs for what API they'll use to
> implement rlimits?

`rlimits` are meaningless on Windows certainly.

wking:
On Tue, May 23, 2017 at 09:23:37AM -0700, John Howard wrote:
> > Do you have Solaris and/or Windows docs for what API they'll use to
> > implement rlimits?
> 
> `rlimits` are meaningless on Windows certainly.

So I can reroll this to make rlimits Linux or Solaris, and just punt
to POSIX for the Solaris values until we hear from some Solaris person
with Solaris-specific docs?

jhowardmsft:
I can only comment on the windows part, but makes sense to me....

wking:
Ok, I've rerolled this with 550db9f → b4d7fae, which does a few things makes `rlimits` Solaris/Linux-specific, adjusting from master where `rlimits` is also nominally legal on Windows.

It also renames the Go `LinuxRlimit` to `POSIXRlimit`, although I'm fine leaving the type `LinuxRlimit` or following #845 and using the generic `Rlimit` (since Windows will hopefully use similar semantics if they add a new feature and decide to call it “rlimit”).

When adding the POSIX link for rlimits, I also added wording for RFC 2119 compliance testing, styled after the approach I'd floated for devices in #829 and some wording we already use for valid `capabilities` values.

If we want to focus on *just* the capabilities change, I'm happy to spin off the first commit into its own PR, and we can get back to this PR and the rlimits change if/when that cap PR lands.

wking:
This is a breaking change for Solaris and Windows (removing
process.capabilities for those configs) and allows Windows runtimes
(which have no API to implement the
currently-required-in-master-for-them process.capabilities) and
Solaris runtimes (which may not have an API to implement the
currently-required-in-master-for-them process.capabilities.ambient) to
be certified compliant.  I think that's sufficient grounds to
milestone it 1.0 freeze [1] unless we preview those platforms [2].

[1]: https://github.com/opencontainers/runtime-spec/milestone/15
[2]: https://github.com/opencontainers/runtime-spec/issues/817#issuecomment-301228367

specs-go/config.go
```diff
@@ -215,8 +215,8 @@ type LinuxIDMapping struct {
 	Size uint32 `json:"size"`
 }
 
-// LinuxRlimit type and restrictions
-type LinuxRlimit struct {
+// POSIXRlimit type and restrictions
+type POSIXRlimit struct {
```
jhowardmsft: Wondering if this should be moved out of the bunch of Linux structures which surround it so that there are POSIX, Linux, Solaris and Windows sections grouped together. Otherwise LGTM (not a maintainer)

wking: > Wondering if this should be moved out of the bunch of Linux structures which surround…

I don't care and am happy to do whatever maintainers want for type ordering.

wking:
Rebased around #846 with b4d7fae → cb8df7b.

crosbymichael:
We should make this change less about linux && solaris and more about unix vs windows as most of the things we have generally apply to unixes and we only need to point out exceptions for windows

wking:
> We should make this change less about linux && solaris and more about unix vs windows…

I'll rebase this one after #838 lands, since I've covered the rest of the POSIX-ification there.

tianon:
I agree with @crosbymichael -- using "POSIX" to mean "not Windows" is a bit off the mark, especially since Windows is largely POSIX-compliant in many areas (or at least has been in the past).

+1 for Unix or Unix-like instead

vbatts:
@wking pls2rebase

vbatts:
no idea when @wking will be back to rebase this. Wednesday the 5th is drawing close ...

dqminh:
if @wking doesnt get to this tomorrow, i can carry this change.

dqminh:
I have opened #880 to replace this.

PR #789, opened 2017-05-09
config: Shift oomScoreAdj from linux.resources to process

wking:
The only discussion related to this is in #236 and #292, where the relationship between `oomScoreAdj` and `disableOOMKiller` is raised.  But since #137 resources has been tied to cgroups, and `oomScoreAdj` is not about cgroups. We [currently have][1]:

> You can configure a container's cgroups via the resources field of the Linux configuration.

In #782, I suggested we move the property from `linux.resources.oomScoreAdj` to `linux.oomScoreAdj` so config authors and runtimes don't have to worry about what `cgroupsPath` means if the only entry in `resources` is `oomScoreAdj`.  @crosbymichael [responded with][2]:

> If anything it should probably go on the process

So that's what this commit does.

Fixes #782.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config-linux.md#L185
[2]: https://github.com/opencontainers/runtime-spec/issues/782#issuecomment-299990075

PR #836, opened 2017-05-18
specs-go/config: Add a 'linux' tag to OOMScoreAdj

wking:
This should have happened in #789 as part of moving the property from a Linux-specific type to a cross-platform type.


mrunalp:
LGTM

crosbymichael:
LGTM

PR #838, opened 2017-05-18
config: Bump Hyper-V condition from root.path to root itself

wking:
Don't require users targetting Hyper-V to set an empty object (`"root": {}`).  This also avoids confusion about whether you can set `root.readonly` without setting `root.path` (you can't).

Following #820.

### DIFF
--- a/config.md
+++ b/config.md
@@ -36,7 +36,7 @@ For all platform-specific configuration values, the scope defined below in the [
 
     * On Windows, `path` MUST be a [volume GUID path][naming-a-volume].
 
-    * On Linux and Solaris, `path` is either an absolute path or a relative path to the bundle.
+    * On POSIX platforms, `path` is either an absolute path or a relative path to the bundle.
         For example, with a bundle at `/to/bundle` and a root filesystem at `/to/bundle/rootfs`, the `path` value can be either `/to/bundle/rootfs` or `rootfs`.
         The value SHOULD be the conventional `rootfs`.
 
@@ -45,7 +45,7 @@ For all platform-specific configuration values, the scope defined below in the [
 * **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false.
     * On Windows, this field MUST be omitted or false.
 
-### Example (POSIX)
+### Example (POSIX platforms)
 
 ```json
 "root": {
@@ -94,9 +94,9 @@ For all platform-specific configuration values, the scope defined below in the [
 ]
 ```
 
-### <a name="configLinuxAndSolarisMounts" />Linux and Solaris Mounts
+### <a name="configPOSIXMounts" />POSIX-platform Mounts
 
-For Linux and Solaris based systems the mounts structure has the following fields:
+For POSIX platforms the `mounts` structure has the following fields:
 
 * **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
   * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
@@ -195,9 +195,9 @@ For Linux-based systems the process structure supports the following process-spe
 
 The user for the process is a platform-specific structure that allows specific control over which user the process runs as.
 
-#### <a name="configLinuxAndSolarisUser" />Linux and Solaris User
+#### <a name="configPOSIXUser" />POSIX-platform User
 
-For Linux and Solaris based systems the user structure has the following fields:
+For POSIX platforms the `user` structure has the following fields:
 
 * **`uid`** (int, REQUIRED) specifies the user ID in the [container namespace](glossary.md#container-namespace).
 * **`gid`** (int, REQUIRED) specifies the group ID in the [container namespace](glossary.md#container-namespace).
@@ -348,9 +348,9 @@ For Windows based systems the user structure has the following fields:
 }
 ```
 
-## <a name="configHooks" />Linux and Solaris Hooks
+## <a name="configHooks" />POSIX-platform Hooks
 
-For Linux- and Solaris-based systems, the configuration structure supports `hooks` for configuring custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
+For POSIX platforms, the configuration structure supports `hooks` for configuring custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
 
 * **`hooks`** (object, OPTIONAL) MAY contain any of the following properties:
     * **`prestart`** (array of objects, OPTIONAL) is an array of [pre-start hooks](#prestart).
--- a/config.md
+++ b/config.md
@@ -26,18 +26,22 @@ For all platform-specific configuration values, the scope defined below in the [
 
 ## <a name="configRoot" />Root
 
-**`root`** (object, REQUIRED) specifies the container's root filesystem.
+**`root`** (object, OPTIONAL) specifies the container's root filesystem.
+    On Windows, for Windows Server Containers, this field is REQUIRED.
+    For [Hyper-V Containers](config-windows.md#hyperv), this field MUST NOT be set.
 
-* **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container.
-    The path is either an absolute path or a relative path to the bundle.
+    On all other platforms, this field is REQUIRED.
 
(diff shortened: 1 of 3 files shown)


specs-go/config.go
```diff
@@ -11,7 +11,7 @@ type Spec struct {
 	// Process configures the container process.
 	Process *Process `json:"process,omitempty"`
 	// Root configures the container's root filesystem.
-	Root Root `json:"root"`
+	Root *Root `json:"root,omitempty"`
```
wking: This may fall under the no-pointers-if-required-on-any-platform policy that @mrunalp [was going to document][1].  If so, I can drop the pointer (and `omitempty`?), but it would be nice to get that policy documented.

[1]: https://github.com/opencontainers/runtime-spec/pull/759#issuecomment-301199323

wking:
Rebased around #839 with c652c06 → de31fae and updated to link to `windows.hyperv` now that #818 has landed with specifics about when a config is for a “Hyper-V Container”.

wking:
Rebased around #849 with de31fae → b9b8a51.  Now instead of just making the requirement clearner I'm also addressing inconsistencies that we'd missed when reviewing #849 (e.g. the [`rootfs` suggestion][1] conflicts with the [volume GUID path requirement][2]).

[1]: https://github.com/opencontainers/runtime-spec/blame/e6c96895260bb9615a347971e0c658048f8ac9d3/config.md#L32
[2]: https://github.com/opencontainers/runtime-spec/blame/e6c96895260bb9615a347971e0c658048f8ac9d3/config.md#L34

wking:
Also, if we want the condition-shift to land before 2.0, either it
needs to land by 1.0 unless we preview Windows [1], because before
this commit Hyper-V containers MUST set ‘"root": {}’ and after it they
MUST NOT set ‘root’.  My impression is that the preview is unlikely to
happen, and I think this PR makes the Windows experience more
straightforward, so I recommend putting this PR in the 1.0 freeze
milestone [2].

Ping @jhowardmsft, @darrenstahlmsft, @sunjayBhatia, and other Windows
folks in case they want to weigh in.

[1]: https://github.com/opencontainers/runtime-spec/issues/817#issuecomment-301228367
[2]: https://github.com/opencontainers/runtime-spec/milestone/15

specs-go/config.go
sunjayBhatia: Does it make sense to have a `pointer-required-if-field-must-not-be-set-on-a-platform` policy instead?

In this case: keep `*Root` and `omitempty` because it `MUST NOT` be set for Hyper-V containers

wking: > Does it make sense to have a pointer-required-if-field-must-not-be-set-on-a-platform policy instead?

We already [have some policy wording for pointers][1], but they focus on input requirements.  I think we also want to consider output requirements with Go's default JSON serializer, and would recommend:

> Use `omitempty` on properties which  MAY be unset for a platform that uses the parent type.  Of `omitempty` properties, use pointers if the Go zero value is legal (so we can serialize `"uid": 0` and such) or if the property is a Go type (to work around golang/go#11939).

as a new hole we'd want to poke in the [current no-pointer preference][1].  That would mean that we'd want `omitempty` here (because `root` MUST NOT be set for Hyper-V containers, which passes “MAY be unset for a platform that uses the parent type”) and we'd want a pointer here (because property is a Go type).  However, it would also mean we'd want `omitempty` for `User.UID` and `User.GID` (because they MAY be unset on Windows, and Windows uses `User` too for `User.Username`) and we'd want pointers for them (because `0` is a valid ID), and those changes were rejected in #759.  So instead of proposing a pointer policy, I'm just trying to flag changes that may be impacted by the maintainers apparent policy and waiting for them to us how they want that case handled.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/style.md#optional-settings-should-not-have-pointer-go-types

wking:
Rebased around #846 with b9b8a51 → 8a3b20b.

crosbymichael:
@jhowardmsft Please review this one

jhowardmsft:
LGTM (not a maintainer)

config.md
```diff
@@ -26,18 +26,22 @@ For all platform-specific configuration values, the scope defined below in the [
...
+    On all other platforms, this field is REQUIRED.
 
-    * On Windows, for Windows Server Containers, this field is REQUIRED and MUST be specified as a [volume GUID path][naming-a-volume].
-      For Hyper-V Containers, this field MUST be omitted.
-    * On all other platforms, this field is REQUIRED.
+* **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
+
+    * On Windows, `path` MUST be a [volume GUID path][naming-a-volume].
+
+    * On Linux and Solaris, `path` is either an absolute path or a relative path to the bundle.
```
crosbymichael: Can you remove the linux && solaris here like we discussed in the call

wking: > Can you remove the linux && solaris here like we discussed in the call

Done with dfe9cb9 → de14115.

wking:
I've rebased around the just-landed #850 with de14115 → 589d2eb.

config.md
```diff
@@ -26,18 +26,22 @@ For all platform-specific configuration values, the scope defined below in the [
...
+    On all other platforms, this field is REQUIRED.
 
-    * On Windows, for Windows Server Containers, this field is REQUIRED and MUST be specified as a [volume GUID path][naming-a-volume].
-      For Hyper-V Containers, this field MUST be omitted.
-    * On all other platforms, this field is REQUIRED.
+* **`path`** (string, REQUIRED) Specifies the path to the root filesystem for the container.
+
+    * On Windows, `path` MUST be a [volume GUID path][naming-a-volume].
+
+    * On POSIX, `path` is either an absolute path or a relative path to the bundle.
```
tianon: "On POSIX" doesn't make any sense -- POSIX isn't a platform, it's a specification, and it's one that Windows is compliant with in many cases.

I'd perfer to see this say either "On Unix" or "On Unix-like platforms".

tianon: See also https://github.com/opencontainers/runtime-spec/pull/835#issuecomment-305576761 (from @crosbymichael).

wking: > "On POSIX" doesn't make any sense -- POSIX isn't a platform, it's a specification, and it's one that Windows is compliant with in many cases.

I've changed “On POSIX” to “On POSIX platforms” in 589d2eb → 767eeaf.

> I'd perfer to see this say either "On Unix" or "On Unix-like platforms".

What is the gap between “Unix-like platforms” and “POSIX platforms”?  “Unix-like platforms” seems too vague for a spec, and if we want a name for non-Windows that is distinct from POSIX, I think we need to define that name [when we define our compliance tracks][1].  I'd suggested that on the 1st, but @crosbymichael [prefered to punt on that for now][2].

Looking at the properties I'm associating with POSIX here:

* For `root.path`, POSIX platforms clearly do not include Windows, because [POSIX paths are slash-separated][3].
* `mounts[].type` is less clearly POSIX-specific (or even Unix-like-specific), since POSIX defines neither `mount(8)` nor `mount(2)`.  However, it's also unclear to me why Windows only supports volume GUID paths.  Is that a limitation of the Windows platform, or just a limitation of the current Windows runtime?
* `process.user.uid` and friends are POSIX-platform-specific, since numeric [user][4] and [group][5] IDs are defined by POSIX.  I'm not aware of numeric user/group IDs on Windows, but if it does support them, I'd expect Windows runtimes to support `process.user.uid`, etc. as well.
* `hooks` is less clearly POSIX-specific (or even Unix-like-specific), since these seem easy to implement regardless of your platform.  I suspect this is a limitation of the current Windows runtime, and not a limitation of the Windows platform, so tying it to POSIX platforms seems no worse than tying it to Unix-like platforms.  Maybe we want to leave this one as an explicit “Linux and Solaris Hooks” or “Hooks (not Windows)” or some such?

[1]: https://github.com/opencontainers/runtime-spec/blob/fd0b0b6f7505ba8d24ee1fe59e59ca969b26843a/spec.md#platforms
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-06-01-17.41.log.html#l-60
[4]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_436
[5]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_189

dqminh:
LGTM

mrunalp:
ping @tianon

vbatts:
(@tianon is on vacation :sparkles: )

LGTM

tianon:
Generally LGTM, although I still have a big eyebrow-raising at our use of POSIX to mean "non-Windows platforms".

PR #840, opened 2017-05-18
bundle: DRYer root.path entry

wking:
Instead of repeating the config.md conditions, just make the linkage more clear.  Now the chain is:

1. Bundles REQUIRE a `config.json` which is a bundle artifact.
2. If that `config.json` has a `root.path` entry (as specified in `config.md`), then add the referenced directory to the set of bundle artifacts.  The `config.md` requirements include “If defined, a directory MUST exist at the path declared by the field”.
3. Apply the “MUST all be present in a single directory” condition to all bundle artifacts.  I [don't like that direct-child restriction][1], but I'm not touching it in this commit.

So these are the same requirements as before this commit, but with less redundancy and fewer words.

[1]: https://github.com/opencontainers/runtime-spec/pull/469

### DIFF
--- a/bundle.md
+++ b/bundle.md
@@ -14,13 +14,7 @@ This includes the following artifacts:
     This REQUIRED file MUST reside in the root of the bundle directory and MUST be named `config.json`.
     See [`config.json`](config.md) for more details.
 
-2. <a name="containerFormat02" />A directory representing the root filesystem of the container.
-    On Windows, for Windows Server containers, this directory is REQUIRED.
-    For Hyper-V containers, it MUST be omitted.
-
-    On all other platforms, this field is REQUIRED.
-
-    If set, this directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
+2. <a name="containerFormat02" />container's root filesystem: the directory referenced by [`root.path`](config.md#root), if that property is set in `config.json`.
 
 When supplied, while these artifacts MUST all be present in a single directory on the local filesystem, that directory itself is not part of the bundle.
 In other words, a tar archive of a *bundle* will have these artifacts at the root of the archive, not nested within a top-level directory.


wking:
Rebased around #839 with 6cedef2 → 22836ee.

crosbymichael:
Please rebase

wking:
> Please rebase

Rebased around #846 with 22836eec → b420c96c.

The Travis error looks like a [network hiccup][1].  If you want that green you can probably just kick the build again.

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/238446813#L213

bundle.md
```diff
@@ -14,13 +14,7 @@ This includes the following artifacts:
...
     See [`config.json`](config.md) for more details.
 
-2. <a name="containerFormat02" />A directory representing the root filesystem of the container.
-    On Windows, for Windows Server containers, this directory is REQUIRED.
-    For Hyper-V containers, it MUST be omitted.
-
-    On all other platforms, this field is REQUIRED.
-
-    If set, this directory MUST be referenced by [`root`](config.md#root) within the `config.json` file.
+2. <a name="containerFormat02" />The directory referenced by [`root.path`](config.md#root), if that property is set in `config.json`.
```
Mashimiao: `A directory referenced by ...` will be better?

wking: > `A directory referenced by ...` wipl be better?

There is only one directory referenced by a given `root.path` value, so I prefer [the definite article][1].  But if you can get a second opinion in favor of the indefinite "A", I'll change it.

[1]: https://en.wikipedia.org/wiki/Article_(grammar)#Definite_article

Mashimiao: I think there is doubt, `directory` is the subject of this sentence.  Here, we want to express there is a directory which is referenced by config.md#root. `The directory` will stress that `directory` is a specific directory, that's OK, but the base is everyone know which and what the directory is. I think there is not such base in context, it looks unreasonable.
If I'm wrong, please feel free to correct me.

wking: > Here, we want to express there is a directory which is referenced by config.md#root.

Yes, and it is *that directory* which is the bundle artifact.

> `The directory` will stress that directory is a specific directory, that's OK, but the base is everyone know which and what the directory is. I think there is not such base in context, it looks unreasonable.

But everyone should know which directory it is, since we say "referenced by `root.path`, if that property is set in `config.json`".  So the previous entry establishes the `config.json`artifact, and there can be only one file at that path.  And this entry establishes (when `root.path` is set in *that* configuration) *the* referenced directory as an artifact too.  Once you have the `confog.json`, there is no further choice about which directory you use for this artifact.

Mashimiao: OK, How about as following:
2. <a name="containerFormat02" />container's root filesystem: the directory referenced by [`root.path`](config.md#root), if that property is set in `config.json`.

keep same format with previous entry, just a suggestion.

wking: > keep same format with previous entry...

Sure.  I can oush that up tomorrow.

wking: > > keep same format with previous entry...
>
> Sure. I can push that up tomorrow.

I've pushed your suggested wording with b420c96 → dbfae14.

tianon:
LGTM

dqminh:
LGTM

PR #841, opened 2017-05-19
config-linux: RFC 2119 wording for linux.resources.devices

wking:
Since #780, the spec has been more aggressively punting to the kernel APIs (vs. carrying local versions of kernel limitations).  For the properties touched by this commit, the JSON Schema changes in #690 were [rejected as part of this punting approach][1].  However, before this commit, it wasn't clear exactly what kernel interface was being punted to.

With this commit, we replace the old inline docs with an explicit punt to the device whitelist controller, listing the exact actions that the runtime MUST take for given config values.  This allows for compliance-testing runtimes (#746, ensuring config portability between compliant runtimes) and makes it possible to validate a given config against a given kernel (e.g. [Linux 4.11.1 only accepts `a`, `b`, and `c` as type characters][3]).

[1]: https://github.com/opencontainers/runtime-spec/pull/690#issuecomment-282824774
[2]: https://github.com/opencontainers/runtime-spec/issues/746
[3]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/security/device_cgroup.c?h=v4.11.1#n618

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -211,20 +211,29 @@ Runtimes MAY attach the container process to additional cgroup controllers beyon
 
 ### <a name="configLinuxDeviceWhitelist" />Device whitelist
 
-**`devices`** (array of objects, OPTIONAL) configures the [device whitelist][cgroup-v1-devices].
-The runtime MUST apply entries in the listed order.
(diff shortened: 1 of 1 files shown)


wking:
Rebased around #846 with f43520d → 5a7ee3a.

crosbymichael:
Looking at this change, its clear we don't want the defaults that default to something insecure.  We should just make these things explicit by the user and provide things like -1 for wildcard.

PR #844, opened 2017-05-23
config-linux: specify mountlabel requirement for runtime

Mashimiao:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -641,7 +641,9 @@ The values MUST be absolute paths in the [container namespace](glossary.md#conta
 
 ## <a name="configLinuxMountLabel" />Mount Label
 
-**`mountLabel`** (string, OPTIONAL) will set the Selinux context for the mounts in the container.
+**`mountLabel`** (string, OPTIONAL) will set the SELinux context for the mounts in the container.
+If `mountLabel` is specified and SELinux is enabled, the runtime MUST set `mountLabel` for the mounts in the container.
+For more information about SELinux, see  [SELinux documentation][selinux].
 
 ### Example
 


config-linux.md
```diff
@@ -641,7 +641,9 @@ The values MUST be absolute paths in the [container namespace](glossary.md#conta
 
 ## <a name="configLinuxMountLabel" />Mount Label
 
-**`mountLabel`** (string, OPTIONAL) will set the Selinux context for the mounts in the container.
+**`mountLabel`** (string, OPTIONAL) will set the SELinux context for the mounts in the container.
+If `mountLabel` is specified and SELinux is enabled, the runtime MUST set `mountLabel` for the mounts in the container.
```
wking: I don't think we want “and SELinux is enabled”.  Runtimes that don't support SELinux [will not be compliant][1], just like runtimes that don't support cgroups will not be compliant.  These runtimes might still be *useful* in certain contexts, and not all hosts may support SELinux (in which case configs that set `mountLabel` will likely crash in `create`), but in order to be a compliant runtime, you *MUST* support this property.  If we want compliance labeling for runtimes that don't support this property, we'd need a new [`selinux` compliance track][2].

And instead of the still-unclear-to-me “MUST set `mountLabel` for the mounts”, I'd rather focus on how we test that this happened.  For example:

    If `mountLabel` is set, `lgetxattr(2)` with `security.selinux` for `name` MUST return this value when called from the container mount namespace (or context, or whatever) on `/` and all `mounts[].destination`.

[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#platforms

Mashimiao: What you said makes me confused.
SELinux is enabled or not is not runtimes' business, that's host's business. If SELinux is enabled on host and users specified mountLabel, rutimes MUST apply it, that's runtimes' business.
I'm not sure how you think about "runtime support SELinux or not"? In my opinion, the `support` means "When SELinux label is specified, and SELinux is enabled on host, runtime can apply the label correctly", and also we can say the runtime is compliant.

wking: > I'm not sure how you think about "runtime support SELinux or not"? In my opinion, the support means "When SELinux label is specified, and SELinux is enabled on host, runtime can apply the label correctly", and also we can say the runtime is compliant.

That makes sense to me.  I was thinking of runtime support like [runC's `selinux` build tag][1].  I think we need to require runtimes to respect this property on hosts where SELinux is enabled.  I care less about hosts where SELinux is not available, but would like runtimes to error out during `create` if `mountLabel` (or other SELinux-based option) was set in a config on those hosts.  I think my [proposed wording][2] is fine for this already.  Or do you want to qualify it based on having a [mounted `selinuxfs` in the host mount namespace][3] or some such?

[1]: https://github.com/opencontainers/runc/tree/v1.0.0-rc3#build-tags
[2]: https://github.com/opencontainers/runtime-spec/pull/844#discussion_r118042435
[3]: https://github.com/opencontainers/selinux/blob/v1.0.0-rc1/go-selinux/selinux.go#L118

mrunalp:
It is better to leave this as it is. We need to consider the platform, selinux modes and whether runtime has been compiled with SELinux.

PR #845, opened 2017-05-23
specs-go: Rlimits is not Linux-specific

Mashimiao:


wking:
Related discussion in #810 and #835.  The Markdown spec removed the
Linux-specific-ness of this property in [1], but I'm still not
convinced that the property makes sense outside of POSIX.

wking:
Does anyone have docs for the API used to implement this in
Windows/Solaris?  I expect Solaris implements POSIX setrlimit and
getrlimit [1], but does it support resources beyond those defined by
POSIX?  Does Windows support an analogous operation?

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html

Mashimiao:
As you already fix it in #835, closing this

PR #846, opened 2017-05-23
format specs with 4 spaces indent

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

PR #847, opened 2017-05-23
config: Replace "SHOULD consider" for 'rootfs' path value

wking:
SHOULD [means][1]:

> … there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.

So consideration is already baked into the definition, and we can just say “SHOULD be” and warn folks whenever they use a value other than `rootfs`.

Also break this into its own paragraph, because the value recommendation is orthogonal to and less important than the value semantics covered in the preceding sentences.

[1]: https://tools.ietf.org/html/rfc2119#section-3

### DIFF
--- a/config.md
+++ b/config.md
@@ -30,11 +30,11 @@ For all platform-specific configuration values, the scope defined below in the [
 
 * **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container.
     The path is either an absolute path or a relative path to the bundle.
-    Users SHOULD consider using a conventional name, such as `rootfs`.
 
     * On Windows, for Windows Server Containers, this field is REQUIRED and MUST be specified as a [volume GUID path][naming-a-volume].
       For Hyper-V Containers, this field MUST be omitted.
     * On all other platforms, this field is REQUIRED.
+        The value SHOULD be the conventional `rootfs`.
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

config.md
```diff
@@ -29,7 +29,8 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 **`root`** (object, REQUIRED) specifies the container's root filesystem.
 
 * **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container. The path is either an absolute path or a relative path to the bundle.
-    Users SHOULD consider using a conventional name, such as `rootfs`.
+
+    The value SHOULD be the conventional `rootfs`.
```
wking: Neither the old nor the new wording applies to Windows (since #849, because `rootfs` is not a [GUID path][1]).  I [address that][2] in the in-flight #838, and am leaving it alone here since it seems orthogonal to dropping “consider”.  If folks do want me to address the “does not apply to Windows” issue in this PR as well, I'm happy to pull over some of the #838 wording to do that.

[1]: https://github.com/opencontainers/runtime-spec/blame/aaa24815e0ab632c424e18d726fd47f403ce2981/config.md#L34
[2]: https://github.com/opencontainers/runtime-spec/pull/838#issuecomment-304129858

wking:
Rebased around #846 with 5d273c0 → 7de3cb8.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #849, opened 2017-05-23
Windows: Remove Sandbox, additional tweaks

jhowardmsft:

This relates to conversations internally here finalising and verifying the changes necessary to support Windows in the v1.0 OCI spec. Specifically the first part of this comment: https://github.com/opencontainers/runtime-spec/pull/828#issuecomment-303537717.

It turns out that between Technical Preview 5 (TP5) and RTM of Windows Server 2016, there was a platform change whereby the SandboxPath no longer needed to be passed, but was kept in for backwards compatibility as long as the layerfolderpath is populated between the runtime and HCS. I wasn't aware of this. Hence it's no longer needed in the spec. Apologies for the noise introducing it.

There are three other clarifications in this PR:

a) Clarify the root.Path for Windows Server containers and give an explicit example.
b) Fix the double-escaping in the utilityvmpath in `config-windows.md`.
c) Removed old unused hyperlinks

### DIFF
--- a/config.md
+++ b/config.md
@@ -31,7 +31,7 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 * **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container. The path is either an absolute path or a relative path to the bundle.
     Users SHOULD consider using a conventional name, such as `rootfs`.
 
-    On Windows, for Windows Server Containers, this field is REQUIRED. For Hyper-V Containers, this field MUST be omitted.
+    On Windows, for Windows Server Containers, this field is REQUIRED and MUST be specified as a [volume GUID path][naming-a-volume]. For Hyper-V Containers, this field MUST be omitted.
 
     On all other platforms, this field is REQUIRED.
 
@@ -40,12 +40,20 @@ For example, if a configuration is compliant with version 1.1 of this specificat
     If defined, a directory MUST exist at the path declared by the field.
 * **`readonly`** (bool, OPTIONAL) If true then the root filesystem MUST be read-only inside the container, defaults to false. On Windows, this field must be omitted or false.
 
-### Example
+### Example (POSIX)
 
 ```json
 "root": {
     "path": "rootfs",
     "readonly": true
+}
+```
+
+### Example (Windows)
+
+```json
+"root": {
+    "path": "\\\\?\\Volume{ec84d99e-3f02-11e7-ac6c-00155d7682cf}\\"
 }
 ```
 
@@ -841,8 +849,7 @@ Here is a full example `config.json` for reference.
 [go-environment]: https://golang.org/doc/install/source#environment
 [ieee-1003.1-2001-xbd-c8.1]: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html#tag_08_01
 [ieee-1003.1-2001-xsh-exec]: http://pubs.opengroup.org/onlinepubs/009695399/functions/exec.html
-[mountvol]: http://ss64.com/nt/mountvol.html
-[set-volume-mountpoint]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365561(v=vs.85).aspx
+[naming-a-volume]: https://aka.ms/nb3hqb
 
 [capabilities.7]: http://man7.org/linux/man-pages/man7/capabilities.7.html
 [mount.2]: http://man7.org/linux/man-pages/man2/mount.2.html
--- a/config-windows.md
+++ b/config-windows.md
@@ -148,15 +148,12 @@ The following parameters can be specified:
 
 * **`utilityvmpath`** *(string, OPTIONAL)* - specifies the path to the image used for the utility VM. This would be specified if using a base image which does not contain a utility VM image. If not supplied, the runtime will search the container filesystem layers from the bottom-most layer upwards, until it locates "UtilityVM", and default to that path.
 
-* **`sandboxpath`** *(string, REQUIRED)* - specifies the root of the path to the sandbox to be used for the container.
-
 ### Example
 
 ```json
     "windows": {
         "hyperv": {
-            "utilityvmpath": "C:\\\\path\\\\to\\utilityvm",
-            "sandboxpath": "C:\\\\programdata\\\\docker\\\\windowsfilter
+            "utilityvmpath": "C:\\path\\to\\utilityvm"
         }
     }
 ```
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -97,10 +97,6 @@
                     "utilityvmpath": {
                         "id": "https://opencontainers.org/schema/bundle/windows/hyperv/utilityvmpath",
                         "type": "string"
-                    },
-                    "sandboxpath": {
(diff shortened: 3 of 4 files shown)


config.md
```diff
@@ -31,7 +31,7 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 * **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container. The path is either an absolute path or a relative path to the bundle.
     Users SHOULD consider using a conventional name, such as `rootfs`.
 
-    On Windows, for Windows Server Containers, this field is REQUIRED. For Hyper-V Containers, this field MUST be omitted.
+    On Windows, for Windows Server Containers, this field is REQUIRED and MUST be specified as a Win32 file namespace volume path. For Hyper-V Containers, this field MUST be omitted.
```
wking: What is “Win32 file namespace volume path”?  Can we get a doc link (maybe [this][1]?)?  Ideally this MUST would be backed up by a stable link to a spec defining that term with sufficient detail for runtime-tools validation to determine if a given `root.path` value met the requirement or not.

And I'm also touching this line in #838.  That change is orthogonal to your addition here, but whichever PR lands second will likely need a rebase.

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx#win32_file_namespaces

config.md
```diff
@@ -49,6 +49,16 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 }
 ```
 
+### Example (Windows)
+
+```json
+"root": {
+    "path": "\\\\?\\Volume{ec84d99e-3f02-11e7-ac6c-00155d7682cf}",
```
wking: You should remove the trailing comma here to get a legal JSON fragment.

config.md
```diff
@@ -49,6 +49,16 @@ For example, if a configuration is compliant with version 1.1 of this specificat
...
+}
+```
+
+
+
```
wking: Almost all of the rest of the spec has a single empty line before a new header, so I'd recommend reducing your number of empty lines here from three to one.

tianon:
LGTM

config.md
jhowardmsft: Good catch. Fixed.

config.md
jhowardmsft: Yup, fixed too.

tianon:
LGTM

config.md
jhowardmsft: Yes, that is the link. Will add.

jhowardmsft: Actually, found a better link at https://msdn.microsoft.com/en-us/library/windows/desktop/aa365248(v=vs.85).aspx. Updated.

tianon:
LGTM

config.md
```diff
@@ -31,7 +31,7 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 * **`path`** (string, OPTIONAL) Specifies the path to the root filesystem for the container. The path is either an absolute path or a relative path to the bundle.
     Users SHOULD consider using a conventional name, such as `rootfs`.
 
-    On Windows, for Windows Server Containers, this field is REQUIRED. For Hyper-V Containers, this field MUST be omitted.
+    On Windows, for Windows Server Containers, this field is REQUIRED and MUST be specified as a volume GUID path. For more information, see [MSDN documentation][naming-a-volume]. For Hyper-V Containers, this field MUST be omitted.
```
wking: Maybe “see MSDN documentation” → “see the MSDN documentation”?  Or use:

> … MUST be specified as a [volume GUID path][naming-a-volume].

and drop the “For more information” sentence?

[naming-a-volume]: https://aka.ms/nb3hqb

jhowardmsft: Sure. Went with the latter. Updated.

config.md
```diff
@@ -49,6 +49,14 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 }
 ```
 
+### Example (Windows)
+
+```json
+"root": {
+    "path": "\\\\?\\Volume{ec84d99e-3f02-11e7-ac6c-00155d7682cf}"
```
wking: The [docs you link][1] give the pattern as `"\\?\Volume{GUID}\"` and say:

> All volume and mounted folder functions that take a volume GUID path as an input parameter require the trailing backslash.

before talking about a `CreateFile` caveat.  I'm not sure if the Windows runtime will always use `CreateFile` to back this property, or if `CreateFile` distinguishes between backslash-terminated or not volume GUID paths, but I expect you either want to add a trailing backslash here or add some words up above about why you don't need it.

[1]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365248%28v=vs.85%29.aspx

jhowardmsft: The trailing backslash doesn't actually matter. The example I gave was pasted from a breakpoint on a live running container.

jhowardmsft: That said, I'm seeing whether HCS itself can cope if a trailing backslash is supplied.

wking: > The trailing backslash doesn't actually matter.

So say that in the docs?  Perhaps:

> … MUST be specified as a [volume GUID path][naming-a-volume] (a trailing backslash is OPTIONAL)…

[naming-a-volume]: https://aka.ms/nb3hqb

jhowardmsft: It doesn't. Will update here.

jhowardmsft: Actually, it's better that we follow MSDN from the spec. I will update the runtime to ensure the trailing backslash is removed if present before passing onto HCS. Updated here accordingly.

tianon: It's a path to a directory/volume, right?  Trailing slash on those is almost always optional, and IMO there's no additional value in being explicit about that here.

wking: >  Trailing slash on those is almost always optional, and IMO there's no additional value in being explicit about that here.

I'd agree with that if the docs for volume GUID paths didn't [require a trailing backslash in most cases][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/849#discussion_r118129096

config.md
```diff
@@ -49,6 +49,14 @@ For example, if a configuration is compliant with version 1.1 of this specificat
 }
 ```
 
+### Example (Windows)
+
+```json
+"root": {
+    "path": "\\\\?\\Volume{ec84d99e-3f02-11e7-ac6c-00155d7682cf}\"
```
wking: You need to double the trailing backslash to avoid escaping the closing quote.

TomSweeneyRedHat: 👍

jhowardmsft: The four/two is correct :)

jhowardmsft: But there should be two at the end.

TomSweeneyRedHat: Okey Doke, I'll take @jhowardmsft word on this one.  Thanks for the follow up.

jhowardmsft: I've pushed an update.

jhowardmsft:
All comments addressed. Ready for review.

tianon:
LGTM

wking:
On Wed, May 24, 2017 at 10:08:22AM -0700, John Howard wrote:
> All comments addressed. Ready for review.

I'm not familiar enough with the underlying technology to comment on
that, but 2283e63 looks internally consistent to me.

jhowardmsft:
@vbatts @mrunalp @crosbymichael OK to get this one merged now for the 1.0 RC cut-off?

mrunalp:
LGTM

PR #850, opened 2017-05-24
Remove platform field

hqhq:
Close: #725

See discussion in #830 and dev-weekly-meeting discussion
in http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2017/opencontainers.2017-06-14-21.03.log.txt

We don't use this field in runc, and mostly only
image-spec cares about this, we can assume tools know
what specific platform this spec is built for.

### DIFF
--- a/config.md
+++ b/config.md
@@ -321,47 +321,19 @@ For Windows based systems the user structure has the following fields:
 "hostname": "mrsdalloway"
 ```
 
-## <a name="configPlatform" />Platform
-
-**`platform`** (object, REQUIRED) specifies the configuration's target platform.
-
-* **`os`** (string, REQUIRED) specifies the operating system family of the container configuration's specified [`root`](#root) filesystem bundle.
-    The runtime MUST generate an error if it does not support the specified **`os`**.
-    Values SHOULD be, and runtimes SHOULD understand, **`os`** entries listed in the Go Language document for [`GOOS`][go-environment].
-    If an operating system is not included in the `GOOS` documentation, it SHOULD be submitted to this specification for standardization.
-* **`arch`** (string, REQUIRED) specifies the instruction set for which the binaries in the specified [`root`](#root) filesystem bundle have been compiled.
-    The runtime MUST generate an error if it does not support the specified **`arch`**.
-    Values SHOULD be, and runtimes SHOULD understand, **`arch`** entries listed in the Go Language document for [`GOARCH`][go-environment].
-    If an architecture is not included in the `GOARCH` documentation, it SHOULD be submitted to this specification for standardization.
-
-### Example
-
-```json
-"platform": {
-    "os": "linux",
-    "arch": "amd64"
-}
-```
-
 ## <a name="configPlatformSpecificConfiguration" />Platform-specific configuration
 
-[**`platform.os`**](#platform) is used to specify platform-specific configuration.
-
 * **`linux`** (object, OPTIONAL) [Linux-specific configuration](config-linux.md).
-    This MAY be set if **`platform.os`** is `linux` and MUST NOT be set otherwise.
+    This MAY be set if the target platform of this spec is `linux`.
 * **`windows`** (object, OPTIONAL) [Windows-specific configuration](config-windows.md).
-    This MUST be set if **`platform.os`** is `windows` and MUST NOT be set otherwise.
+    This MUST be set if the target platform of this spec is `windows`.
 * **`solaris`** (object, OPTIONAL) [Solaris-specific configuration](config-solaris.md).
-    This MAY be set if **`platform.os`** is `solaris` and MUST NOT be set otherwise.
+    This MAY be set if the target platform of this spec is `solaris`.
 
 ### Example (Linux)
 
 ```json
 {
-    "platform": {
-        "os": "linux",
-        "arch": "amd64"
-    },
     "linux": {
         "namespaces": [
             {
@@ -477,10 +449,6 @@ Here is a full example `config.json` for reference.
 ```json
 {
     "ociVersion": "0.5.0-dev",
-    "platform": {
-        "os": "linux",
-        "arch": "amd64"
-    },
     "process": {
         "terminal": true,
         "user": {
@@ -856,7 +824,6 @@ Here is a full example `config.json` for reference.
 [no-new-privs]: https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt
 [procfs_2]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 [semver-v2.0.0]: http://semver.org/spec/v2.0.0.html
-[go-environment]: https://golang.org/doc/install/source#environment
 [ieee-1003.1-2008-xbd-c8.1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_01
 [ieee-1003.1-2008-xsh-exec]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html
 [naming-a-volume]: https://aka.ms/nb3hqb
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -37,24 +37,6 @@
                 "$ref": "defs.json#/definitions/Mount"
             }
         },
-        "platform": {
-            "id": "https://opencontainers.org/schema/bundle/platform",
-            "type": "object",
-            "required": [
-                "arch",
-                "os"
-            ],
-            "properties": {
-                "arch": {
-                    "id": "https://opencontainers.org/schema/bundle/platform/arch",
-                    "type": "string"
-                },
-                "os": {
-                    "id": "https://opencontainers.org/schema/bundle/platform/os",
-                    "type": "string"
-                }
-            }
-        },
         "root": {
             "description": "Configures the container's root filesystem.",
             "id": "https://opencontainers.org/schema/bundle/root",
@@ -233,7 +215,6 @@
     },
     "required": [
         "ociVersion",
-        "platform",
         "root"
     ]
 }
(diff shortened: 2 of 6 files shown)


config.md
```diff
@@ -304,46 +304,32 @@ For Windows based systems the user structure has the following fields:
...
-* **`os`** (string, REQUIRED) specifies the operating system family of the container configuration's specified [`root`](#root) file system bundle.
-  The runtime MUST generate an error if it does not support the specified **`os`**.
-  Bundles SHOULD use, and runtimes SHOULD understand, **`os`** entries listed in the Go Language document for [`GOOS`][go-environment].
-  If an operating system is not included in the `GOOS` documentation, it SHOULD be submitted to this specification for standardization.
-* **`arch`** (string, REQUIRED) specifies the instruction set for which the binaries in the specified [`root`](#root) file system bundle have been compiled.
-  The runtime MUST generate an error if it does not support the specified **`arch`**.
-  Values for **`arch`** SHOULD use, and runtimes SHOULD understand, **`arch`** entries listed in the Go Language document for [`GOARCH`][go-environment].
-  If an architecture is not included in the `GOARCH` documentation, it SHOULD be submitted to this specification for standardization.
+**`platform`** (string, REQUIRED) specifies the configuration's target platform.
+  The value MUST be a slug from [the platform list](spec.md#platforms).
```
jhowardmsft: Not sure I like the word `slug`...

config.md
```diff
@@ -304,46 +304,32 @@ For Windows based systems the user structure has the following fields:
...
 
-[**`platform.os`**](#platform) is used to specify platform-specific configuration.
+[**`platform`**](#platform) is used to specify platform-specific configuration.
 Runtime implementations MAY support any valid values for platform-specific fields as part of this configuration.
 
 * **`linux`** (object, OPTIONAL) [Linux-specific configuration](config-linux.md).
-  This MAY be set if **`platform.os`** is `linux` and MUST NOT be set otherwise.
+  This MAY be set if **`platform`** is `linux` and MUST NOT be set otherwise.
 * **`windows`** (object, OPTIONAL) [Windows-specific configuration](config-windows.md).
-  This MAY be set if **`platform.os`** is `windows` and MUST NOT be set otherwise.
```
jhowardmsft: Actually, this MUST be set if the platform is windows. A container isn't going to run at all without it. Not sure whether this applies to other platforms though.

config.md
hqhq: Updated, thanks.

config.md
hqhq: Maybe we can open a new PR to fix this? It's non-relevant  to this PR.

wking: > Maybe we can open a new PR to fix this?

#828 is covering this change, so we can leave this as MAY for this PR.

tianon:
Since this appears to map `platform` to image-spec's `platform.os`, would it make sense to instead call this `os` or `platformOs` (and perhaps even reference image-spec)?

IMO dropping our `platform` object down to just `os` is a decent compromise (rather than removing `platform` entirely), but I'd love to hear the opinions of other @opencontainers/runtime-spec-maintainers.

wking:
On Wed, May 24, 2017 at 08:33:29AM -0700, Tianon Gravi wrote:
> … and perhaps even reference image-spec…

There's no need for this, because the values come from the
runtime-spec platform list [1].

[1]: https://github.com/opencontainers/runtime-spec/blob/2ec18c02474dd49ce20aa30236959a8a80763042/spec.md#platforms

stevvooe:
This seems like a step backwards.

The only change needed to make it compatible with the image spec is `arch` -> `architecture`.

config.md
wking: Now that #828 has landed, this PR will need a rebase.

hqhq:
@opencontainers/runtime-spec-maintainers More comments on this? I see many different suggestions and not sure which way to go.

wking:
On Thu, Jun 01, 2017 at 08:11:13PM -0700, Qiang Huang wrote:
> More comments on this? I see many different suggestions and not sure
> which way to go.

I'm not a maintainer, but I think we want to gut the current platform
(like you're doing in dc88c9e) unless someone wants support for the
“I'm not sure if my runtime will execute this container on my current
host or some other platform…” use case [1].  I haven't heard anyone
advocating for that.

We'll need some way to declare your compliance track [2], and for that
I'm fine with either of:

a. Your ‘platform’ property where the value “MUST be an item from the
   platform list [2]”.
b. @crosbymichael's ‘--platform’ option [3], with the same
   compliance-track values [2].

I don't see much difference between a or b myself, but @crosbymichael
sees his proposal as more robust [3].

You could swing something more complicated like @tianon's proposal
[4], but I feel like using the platform-specific sections to both
select the compliance track *and* hold some subset of the
platform-specific properties is too overloaded [5].

[1]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-301841186
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#platforms
[3]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303463569
[4]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303852009
[5]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303868422

jhowardmsft:
Actually, just to throw one out there.... Windows will imminently be able to run both Windows and Linux containers. So it might matter.

wking:
On Thu, Jun 01, 2017 at 08:47:56PM -0700, John Howard wrote:
> Windows will imminently be able to run both Windows and Linux
> containers. So it might matter.

The arch will matter?

I think all the proposals include some way to specify the compliance
track [1], so you can still send an explicitly-Linux config into a
runtime on a Windows host.  Will the Windows runtime need any more
information than “this is a windows config” or “this is a linux
config”?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/spec.md#platforms

jhowardmsft:
I'm still figuring that out. It will need bits of the Windows and Linux sections - for example the LayerFolders in Windows are going to be required for the runtime, but much of the Linux section will need to be populated as that's where the Linux settings are. For 1.0, I'd probably ignore my comment, it was just in passing. However, there will definitely be quite a few spec updates required for multi-platform on a single OS.

wking:
On Thu, Jun 01, 2017 at 08:55:41PM -0700, John Howard wrote:
> It will need bits of the Windows and Linux sections - for example
> the LayerFolders in Windows are going to be required for the
> runtime, but much of the Linux section will need to be populated as
> that's where the Linux settings are.

*shudder* Some of these host-vs.-container issues have come up in
#405, although I don't think compliance testing plans for those VM
settings have been worked out yet.

> For 1.0, I'd probably ignore my comment…

Ok, so we're back to “nobody is asking for anything more than the
compliance track at the moment”.

config.md
```diff
@@ -304,46 +304,32 @@ For Windows based systems the user structure has the following fields:
 
 ## <a name="configPlatform" />Platform
 
-**`platform`** (object, REQUIRED) specifies the configuration's target platform.
-
-* **`os`** (string, REQUIRED) specifies the operating system family of the container configuration's specified [`root`](#root) file system bundle.
-  The runtime MUST generate an error if it does not support the specified **`os`**.
-  Bundles SHOULD use, and runtimes SHOULD understand, **`os`** entries listed in the Go Language document for [`GOOS`][go-environment].
```
wking: You'll want to drop [the reference at the ebd of this file][1] too, since you remove the only consumers.

[1]: https://github.com/opencontainers/runtime-spec/blame/e95b4f8ead3dd0a0ce2676339490f0ce126ca898/config.md#L859

crosbymichael:
We discussed this on the call today and feel that it would be better to fully remove it from the spec.  The tool that interprets the spec knows how/what platform it is built for or what platform it should use in the case of windows or VMs.  

@hqhq can you update this PR to remove the platform field fully?  Validation tools will be told what they are validating a specific spec against.

hqhq:
@crosbymichael Updated.

config.md
```diff
@@ -855,8 +823,7 @@ Here is a full example `config.json` for reference.
 [selinux]:http://selinuxproject.org/page/Main_Page
 [no-new-privs]: https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt
 [procfs_2]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt
-[semver-v2.0.0]: http://semver.org/spec/v2.0.0.html
-[go-environment]: https://golang.org/doc/install/source#environment
+[semver-v3.0.0]: http://semver.org/spec/v2.0.0.html
```
wking: I don't think you want to bump `semver-v2.0.0` → `semver-v3.0.0` (especially since 2.0.0 is still the most recent version ;).

crosbymichael:
platform changes 

LGTM

config.md
tianon: Indeed -- minus this change, the rest of this PR LGTM now. :+1:

hqhq: Oops, no idea where that change comes from, updated.

Mashimiao:
Sorry for asking so late. I have a question.
If we set more than one of linux, windows and solaris, how can runtime distinguish which platform is the target platform?

wking:
> If we set more than one of linux, windows and solaris, how can runtime distinguish which platform is the target platform?

The runtime will have a compiled-in favorite.  For example, a Linux build of `runc` will assume a Linux config, regardless of which properties are set.

Tools that can handle configs for multiple platforms (e.g. `runtime-tools`) will need a [new `--platform` option][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/830#issuecomment-303463569

tianon:
LGTM

crosbymichael:
LGTM

Issue #813, opened 2017-05-12
question about valid values  runtime choose to support


Mashimiao:
In config.md, spec says `Unless support for a valid value is explicitly required, runtimes MAY choose which subset of the valid values it will support.`
I think there may will be a problem, if different runtime support differen subset. If so, we can't guarantee the portability. 
I think this is against our goal of standardization.

wking:
I think the idea was to avoid failing a runtime as non-compliant if
the host system didn't support a feature.  But with the current
wording, you can have compliant runtimes that use that language to not
support most values described in the spec (everything except the ‘ro’
mount option on Windows [1] and ‘l3CacheSchema’ on Linux [2]?) even
when the host system supports the values.

I'd rather require runtimes to support at least the values the spec
lists as valid (with runtimes having the option to support additional
values as they see fit).  I believe the “kernel doesn't support the
feature you've asked for case” is covered by [3]:

  If the runtime cannot apply a property as specified in the
  configuration, it MUST generate an error and a new container MUST
  NOT be created.

Dropping the “runtimes MAY choose which subset” line would make using
compliant runtimes more reliable, although we'd probably want to
either inline our valid-values list (like we do for ‘ro’ in Windows
mounts [1]) or link to a specific version of an external doc (like we
do for Solaris rlimits [4,5]).  Linking to a floating external doc
(like we do for Linux rlimits [6,7]) will lead to issues as values are
added to and removed from the external doc.

There have also been concerns about linking to man pages for lists of
valid values, because man pages may be an incomplete or incorrect
reflection of what the kernel actually supports.  I'd address those
cases by either:

a. Submitting upstream patches to fix the man pages, and then linking
   the spec to the patched version, or
b. Inlining the valid values in the spec so we don't need a normative
   link to the buggy man page.

With both of those being fairly straightforward, I don't think “man
pages are unreliable” is a sufficient reason to avoid requiring
runtime support for values that the spec lists as valid.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L84
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L459
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/runtime.md#L106
[4]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L169
[5]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L857
[...]

PR #851, opened 2017-05-24
config: Remove "MAY support any valid values" sentence

wking:
This line landed in #673, but I'm not clear on the motivation.  The wording reads to me like “you don't have to support valid values for platform-specific fields if you don't want to”, but we already cover unsupported value handling with the [valid values section][1] separated out in #681.  This commit removes the ambiguous line.

Related issue in #813, although this commit will not close that issue.

[1]: https://github.com/opencontainers/runtime-spec/blob/837ee7659b0c115bb32598e4454a1f6d2460b1ea/config.md#valid-values

### DIFF
--- a/config.md
+++ b/config.md
@@ -340,7 +340,6 @@ For Windows based systems the user structure has the following fields:
 ## <a name="configPlatformSpecificConfiguration" />Platform-specific configuration
 
 [**`platform.os`**](#platform) is used to specify platform-specific configuration.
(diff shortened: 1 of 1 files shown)


crosbymichael:
@wking can you please rebase this

mrunalp:
Needs rebase

mrunalp:
LGTM

crosbymichael:
LGTM

PR #854, opened 2017-05-24
Mount correction

jhowardmsft:

a) Tagged the `type` of mount with the correct platform tag (not applicable on `Windows`.

b) Generally, there are not platform specific comments inside the `.go` files - that why we have the `.md` files. Hence moved the `Linux` comments for mount from `config.go` to `config.md`.

c) The comment about `the path and child directories MUST exist, a runtime MUST NOT create directories automatically to a mount point.` is incorrect in the way mapped directories (which is what the runtime uses through HCS to provide "pseudo"-mounts on Windows) work. Hence made that a `Posix` comment.

### DIFF
--- a/config.md
+++ b/config.md
@@ -65,21 +65,36 @@ For all platform-specific configuration values, the scope defined below in the [
     For Linux, the parameters are as documented in [mount(2)][mount.2] system call man page.
     For Solaris, the mount entry corresponds to the 'fs' resource in the [zonecfg(1M)][zonecfg.1m] man page.
 
-
 * **`destination`** (string, REQUIRED) Destination of mount point: path inside container.
     This value MUST be an absolute path.
     * Windows: one mount destination MUST NOT be nested within another mount (e.g., c:\\foo and c:\\foo\\bar).
     * Solaris: corresponds to "dir" of the fs resource in [zonecfg(1M)][zonecfg.1m].
-* **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
-    * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
-    * Windows: this field MUST NOT be supplied.
-    * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`source`** (string, OPTIONAL) A device name, but can also be a directory name or a dummy.
     * Windows: a local directory on the filesystem of the container host. UNC paths and mapped drives are not supported.
     * Solaris: corresponds to "special" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`options`** (array of strings, OPTIONAL) Mount options of the filesystem to be used.
     * Linux: supported options are listed in the [mount(8)][mount.8] man page. Note both [filesystem-independent][mount.8-filesystem-independent] and [filesystem-specific][mount.8-filesystem-specific] options are listed.
     * Solaris: corresponds to "options" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+
+### Example (Windows)
+
+```json
+"mounts": [
+    {
+        "destination": "C:\\folder-inside-container",
+        "source": "C:\\folder-on-host",
+        "options": []
+    }
+]
+```
+
+### <a name="configLinuxAndSolarisMounts" />Linux and Solaris Mounts
+
+For Linux and Solaris based systems the mounts structure has the following fields:
+
+* **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
+  * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 
 ### Example (Linux)
 
@@ -96,18 +111,6 @@ For all platform-specific configuration values, the scope defined below in the [
         "type": "bind",
         "source": "/volumes/testing",
         "options": ["rbind","rw"]
-    }
-]
-```
-
-### Example (Windows)
-
-```json
-"mounts": [
-    {
-        "destination": "C:\\folder-inside-container",
-        "source": "C:\\folder-on-host",
-        "options": []
     }
 ]
 ```
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -112,12 +112,11 @@ type Platform struct {
 
 // Mount specifies a mount for a container.
 type Mount struct {
-	// Destination is the path where the mount will be placed relative to the container's root.  The path and child directories MUST exist, a runtime MUST NOT create directories automatically to a mount point.
+	// Destination is the absolute path where the mount will be placed in the container.
 	Destination string `json:"destination"`
 	// Type specifies the mount kind.
-	Type string `json:"type,omitempty"`
-	// Source specifies the source path of the mount.  In the case of bind mounts on
-	// Linux based systems this would be the file on the host.
+	Type string `json:"type,omitempty" platform:"linux,solaris"`
+	// Source specifies the source path of the mount.
 	Source string `json:"source,omitempty"`
 	// Options are fstab style mount options.
 	Options []string `json:"options,omitempty"`


jhowardmsft:
Fixed white-space errors.

config.md
```diff
@@ -59,6 +59,7 @@ For Solaris, the mount entry corresponds to the 'fs' resource in the [zonecfg(1M
 
 * **`destination`** (string, REQUIRED) Destination of mount point: path inside container.
   This value MUST be an absolute path.
+  * Posix: the path and child directories MUST exist, a runtime MUST NOT create directories automatically to a mount point.
```
wking: Something very like this change was rejected in #591.

config.md
```diff
@@ -69,7 +70,7 @@ For Solaris, the mount entry corresponds to the 'fs' resource in the [zonecfg(1M
   * Windows: a local directory on the filesystem of the container host. UNC paths and mapped drives are not supported.
   * Solaris: corresponds to "special" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`options`** (array of strings, OPTIONAL) Mount options of the filesystem to be used.
-  * Linux: supported options are listed in the [mount(8)][mount.8] man page. Note both [filesystem-independent][mount.8-filesystem-independent] and [filesystem-specific][mount.8-filesystem-specific] options are listed.
+  * Linux: supported options are listed in the [mount(8)][mount.8] man page. Note both [filesystem-independent][mount.8-filesystem-independent] and [filesystem-specific][mount.8-filesystem-specific] options are listed. In the case of bind mounts, this would be the file on the host.
```
wking: `options` does not include a file from the host for bind mounts.  More on bind mounts in #771, which I hope to revive after opencontainers/runc#1442 lands and I can PR runC support for `(no)lazytime`, etc.

specs-go/config.go
```diff
@@ -112,12 +112,11 @@ type Platform struct {
...
 // Mount specifies a mount for a container.
 type Mount struct {
-	// Destination is the path where the mount will be placed relative to the container's root.  The path and child directories MUST exist, a runtime MUST NOT create directories automatically to a mount point.
+	// Destination is the absolute path where the mount will be placed in the container.
 	Destination string `json:"destination"`
 	// Type specifies the mount kind.
-	Type string `json:"type,omitempty"`
-	// Source specifies the source path of the mount.  In the case of bind mounts on
-	// Linux based systems this would be the file on the host.
+	Type string `json:"type,omitempty" platform:"linux,solaris"`
```
wking: Currently `type` is still *defined* for Windows, we just [require it to be unset][1].  I'd rather move it to a POSIX-specific subsection ([like we already do for UID/GID][2]) so the field would be undefined in the Windows-track spec.  More on this distinction [here][3] and in #680.  But as long as the Markdown docs define the property for Windows (even if they forbid it), I don't think we want to add this `platform` tag here.

[1]: https://github.com/opencontainers/runtime-spec/blame/2ec18c02474dd49ce20aa30236959a8a80763042/config.md#L66
[2]: https://github.com/opencontainers/runtime-spec/blob/2ec18c02474dd49ce20aa30236959a8a80763042/config.md#linux-and-solaris-user
[3]: https://github.com/opencontainers/runtime-spec/pull/747/files#diff-958e7270f96f5407d7d980f500805b1bR58

config.md
jhowardmsft: Right, but currently it is incorrect, as it states this in config.go for Windows where that statement is incorrect.

config.md
jhowardmsft: The change here (moving from config.go to config.md is purely for consistency. I haven't changed anything semantically as the statement already exists.

config.md
wking: > Right, but currently it is incorrect, as it states this in config.go for Windows where that statement is incorrect.

The [Markdown is normative, and the Go comments are not][1].  I'm in favor of removing the incorrect Go comment (like you do in this PR), and in general making the Go comments as simple as possible to avoid this sort of problem.  But I don't want to carry incorrect Go comments back into the normative Markdown (like you do in this hunk).

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/config.md#L4

specs-go/config.go
jhowardmsft: I'm not sure I understand this comment. Is this change not entirely consistent with the `User` struct, and the `UID`, `GID`, `AdditionalGids` and `Username` fields of it?

config.md
jhowardmsft: To be clear then, you recommend removing the addition both here, and keep the removal in config.go.

config.md
wking: > I haven't changed anything semantically as the statement already exists.

Moving text from the non-normative Go comments to the normative Markdown spec is a semantic change.  I'd rather this PR just adjusted the Go comments to not conflict with the Markdown spec while leaving the Markdown spec alone (unless you intend to make a semantic change, in which case you have to adjust the Markdown spec).

jhowardmsft: > I'd rather this PR just adjusted the Go comments to not conflict with the Markdown spec

I'm still confused. What exactly do you believe is incorrect here. Are you saying the statement `In the case of bind mounts, this would be the file on the host.` is invalid? It's inconsistent to have it in the go code, so where should it go? Or should it be entirely removed?

specs-go/config.go
wking: > I'm not sure I understand this comment.

I'm saying I'd like something closer to:

```Markdown
## Mounts

…current section contents without `type`…

### Linux and Solaris Mounts

For Linux- and Solaris-based systems the mount structure has the following additional properties:

* **`type`** (int, OPTIONAL) …
```

Then a Windows config setting `mounts[].type` would be in the same boat as a Windows config setting `process.users.uid` (although the discussion in #680 shows that what that boat means could still use clarification).

config.md
wking: > To be clear then, you recommend removing the addition both here, and keep the removal in config.go.

Yes.

config.md
wking: > Are you saying the statement In the case of bind mounts, this would be the file on the host. is invalid?

Yes.

> It's inconsistent to have it in the go code, so where should it go?

I think the Go comments should be as short and sweet as possible to avoid these inconsistencies.  The more of the spec that gets echoed in the Go comments, the less DRY we are and the more likely we have these inconsistencies.

> Or should it be entirely removed?

Yes, I'd like this PR to remove it from the Go comment but not inject it into the Markdown.

config.md
jhowardmsft: Done and update pushed.

config.md
jhowardmsft: Done. Update pushed.

wking: > > Are you saying the statement In the case of bind mounts, this would be the file on the host. is invalid?
> Yes.

Actually, the current Go comment [belongs to `source`][1], not `options` (although you're adding it under `options` here).  So the current Go comment is correct in master, and might conceivably be useful in Markdown (under `source`).  Although with the current spec having nothing to say about bind mounts and with #771 rejected at the moment, I think copying this one line over to Markdown without filling in the rest of a bind-mount spec isn't particularly useful.

I'm still in favor of removing it from the Go comment to avoid having this line go stale or conflict as the Markdown evolves.

[1]: https://github.com/opencontainers/runtime-spec/blob/2ec18c02474dd49ce20aa30236959a8a80763042/specs-go/config.go#L120

specs-go/config.go
jhowardmsft: Ah, I see. It confused me because the comment was against `config.go` when really you're referring to `config.md`....

wking: > It confused me because the comment was against `config.go` when really you're referring to `config.md`....

I'm saying that your current `platform:"linux,solaris"` addition here is not as clearly backed by the Markdown spec as I'd like.  I agree that it's where we want the Go type to end up, so I'd like to see your Go change, but I'd *also* like to see the Markdown adjusted to more clearly support the Go change.  In the absence of the Markdown change, I'd rather wait on the Go change.

jhowardmsft: OK, that's done. But I've been consistent with the existing `User` markdown. Specifically, it does not have a training `-` after Linux and Solaris as per your example. No use of the word `additional`. And uses the term `fields` rather than `properties`. 

Update pushed.

wking: > No use of the word `additional`.

In the `process.user` case, the platform-specific sections cover the *only* properties for that object.  In this case, the Linux/Solaris section will be adding a new property to the existing platform-independent properties.  The best precedent we have for that is probably [here][1], which doesn't use “additional” either.  I'd be happier with “additional” or some other language that makes it clear that both the platform-independent properties and the POSIX-specific `type` are valid entries, but I'm ok punting that clarification to follow-up work.

I'd rather have the other bits changed too (e.g. we do use the hypenated “Linux-based” [here][1]), but I'm fine punting those to follow-up work as well.

[1]: https://github.com/opencontainers/runtime-spec/blame/2ec18c02474dd49ce20aa30236959a8a80763042/config.md#L158

config.md
```diff
@@ -56,22 +56,37 @@ The runtime MUST mount entries in the listed order.
...
+
+### <a name="configLinuxAndSolarisMounts" />Linux and Solaris Mounts
+
+For Linux and Solaris based systems the mounts structure has the following fields:
+
+* **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
+  * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+
 ### Example (Linux)
```
wking: This (and the Solaris example) may want another `#` to nest them under `Linux and Solaris Mounts`.  But since the examples are not *just* about `type`, I'm also ok leaving them with `###` headers.

jhowardmsft: Agreed. I thought about adding another `#`, but decided against it for that reason.

specs-go/config.go
jhowardmsft: Agreed on a a follow-up consistency PR covering the markdown generally.

config.md
```diff
@@ -56,22 +56,37 @@ The runtime MUST mount entries in the listed order.
...
+]
+```
+
+### <a name="configLinuxAndSolarisMounts" />Linux and Solaris Mounts
+
+For Linux and Solaris based systems the mounts structure has the following fields:
+
+* **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
+  * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+  * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
```
wking: We can reduce future conflicts with #846 by adjusting these to use four-space indents when we touch the lines.  That doesn't have to happen in this PR, but I think it's useful for reducing the rebase load on @Mashimiao if this PR lands first.

jhowardmsft:
Should this be tagged for 1.0?

crosbymichael:
@jhowardmsft can you rebase this?

jhowardmsft:
@crosbymichael done

tianon:
LGTM

PR #855, opened 2017-05-24
Hooks to linux,solaris and doc clarification

jhowardmsft:

Made `hooks` Linux and solaris. Windows does not currently support hooks. While we might at a later date, have clarified that in the .go and .md file.

Also, as I mentioned in the description of https://github.com/opencontainers/runtime-spec/pull/854, platform-specific things are generally in the `.md` file, not the `.go` file. Hence moved the Linux comment.

### DIFF
--- a/config.md
+++ b/config.md
@@ -356,7 +356,8 @@ Runtime implementations MAY support any valid values for platform-specific field
 
 ## <a name="configHooks" />Hooks
 
-Hooks allow for the configuration of custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
+Hooks allow for the configuration of custom actions related to the [lifecycle](runtime.md#lifecycle) of the container if supported by the platform.
+On Linux, they are run after the container namespaces are created.
 
 * **`hooks`** (object, OPTIONAL) MAY contain any of the following properties:
     * **`prestart`** (array of objects, OPTIONAL) is an array of [pre-start hooks](#prestart).
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -17,7 +17,7 @@ type Spec struct {
 	// Mounts configures additional mounts (on top of Root).
 	Mounts []Mount `json:"mounts,omitempty"`
 	// Hooks configures callbacks for container lifecycle events.
-	Hooks *Hooks `json:"hooks,omitempty"`
+	Hooks *Hooks `json:"hooks,omitempty" platform:"linux,solaris"`
 	// Annotations contains arbitrary metadata for the container.
 	Annotations map[string]string `json:"annotations,omitempty"`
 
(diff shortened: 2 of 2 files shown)


config.md
```diff
@@ -356,7 +356,8 @@ Runtime implementations MAY support any valid values for platform-specific field
 
 ## <a name="configHooks" />Hooks
 
-Hooks allow for the configuration of custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
+Hooks allow for the configuration of custom actions related to the [lifecycle](runtime.md#lifecycle) of the container if supported by the platform.
+On Linux, they are run after the container namespaces are created.
```
wking: Not all hooks are run after creation.  In fact, there are currently no `postcreate` hooks (`prestart` hooks are called “[after the `start` operation is called but before the user-specified program command is executed][1]”.  I suggest not adding this line (and removing it from the Go comment).

[1]: https://github.com/opencontainers/runtime-spec/blob/2ec18c02474dd49ce20aa30236959a8a80763042/config.md#prestart

config.md
```diff
@@ -356,7 +356,8 @@ Runtime implementations MAY support any valid values for platform-specific field
 
 ## <a name="configHooks" />Hooks
 
-Hooks allow for the configuration of custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
+Hooks allow for the configuration of custom actions related to the [lifecycle](runtime.md#lifecycle) of the container if supported by the platform.
```
wking: I'd rather follow the [example set for `process.user.uid` and friends][1] and make this:

```Markdown
## Linux and Solaris Hooks

For Linux- and Solaris-based systems the configuration structure supports `hooks` for configuring custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
```

Although I'd rather have [this ABNF][2] so we could say:

* **`hooks`** (object, OPTIONAL, linux solaris) …

to protect readers who skipped over this lead-in paragraph and went straight to the `hooks` list entry.

[1]: https://github.com/opencontainers/runtime-spec/blob/2ec18c02474dd49ce20aa30236959a8a80763042/config.md#linux-and-solaris-user
[2]: https://github.com/opencontainers/runtime-spec/pull/747/files#diff-958e7270f96f5407d7d980f500805b1bR43

wking: If we *don't* clearly say which platforms can rely on `hooks`, we're back to the rejected #472.

vbatts:
SGTM

config.md
vbatts: correct.

crosbymichael:
LGTM

tianon:
LGTM

wking:
On Wed, May 31, 2017 at 03:44:08PM -0700, Tianon Gravi wrote:
> Merged #855.

Hmm.  I'll file follow-up PRs for at least [this point][1] and something about [this one][2].

[1]: https://github.com/opencontainers/runtime-spec/pull/855#discussion_r118341546
[2]: https://github.com/opencontainers/runtime-spec/pull/855#discussion_r118343052

Issue #857, opened 2017-05-25
Camel casing needs fixing in config.go and matching schema


jhowardmsft:
As per this comment: https://github.com/opencontainers/runtime-spec/pull/828#discussion_r118375184

wking:
If we're going to adjust any of this casing, I'd *really* like to have
that happen before the next tag.  #842 (the proposed rc6, currently at
4e3635b) has:

  $ git checkout origin/pr/842
  $ git grep '`' -- config-windows.md
  …
  config-windows.md:* **`sandboxSize`** *(uint64, OPTIONAL)* - specifies the minimum size of the system drive in bytes.
  …
  config-windows.md:* **`egressBandwidth`** *(uint64, OPTIONAL)* - specified the maximum egress bandwidth in bytes per second for the container.
  …
  config-windows.md:The `credentialspec` is a JSON object whose properties are implementation-defined.
  …

And master has:

  $ git checkout origin/master
  $ git describe
  v1.0.0-rc5-194-g9151c77
  $ git grep '`' -- config-windows.md
  …
  config-windows.md:* **`sandboxSize`** *(uint64, OPTIONAL)* - specifies the minimum size of the system drive in bytes.
  …
  config-windows.md:* **`endpointList`** *(array of strings, OPTIONAL)* - list of HNS (Host Network Service) endpoints that the container should connect to.
  config-windows.md:* **`allowUnqualifiedDNSQuery`** *(bool, OPTIONAL)* - specifies if unqualified DNS name resolution is allowed.
  config-windows.md:* **`DNSSearchList`** *(array of strings, OPTIONAL)* - comma seperated list of DNS suffixes to use for name resolution.
  config-windows.md:* **`networkSharedContainerName`** *(string, OPTIONAL)* - name (ID) of the container that we will share with the network stack.
  …
  config-windows.md:The `credentialspec` is a JSON object whose properties are implementation-defined.
  …
  config-windows.md:* **`utilityvmpath`** *(string, OPTIONAL)* - specifies the path to the image used for the utility VM. This would be specified if using a base image which does not contain a utility VM image. If not supplied, the runtime will search the container filesystem layers from the bottom-most layer upwards, until it locates "UtilityVM", and default to that path.
  config-windows.md:* **`sandboxpath`** *(string, REQUIRED)* - specifies the root of the path to the sandbox to be used for the container.
  …

which both have camelCase (sandboxSize, etc.) and lowercase
[...]

PR #859, opened 2017-05-25
Windows: Fix camelCase inconsistencies

jhowardmsft:

As per https://github.com/opencontainers/runtime-spec/pull/828#discussion_r118537284, fixes https://github.com/opencontainers/runtime-spec/issues/857.

Addresses `credentialSpec`, `utilityVMPath`, and `ignoreFlushesDuringBoot`.

REMOVED ~~~I noticed there were discrepancies also, particularly in the `SolarisAnet` structure (camelCase and non-matching field names to json names), so I fixed these up at the same time.~~~

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -433,11 +433,11 @@ type Windows struct {
 	// Resources contains information for handling resource constraints for the container.
 	Resources *WindowsResources `json:"resources,omitempty"`
 	// CredentialSpec contains a JSON object describing a group Managed Service Account (gMSA) specification.
-	CredentialSpec interface{} `json:"credentialspec,omitempty"`
+	CredentialSpec interface{} `json:"credentialSpec,omitempty"`
 	// Servicing indicates if the container is being started in a mode to apply a Windows Update servicing operation.
 	Servicing bool `json:"servicing,omitempty"`
 	// IgnoreFlushesDuringBoot indicates if the container is being started in a mode where disk writes are not flushed during its boot process.
-	IgnoreFlushesDuringBoot bool `json:"ignoreflushesduringboot,omitempty"`
+	IgnoreFlushesDuringBoot bool `json:"ignoreFlushesDuringBoot,omitempty"`
 	// HyperV contains information for running a container with Hyper-V isolation.
 	HyperV *WindowsHyperV `json:"hyperv,omitempty"`
 	// Network restriction configuration.
@@ -495,7 +495,7 @@ type WindowsNetwork struct {
 // WindowsHyperV contains information for configuring a container to run with Hyper-V isolation.
 type WindowsHyperV struct {
 	// UtilityVMPath is an optional path to the image used for the Utility VM.
-	UtilityVMPath string `json:"utilityvmpath,omitempty"`
+	UtilityVMPath string `json:"utilityVMPath,omitempty"`
 }
 
 // LinuxSeccomp represents syscall restrictions
--- a/config-windows.md
+++ b/config-windows.md
@@ -106,8 +106,8 @@ The following parameters can be specified:
 
 ## <a name="configWindowsCredentialSpec" />Credential Spec
 
-You can configure a container's group Managed Service Account (gMSA) via the OPTIONAL `credentialspec` field of the Windows configuration.
-The `credentialspec` is a JSON object whose properties are implementation-defined.
+You can configure a container's group Managed Service Account (gMSA) via the OPTIONAL `credentialSpec` field of the Windows configuration.
+The `credentialSpec` is a JSON object whose properties are implementation-defined.
 For more information about gMSAs, see [Active Directory Service Accounts for Windows Containers][gMSAOverview].
 For more information about tooling to generate a gMSA, see [Deployment Overview][gMSATooling].
 
@@ -130,13 +130,13 @@ You can indicate that a container should be started in a mode to apply pending s
 
 ## <a name="configWindowsIgnoreFlushesDuringBoot" />IgnoreFlushesDuringBoot
 
-You can indicate that a container should be started in an a mode where disk flushes are not performed during container boot via the OPTIONAL `ignoreflushesduringboot` field of the Windows configuration.
+You can indicate that a container should be started in an a mode where disk flushes are not performed during container boot via the OPTIONAL `ignoreFlushesDuringBoot` field of the Windows configuration.
 
 ### Example
 
 ```json
     "windows": {
-        "ignoreflushesduringboot": true
+        "ignoreFlushesDuringBoot": true
     }
 ```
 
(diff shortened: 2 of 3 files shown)


config-solaris.md
```diff
@@ -12,15 +12,15 @@ The SMF(Service Management Facility) FMRI which should go to "online" state befo
 "milestone": "svc:/milestone/container:default"
 ```
 
-## <a name="configSolarisLimitpriv" />limitpriv
+## <a name="configSolarisLimitPriv" />limitpriv
```
wking: If you're changing `limitpriv` to `limitPriv` below, I expect you want to change it in the header here as well.  You've currently changed the anchor, but not the visible text.

config-solaris.md
```diff
@@ -90,7 +90,7 @@ Mapped to `allowed-address` in the [zonecfg(1M)][zonecfg.1m_2] man page.
 * **`configureAllowedAddress`** *(string, OPTIONAL)* If configureAllowedAddress is set to true, the addresses specified by allowedAddress are automatically configured on the interface each time the container starts.
 When it is set to false, the allowedAddress will not be configured on container start.
 Mapped to `configure-allowed-address` in the [zonecfg(1M)][zonecfg.1m_2] man page.
-* **`defrouter`** *(string, OPTIONAL)* The value for the OPTIONAL default router.
+* **`defaultRouter`** *(string, OPTIONAL)* The value for the OPTIONAL default router.
```
wking: It seems inconsistent to me to expand `def` → `default` here but [*not* expand `Priv` → `Privilege`][1].  I personally prefer the abbreviated names, but am not sure how the Solaris folks want to handle this.  I'm fine with this expansion and/or the `Priv` → `Privilege` expansion happening in a follow-up PR if we want to isolate those changes from the camelCase changes.

[1]: https://github.com/opencontainers/runtime-spec/pull/859#issue-231471046

specs-go/config.go
```diff
@@ -495,7 +495,7 @@ type WindowsNetwork struct {
 // WindowsHyperV contains information for configuring a container to run with Hyper-V isolation.
 type WindowsHyperV struct {
 	// UtilityVMPath is an optional path to the image used for the Utility VM.
-	UtilityVMPath string `json:"utilityvmpath,omitempty"`
+	UtilityVMPath string `json:"utilityVMPath,omitempty"`
```
wking: On the unabbreviation front, this could be `utilityVirtualMachinePath`.  I don't see a way to draw a hard line for this, and it's orthogonal to camelCasing, but I thought I'd point out that `def` → `default` style expansion is something of a slippery slope.  My own fuzzy possition is that I like `def` → `default` but am less enthusiastic about `VM` → `VirtualMachine`.  And while I think `blkioThrottleWriteIOPSDevice` is a bit hard to read, it may be a better choice than `throttleWriteOperationsPerSecondPerDevice`.  But if anyone does see a way to draw a good line, it would be nice to document that line in `style.md`.

config-solaris.md
jhowardmsft: Ugh, sorry, I missed that one. Force update coming in a minute.

config-solaris.md
jhowardmsft: The link in the solaris markdown was wrong, but I'm updating it to http://docs.oracle.com/cd/E19253-01/816-5168/priv-str-to-set-3c/index.html as that seems the correct link.

However, as from that link the documentation seems to use `Priv`, I'm leaving it as-is.

As for `def` or `default`, `def` seems a very unusual abbreviation. Using `default` is also consistent with `defaultAction`. There seems no point that I can see for abbreviating it.

specs-go/config.go
jhowardmsft: Yes, I agree it's a fine line, but IMO, I would say `VM` is a very common abbreviation for 'Virtual Machine' and think it's OK as is.

config-solaris.md
jhowardmsft: I take it back, simply on account of it being `def` in the Solaris documentation. Reverted that change.

wking: > I take it back, simply on account of it being def in the Solaris documentation.

“Expose the platform abbreviation” sounds like a sane way to frame the policy.  But then perhaps we want “expose the platform casing” as well, in which case we'd want to leave this entry as [`defrouter`][1] and not touch it in this PR.

[1]: https://docs.oracle.com/cd/E36784_01/html/E36871/zonecfg-1m.html

jhowardmsft:
OK, I don't want to rathole for 1.0 with what I was originally intending fixing which was the camelCasing in Windows. Reverted all Solaris changes and that can be taken up separately if necessary.

wking:
On Thu, May 25, 2017 at 04:13:20PM -0700, John Howard wrote:
> Reverted all Solaris changes and that can be taken up separately if
> necessary.

That sounds reasonable to me, with the policy being something like
“use whatever contributors interested in platform can get the
maintainers to agree to” ;).

And 451856a2 looks like it covers all the camelCasing for
config-windows.md, although there are still references in the JSON
Schema that need updating.

jhowardmsft:
> And 451856a looks like it covers all the camelCasing for
> config-windows.md, although there are still references in the JSON
> Schema that need updating.

Thanks. Those too are updated now at c6338b37af32420507ca31841dd3974e8fc2899e

specs-go/config.go
anuthan: FYI, I no longer work on containers, I will find a point of contact from Solaris for you guys.
Sorry about this.

TomSweeneyRedHat:
LGTM

config-solaris.md
epilatow: hey @wking and @jhowardmsft.  i'll be taking over responsibility for the solaris components from @anuthan.  i've reviewed our use of camel case and i don't actually want to make any changes to it.  in most cases our property names mirror what's found in zonecfg ("defrouter" being an example), and camel case was only used for property names that had a hyphen in them.  for example, the zonecfg "allowed-address" property is mapped to "allowedAddress".  hopefully that sounds reasonable?

that said, some of our docs.oracle.com links are broken so i'll file and issue for that and get it fixed.

specs-go/config.go
epilatow: hey @wking and @jhowardmsft.  i'll be taking over responsibility for the solaris containers work from @anuthan.  i did an audit of our use of camel case and i'd prefer not to change it.  in general we've tried to have our property names map to zonecfg.1m properties.  "defrouter" being a good example.  the places where we have used camel case is for zonecfg properties that have hyphens in them.  so for example the zonecfg "allowed-address" property has been mapped to allowedAddress.  hopefully this seems sensible to folks.

that said, john correctly noted that some of our doc links are broken so i filed issue #864 on that.

mrunalp:
LGTM

vbatts:
LGTM

PR #860, opened 2017-05-25
config-linux: Drop redundant 'blkio' prefix from blockIO properties

wking:
These are long enough without the prefix, and `linux.resources.blockIO.blkioWeight`, etc. are just as specific as `linux.resources.blockIO.weight`.  Generated with:

```
$ sed -i s/blkioWeight/weight/g $(git grep -l blkioWeight)
$ sed -i s/blkioLeaf/leaf/g $(git grep -l blkioLeaf)
$ sed -i s/blkioThrottle/throttle/g $(git grep -l blkioThrottle)
```

Property-name polishing inspired by #859, in case we want to clean this sort of thing up before 1.0.  If not, we can just close this PR.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -328,16 +328,16 @@ For more information, see the kernel cgroups documentation about [blkio][cgroup-
 
 The following parameters can be specified to set up the controller:
 
-* **`blkioWeight`** *(uint16, OPTIONAL)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules.
-* **`blkioLeafWeight`** *(uint16, OPTIONAL)* - equivalents of `blkioWeight` for the purpose of deciding how much weight tasks in the given cgroup has while competing with the cgroup's child cgroups.
-* **`blkioWeightDevice`** *(array of objects, OPTIONAL)* - specifies the list of devices which will be bandwidth rate limited. The following parameters can be specified per-device:
+* **`weight`** *(uint16, OPTIONAL)* - specifies per-cgroup weight. This is default weight of the group on all devices until and unless overridden by per-device rules.
+* **`leafWeight`** *(uint16, OPTIONAL)* - equivalents of `weight` for the purpose of deciding how much weight tasks in the given cgroup has while competing with the cgroup's child cgroups.
+* **`weightDevice`** *(array of objects, OPTIONAL)* - specifies the list of devices which will be bandwidth rate limited. The following parameters can be specified per-device:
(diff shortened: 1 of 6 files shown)


hqhq:
I tend to keep them close to cgroup APIs, so +1 for closing this PR.

wking:
On Mon, May 29, 2017 at 07:07:36PM -0700, Qiang Huang wrote:
> I tend to keep them close to cgroup APIs, so +1 for closing this PR.

We don't use cgroup-namespacing for any other controller
[1,2,3,4,5,6,7].  Do you think we should add them there, with changes
like linux.resources.pids.limit → linux.resources.pids.pidsLimit (or
pidsMax to stay even closer to the pids.max kernel API)?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#device-whitelist
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#disable-out-of-memory-killer
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#memory
[4]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#cpu
[5]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#huge-page-limits
[6]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#network
[7]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config-linux.md#pids

wking:
If we want this before 2.0, it needs to go in the [1.0.0 milestone][1].  If we don't want this before 2.0, and are ok with the currently-inconsistent controller-prefixing, then maintainers should close this issue with a note to that effect.

The [current maintainer comment][2] seems to be in favor of controller-prefixing, and if we want to do that consistently (we [don't at the moment][3], currently only for the `blkio*` properties), then updating the other properties to include controller-prefixing would also have to happen before 1.0 if we want it to happen before 2.0.

[1]: https://github.com/opencontainers/runtime-spec/milestone/7
[2]: https://github.com/opencontainers/runtime-spec/pull/860#issuecomment-304757396
[3]: https://github.com/opencontainers/runtime-spec/pull/860#issuecomment-304768085

wking:
Rebased around #846 with a70bd8f → d8ca338.

crosbymichael:
LGTM

mrunalp:
LGTM

PR #861, opened 2017-05-26
glossary: Define "set" and "unset"

wking:
There [seems to be confusion about whether the empty string (and possible other values that match Go's zero values) qualifies as "set" or not][1].  This commit clarifies that Go zero values are not relevant to the spec, but it does not address how they should be handled (I'm leaving that to follow-up work).

[1]: https://github.com/opencontainers/runtime-spec/pull/823#discussion_r118620374

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -39,7 +39,7 @@ The following parameters can be specified to set up namespaces:
     The runtime MUST place the container process in the namespace associated with that `path`.
     The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
 
-    If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
+    If `path` is [unset](glossary.md#set), the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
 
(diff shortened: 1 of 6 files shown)


wking:
Something like this needs to be part of [1.0.0][1] if we want to clarify empty-string handling for `cgroupsPath` (about which @Mashimiao and I are [currently][2] [disagreeing][3]) in the 1.x releases.  Adding clear “set to a Go zero value” wording post-1.0 would be a breaking change for either @Mashimiao or me depending on the post-1.0 clarification.

[1]: https://github.com/opencontainers/runtime-spec/milestone/7
[2]: https://github.com/opencontainers/runtime-spec/pull/823#discussion_r118620374
[3]: https://github.com/opencontainers/runtime-spec/pull/823#discussion_r118752878

wking:
Rebased around #846 with 4a8ac77 → 87f66dc.

mrunalp:
I don't think we want to define set/unset.

wking:
On Thu, Jun 01, 2017 at 10:46:30AM -0700, Mrunal Patel wrote:
> I don't think we want to define set/unset. 

Do you have an opinion on whether:

  "cgroupsPath": ""

is set or unset?

PR #862, opened 2017-05-26
config: Drop redundant "unique within this map" annotation requirement

wking:
This condition landed in #510 with subsequent wording tweaks in #645 and #654.  However, since #584 we forbid duplicate keys on *all* objects (not just annotations), so this PR removes the redundant annotation-specific condition.

### DIFF
--- a/config.md
+++ b/config.md
@@ -442,7 +442,6 @@ Cleanup or debugging functions are examples of such a hook.
     If there are no annotations then this property MAY either be absent or an empty map.
 
     Keys MUST be strings.
-    Keys MUST be unique within this map.
     Keys MUST NOT be an empty string.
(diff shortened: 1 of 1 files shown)


wking:
This polishing is not changing the spec requirements (just making the spec more efficient about declaring those requirements), so it's not a problem if it lands post-1.0.  Can a maintainer tag it [v1.NEXT-maybe][1] if it's too much trouble to review now?

[1]: https://github.com/opencontainers/runtime-spec/labels/v1.NEXT-maybe

wking:
Rebased around #846 with aa115ac → eba7304.

mrunalp:
LGTM

crosbymichael:
LGTM

PR #863, opened 2017-05-26
config: Explicily make consoleSize unspecified if terminal is false or unset

wking:
The old language is from #563, where nobody commented on the “if attached” wording.  But reading the old line now, it's not clear to me what `consoleSize` means when `terminal` is not `true`.

This commit explicitly declares `consoleSize` unspecified in that condition, so runtimes are free to do what they want short of erroring out.  I considered making the property undefined or requiring it to be unset, but those seemed too strict given our permissive “[MUST ignore unknown properties][1]” extensibility requirement.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md#extensibility

### DIFF
--- a/config.md
+++ b/config.md
@@ -137,7 +137,8 @@ For all platform-specific configuration values, the scope defined below in the [
 
 * **`terminal`** (bool, OPTIONAL) specifies whether a terminal is attached to that process, defaults to false.
     As an example, if set to true on Linux a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
-* **`consoleSize`** (object, OPTIONAL) specifies the console size in characters of the terminal if attached, containing the following properties:
+* **`consoleSize`** (object, OPTIONAL) specifies the console size in characters of the terminal.
+    Runtimes MUST ignore `consoleSize` if `terminal` is `false` or unset.
     * **`height`** (uint, REQUIRED)
     * **`width`** (uint, REQUIRED)
 * **`cwd`** (string, REQUIRED) is the working directory that will be set for the executable.


wking:
I don't think the old language is clear enough for me to make a call about whether this is a breaking change (and therefore should happen before 1.0) or not.  If maintainers feel like the old language *is* clear enough to make that call, they can probably just close this with a “this is already clear enough” comment (although I'd personally like them to also include a longer description about what the old language means, since it's not clear to me).

If maintainers agree that the old language is unclear, then I think we want to land *something* to clear this up before 1.0, whether it's this PR or a maintainer-authored PR.

wking:
Rebased around #846 with f4255fb → 29074b4.

config.md
```diff
@@ -137,7 +137,8 @@ For all platform-specific configuration values, the scope defined below in the [
 
 * **`terminal`** (bool, OPTIONAL) specifies whether a terminal is attached to that process, defaults to false.
     As an example, if set to true on Linux a pseudoterminal pair is allocated for the container process and the pseudoterminal slave is duplicated on the container process's [standard streams][stdin.3].
-* **`consoleSize`** (object, OPTIONAL) specifies the console size in characters of the terminal if attached, containing the following properties:
+* **`consoleSize`** (object, OPTIONAL) specifies the console size in characters of the terminal.
+    `consoleSize` is unspecified if `terminal` is `false` or unset.
```
crosbymichael: Please change `unspecified` to `ignored`

wking: > Please change `unspecified` to `ignored`

Done with 29074b4 → ed6c9ef, which shuffles some words to make it clear that it is the runtime that must be doing the ignoring.

mrunalp:
LGTM

crosbymichael:
LGTM

Issue #864, opened 2017-05-26
Fix docs.oracle.com links


epilatow:
the zoneadmd.1m and priv-str-to-set.3c docs.oracle.com links in config-solaris.md are broken.  the zonecfg.1m link (in config.md and config-solaris.md) also points to the solaris 11.2 man page.  all the links should be updated to point to the latest solaris 11.3 man pages.

wking:
Do we want a PR fixing these links to be part of 1.0?  I'm happy to work one up, and it seems like it would be easy enough to review.

wking:
Actually, it looks like this was fixed by #865.  We can probably close this issue.

PR #865, opened 2017-05-26
fix docs.oracle.com links #864

epilatow:
hey all.  i recently started working on the solaris docker port and that means i'm also responsible for the solaris components of the container runtime spec.  while looking at the spec i noticed that some of our doc links were broken.  i've never worked directly on github.com before, so i figured that fixing this would be a good way to learn the ropes and introduce myself.  so apologies in advance if i've made any noob mistakes.  cheers.


vbatts:
yay. a solaris contributor.

LGTM

mrunalp:
LGTM

PR #866, opened 2017-05-28
config: RFC 2119 language for uid, gid, and additionalGids

wking:
Also link to POSIX for the associated definitions.

This adds runtime-oriented requirements similar to what runtime-tools is [already checking][1].  However, runtime-tools is currently allowing additional groups in the `getgroups()` response beyond those specified in `additionalGids`, with “superset” wording documented in opencontainers/runtime-tools@0ad3c80a.  I'm not sure if that superset logic is intended to be an open door, or if it is a workaround for the implementation-defined inclusion of the effective group ID in the [`getgroups()` response][2]:

> It is implementation-defined whether `getgroups()` also returns the effective group ID in the grouplist array.
>
> …
>
> Duplication may exist, but the application needs to call `getegid()` to be sure of getting all of the information.

@mrunalp, 2015 was a while ago, but do you remember why you went with the superset approach?

The wording I'm using glosses over a potential distinction between the real and effective group ID, since it's using the `gid` value as the effective group ID in the `additionalGids` comparison and as the real group ID in the `gid` comparison.  That could conceivably be an issue if `process.args` points to a setgid binary, but I think that is unlikely to be a major issue in the wild, and it is certainly avoidable within runtime-tools' compliance testing, so I haven't addressed it in this commit.

[1]: https://github.com/opencontainers/runtime-tools/blob/f49d861b69e359dc9278fcfa6948034a4af58705/cmd/runtimetest/main.go#L100-L123
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getgroups.html

### DIFF
--- a/config.md
+++ b/config.md
@@ -185,8 +185,11 @@ The user for the process is a platform-specific structure that allows specific c
 For Linux and Solaris based systems the user structure has the following fields:
 
 * **`uid`** (int, REQUIRED) specifies the user ID in the [container namespace](glossary.md#container-namespace).
+    The runtime MUST set the [real user ID][real-user-id] of the process such that a call to [`getuid(3)`][getuid.3] returns the configured value.
 * **`gid`** (int, REQUIRED) specifies the group ID in the [container namespace](glossary.md#container-namespace).
-* **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs (in the [container namespace](glossary.md#container-namespace) to be added to the process.
+    The runtime MUST set the [real group ID][real-group-id] of the process such that a call to [`getgid(3)`][getgid.3] returns the configured value.
+* **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs (in the [container namespace](glossary.md#container-namespace)) to be added to the process.
+    The runtime MUST set the [supplementary group IDs][supplementary-group-id] of the process such that the union of group IDs from a call to [`getgroups(3)`][getgroups.3] and a call to [`getegid(3)`][getegid] equals the union of the configured `additionalGroups` and `gid` values.
 
 _Note: symbolic name for uid and gid, such as uname and gname respectively, are left to upper levels to derive (i.e. `/etc/passwd` parsing, NSS, etc)_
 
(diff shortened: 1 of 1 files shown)


crosbymichael:
We don't want to bind the spec to platform specific implementations especially when it ends up having specific syscalls and documents referenced in the spec for fields.  We don't want to bind the spec to the fate of a kernel's API.

wking:
> We don't want to bind the spec to the fate of a kernel's API.

[`getuid(3)`][1] is a POSIX function and the behavior I am using here [has not changed][2] since [SVID][3].  [`getgid(3)`][4] and [`getgroups(3)`][5] have similarly long, stable histories, with [`getgroups(3)` being the youngest at issue 3][5] ([1988][6]).  I don't expect changes in the next 30 years either.  In the extremely unlikely event that Linux, Solaris, or POSIX drop support for any of those functions, I'd be very surprised if the `uid`, `gid`, or `additionalGids` properties survived the change to be ported over to newer APIs.

Without tying the properties to explicit tests, corner cases of the required runtime behavior are not clear to me.  For example, is the runtime free to inject as many additional groups as it sees fit?  Runtime-tools is currently allowing that (as I mentioned in the topic post for this PR).  Are the `uid` and `gid` values real or effective IDs (runtime-tools and this PR are both treating them as real IDs)?  What are the effective IDs of a new process (runtime-tools is not checking, and this PR is not placing any requirements on the runtime for the effective IDs)?  I'd really rather not inline documentation for all of that in the spec when we have well-established, stable POSIX APIs that we can punt to.  And I don't see how we can make those decisions in runtime-tools without either inline documentation in the spec or the kind of punt-to-testing-APIs that I define in this PR.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getuid.html
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getuid.html#tag_16_223_11
[3]: https://en.wikipedia.org/wiki/System_V_Interface_Definition
[4]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getgid.html#tag_16_196_11
[5]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getgroups.html#tag_16_199_11
[6]: http://pubs.opengroup.org/onlinepubs/9699919799/frontmatter/refdocs.html

PR #871, opened 2017-06-01
config-linux: "may" -> "MAY" for supplying linux.devices

wking:
The way that the runtime supplies these is “[truly optional][1]”, as long as they get supplied.  Also backtick `mknod`.

An uncontentious change spun off from #829.

[1]: https://tools.ietf.org/html/rfc2119#section-5


crosbymichael:
LGTM

mrunalp:
LGTM

PR #875, opened 2017-06-23
Allow created containers to be killed

mrunalp:

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -114,8 +114,7 @@ Any changes made to the [`config.json`](config.md) file after this operation wil
 `start <container-id>`
 
 This operation MUST [generate an error](#errors) if it is not provided the container ID.
-Attempting to start a container that does not exist MUST [generate an error](#errors).
-Attempting to start an already started container MUST have no effect on the container and MUST [generate an error](#errors).
+Attempting to `start` a container that is not [`created`](#state) MUST have no effect on the container and MUST [generate an error](#errors).
 This operation MUST run the user-specified program as specified by [`process`](config.md#process).
 This operation MUST generate an error if `process` was not set.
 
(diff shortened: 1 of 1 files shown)


crosbymichael:
LGTM

runtime.md
```diff
@@ -123,7 +123,7 @@ This operation MUST generate an error if `process` was not set.
 `kill <container-id> <signal>`
 
 This operation MUST [generate an error](#errors) if it is not provided the container ID.
-Attempting to send a signal to a container that is not running MUST have no effect on the container and MUST [generate an error](#errors).
+Attempting to send a signal to a container that is not running or not created MUST have no effect on the container and MUST [generate an error](#errors).
```
wking: I'd rather tie this more clearly to the already-well-defined state values.  Something like:

``Attempting to send a signal to a container that is neither [`created` nor `running`](#state) MUST have no effect…``

That's pretty close to what you're suggesting, although it:

* Uses “neither”/“nor” instead of repeating “not”.
* Links to the “State” section for formal definitions of `created` and `running` states.
* Uses backticks to strengthen the “these are formally defined terms” semantics of `running` and `created`.

[1]: https://github.com/opencontainers/runtime-spec/blame/fd0b0b6f7505ba8d24ee1fe59e59ca969b26843a/runtime.md#L21

tianon:
What's the use-case for killing a container that isn't running?  What's the expected behavior?

wking:
> What's the use-case for killing a container that isn't running?

Getting ready to `delete` a container that you hadn't `started`.  More on containers not getting `started` in #701.

> What's the expected behavior?

The signal gets sent to the container process, same as in the `running` case.  The container process will still be running a runtime-supplied program, but that doesn't effect sending signals to it.

wking:
While we're cleaning up the relationship between operations and container state, I think we can replace [these two `start` lines][1] with:

``Attempting to `start` a container that is not [`created`](#state) MUST have no effect on the container and MUST [generate an error](#errors).``

And [these two `delete` lines][2] with:

``Attempting to `delete` a container that is not [`stopped`](#state) MUST have no effect on the container and MUST [generate an error](#errors).``

[1]: https://github.com/opencontainers/runtime-spec/blame/fd0b0b6f7505ba8d24ee1fe59e59ca969b26843a/runtime.md#L117-L118
[2]: https://github.com/opencontainers/runtime-spec/blame/fd0b0b6f7505ba8d24ee1fe59e59ca969b26843a/runtime.md#L133-L134

runtime.md
dqminh: LGMT with @wking's suggestion,

mrunalp:
Updated.

tianon:
LGTM

mrunalp:
@tianon Orchestrator could have a sequence where it is first creating a bunch of containers and then starting them. If there is a failure in the start sequence it could just kill the remaining created containers.

crosbymichael:
LGTM

PR #878, opened 2017-06-30
config.md: document what type should be when bind-mount

Mashimiao:
I re-read config.md and found we really didn't define what
type should be if we want to do bind-mount.
So, I decided to add it.

### DIFF
--- a/config.md
+++ b/config.md
@@ -101,6 +101,7 @@ For POSIX platforms the `mounts` structure has the following fields:
 
 * **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
   * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660").
+    Note: `type` is empty or SHOULD be ignored when `options` contains "bind" or "rbind", which means bind-mount directory `source` from host into `destination` in container. This is similar with "--bind" or "--rbind" optins in [mount(8)][mount.8].
   * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 
 ### Example (Linux)


cyphar:
I think these semantics are implied already by referencing `mount(8)` and "Linux filesystems". But maybe it would be a useful note for users.

tianon:
Indeed, `type` is explicitly optional and doesn't include language about `bind` or `rbind`, so while I think this case is worth having more language around, I'm pushing it to post-1.0.

wking:
On Wed, Jul 05, 2017 at 08:01:49AM -0700, Tianon Gravi wrote:
> Indeed, `type` is explicitly optional and doesn't include language
> about `bind` or `rbind`, so while I think this case is worth having
> more language around, I'm pushing it to post-1.0.

But there's currently no spec wording about the (r)bind use case at
all, despite an example which uses it [1].  I'd rather re-open #771
and actually define those options now that opencontainers/runc#1460
has landed (removing the reason #771 was closed, which I think was
“this doesn't match the current runC implementation” although [2]
doesn't say).  #771 will need some re-rolling to bring it up to speed
with the current master, but I'm leaving the branch alone until the PR
is re-opened.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L120
[2]: https://github.com/opencontainers/runtime-spec/pull/771#event-1077200295

flx42:
I'm a little bit late to the party... but I agree it would be useful for runtime developers to clarify the expected semantic here.

For instance, it seems like `runc` doesn't ignore the `bind` type, as it should here (?)
https://github.com/opencontainers/runc/blob/f753f300ae6a725becddbacf15955a0a03b895cb/libcontainer/specconv/spec_linux.go#L272-L276

wking:
On Thu, Apr 12, 2018 at 10:11:32PM +0000, Felix Abecassis wrote:
> For instance, it seems like `runc` doesn't ignore the `bind` type, as it should here (?)
> https://github.com/opencontainers/runc/blob/f753f300ae6a725becddbacf15955a0a03b895cb/libcontainer/specconv/spec_linux.go#L272-L276

I've filed opencontainers/runc#1753 for that, and @cyphar sounded
positive in early review [1].  However, the associated spec PR (#954)
was recently closed by @vbatts [2].  I'm not sure where that leaves us
on the “should” front.  But however the spec maintainers want bind
mounts to work, I think they should get some docs about them landed so
the rest of us can do the right thing ;).

[1]: https://github.com/opencontainers/runc/pull/1753#issuecomment-370988991
[2]: https://github.com/opencontainers/runtime-spec/pull/954#issuecomment-378758610

flx42:
`bind` is clearly not a valid type. In general, if an unknown type is present, I think it MUST be ignored (which is what fstab does, AFAIK).

wking:
On Thu, Apr 12, 2018 at 10:51:36PM +0000, Felix Abecassis wrote:
> `bind` is clearly not a valid type.

I dunno about “clearly” ;).  The current spec has nothing about “bind”
outside of an informative example [1].  *I* agree that the current
normative spec wording [2] makes it sound like values should be
entries in /proc/filesystems, but @vbatts explicitly disagrees [3].
And there's certainly no MUST or other RFC 2119 wording in the ‘type’
specification to restrict either the configured values or the runtime
interpretation of those values.

> In general, if an unknown type is present, I think it MUST be
> ignored (which is what fstab does, AFAIK).

That's fine with me, and the wording in this PR is a step in that
direction (although it's currently only SHOULDing the runtime to
ignore the value).  I'm just saying that the current situation is
unclear, and that I wish the spec had more language that went into
detail on how bind configuration works.  Which is basically what
@tianon said back in [4].  Now that we're 274 days post 1.0.0, perhaps
some maintainers can circle back around to this PR?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#example-linux
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#posix-platform-mounts
[3]: https://github.com/opencontainers/runtime-spec/pull/954#issuecomment-378761811
[4]: https://github.com/opencontainers/runtime-spec/pull/878#issuecomment-313129886

vbatts:
Bind is an odd one. There are places on Linux that it's treated like type,
but yes it's not an absolute type. It seems like a clarity or usability to
keep it as a type for OCI.

wking:
> Bind is an odd one.

So do you agree with @tianon that [it could use more spec wording][1]?  If so, what sort of wording would you like to see?

[1]: https://github.com/opencontainers/runtime-spec/pull/878#issuecomment-313129886

cyphar:
@flx42 

> For instance, it seems like runc doesn't ignore the bind type, as it should here (?)
> https://github.com/opencontainers/runc/blob/f753f300ae6a725becddbacf15955a0a03b895cb/libcontainer/specconv/spec_linux.go#L272-L276

This is a fairly long-standing bug in `runc` and is something that will require quite a lot of work to fix (in fact most of `rootfs_linux.go` needs a good clean up to be honest). However, it is definitely *incorrect* currently, so I wouldn't use that code as an example of what should be done.

wking:
> This is a fairly long-standing bug in `runc` and is something that will require quite a lot of work to fix (in fact most of `rootfs_linux.go` needs a good clean up to be honest).

A general `rootfs_linux.go` may take some work, but I think opencontainers/runc#1753 already has all we need for *this* issue (including the integration tedsts you'd asked for earlier).  If you need anything else there, could you comment in that PR?

cyphar:
Ah, I forgot about that patch. Yes that works and will fix the issue (and we should probably merge it), but it's just taking advantage of the already broken code -- fixing the broken-ness will be much more annoying to do. :wink:

PR #883, opened 2017-07-05
Add reference to railcar impl

caniszczyk:
Add https://github.com/oracle/railcar

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

### DIFF
--- a/implementations.md
+++ b/implementations.md
@@ -6,6 +6,7 @@ If you know of any associated projects that are not listed here, please file a p
 ## <a name="implementationsRuntimeContainer" />Runtime (Container)
 
 * [opencontainers/runc][runc] - Reference implementation of OCI runtime
+* [oracle/railcar][railcar] - A Rust implementation of OCI Runtime
 
 ## <a name="implementationsRuntimeVirtualMachine" />Runtime (Virtual Machine)
 


caniszczyk:
This can be post v1.0 obviously

implementations.md
```diff
@@ -6,6 +6,7 @@ If you know of any associated projects that are not listed here, please file a p
 ## <a name="implementationsRuntimeContainer" />Runtime (Container)
 
 * [opencontainers/runc][runc] - Reference implementation of OCI runtime
+* [oracle/railcar][railcar] - A Rust implementation of OCI Runtime
```
wking: You need to define the `railcar` [link label][1] [at the bottom of this file][2].

[1]: https://daringfireball.net/projects/markdown/syntax#link
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc6/style.md#links

cyphar:
:+1:

tianon:
IMO if we add this, we should also make an effort to find other implementations to list.  Doesn't `rkt` also have something in-progress, for example?

cyphar:
`rkt` does not yet have support (internal or external) for OCI runtime-spec. They do have support for image-spec. But this is something that I've talked to @jonboulle about working on in the _very_ near future (effectively implementing an OCI `stage1` and `stage2` as well as including OCI support for internal configuration).

PR #885, opened 2017-07-10
defs-linux: add action to to syscalls required

q384566678:
According to https://github.com/opencontainers/runtime-spec/blame/master/config-linux.md#L544.

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -97,7 +97,8 @@
                 }
             },
             "required": [
-                "names"
+                "names",
+                "action"
             ]
         },
         "Major": {


vrothberg:
:+1:, but the commit message could be more verbose by, for instance, pointing to https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md which requires the action to be set.

q384566678:
> :+1:, but the commit message could be more verbose by, for instance, pointing to https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md which requires the action to be set.

updated, thanks.

hqhq:
LGTM

crosbymichael:
LGTM

TomSweeneyRedHat:
LGTM, but it would be nice to at least remove one "to" from the commit message.

PR #887, opened 2017-07-10
schema: add required to user

q384566678:
https://github.com/opencontainers/runtime-spec/blame/master/config.md#L218

### DIFF
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -100,6 +100,10 @@
                 "user": {
                     "id": "https://opencontainers.org/schema/bundle/process/user",
                     "type": "object",
+                    "required": [
+                        "uid",
+                        "gid"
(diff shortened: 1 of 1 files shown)


schema/config-schema.json
```diff
@@ -100,6 +100,10 @@
                 "user": {
                     "id": "https://opencontainers.org/schema/bundle/process/user",
                     "type": "object",
+                    "required": [
+                        "uid",
```
wking: These are not required, because [they are POSIX-only][1] and `config-schema.json` also has to cover Windows.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc6/config.md#posix-platform-user

hqhq:
Close as they are not required.

PR #894, opened 2017-07-12
[Runtime] Allow for excluding pid from state

RobDolinMS:
Re-base of #459; does not impact functionality but may be helpful for certification

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -23,6 +23,7 @@ The state of a container includes the following properties:
 
     Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`** (int, REQUIRED when `status` is `created` or `running`) is the ID of the container process, as seen by the host.
+    The pid MAY be excluded or set to a value indicating empty or null if it is not seen by the host. 
 * **`bundle`** (string, REQUIRED) is the absolute path to the container's bundle directory.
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -23,6 +23,7 @@ The state of a container includes the following properties:
 
     Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
 * **`pid`** (int, REQUIRED when `status` is `created` or `running`) is the ID of the container process, as seen by the host.
+    The pid MAY be excluded or set to a value indicating empty or null if it is not seen by the host.
```
wking: As you have it, this conflicts with the REQUIRED in the line above.  Maybe pull this entry down below this list and guard it (like we do [here][1]).

    For systems where the container process ID is visible from the [runtime namespace](glossary.md#runtime-namespace), the state object supports the following additional properties:

    * **`pid`** (int, REQUIRED when `status` is `created` or `running`) is the ID of the container process, as seen by the host.

That also gives us an opportunity to address #731 if we want to.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc6/config.md#posix-process

crosbymichael:
We are going to take #895

wking:
With the PR closed, someone with write access should probably delete the local `RobDolinMS-pid-not-visible` branch.

Issue #772, opened 2017-04-15
specs-go: Consistent pointer policy for invalid-value detection?


wking:
The nominal original policy for this repository was to use pointers for Go properties that were not always REQUIRED.  This was codified in #287, and then reversed in #656 to be anti-pointer.  However, in #764, maintainers are [again asking for pointers][1].  I think we should have a consistent position on this issue, so here's an attempt to summarize the arguments on each side:

## Anti-pointer

* Pointers make the Go code more complicated (see @vbatts [here][2] and @vishh [here][10]).  And as evidence that this is a thing, see @cyphar's recent bugfix in opencontainers/runtime-tools#361.

* Distinguishing `null` / unset from the zero value is not useful when both have the same semantics (see @vishh [here][13] and [here][17] and @mrunalp [here][14])

    In some specific cases:

    * An explicitly empty-string `cgroupsPath` can mean “let the runtime handle it” which is [what unset means][20] (see @mrunap [here][21]).

* Distinguishing `null` / unset from the zero value is not useful when the zero value is invalid (see @crosbymichael [here][22]).

    In some specific cases:

    * No need for pointers on `Cpus` or `Mems`, since we can have a runtime no-op for all three cases (`null` / unset / empty-string) (see @crosbymichael [here][22])

* String pointers are never useful (see @crosbymichael [here][18]).

## Pro-pointer

* Pointers provide a consistent way for Go consumers to determine if the property was `null` / unset (see me [here][3]; @cyphar [here][4]; @crosbymichael “where the zero value means something for our settings” [here][5], @hqhq [here][11], @munralp [here][12], @duglin [here][19]).

    In some specific cases:

    * @mrunalp [here][6], @hqhq [here][7], and @vishh [here][8] and [here][15], and me [here][16] on making `DisableOOMKiller` a pointer to distinguish “set true”, “set false”, and “leave alone”.

    * @cyphar [here][9] on a pointer for `Cpus` so we can fail on an invalid empty-string, backed up by @mrunalp [here][27]

    * @crosbymichael [here][23] on a pointer for `Timeout` so we can fail on an invalid zero.

* golang/go#11939 means that `omitempty` requires pointer properties for structs.  More on this in opencontainers/image-spec#625.

## Consistent policy

While “pointers for every `omitempty` property is nice and consistent, the consensus position seems to be that the increased complication of pointer handling outweighs the desire for consistency there.  That's fine.

[...]

PR #895, opened 2017-07-12
runtime: do not require pid while created

darrenstahlmsft:
It seems odd to me that the `pid` is REQUIRED when the `status` is `created`, as at that time, the container process is not running. How would one know the `pid` for a process that is not yet running (or created, as the lifecycle only requires that `create` creates the namespace, and prohibits it from creating the process)?

If this is not the case, then the language for container process should be clarified.

Signed-off-by: Darren Stahl <darst@microsoft.com>

### DIFF
--- a/specs-go/state.go
+++ b/specs-go/state.go
@@ -9,7 +9,7 @@ type State struct {
 	// Status is the runtime status of the container.
 	Status string `json:"status"`
 	// Pid is the process ID for the container process.
-	Pid int `json:"pid"`
+	Pid *int `json:"pid"`
 	// Bundle is the path to the container's bundle directory.
 	Bundle string `json:"bundle"`
 	// Annotations are key values associated with the container.
--- a/runtime.md
+++ b/runtime.md
@@ -22,7 +22,7 @@ The state of a container includes the following properties:
     * `stopped`: the container process has exited (step 7 in the [lifecycle](#lifecycle))
 
     Additional values MAY be defined by the runtime, however, they MUST be used to represent new runtime states not defined above.
-* **`pid`** (int, REQUIRED when `status` is `created` or `running`) is the ID of the container process, as seen by the host.
+* **`pid`** (int, REQUIRED when `status` is `created` or `running` under Linux, OPTIONAL on other platforms) is the ID of the container process, as seen by the host.
 * **`bundle`** (string, REQUIRED) is the absolute path to the container's bundle directory.
     This is provided so that consumers can find the container's configuration and root filesystem on the host.
 * **`annotations`** (map, OPTIONAL) contains the list of annotations associated with the container.
--- a/schema/state-schema.json
+++ b/schema/state-schema.json
@@ -40,7 +40,6 @@
         "ociVersion",
         "id",
         "status",
-        "pid",
         "bundle"
     ]
 }


wking:
On Wed, Jul 12, 2017 at 08:26:03PM +0000, Darren Stahl wrote:
> It seems odd to me that the `pid` is REQUIRED when the `status` is
> `created`, as at that time, the container process is not running.

It *is* running.  The container process MUST be running by the time
‘create’ finishes, for example, to hold open a new PID namespace.  The
difference between ‘created’ and ‘running’ is just whether that
container process has executed the user-specified program or not [1].
So the clone(2) (or fork(2) or whatever) that creates the *process*
happens in ‘create’.  And the execvpe(3) (or whatever) that executes
the user-specified program happens in ‘start’.

Without a container process (running a runtime-supplied program)
running between ‘create’ and ‘start’, doing things like custom cgroup
management is impossible (no container PID to add to the cgroup your
setting up).

I think we can close this PR.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/runtime.md#L20

wking:
I gave a Linux-centrix answer to a potentially Windows-motivated issue
[1].  Getting a bit more generic, the reason we supply ‘pid’ in the
state JSON is so runtime-callers can reach around the runtime to do
whatever setup they need without having to go through a hook (lots
more on this in #384).  I'm not sure how that decomposes into Windows
primitives, but without something PID-ish, I don't see how Windows
users are supposed to manipulate their container when they need/want
to reach around the runtime and touch things directly.  Maybe that
reaching around is not going to be supported on Windows?  If so, we
probably just want to make the ‘pid’ entry here POSIX-platform
specific.  Or maybe Linux specific, since the Solaris folks had
previously said something about a start-time process as well [2].

[1]: https://github.com/opencontainers/runtime-spec/pull/895#issuecomment-314888721
[2]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-07-13-17.03.log.html#l-24

darrenstahlmsft:
Yes, this was a Windows motivated PR, but without the context of the separation of create and start, it seemed odd that a PID would exist before the process. There is no runtime supplied `container process` on Windows. The user-specified program in Windows containers (both Hyper-V and Windows Server) must be a new process created at the time of `start` as a Windows containers do not have a analogy to execvpe(3).

I can't comment on if going behind the back of the Windows runtime will be supported or not, but that would be done using the container `id` rather than `pid`.

wking:
On Wed, Jul 12, 2017 at 08:55:13PM +0000, Darren Stahl wrote:
> I can't comment on if going behind the back of the Windows runtime
> will be supported or not, but that would be done using the container
> `id` rather than `pid`.

On Linux, the container ID is a userspace concept maintained by the
runtime.  On Windows and Solaris, container IDs seem to be kernel
primitives.  I think that's a reasonable argument for making ‘pid’
Linux-only (which would also obsolete #894).  It would make some
things (like prctl calls to configure Solaris container processes
between ‘create’ and ‘start’ [1]) impossible, but we can always add it
back in later on.  On the other hand, if we keep ‘pid’ required for
1.0, then relaxing that requirement later is more likely to annoy
state consumers (although the state JSON is not formally covered by
runtime-spec's SemVer [2]).

[1]: https://docs.oracle.com/cd/E23824_01/html/821-1460/rmctrls.task-33.html
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L17

vbatts:
LGTM

vbatts:
Oh i forgot the update needed to be platform specifc.

vbatts:
rejected

vbatts:
NLGTM

mrunalp:
We could get this into a 1.0.1 per the discussion on the call today.

darrenstahlmsft:
What would be the suggested way to have a field that is not required in the `created` state on Windows, but is still required in the `running` state?

Would it be better to just exclude this field on Windows instead as suggested above?

crosbymichael:
Maybe:

`(int, REQUIRED when `status` is `created` under linux, OPTIONAL on other platforms) is the ID of the container process, as seen by the host.`

wking:
> Maybe:
>
>    (int, REQUIRED when `status` is `created` under linux, OPTIONAL on other platforms) is the ID of the container process, as seen by the host.

Our current pattern is to use either “Linux” (e.g. [here][1]) or `linux` (e.g. [here][2]).  I'd rather not use a lowercase “linux” without backticks.  Otherwise that wording looks good to me.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L69
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc6/config.md#L347

crosbymichael:
@wking fine with me.  @darrenstahlmsft please make this change ASAP for 1.0

wking:
Other changes we probably want to make (from [here][1]):

* add a pointer to [the Go property][2].
* remove [`pid` from the required section of the JSON Schema][3].

[1]: https://github.com/opencontainers/runtime-spec/pull/894#pullrequestreview-49609126
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc6/specs-go/state.go#L12
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc6/schema/state-schema.json#L43

darrenstahlmsft:
Updated and added changes suggested by @wking

darrenstahlmsft:
@crosbymichael I'm fine if you want to take the carry over this one.

specs-go/state.go
```diff
@@ -9,7 +9,7 @@ type State struct {
 	// Status is the runtime status of the container.
 	Status string `json:"status"`
 	// Pid is the process ID for the container process.
-	Pid int `json:"pid"`
+	Pid *int `json:"pid"`
```
wking: An alternative would be to leave off the pointer and use `omitempty`, since zero isn't a valid PID.  Which way we should go depends on the unresolved #772, but the maintainers may have a one-off opinion on this.

wking: > An alternative would be to leave off the pointer and use `omitempty`…

#897 is currently going this way.

darrenstahlmsft: I talked to @crosbymichael offline. He's going to carry this. I'll close this PR in favour of #897

darrenstahlmsft:
Closing in favour of #897

PR #896, opened 2017-07-12
config-linux: Move 'disableOOMKiller' under 'memory'

### LINKED ISSUE #806: config-linux: Move 'disableOOMKiller' under 'memory'?

wking:
Fixes #806.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -253,22 +253,6 @@ Each entry has the following structure:
     ]
 ```
 
-### <a name="configLinuxDisableOutOfMemoryKiller" />Disable out-of-memory killer
-
-`disableOOMKiller` contains a boolean (`true` or `false`) that enables or disables the Out of Memory killer for a cgroup.
-If enabled (`false`), tasks that attempt to consume more memory than they are allowed are immediately killed by the OOM killer.
-The OOM killer is enabled by default in every cgroup using the `memory` subsystem.
-To disable it, specify a value of `true`.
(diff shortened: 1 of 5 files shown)


crosbymichael:
LGTM

mrunalp:
LGTM

PR #898, opened 2017-07-14
schema/defs-linux: change weight type to uint16

0x0916:
Signed-off-by: 0x0916 <w@laoqinren.net>

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -151,7 +151,7 @@
             }
         },
         "weight": {
-            "type": "integer"
+            "$ref": "defs.json#/definitions/uint16"
         },
         "blockIODevice": {
             "type": "object",


dqminh:
LGTM

stevvooe:
This change is probably not backwards compatible and against best practices. Do you have any documentation citing this as a `uint16`?

cc @tonistiigi @crosbymichael

0x0916:
@stevvooe Please refer to https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#block-io

also, the kernel document: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt

```
- blkio.weight
	- Specifies per cgroup weight. This is default weight of the group
	  on all the devices until and unless overridden by per device rule.
	  (See blkio.weight_device).
	  Currently allowed range of weights is from 10 to 1000.

```

stevvooe:
@0x0916 How does that counter any of the concerns? The kernel API doesn't mention that this is restricted to a `uint16`. The specification currently calls for an integer and this requires a `uint16`. If others were expecting this to take something outside of a `uint16`, it would be considered backwards incompatible.

It is unlikely that this will break anyone, but I'd urge caution.

dqminh:
@stevvooe this should not break anyone, it also matches the type in spec-go which has been uint16 since forever. I think this is more like fixing a typo.

stevvooe:
@dqminh Didn't we recently [learn our lesson](https://github.com/opencontainers/runtime-spec/pull/876) about liberal use of unsigned integers? The only use case of unsigned integers is when they map a field to a specific bit width. Unsigned integers should not be used to validate that something is positive or within a certain range of values. That is the job of validation.

wking:
To help this PR along, here are the current types for `weight` in the current master (34170989dc28):

* `uint16` [in `config-linux.md` under `blockIO`][1].
* `uint16` [in `config-linux.md` under `weightDevice`][2].
* `*uint16` [in `config.go` under `LinuxBlockIO`][3].
* `*uint16` [in `config.go` under `LinuxWeightDevice`][4].
* `integer` [in `defs-linux.json`][5] via [`blockIO` in `config-linux.json`][6].
* `integer` [in `defs-linux.json`][5] via [`weightDevice` in `config-linux.json`][7] and [`blockIODeviceWeight` in `defs-linux.json`][8].

The `weight` entry in `defs-linux.json` (which is what this PR is touching as of 574182a3fed7) is *also* consumed by the `leafWeight` entries:

```
$ git describe
v1.0.0-39-g3417098
$ git grep -B2 definitions/weight schema/
schema/config-linux.json-                            "weight": {
schema/config-linux.json-                                "id": "https://opencontainers.org/schema/bundle/linux/resources/blockIO/weight",
schema/config-linux.json:                                "$ref": "defs-linux.json#/definitions/weight"
--
schema/config-linux.json-                            "leafWeight": {
schema/config-linux.json-                                "id": "https://opencontainers.org/schema/bundle/linux/resources/blockIO/leafWeight",
schema/config-linux.json:                                "$ref": "defs-linux.json#/definitions/weight"
--
schema/defs-linux.json-                    "properties": {
schema/defs-linux.json-                        "weight": {
schema/defs-linux.json:                            "$ref": "#/definitions/weight"
schema/defs-linux.json-                        },
schema/defs-linux.json-                        "leafWeight": {
schema/defs-linux.json:                            "$ref": "#/definitions/weight"
```

And `leafWeight` is also a `unit16` in master:

```
$ git grep 'leafWeight.*OPTIONAL' config-linux.md
config-linux.md:* **`leafWeight`** *(uint16, OPTIONAL)* …
config-linux.md:    * **`leafWeight`** *(uint16, OPTIONAL)* …
$ git grep 'leafWeight' specs-go/
specs-go/config.go:     LeafWeight *uint16 `json:"leafWeight,omitempty"`
specs-go/config.go:     LeafWeight *uint16 `json:"leafWeight,omitempty"`
```

So the *only* thing in that whole chain that is not `uint16` (modulo pointer) is the `defs-linux.json` entry being fixed in 574182a3fed7.

@stevvooe [brings up][9] #876, and I have no problem with runtime-spec maintainers deciding that `weight` and/or `leafWeight` should really have been `int64` (or whatever).  But whatever they have chosen as the type in the normative Markdown spec should be reflected in the informative, provided as-is JSON Schema and Go types provided alongside the spec.  574182a3fed7 is making things consistent assuming the runtime-spec maintainers want to stick with the [Markdown][1] [`uint16`][2].  If they decide to switch to a different type, we'd need a separate PR.  But just leaving things as they stand in master leaves the JSON Schema out of sync with the normative Markdown for no purpose.

[1]: https://github.com/opencontainers/runtime-spec/blame/34170989dc28be797f7078f88df156a474d7fd11/config-linux.md#L328
[2]: https://github.com/opencontainers/runtime-spec/blame/34170989dc28be797f7078f88df156a474d7fd11/config-linux.md#L334
[3]: https://github.com/opencontainers/runtime-spec/blob/34170989dc28be797f7078f88df156a474d7fd11/specs-go/config.go#L258
[4]: https://github.com/opencontainers/runtime-spec/blob/34170989dc28be797f7078f88df156a474d7fd11/specs-go/config.go#L243
[5]: https://github.com/opencontainers/runtime-spec/blob/34170989dc28be797f7078f88df156a474d7fd11/schema/defs-linux.json#L153-L154
[6]: https://github.com/opencontainers/runtime-spec/blob/34170989dc28be797f7078f88df156a474d7fd11/schema/config-linux.json#L69
[7]: https://github.com/opencontainers/runtime-spec/blob/34170989dc28be797f7078f88df156a474d7fd11/schema/config-linux.json#L107
[8]: https://github.com/opencontainers/runtime-spec/blob/34170989dc28be797f7078f88df156a474d7fd11/schema/defs-linux.json#L181
[9]: https://github.com/opencontainers/runtime-spec/pull/898#issuecomment-316187331

vbatts:
_house keeping:_ as this is in just the jsonschema and matches what the docs say, i'm cool with it

LGTM

tianon:
LGTM

This also matches the Go implementation, which is convincing IMO.

PR #901, opened 2017-07-19
config: Simplify title to "Configuration"

wking:
The capitalization in “Container Configuration file” (which we've used since #176) was halfway between “Title Case” and “Sentence case”.  The current spec isn't particularly consistent (e.g. we have both “Specification version” and “POSIX-platform Mounts”), but the ToC has used “Configuration” for this file since #626, so dodge the sentence/title issue and use that here too.

### DIFF
--- a/config.md
+++ b/config.md
@@ -1,4 +1,4 @@
-# <a name="containerConfigurationFile" />Container Configuration file
+# <a name="configuration" />Configuration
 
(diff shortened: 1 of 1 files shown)


hqhq:
LGTM

crosbymichael:
LGTM

PR #904, opened 2017-07-19
schema/config-linux: Check swappiness for zero to 100 range

wking:
Catch the JSON Schema up with #876, which carried the kernel's 0 to 100 range into the spec.  I'd [argued against carrying those limits locally][1], but as long as we have them in the normative Markdown spec we should be enforcing them in the JSON Schema.

As of v1.0.0, the Markdown line carrying the kernel's limits is [here][2].

[1]: https://github.com/opencontainers/runtime-spec/pull/876#discussion_r123856763
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L272

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -188,7 +188,7 @@
                             },
                             "swappiness": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/swappiness",
-                                "$ref": "defs.json#/definitions/uint64"
+                                "$ref": "defs.json#/definitions/percent"
                             },
                             "disableOOMKiller": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/disableOOMKiller",


wking:
I think this belongs in the [v1.0.Z milestone][1], since it brings the JSON Schema closer to the normative Markdown.  It will be a breaking change for anyone who had been using the JSON Schema to validate fields other than `swappiness` but *didn't* want to catch invalid-acording-to-our-Markdown `swappiness` values, and I doubt any such users exist.

[1]: https://github.com/opencontainers/runtime-spec/milestone/16

schema/config-linux.json
```diff
@@ -188,7 +188,7 @@
                             },
                             "swappiness": {
                                 "id": "https://opencontainers.org/schema/bundle/linux/resources/memory/swappiness",
-                                "$ref": "defs.json#/definitions/uint64"
+                                "$ref": "defs.json#/definitions/percent"
```
hqhq: But it's not actually a percent, just happen to have the same range.

crosbymichael:
REJECT

Like @hqhq said, just because the numbers are from 0-100 do not make it a percentage and its not in the kernel either.

wking:
> Like @hqhq said, just because the numbers are from 0-100 do not make it a percentage...

So I can rename our current `percent` range to `zeroThroughOneHundred`.

> ... and it's not in the kernel either.

Our JSON Schema is not about validating things that are in the kernel (although as @justincormack [pointed out][2], this [is in the kernel][3]), it's about validating against our Markdown spec.  And as I poimted out in this PR's opening post [we limit `swappiness` to 0 through 100 there][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L272
[2]: https://github.com/opencontainers/runtime-spec/pull/876#discussion_r123857396
[3]: https://github.com/torvalds/linux/blob/v4.12/mm/vmscan.c#L149

PR #911, opened 2017-08-18
Fix indentation in example

gmjosack:


wking:
df69020b1 is missing a Signed-off-by [1], but other than that it looks
good to me.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/README.md#sign-your-work

PR #858, opened 2017-05-25
config: Bump POSIX spec references to IEEE Std 1003.1-2008, 2016 Edition

wking:
These references had been using IEEE Std 1003.1, 2004 Edition, but:

    $ curl -s http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html | grep -B2 'newer edition'
    <center><font size="2">The Open Group Base Specifications Issue 6<br>
    IEEE Std 1003.1, 2004 Edition<br>
    Copyright &copy; 2001-2004 The IEEE and The Open Group, All Rights reserved.</font></center><center><font color="red">A newer edition of this document exists <a href="http://pubs.opengroup.org/onlinepubs/9699919799/" target="_parent">here</a></font></center>

Shifting to 2016 also syncs us with the [`file` reference][0].  The initial reasoning for the 2004 edition is unclear to me (more on that in 70858bc4).

The [change-log for the `exec` page][1] doesn't list any relevant changes, and skimming a diff of the two HTML pages didn't turn up anything significant.

Diffing the two HTML pages for environment variables also turned up no significant changes.  Both definition reference the Portable Character Set for uppercase letters, digits, and underscore, but the [`U####`][2] [values][3] for those have not changed.

[0]: https://github.com/opencontainers/runtime-spec/blame/9151c778a59e281f8c55e17c34245ea0f6a39c37/config-linux.md#L665
[1]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html#tag_16_111_14
[2]: http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap06.html#tag_06_01
[3]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tag_06_01

PR #913, opened 2017-08-20
Fix link on IEEE Std 1003.1-2008

rutsky:
/cc: @wking


wking:
On Sun, Aug 20, 2017 at 11:01:43PM +0000, Vladimir Rutsky wrote:
> /cc: @wking

Oops, it looks like I broke these in #858.  Sorry about that :/.

Personally, I find “functions” more understandable than “xsh”, so I'd
like to have the link label definition updated to match the reference:

  [ieee-1003.1-2008-functions-exec]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html

(vs. db58e01e5d which updates the reference to match the link label
definition).  But either way would clearly be better than the current
broken link ;).

rutsky:
@wking updated according to your suggestion.

hqhq:
LGTM

mrunalp:
LGTM

tianon:
LGTM

crosbymichael:
LGTM

PR #914, opened 2017-08-25
access of device cgroup is optional depends on type

Mashimiao:
In Linux, only when type is a, access can be unset. So, I think if we have to make access as OPTIONAL, let it to be `rwm` as default may be better.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -222,10 +222,10 @@ Each entry has the following structure:
 
 * **`allow`** *(boolean, REQUIRED)* - whether the entry is allowed or denied.
 * **`type`** *(string, OPTIONAL)* - type of device: `a` (all), `c` (char), or `b` (block).
-    Unset values mean "all", mapping to `a`.
+    Unset value means "all", mapping to `a`.
 * **`major, minor`** *(int64, OPTIONAL)* - [major, minor numbers][devices] for the device.
-    Unset values mean "all", mapping to [`*` in the filesystem API][cgroup-v1-devices].
-* **`access`** *(string, OPTIONAL)* - cgroup permissions for device.
+    Unset value means "all", mapping to [`*` in the filesystem API][cgroup-v1-devices].
+* **`access`** *(string, REQUIRED unless `type` is `a`)* - cgroup permissions for device.
     A composition of `r` (read), `w` (write), and `m` (mknod).
 
 #### Example


crosbymichael:
We usually error towards secure and explicit.  This change would be the opposite of both.

wking:
On Fri, Aug 25, 2017 at 07:37:59AM +0000, Ma Shimiao wrote:
> In Linux, only when type is a, access can be unset. So, I think if
> we have to make access as OPTIONAL, let it to be `rwm` as default
> may be better.

Or REQUIRE it when type is equivalent to ‘a’?

Mashimiao:
On 08/26/2017 12:32 AM, W. Trevor King wrote:
> Or REQUIRE it when type is equivalent to ‘a’?

Just the opposite.
When type is `a`, access can be OPTIONAL.
Otherwise, access must be REQURIE.

Mashimiao:
updated, PTAL

config-linux.md
```diff
@@ -222,11 +222,12 @@ Each entry has the following structure:
...
 * **`allow`** *(boolean, REQUIRED)* - whether the entry is allowed or denied.
 * **`type`** *(string, OPTIONAL)* - type of device: `a` (all), `c` (char), or `b` (block).
-    Unset values mean "all", mapping to `a`.
+    Unset value means "all", mapping to `a`.
 * **`major, minor`** *(int64, OPTIONAL)* - [major, minor numbers][devices] for the device.
-    Unset values mean "all", mapping to [`*` in the filesystem API][cgroup-v1-devices].
+    Unset value means "all", mapping to [`*` in the filesystem API][cgroup-v1-devices].
 * **`access`** *(string, OPTIONAL)* - cgroup permissions for device.
     A composition of `r` (read), `w` (write), and `m` (mknod).
+    * When type is not `a`, `access` is REQUIRED.
```
wking: We don't define our property-specification syntax (I'd taken a stab at that in #747, see `requirement-condition` [here][1]), but our current pattern for conditionally-required properties would be:

* **`access`** *(string, REQUIRED when `type` is `b` or `c`)* - cgroup permissions for device.
    A composition…

Existing examples are:

```
$ git grep 'REQUIRED \(when\|unless\)' v1.0.0 --
v1.0.0:config-linux.md:* **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - [major, minor numbers][devices] for the device.
v1.0.0:config.md:    This property is REQUIRED when [`start`](runtime.md#start) is called.
v1.0.0:runtime.md:* **`pid`** (int, REQUIRED when `status` is `created` or `running` on Linux, OPTIONAL on other platforms) is the ID of the container process, as seen by the host.
```

Mashimiao:
On 09/01/2017 04:38 AM, W. Trevor King wrote:
> |v1.0.0:config-linux.md:* **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - [major, minor numbers][devices] for the device.|

I think the first one seems more elegant.

config-linux.md
```diff
@@ -222,10 +222,10 @@ Each entry has the following structure:
 
 * **`allow`** *(boolean, REQUIRED)* - whether the entry is allowed or denied.
 * **`type`** *(string, OPTIONAL)* - type of device: `a` (all), `c` (char), or `b` (block).
-    Unset values mean "all", mapping to `a`.
+    Unset value means "all", mapping to `a`.
 * **`major, minor`** *(int64, OPTIONAL)* - [major, minor numbers][devices] for the device.
-    Unset values mean "all", mapping to [`*` in the filesystem API][cgroup-v1-devices].
-* **`access`** *(string, OPTIONAL)* - cgroup permissions for device.
+    Unset value means "all", mapping to [`*` in the filesystem API][cgroup-v1-devices].
+* **`access`** *(string, REQUIRED unless `type` is `a`)* - cgroup permissions for device.
```
wking: I think this phrasing leaves the `type`-unset case (where type is only *effectively* `a`) less clear than my [earlier recomendation][1].  But whatever, I expect folks will get the idea.

[1]: https://github.com/opencontainers/runtime-spec/pull/914#discussion_r136442850

Mashimiao: I prefer this one, because when type is unset it mapped to `a`, so we may can say type is always set.

Mashimiao:

Mashimiao:

vbatts:
:-1: 
The fewer conditional `REQUIRED`'s, the better.

PR #917, opened 2017-09-01
config.go add omitempty to device.Major and device.Minor

q384566678:

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -345,9 +345,9 @@ type LinuxDevice struct {
 	// Device type, block, char, etc.
 	Type string `json:"type"`
 	// Major is the device's major number.
-	Major int64 `json:"major"`
+	Major *int64 `json:"major,omitempty"`
 	// Minor is the device's minor number.
-	Minor int64 `json:"minor"`
+	Minor *int64 `json:"minor,omitempty"`
 	// FileMode permission bits for the device.
 	FileMode *os.FileMode `json:"fileMode,omitempty"`
 	// UID of the device.


crosbymichael:
Can you please explain why you think this should change?

It is hard to provide a reason to reject or approve this change without knowing first why you think it should be different.

q384566678:
Modify according to the following spec：
1`major, minor`** *(int64, REQUIRED unless `type` is `p`)* 
2.Both MAJOR and MINOR must be specified when TYPE is b, c, or u, and they must be omitted when TYPE is p.

crosbymichael:
I don't think this should be changed.  A device type of `p` is not valid for the devices cgroup.  I think adding omitempty is enough

wking:
> I don't think this should be changed. A device type of `p` is not valid for the devices cgroup.

cb01bcd86c is updating the `LinuxDevice` struct, which is [used for `linux.devices`][1].  The cgroup uses `linux.resources.devices`, which is [backed by `LinuxDeviceCgroup`][2].  The current specs for both (unchanged since v1.0.0) are [here][3] (for `linux.devices`):

> * **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - [major, minor numbers][devices] for the device.

and [here][4] (for `linux.resources.devices`):

> **`major, minor`** *(int64, OPTIONAL)* - [major, minor numbers][devices] for the device.

That means that both cases should have `omitempty` for `Major` and `Minor` and `LinuxDeviceCgroup` [already does][5].

Once we add an `omitempty`, we need to decide on pointer vs. non-pointer.  I don't think we have a clear policy on this at the moment (#772), but `LinuxDeviceCgroup` uses pointers for the same (occasionally) OPTIONAL situation, so cb01bcd8 keeps us consistent with that.  Keeping non-pointers would also work for `Major`, since major zero is reserved][6], and that would make migration easier for Go consumers.  But using a non-pointer for `Minor` is a no-go, because 0 is a valid minor value (e.g. [here][7]).  And the slightly easier migration for:

```go
Major int64 `json:"major,omitempty"`
Minor *int64 `json:"minor,omitempty"`
```

does  not seem worth breaking consistency with `LinuxDeviceCgroup`.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/specs-go/config.go#L149-L150
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/specs-go/config.go#L325-L326
[3]: https://github.com/opencontainers/runtime-spec/blame/689521f694a44ef01d4d39d17c638271655d58fd/config-linux.md#L121
[4]: https://github.com/opencontainers/runtime-spec/blame/689521f694a44ef01d4d39d17c638271655d58fd/config-linux.md#L226
[5]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/specs-go/config.go#L365-L368
[6]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/devices.txt?h=v4.13#n1
[7]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/admin-guide/devices.txt?h=v4.13#n32

vbatts:
NACK
I don't think this is needed. This sets up implementations to now get a `nil` where that was not possible before, so this would need a version bump for not a particular gain or fix.

PR #918, opened 2017-09-07
schema: add devicecgroup type pattern

Mashimiao:
There is type limitation for deviceCgroup, I still think we should add such check in schema

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -209,7 +209,8 @@
                     "type": "boolean"
                 },
                 "type": {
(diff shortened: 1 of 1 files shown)


wking:
I'd rather re-open and merge #690.  But if a new PR with a subset of those changes is the easiest way to move forward, this is fine.

Mashimiao:

hqhq:
@Mashimiao I've reopened #690 .

Mashimiao:
OK

PR #920, opened 2017-09-08
Add vTPM specification

stefanberger:
Add the vTPM specification to the documentation, config.go, and
schema description. The following is an example of a vTPM description
that is found under the path /linux/resources/vtpms:

    "vtpms": [
        {
            "VTPMVersion": "1.2",
            "CreateCertificates" : false
        }
    ]

Signed-off-by: Stefan Berger <stefanb@linux.vnet.ibm.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -387,6 +387,36 @@ The following parameters can be specified to set up the controller:
 }
 ```
 
+### <a name="configLinuxVTPMs" />vTPMs
+
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that will be made available to the container.
+
+Each entry has the following structure:
+
+* **`statePath`** *(string, REQUIRED)* - Unique path where vTPM writes its state into.
+* **`statePathIsManaged`** *(string, OPTIONAL)* - Whether runc is allowed to delete the TPM's state path upon destroying the TPM, defaults to false.
+* **`vtpmVersion`** *(string, OPTIONAL)* - The version of TPM to emulate, either 1.2 or 2, defaults to 1.2.
+* **`createCerts`** *(boolean, OPTIONAL)* - If true then create certificates for the vTPM, defaults to false.
+* **`runAs`** *(string, OPTIONAL)* - Under which user to run the vTPM, e.g.  'tss'.
+* **`pcrBanks`** *(string, OPTIONAL)* - Comma-separated list of PCR banks to activate, default depends on `swtpm`.
+* **`encryptionPassword`** *(string, OPTIONAL)* - Write state encrypted with a key derived from the password, defaults to not encrypted.
+
+#### Example
+
+```json
+    "vtpms": [
+        {
+            "statePath": "/var/lib/runc/myvtpm1",
+            "statePathIsManaged": false,
+            "vtpmVersion": "2",
+            "createCerts": false,
+            "runAs": "tss",
+            "pcrBanks": "sha1,sha512",
+            "encryptionPassword": "mysecret"
+        }
+    ]
+```
+
 ### <a name="configLinuxHugePageLimits" />Huge page limits
 
 **`hugepageLimits`** (array of objects, OPTIONAL) represents the `hugetlb` controller which allows to limit the
--- a/config.md
+++ b/config.md
@@ -886,7 +886,16 @@ Here is a full example `config.json` for reference.
                         "rate": 300
                     }
                 ]
-            }
+            },
+            "vtpms": [
+                {
+                    "statePath": "/var/lib/runc/myvtpm1",
+                    "vtpmVersion": "2",
+                    "createCerts": false,
+                    "runAs": "tss",
+                    "pcrBanks": "sha1,sha512"
+                }
+            ]
         },
         "rootfsPropagation": "slave",
         "seccomp": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -398,6 +398,8 @@ type LinuxResources struct {
 type LinuxDevice struct {
 	// Path to the device.
 	Path string `json:"path"`
+	// Path of passed-through device on host
+	Devpath string `json:"devpath"`
 	// Device type, block, char, etc.
 	Type string `json:"type"`
 	// Major is the device's major number.
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -352,6 +352,24 @@ type LinuxRdma struct {
 	HcaObjects *uint32 `json:"hcaObjects,omitempty"`
 }
 
+// LinuxVTPM for vTPM definition
+type LinuxVTPM struct {
+	// Path on host where vTPM writes state to
+	StatePath string `json:"statePath,omitempty"`
+	// Whether runc is allowed to delete the 'Statepath' once the TPM is destroyed
+	StatePathIsManaged bool `json:"statePathIsManaged,omitempty"`
+	// Version of the TPM that is emulated
+	TPMVersion string `json:"vtpmVersion,omitempty"`
+	// Whether to create certificates upon first start of vTPM
+	CreateCertificates bool `json:"createCerts,omitempty"`
+	// The PCR banks to enable
+	PcrBanks string `json:"pcrBanks,omitempty"`
+	// Under what user to run the vTPM process
+	RunAs string `json:"runAs,omitempty"`
+	// The password to derive the encryption key from
+	EncryptionPassword string `json:"encryptionPassword,omitempty"`
+}
+
 // LinuxResources has container runtime resource constraints
 type LinuxResources struct {
 	// Devices configures the device whitelist.
@@ -372,6 +390,8 @@ type LinuxResources struct {
 	// Limits are a set of key value pairs that define RDMA resource limits,
 	// where the key is device name and value is resource limits.
 	Rdma map[string]LinuxRdma `json:"rdma,omitempty"`
+	// VTPM configuration
+	VTPMs []LinuxVTPM `json:"vtpms,omitempty"`
 }
 
 // LinuxDevice represents the mknod information for a Linux special device file
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -40,6 +40,12 @@
                             "$ref": "defs-linux.json#/definitions/DeviceCgroup"
                         }
                     },
+                    "vtpms" : {
+                        "type": "array",
+                        "items": {
+                            "$ref": "defs-linux.json#/definitions/VTPM"
+                        }
+                    },
                     "pids": {
                         "type": "object",
                         "properties": {
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -140,6 +140,14 @@
             "description": "minor device number",
             "$ref": "defs.json#/definitions/int64"
         },
+        "TPMVersion": {
+            "description": "The TPM version",
+            "type": "string",
+            "enum": [
+                "1.2",
+                "2"
+            ]
+        },
         "FileMode": {
             "description": "File permissions mode (typically an octal value)",
             "type": "integer",
@@ -231,6 +239,35 @@
                         }
                     }
                 }
+            ]
+        },
+        "VTPM" : {
+            "type": "object",
+            "properties" : {
+                "statePath": {
+                    "type": "string"
+                },
+                "statePathIsManaged": {
+                    "type": "boolean"
+                },
+                "vtpmVersion": {
+                    "$ref": "#/definitions/TPMVersion"
+                },
+                "createCerts": {
+                    "type": "boolean"
+                },
+                "runAs": {
+                    "type": "string"
+                },
+                "pcrBanks": {
+                    "type": "string"
+                },
+                "encryptionPassword": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "statePath"
             ]
         },
         "DeviceCgroup": {
--- a/schema/test/config/good/spec-example.json
+++ b/schema/test/config/good/spec-example.json
@@ -330,7 +330,25 @@
                         "rate": 300
                     }
                 ]
-            }
+            },
+            "vtpms": [
+                {
+                    "statePath": "/var/lib/runc/myvtpm1",
+                    "vtpmVersion": "2",
+                    "createCerts": false,
+                    "runAs": "tss",
+                    "pcrBanks": "sha1,sha512"
+                },
+                {
+                    "statePath": "/var/lib/runc/myvtpm2",
+                    "statePathIsManaged": true,
+                    "vtpmVersion": "1.2",
+                    "createCerts": true,
+                    "runAs": "root",
+                    "pcrBanks": "sha1,sha512",
+                    "encryptionPassword": "mysecret"
+                }
+            ]
         },
         "rootfsPropagation": "slave",
         "seccomp": {


crosbymichael:
@stefanberger is this interface stable from the kernel POV?

specs-go/config.go
```diff
@@ -568,3 +570,13 @@ type LinuxIntelRdt struct {
 	// Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
 	L3CacheSchema string `json:"l3CacheSchema,omitempty"`
 }
+
+// VTPM is used to hold the configuration state of a VTPM
+type VTPM struct {
+	// The directory where the TPM emulator writes the TPM state to
+	Statepath string `json:"statepath"`
```
crosbymichael: StatePath string `json:"statePath"`

specs-go/config.go
```diff
@@ -568,3 +570,13 @@ type LinuxIntelRdt struct {
...
+	// Whether to create a certificate for the VTPM
+	Createcerts bool `json:"createcerts"`
```
crosbymichael: CreateCerts bool `json:"createCerts"`

specs-go/config.go
```diff
@@ -568,3 +570,13 @@ type LinuxIntelRdt struct {
...
+	// Version of the TPM
+	Vtpmversion string `json:"vtpmversion"`
```
crosbymichael: VTPMVersion string `json:"vtpmVersion"`

config.md
```diff
@@ -772,7 +772,14 @@ Here is a full example `config.json` for reference.
                         "rate": 300
                     }
                 ]
-            }
+            },
+            "vtpms": [
+                {
+                    "Statepath": "/var/run/runc/ubuntu/tpm12_1",
```
crosbymichael: these should be the same as the json tags on the go struct

config-linux.md
```diff
@@ -384,6 +384,31 @@ The following parameters can be specified to set up the controller:
...
+* **`CreateCertificates`** *(boolean, OPTIONAL)* - Whether to create certificates for the vTPM
+
+The `Statepath` MUST be unique per container.
+
+### Example
+
+```json
+    "vtpms": [
+        {
+            "Statepath": "/var/run/runc/ubuntu/tpm12_1",
```
crosbymichael: lowercase like the json tags

stefanberger:
@crosbymichael I updated the patches to reflect the requested changes.

I added 3 fields to the VTPM struct in config.go. In runc I have a lot more fields. I suppose this is ok.

stefanberger:
@crosbymichael The kernel has a vTPM proxy driver that we would be using in runc. Its interface is stable. New ioctl's may be added, but that shouldn't be a problem. 

The interface of the TPM emulator implementation runc is using, swtpm, is also stable.

crosbymichael:
@stefanberger why does runc have more fields than the spec?

stefanberger:
@crosbymichael It's keeping data such as the created major/minor numbers of the character device, file descriptor, etc. Here's where it resides then. https://github.com/stefanberger/runc/blob/vtpm/libcontainer/vtpm/vtpm.go#L21

crosbymichael:
@stefanberger ok, so those are implementation details, not input from the user

config-linux.md
```diff
@@ -384,6 +384,31 @@ The following parameters can be specified to set up the controller:
...
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that
+will be made available to the container.
+
+Each entry has the following structure:
+
+* **`statePath`** *(string, REQUIRED)* - full path to a directory where the vTPM is to write its persistent state into
+* **`vtpmVersion`** *(string, OPTIONAL)* - The version of TPM to emulate; either 1.2 or 2; default is 1.2
+* **`createCerts`** *(boolean, OPTIONAL)* - Whether to create certificates for the vTPM
+
+The `statePath` MUST be unique per container.
```
wking: nit: suggested copy edit:

> If the `vtpms` array contains duplicated entries with the same `statePath`, the runtime MUST [generate an error](runtime.md#errors).

to echo (ish) our existing language [here][1].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L45

specs-go/config.go
```diff
@@ -568,3 +570,13 @@ type LinuxIntelRdt struct {
...
 }
+
+// VTPM is used to hold the configuration state of a VTPM
+type VTPM struct {
+	// The directory where the TPM emulator writes the TPM state to
+	StatePath string `json:"statePath"`
+	// Whether to create a certificate for the VTPM
+	CreateCerts bool `json:"createCerts"`
+	// Version of the TPM
+	VTPMversion string `json:"vtpmVersion"`
```
wking: This (and `CreatCerts`) should be `omitempty` because they are optional.

specs-go/config.go
```diff
@@ -161,6 +161,8 @@ type Linux struct {
 	// IntelRdt contains Intel Resource Director Technology (RDT) information
 	// for handling resource constraints (e.g., L3 cache) for the container
 	IntelRdt *LinuxIntelRdt `json:"intelRdt,omitempty"`
+	// VTPM configuration
+	VTPMS []*VTPM `json:"vtpms"`
```
wking: None of our other arrays are arrays of pointers.  Can we use `[]VTPM` here?

And with aggressive namespacing (#567) we probably want `[]LinuxVTPM` or some such.

config-linux.md
```diff
@@ -384,6 +384,31 @@ The following parameters can be specified to set up the controller:
...
+## <a name="configLinuxVTPMs" />vTPMs
+
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that
+will be made available to the container.
+
+Each entry has the following structure:
+
+* **`statePath`** *(string, REQUIRED)* - full path to a directory where the vTPM is to write its persistent state into
+* **`vtpmVersion`** *(string, OPTIONAL)* - The version of TPM to emulate; either 1.2 or 2; default is 1.2
+* **`createCerts`** *(boolean, OPTIONAL)* - Whether to create certificates for the vTPM
```
wking: I think we also want to say “Defaults to `false`.”, although existing uses ([here][1] and [here][2]) have tacked that onto the first sentence-ish and not backticked the JSON literal.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L45
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L148

config-linux.md
```diff
@@ -384,6 +384,31 @@ The following parameters can be specified to set up the controller:
...
 ```
 
+## <a name="configLinuxVTPMs" />vTPMs
+
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that
+will be made available to the container.
+
+
+* **`statePath`** *(string, REQUIRED)* - full path to a directory where the vTPM is to write its persistent state into
```
wking: From opencontainers/runc#1591, it looks like you're writing a `swtpm` [PID][1] and [log][2] files to this directory.  But this spec PR doesn't explain that, so folks consuming a generic runtime (which may or may not be runc) won't be able to rely on that.  Do you have a use-case for the runtime-caller reaching into this directory to check those files?  If so, I think you should specify the files here.  If not, I think we can drop this property from runtime-spec, and generate it on the fly (or based on runtime-specific options, like runc's existing `--root` option) instead of having it be configured here.

[1]: https://github.com/opencontainers/runc/pull/1591/files#diff-3076cfa104fbee0471fea515edf4f255R154
[2]: https://github.com/opencontainers/runc/pull/1591/files#diff-3076cfa104fbee0471fea515edf4f255R158

config-linux.md
```diff
@@ -384,6 +384,31 @@ The following parameters can be specified to set up the controller:
     }
 ```
 
+## <a name="configLinuxVTPMs" />vTPMs
```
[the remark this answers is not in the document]
stefanberger: fixed.

config-linux.md
```diff
@@ -384,6 +384,31 @@ The following parameters can be specified to set up the controller:
...
+
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that
+will be made available to the container.
```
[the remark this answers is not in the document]
stefanberger: fixed.

config-linux.md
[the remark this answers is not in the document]
stefanberger: fixed.

config-linux.md
stefanberger: I reformulated that a bit to "If true then create certificates for the vTPM, default to false."

specs-go/config.go
stefanberger: fixed

specs-go/config.go
stefanberger: fixed

specs-go/config.go
```diff
@@ -568,3 +570,13 @@ type LinuxIntelRdt struct {
 	// Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
 	L3CacheSchema string `json:"l3CacheSchema,omitempty"`
 }
+
+// VTPM is used to hold the configuration state of a VTPM
+type LinuxVTPM struct {
+	// The directory where the TPM emulator writes the TPM state to
+	StatePath string `json:"statePath"`
```
crosbymichael: Do you really need a state path at all? Can this just be an implementation detail of the runtime and have it handle it internally?

config-linux.md
```diff
@@ -384,6 +384,30 @@ The following parameters can be specified to set up the controller:
...
+
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that will be made available to the container.
+
+Each entry has the following structure:
+
+* **`statePath`** *(string, REQUIRED)* - a directory for persisting vTPM state. This value MUST be an absolute path.
+* **`vtpmVersion`** *(string, OPTIONAL)* - The version of TPM to emulate, either 1.2 or 2; default is 1.2.
+* **`createCerts`** *(boolean, OPTIONAL)* - If true then create certificates for the vTPM, defaults to false.
+
+The `statePath` MUST be unique per container. If the `vtpms` array contains duplicate entries with the same `statePath`, the runtime MUST generate an error.
```
wking: These two sentences [should be on two lines][1].

And I think the two sentences you have for the `statePath` list entry above should be as well:

```
* **`statePath`** *(string, REQUIRED)* - a directory for persisting vTPM state.
```

(with a four-space indent, #495, #846), although the current spec is not always consistent about list entries (e.g. [here we have two sentences in one list-entry line][2]).

Also, existing examples use a link for “generate an error” (e.g. see [here][3]).

[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L272
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L452

specs-go/config.go
stefanberger: Also added 'omitempty' now.

specs-go/config.go
stefanberger: Yes, it should be possible to remove it.

config-linux.md
stefanberger: Thanks for your patience. Fixed.

crosbymichael:
@stefanberger 

Is this needed at the runtime level because of interactions with the tmp and user namespaces?  Sorry, I'm just trying to understand this more and hope I'm not asking stupid questions.

specs-go/config.go
stefanberger: The runtime-spec and runc differ significantly in this regard. In runc there are arrays with pointers, such as `Mounts []*Mount` and `Devices []*Device`. How do the runtime-spec and runc relate to each other in regards to these files ? If I don't make these changes in runc but leave it at *VTPM, I suppose this will be alright (not to create more churn...)?

stefanberger:
@crosbymichael I am not sure what your latest comment ('Is this needed at the runtime level...')is referring to.

crosbymichael:
@stefanberger why couldn't this be done outside of the runtime(runc) and just configured the spec to add the new device to the container?

wking:
Looking over this PR and opencontainers/runc#1591, I have some slight concerns about making the `vtmps` object so opaque.  As it stands in the runc PR, these parameters will be backed by [swtpm][1], and the API we're exposing is basically “things swtpm needs the runtime-caller to decide”.  But the swtpm config is different from the kernel API.  We could have a thinner runtime-spec interface if we stayed closer to the [kernel API][2].  You could just configure the number of vtpms you want (`"vtpms": 2`) and have the runtime create the appropriate character devices for the container and pipe out the file descriptors to the caller (like we already do for [pseudoterminal masters with `--console-socket`][3], although we have yet to land a spec for that).  We could re-use the existing console socket (with a more generic name?) and just define [a new request-message type][4] for the vTPM descriptors.  Then runtime callers could pass that file descriptor to their emulator of choice and manage the emulator's lifecycle independently.

On the other hand, with that increased flexibility, you'd have more moving parts.  If we're ok claiming that the swtpm or other runtime-chosen emulator is a good enough for anyone using `vtpms`, then that's fine too.  I guess folks who were not satisfied with their runtime's choice could always create the vTPM and push the appropriate device into their rootfs before `create`-ing the container if they needed more flexibility.

[1]: https://github.com/stefanberger/swtpm
[2]: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/tpm/tpm_vtpm_proxy.txt?h=v4.9
[3]: https://github.com/wking/ocitools-v2/blob/a2725b42d7b312681ed5315cf7e663a13c487445/docs/command-line-interface.md#console-socket
[4]: https://github.com/wking/ocitools-v2/blob/a2725b42d7b312681ed5315cf7e663a13c487445/docs/command-line-interface.md#requests

wking:
> Is this needed at the runtime level because of interactions with the tmp and user namespaces?

Is there a `tpm` namespace?  I agree that handing this outside of the runtime seems cleanest.

stefanberger:
@wking So the kernel API does take a VTPMVersion as a parameter of an ioctl() but doesn't care about creation of certificates, which is a configuration parameter to the emulator. So in the case of vTPM we may not just have kernel parameters but also emulator parameters that need to be passed. Another possible parameter to the emulator would be whether the emulator is supposed to encrypt the files the vTPM writes out. On input this could be a boolean and runc creates a random key or one could pass an (AES) key directly. 
I suppose passing emulator parameters via the JSON is needed and allowed in this case, even if these parameters are not directly kernel parameters.

stefanberger:
@wking There's no `tpm` namespace.

wking:
On Mon, Sep 11, 2017 at 09:20:57PM +0000, Stefan Berger wrote:
> So in the case of vTPM we may not just have kernel parameters but
> also emulator parameters that need to be passed.

What do we gain by making the runtime a middleman between the emulator
and the runtime-caller?  It seems simpler to have the caller:

1. Setup whichever emulator they want however they want.
2. Put the resulting character device under the container's root.path
   (or add it to linux.devices).
3. Call ‘create’ to launch the container.

Then there's no need to tell the runtime about vTPM at all, it's just
passing a device through to the container like all the other devices
it passes through.

wking:
> What do we gain by making the runtime a middleman between the emulator and the runtime-caller?

As an example of this in another context, making the runtime a middleman for pseudoterminal creation allows you to create the pseudoterminal pair from a `newinstance` `devpts` mount.  If you're an unprivileged user, you need the container mount namespace to be created before you have permission for the `devpts` mount, and while you *could* create those namespaces outside of the runtime and pass them in via file descriptor paths, that duplicates a lot of the work that the runtime is supposed to be helping you do.  So having the runtime handle the [`newinstance` `devpts` mount][1] and then [create a new pseudoterminal pair from that `devpts`][2] makes sense.  It's not clear to me (yet?) what efficiencies we get by similarly pushing vTPM into the runtime.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L553-L565
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config.md#L473

stefanberger:
@wking @crosbymichael dumb question: if we take the runtime, which I suppose you are referring to is represented by this code base here, out, does that mean we wouldn't have anything vTPM related in this repository? Maybe I am not following the discussion correctly, but I suppose that at least the config.json would have to contain the `vtpms` array and at least the version of TPM to emulate (and possibly the other parameters necessary for the emulator)?

wking:
> … if we take the runtime, which I suppose you are referring to is represented by this code base here, out, does that mean we wouldn't have anything vTPM related in this repository?

Well, I could see you having:

```json
{
  "linux": {
    "devices": [
      {
        "path": "/dev/tpm0", (or whatever)
        "type": "c",
        "major": …, (whatever you got back from the ioctl)
        "minor": …, (whatever you got back from the ioctl)
        "fileMode": 384, (0o600, or whatever you like),
        "uid": 0,  (or whatever you like),
        "gid": 0  (or whatever you like),
      },
      …
    ],
  }
}
```

But you can also accomplish the same thing by `mknod`-ing the device yourself at any point before you call `start`, in which case you wouldn't need a config entry at all.  Or you could mknod the device in a [pre-start hook][1].  Lots of options.

> … but I suppose that at least the config.json would have to contain the `vtpms` array and at least the version of TPM to emulate (and possibly the other parameters necessary for the emulator)?

But you wouldn't need that if [the caller was setting up the emulator][2].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#prestart
[2]: https://github.com/opencontainers/runtime-spec/pull/920#issuecomment-328664527

stefanberger:
@wking If we express a vTPM instance as shown above with a device on the level of the runtime-spec, then how do we represent it at the level of runc ? Can runc extend the JSON?

wking:
> If we express a vTPM instance as shown above with a device on the level of the runtime-spec, then how do we represent it at the level of runc ? Can runc extend the JSON?

Runc can store additional information in its libcontainer config.  But why would it need to?  If the caller is managing swtpm and the devices outside the runtime, there's nothing for the runtime to do.

If you want a tighter binding, you could reroll opencontainers/runc#1591 into a third-party hook tool, with a setup command for a [pre-start hook][1] and a teardown command for a [post-stop hook][2].

But I don't see anything in a quick skim of the runc PR, except maybe checkpoint blocking, that cannot be handled by hooks or other external-to-the-runtime code.  Am I missing something?  Can you provide more details on the checkpoint breakage?

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#prestart
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/config.md#poststop

stefanberger:
@wking Who is the 'caller' in this case? ('If the caller is managing swtpm and the device...'). Do you want to support vTPM on the runc level or push to even higher levels?

I assumed that checkpointing includes migration as well. With an attached vTPM its state would have to be migrated along with the state of the container, which isn't implemented.

stefanberger:
@wking FYI: I am also working on namespacing of IMA. There I am hooking up an IMA namespace with a virtual TPM instance and the vTPM receives the TPM commands (PCR Extends) from IMA that would normally go to the hardware TPM. If we wanted to support that on the runc level some day, then there should not only be support for an IMA namespace on that level but also to hook up a vTPM to it.

wking:
> Who is the 'caller' in this case?

The runtime-caller (e.g., see the steps I floated [hrre][1]).

> Do you want to support vTPM on the runc level or push to even higher levels?

I want to push it up to higher levels, unless we have a reason for embedding it in the runtime (like we already do for network setup).

> With an attached vTPM its state would have to be migrated along with the state of the container, which isn't implemented.

That makes sense.  But I don't see a reason why emulator checkpointing couldn't happen at higher levels too, so it's not a reason for putting this in the runtime vs. higher levels.

> I am also working on namespacing of IMA.

Depending on the details, this could end up being like network-namespace setup (which we punt to higher levels) or mount-namespace setup (which we handle in the runtime via `mount`).  Which do you think it seems closer to?

[1]: https://github.com/opencontainers/runtime-spec/pull/920#issuecomment-328664527

stefanberger:
@wking To resume this discussion. I integrated vTPM (with IMA namespacing) into Docker-CE 17.12. As part of that I found it to be necessary to support the following runtime spec for supporting a vTPM:

```
+// VTPM definition
+type VTPM struct {
+	Statepath string `json:"statepath,omitempty"`
+	StatepathIsManaged bool `json:"statepathismanaged,omitempty"`
+	TPMVersion string `json:"vtpmversion,omitempty"`
+	CreateCertificates bool `json:"createcerts,omitempty"`
+}
```

I extended RunC with vTPM. When running RunC with vTPM support in Docker-CE one surprise was that when running 'docker restart' the vTPM state path was deleted and thus all the persistent state of the vTPM got lost. This for example required the addition of the 'StatePathIsManaged' field in this patch [here](https://github.com/stefanberger/runc/commit/a9367b64f22637f035f5ae0cb1af3d79fe869827), which basically helps us restart a container on the docker level without erasing the state like we do when we delete the container on the Docker level or using `runc delete xyz` on the RunC level.

Further, I find it necessary to integrate the vTPM into RunC also because of RunC spawning an IMA namespace (namespacing of IMA is not upstreamed) and to hook a vTPM to the IMA namespace I need to be able to transfer the vTPM device to the IMA namespace by calling an ioctl on the vTPM device's file descriptor. I doubt that this would be possible if vTPM was created in a hook like where we can setup  networking using [netns](https://github.com/jessfraz/netns).

wking:
> I extended RunC with vTPM. When running RunC with vTPM support in Docker-CE one surprise was that when running 'docker restart' the vTPM state path was deleted and thus all the persistent state of the vTPM got lost. This for example required the addition of the 'StatePathIsManaged' field in this patch [here][1], which basically helps us restart a container on the docker level without erasing the state like we do when we delete the container on the Docker level or using `runc delete xyz` on the RunC level.

That patch is making runc's removal of the vTPM state path conditional.  But if all vTPM handling happens in higher layers, runc would never be deleting the vTPM state path, so I don't see an issue there.

> Further, I find it necessary to integrate the vTPM into RunC also because of RunC spawning an IMA namespace (namespacing of IMA is not upstreamed)…

I [can't find code for this][2].  Can you link to it?

> … and to hook a vTPM to the IMA namespace I need to be able to transfer the vTPM device to the IMA namespace by calling an ioctl on the vTPM device's file descriptor.

I see no reason why this couldn't happen between `create` and `start` (or in a `prestart` hook).  You have the container PID from the state, so you can find the IMA namespace under `/proc/{pid}/ns/`.  And you can configure your hook with the path to the vTPM device, so the hook can open that when it needs the file descriptor.

[1]: https://github.com/stefanberger/runc/commit/a9367b64f22637f035f5ae0cb1af3d79fe869827
[2]: https://github.com/stefanberger/runc/search?utf8=%E2%9C%93&q=ima&type=

vbatts:
_house keeping:_
TPMs seem like a fine thing to support, but this conversation died off. @stefanberger perhaps a fresh PR would be useful, and https://github.com/opencontainers/runtime-spec/pull/920#issuecomment-328660319 asked if there had been any methods used to accomplish this functionality outside of the runtime-spec.
(i'm inclined to close this for now)

stefanberger:
> _house keeping:_
> TPMs seem like a fine thing to support, but this conversation died off. @stefanberger perhaps a fresh PR would be useful, and [#920 (comment)](https://github.com/opencontainers/runtime-spec/pull/920#issuecomment-328660319) asked if there had been any methods used to accomplish this functionality outside of the runtime-spec.
> (i'm inclined to close this for now)

@crosbymichael @vbatts 

I think I had tried to answer some of the issues [here.](https://github.com/opencontainers/runtime-spec/pull/920#issuecomment-366037611).

I haven't looked at this in a while but my arguments for supporting vTPM on the runc level rather than higher layers doing it would be:
- vTPM is abstracted to a JSON object which makes it easier for any higher layer wanting to use vTPM; one can start vTPM on the runc level as just another device
- vTPM is stateful and for CRIU would either have to block CRIU or have its state blobs written
- back then I was experimenting with IMA namespaces and attaching a vTPM to an IMA namespace required the file descriptor of the vTPM to be passed to the IMA namespace using an ioctl

AkihiroSuda:
needs rebase

specs-go/config.go
stefanberger: Renamed.

specs-go/config.go
stefanberger: Renamed.

config.md
stefanberger: Done

config-linux.md
stefanberger: Fixed.

stefanberger:
The update I pushed today adds all those fields that are needed to support the vTPM features implemented in runc PR https://github.com/opencontainers/runc/pull/1591

lukehinds:
@crosbymichael / @wking are you able to recommence / continue your reviews please. As @vbatts points out, vTPM support would be immensely useful for measuring trust of containers.

specs-go/config.go
```diff
@@ -372,12 +390,16 @@ type LinuxResources struct {
...
+	// VTPM configuration
+	VTPMs []LinuxVTPM `json:"vtpms,omitempty"`
 }
 
 // LinuxDevice represents the mknod information for a Linux special device file
 type LinuxDevice struct {
 	// Path to the device.
 	Path string `json:"path"`
+	// Path of passed-through device on host
+	Devpath string `json:"devpath"`
```
mrunalp, line 402: What is this needed for?

stefanberger: So that /dev/tpm10 on the host can appear as /dev/tpm0 inside the container.

config-linux.md
```diff
@@ -387,6 +387,36 @@ The following parameters can be specified to set up the controller:
...
+
+**`vtpms`** (array of objects, OPTIONAL) lists a number of emulated TPMs that will be made available to the container.
+
+Each entry has the following structure:
+
+* **`statePath`** *(string, REQUIRED)* - Unique path where vTPM writes its state into.
+* **`statePathIsManaged`** *(string, OPTIONAL)* - Whether runc is allowed to delete the TPM's state path upon destroying the TPM, defaults to false.
+* **`vtpmVersion`** *(string, OPTIONAL)* - The version of TPM to emulate, either 1.2 or 2, defaults to 1.2.
+* **`createCerts`** *(boolean, OPTIONAL)* - If true then create certificates for the vTPM, defaults to false.
+* **`runAs`** *(string, OPTIONAL)* - Under which user to run the vTPM, e.g.  'tss'.
```
mrunalp, line 400: Does it make sense to run this as the container user or it is typically set to a separate `tss` user?

specs-go/config.go
mrunalp: Do we need to know the host path or the entity creating the config.json can get the host device major / minor? For e.g. docker/podman have this feature:
```
sudo podman run --rm -it --device=/dev/null:/dev/mynull fedora:32 sh
sh-5.0# ls -l /dev/
total 0
crw--w----. 1 root tty  136, 0 Aug  6 21:35 console
lrwxrwxrwx. 1 root root     11 Aug  6 21:35 core -> /proc/kcore
lrwxrwxrwx. 1 root root     13 Aug  6 21:35 fd -> /proc/self/fd
crw-rw-rw-. 1 root root   1, 7 Aug  6 21:35 full
drwxrwxrwt. 2 root root     40 Aug  6 21:35 mqueue
crw-rw-rw-. 1 root root   1, 3 Aug  6 21:35 mynull
crw-rw-rw-. 1 root root   1, 3 Aug  6 21:35 null
lrwxrwxrwx. 1 root root      8 Aug  6 21:35 ptmx -> pts/ptmx
drwxr-xr-x. 2 root root      0 Aug  6 21:35 pts
crw-rw-rw-. 1 root root   1, 8 Aug  6 21:35 random
drwxrwxrwt. 2 root root     40 Aug  6 21:35 shm
lrwxrwxrwx. 1 root root     15 Aug  6 21:35 stderr -> /proc/self/fd/2
lrwxrwxrwx. 1 root root     15 Aug  6 21:35 stdin -> /proc/self/fd/0
lrwxrwxrwx. 1 root root     15 Aug  6 21:35 stdout -> /proc/self/fd/1
crw-rw-rw-. 1 root root   5, 0 Aug  6 21:35 tty
crw-rw-rw-. 1 root root   1, 9 Aug  6 21:35 urandom
crw-rw-rw-. 1 root root   1, 5 Aug  6 21:35 zero
sh-5.0#
```

The config.json snippet for this:
```
 "devices": [
      {
        "path": "/dev/mynull",
        "type": "c",
        "major": 1,
        "minor": 3,
        "fileMode": 8630,
        "uid": 0,
        "gid": 0
      }
    ],

```

mfranczy:
@stefanberger are you still working on this PR?

stefanberger:
No, I am not working on it.

PR #930, opened 2017-10-17
RFC: config-linux: add an initial spec of RDMA related field

mitake:
This commit adds an initial spec of RDMA related field for Linux
configuration. They will be used for controlling rdmacg of the Linux
kernel.

Signed-off-by: Hitoshi Mitake <mitake.hitoshi@lab.ntt.co.jp>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -637,6 +637,29 @@ The following parameters can be specified to set up seccomp:
     "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
 ```
 
+### <a name="configLinuxRdmaLimits" />RDMA resource limits
+
+**`rdma_limits`** (array of objects, OPTIONAL) limits resources for network interfaces which support RDMA (e.g. HCA handles).
+For more information, see the [the kernel RMDA controller documentation][cgroup-v1-rdma].
+
+Each entry has the following structure:
+
+* **`interface_name`** (string, REQUIRED) - A name of an interface.
+* **`hca_handle_limit`** (int64, REQUIRED) - Maximum number of HCA Handles (use `-1` to request the maximum available).
+* **`hca_object_limit`** (int64, REQUIRED) - Maximum number of HCA Objects (use `-1` to request the maximum available).
+
+#### Example
+
+```json
+    "rdma_limits": [
+        {
+            "interface_name": "mlx5_0",
+            "hca_handle_limit": 1,
+	    "hca_object_limit": 20
+        }
+   ]
+```
+
 
 [cgroup-v1]: https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt
 [cgroup-v1-blkio]: https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt
@@ -647,6 +670,7 @@ The following parameters can be specified to set up seccomp:
 [cgroup-v1-net-cls]: https://www.kernel.org/doc/Documentation/cgroup-v1/net_cls.txt
 [cgroup-v1-net-prio]: https://www.kernel.org/doc/Documentation/cgroup-v1/net_prio.txt
 [cgroup-v1-pids]: https://www.kernel.org/doc/Documentation/cgroup-v1/pids.txt
+[cgroup-v1-rdma]: https://www.kernel.org/doc/Documentation/cgroup-v1/rdma.txt
 [cgroup-v2]: https://www.kernel.org/doc/Documentation/cgroup-v2.txt
 [devices]: https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
 [devpts]: https://www.kernel.org/doc/Documentation/filesystems/devpts.txt


mitake:
/cc @cyphar this is the spec side PR of https://github.com/opencontainers/runc/pull/1612. The main motivation is multiplexing RDMA network interface safely by protecting RDMA resources from buggy applications. Our current target is deep learning framework e.g. TensorFlow and MXNet.

Should I update other files e.g. https://github.com/opencontainers/runtime-spec/blob/master/schema/config-linux.json ?

config-linux.md
```diff
@@ -637,6 +637,30 @@ The following parameters can be specified to set up seccomp:
     "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
 ```
 
+### <a name="configLinuxRdmaLimits" />RDMA resource limits
+
+**`rdma_limits`** (array of objects, OPTIONAL) represents the `rdma` controller which allows to limit the
```
TomSweeneyRedHat: suggest:  "which allows to limit the" to "which allows the limiting of the"

config-linux.md
```diff
@@ -637,6 +637,30 @@ The following parameters can be specified to set up seccomp:
...
+resources related to network interfaces which support RDMA (e.g. HCA handles).
+For more information, see the RDMA section of [the kernel cgroups documentation][cgroup-v2].
```
wking: To show that RDMA is not v2-specific, we may want to link both the [v1][1] and [v2][2] docs.

[1]: https://www.kernel.org/doc/Documentation/cgroup-v1/rdma.txt
[2]: https://www.kernel.org/doc/Documentation/cgroup-v2.txt

mitake:
Updated based on the comment, could you take a look?

config-linux.md
```diff
@@ -637,6 +637,30 @@ The following parameters can be specified to set up seccomp:
     "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
 ```
 
+### <a name="configLinuxRdmaLimits" />RDMA resource limits
+
+**`rdma_limits`** (array of objects, OPTIONAL) represents the `rdma` controller which allows the limiting of the
+resources related to network interfaces which support RDMA (e.g. HCA handles).
```
wking: This doesn't follow the [one sentence per line policy][1].

Also maybe rephrase to the more compact:

> **`rdma_limits`** (array of objects, OPTIONAL) limits resources for network interfaces which support RDMA (e.g. HCA handles).

config-linux.md
```diff
@@ -637,6 +637,30 @@ The following parameters can be specified to set up seccomp:
...
+For more information, see the RDMA section of [the kernel cgroups documentation][cgroup-v1-rdma].
```
wking: If you're just linking the v1 docs, you don't need “the RDMA section of”, because the whole page is about RDMA.  How about:

    For more information, see the [the kernel RMDA controller documentation][cgroup-v1-rdma].

or:

    For more information, see the the kernel RMDA controller documentation for [v1][cgroup-v1-rdma] and [v2][cgroup-v2] cgroups.

config-linux.md
```diff
@@ -637,6 +637,30 @@ The following parameters can be specified to set up seccomp:
...
+
+Each entry has the following structure:
+
+* **`interface_name`** *(string, REQUIRED)* - a name of an interface
```
wking: `config_linux.md` isn't consistent on this, but `config.md` is pretty consistent in not wrapping the type/requirement parenthetical in `*`.  That would make this line:

    * **`interface_name`** (string, REQUIRED) - a name of an interface

Because of the current `config-linux.md` inconsistency, I'd be fine with you using wrapping `*` or not in this commit, although you should probably be internally consistent.  Can you drop them here and below, or keep them here and in the `rdma_limits` entry above?

config-linux.md
```diff
@@ -637,6 +637,30 @@ The following parameters can be specified to set up seccomp:
...
+* **`hca_handle_limit`** *(int64, REQUIRED)* - limit a number of HCA handles which can be used by a container (-1 means max)
```
wking: Maybe follow [the kernel wording][1] and use:

    * **`hca_handle_limit`** (int64, REQUIRED) - Maximum number of HCA Handles (use `-1` to request the maximum available).

I think talking about “maximum number of” is less ambiguous than “limit a number of”, although I can't put my finger on why, so I'm also fine leaving this as it stands.

[1]: https://www.kernel.org/doc/Documentation/cgroup-v1/rdma.txt

wking:
Also, this should be tagged [1.Y.0][1] or 1.1.0 or some such, because it's adding functionality and therefore is not a patch-level change.

[1]: https://github.com/opencontainers/runtime-spec/milestone/9

mitake:
@wking updated for your comments, could you take a look?

mitake:
@wking updated based on your latest comment, could you take a look?

paravmellanox:
I noticed this RFC now during meeting. 
One reference implementation is available at
https://github.com/paravmellanox/cgroups/commit/469ac1c45b37c765411978b4417f21ff30a31ab0

There are few minor differences between two PR specifically with respect the size as int64 and int32

wking:
Ah, I'd forgotten about this one.  Cross-linking #942, which is moving in the same direction.

config-linux.md
```diff
@@ -637,6 +637,29 @@ The following parameters can be specified to set up seccomp:
...
+### <a name="configLinuxRdmaLimits" />RDMA resource limits
+
+**`rdma_limits`** (array of objects, OPTIONAL) limits resources for network interfaces which support RDMA (e.g. HCA handles).
+For more information, see the [the kernel RMDA controller documentation][cgroup-v1-rdma].
+
+Each entry has the following structure:
+
+* **`interface_name`** (string, REQUIRED) - A name of an interface.
+* **`hca_handle_limit`** (int64, REQUIRED) - Maximum number of HCA Handles (use `-1` to request the maximum available).
+* **`hca_object_limit`** (int64, REQUIRED) - Maximum number of HCA Objects (use `-1` to request the maximum available).
```
wking: From the kernel docs, I think `hca_handle_limit` and `hca_object_limit` should both be OPTIONAL, possibly with a requirement that at least one of them be set (this is the same suggestion I made for #942 [here][1]).

[1]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155593193

config-linux.md
```diff
@@ -637,6 +637,29 @@ The following parameters can be specified to set up seccomp:
...
+
+#### Example
+
+```json
+    "rdma_limits": [
```
wking: You'll want to drop the leading indent from this block to catch up with #937.

mitake:
I'm closing this because @paravmellanox has the good alternative and reference implementation.

PR #931, opened 2017-10-18
config-linux: add SHOULD to linux.namespaces.type

q384566678:
In the  https://github.com/opencontainers/runtime-tools/pull/486 appeared in this problem, I think this will be better to modify.


config-linux.md
```diff
@@ -26,7 +26,7 @@ For more information, see the [namespaces(7)][namespaces.7_2] man page.
 Namespaces are specified as an array of entries inside the `namespaces` root field.
 The following parameters can be specified to set up namespaces:
 
-* **`type`** *(string, REQUIRED)* - namespace type. The following namespace types are supported:
+* **`type`** *(string, REQUIRED)* - namespace type. The following namespace types SHOULD be supported:
```
wking: In the runtime-spec PR, [you're claiming a config validation][1], but the "supported" language here sounds like it targets the runtime.  How about:

> Configuration SHOULD use, and runtimes SHOULD support, the following types:

Personally, I'd prefer MUSTing runtime support (more in #813), but even SHOULD would better than the current lack of RFC 2119 language.

[1]: https://github.com/opencontainers/runtime-tools/pull/486/files#diff-8351286f57eb81e245c9c99c07f3b34fR898

vbatts:
_house keeping:_ I think the current vague language is being leveraged to not require certain kernel versions and features. Making this into RFC language `MUST` would not be acceptable. `SHOULD` is ok enough here.

LGTM

tianon:
LGTM

... however, this is conflicting now and needs a rebase :disappointed:

vbatts:
i've just messed this PR up, by trying to "fix" the commit within github ....

vbatts:
see #1025

PR #932, opened 2017-10-18
config-linux: add Intel RDT/MBA Linux support

xiaochenshen:
Add support for Intel Resource Director Technology (RDT) /
Memory Bandwidth Allocation (MBA). Add memory bandwidth resource
constraints in Linux-specific configuration.

In this PR, the spec for memory bandwidth (memBwSchema) keeps
the same format as existed spec for L3 cache (l3CacheSchema)
for consistency and compatibility in runtime-spec 1.x.

Example:
```
"linux": {
    "intelRdt": {
        "l3CacheSchema": "L3:0=7f0;1=1f",
        "memBwSchema": "MB:0=20;1=70"
    }
}
```

This is the prerequisite of this runc proposal:
https://github.com/opencontainers/runc/issues/1596

For more information about Intel RDT/MBA, please refer to:
https://github.com/opencontainers/runc/issues/1596

Signed-off-by: Xiaochen Shen <xiaochen.shen@intel.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -491,7 +491,7 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
     If `intelRdt` is set, the runtime MUST write the container process ID to the `tasks` file in a proper sub-directory in a mounted `resctrl` pseudo-filesystem. That sub-directory name is specified by `closID` parameter.
     If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
 
-    If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` pseudo-filesystems.
+If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` pseudo-filesystems.
 
 The following parameters can be specified for the container:
 
@@ -499,20 +499,33 @@ The following parameters can be specified for the container:
     If `closID` is set, runtimes MUST create `closID` directory in a mounted `resctrl` pseudo-filesystem if it doesn't exist. If not set, runtimes MUST use the container ID from [`start`](runtime.md#start) and create the `<container-id>` directory.
 
 * **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
-    If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in that sub-directory discussed in `closID`. If not set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
+    The value SHOULD start with `L3:` and SHOULD NOT contain newlines.
+* **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth percentage per L3 cache id.
+    The value MUST start with `MB:` and MUST NOT contain newlines.
 
-    If `closID` and `l3CacheSchema` both are set, runtimes MUST compare `l3CacheSchema` value with `schemata` file, and [generate an error](runtime.md#errors) if doesn't match.
+    If both `l3CacheSchema` and `memBwSchema` are set, runtimes MUST write the combined value to the `schemata` file in that sub-directory discussed in `closID`.
+    If `l3CacheSchema` contains a line beginning with `MB:`, the value written to `schemata` file MUST be the non-`MB:` line(s) from `l3CacheSchema` and the line from `memBWSchema`.
+
+    If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the that sub-directory discussed in `closID`.
+
+    If neither `l3CacheSchema` nor `memBwSchema` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
+
+    If `closID` is set, `l3CacheSchema` and/or `memBwSchema` is set, runtimes MUST compare `l3CacheSchema` and/or `memBwSchema` value with `schemata` file, and [generate an error](runtime.md#errors) if doesn't match.
 
 ### Example
 
-Consider a two-socket machine with two L3 caches where the default CBM is 0xfffff and the max CBM length is 20 bits.
-Tasks inside the container only have access to the "upper" 80% of L3 cache id 0 and the "lower" 50% L3 cache id 1:
+Consider a two-socket machine with two L3 caches where the default CBM is 0x7ff and the max CBM length is 11 bits,
+and minimum memory bandwidth of 10% with a memory bandwidth granularity of 10%.
+
+Tasks inside the container only have access to the "upper" 7/11 of L3 cache on socket 0 and the "lower" 5/11 L3 cache on socket 1,
+and may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.
 
 ```json
 "linux": {
     "intelRdt": {
         "closID": "guaranteed_group",
-        "l3CacheSchema": "L3:0=ffff0;1=3ff"
+        "l3CacheSchema": "L3:0=7f0;1=1f",
+        "memBwSchema": "MB:0=20;1=70"
     }
 }
 ```
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -233,6 +233,10 @@
                     },
                     "l3CacheSchema": {
                         "type": "string"
+                    },
+                    "memBwSchema": {
+                        "type": "string",
+                        "pattern": "^MB:[^\\n]*$"
                     }
                 }
             }
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -160,8 +160,8 @@ type Linux struct {
 	ReadonlyPaths []string `json:"readonlyPaths,omitempty"`
 	// MountLabel specifies the selinux context for the mounts in the container.
 	MountLabel string `json:"mountLabel,omitempty"`
-	// IntelRdt contains Intel Resource Director Technology (RDT) information
-	// for handling resource constraints (e.g., L3 cache) for the container
+	// IntelRdt contains Intel Resource Director Technology (RDT) information for
+	// handling resource constraints (e.g., L3 cache, memory bandwidth) for the container
 	IntelRdt *LinuxIntelRdt `json:"intelRdt,omitempty"`
 }
 
@@ -623,12 +623,16 @@ type LinuxSyscall struct {
 	Args   []LinuxSeccompArg  `json:"args,omitempty"`
 }
 
-// LinuxIntelRdt has container runtime resource constraints
-// for Intel RDT/CAT which introduced in Linux 4.10 kernel
+// LinuxIntelRdt has container runtime resource constraints for Intel RDT
+// CAT and MBA features which introduced in Linux 4.10 and 4.12 kernel
 type LinuxIntelRdt struct {
 	// The identity for RDT Class of Service
 	ClosID string `json:"closID,omitempty"`
 	// The schema for L3 cache id and capacity bitmask (CBM)
 	// Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
 	L3CacheSchema string `json:"l3CacheSchema,omitempty"`
+
+	// The schema of memory bandwidth percentage per L3 cache id
+	// Format: "MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;..."
+	MemBwSchema string `json:"memBwSchema,omitempty"`
 }


config-linux.md
```diff
@@ -466,19 +466,37 @@ The following parameters can be specified to set up the controller:
 The following parameters can be specified for the container:
 
 * **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
-    If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
+* **`memBwSchema`** *(array of objects, OPTIONAL)* - specifies the schema of memory bandwidth (b/w) percentage per L3 cache id.
```
crosbymichael: why not make this simpler with just or something of the like:

```
"bandwidth" : {
    "id": 0,
    "value" 70,
}
```

you don't have to prefix everything when they are part of an object, its implied that it is an attribute of the containing object.

xiaochenshen: @crosbymichael 
> why not make this simpler with just or something of the like:

It makes sense. Thank you.

config-linux.md
```diff
@@ -466,19 +466,37 @@ The following parameters can be specified to set up the controller:
...
-    If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
+* **`memBandwidth`** *(array of objects, OPTIONAL)* - an array of memory bandwidth (b/w) percentage per L3 cache id.
+    Each entry has the following structure:
+    * **`id`** *(uint32, REQUIRED)* - L3 cache id
+    * **`value`** *(uint32, REQUIRED)* - memory bandwidth (b/w) percentage
+
+    If `l3CacheSchema` or `memBandwidth` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
 
-    If `l3CacheSchema` is not set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
+    If neither `l3CacheSchema` nor `memBandwidth` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
```
wking: I don't think these two set/unset lines are quite right.  The old `l3CacheSchema` requirements were sufficient because the `l3CacheSchema` values were [expected to contain the `L3:` prefix][1].  However, dd23a4c9 only mentions the `MB:` prefix and related formatting the a non-normative Go comment.

If we go the structured route, I (which is where you seem to be taking `memBandwidth`, I think we need to make the `MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;...` formatting a normative part of the spec.  We'd also want to define the interactions between them (e.g. “the formatted lines MAY be written to `schemata` in any order, either with a single write or a series of writes”).  And I think we'd want to structure `l3CacheSchema` (as `l3CacheSchema2?`), both for consistency, and to avoid the ambiguity of:

```json
{
  "l3CacheSchema": "MB:0=50;1=50",
  "memBandwith": [{"id": 0, "value": 25}, {"id": 1, "value": 75}]
}
```


If we go the unstructured route (where `l3CacheSchema` already is), then we can just have an opaque array of strings.  Or even [a single string][2]:

```
"l3CacheSchema": "L3:0=3;1=c\nMB:0=50;1=50"
```

because [JSON allows escaped newlines][3].  At that point, the `l3CacheSchema` property name would be stale so it would be nice to rename it to something more generic (`schemata`?) in v2.  But I think it already does what you need here.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0/config-linux.md#L471
[2]: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/Documentation/x86/intel_rdt_ui.txt?h=v4.13.8#n197
[3]: https://tools.ietf.org/html/rfc7159#section-7

config-linux.md
```diff
@@ -466,19 +466,37 @@ The following parameters can be specified to set up the controller:
...
 
 ### Example
 
-Consider a two-socket machine with two L3 caches where the default CBM is 0xfffff and the max CBM length is 20 bits.
-Tasks inside the container only have access to the "upper" 80% of L3 cache id 0 and the "lower" 50% L3 cache id 1:
+Consider a two-socket machine with two L3 caches where the default CBM is 0x7ff and the max CBM length is 11 bits,
+and minimum memory b/w of 10% with a memory b/w granularity of 10%.
```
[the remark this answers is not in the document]
xiaochenshen: @wking 
> I don't think we need to abbreviate “bandwidth” in the docs. It won't take much longer to read, and slightly reduces the chance of confusion if we spell it out, instead of using “b/w”.

I agree with you. I will correct this.

config-linux.md
xiaochenshen: @wking 
> I don't think these two set/unset lines are quite right. The old l3CacheSchema requirements were sufficient because the l3CacheSchema values were expected to contain the L3: prefix. However, dd23a4c only mentions the MB: prefix and related formatting the a non-normative Go comment.

Actually, CAT and MBA features are orthogonal, each one doesn't depend on the other. We could see (1) either "L3 cache" or "memory bandwidth" schema available or (2) both "L3 cache" and "memory bandwidth" schemata are available with specific hardware and kernel configuration.

xiaochenshen: @wking 
> If we go the structured route, (which is where you seem to be taking memBandwidth), I think we need to make the MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;... formatting a normative part of the spec. ...

> If we go the unstructured route (where l3CacheSchema already is), then we can just have an opaque array of strings. Or even a single string: ...

Originally I proposed unstructured JSON format (opaque string: e.g., `"memBwSchema": "MB:0=20;1=70"`) for memory bandwidth schema in https://github.com/opencontainers/runc/issues/1596. 

@cyphar prefers structured JSON format (array of object: e.g., `"memBandwith": [{"id": 0, "value": 20}, {"id": 1, "value": 70}]`). @cyphar's justification in https://github.com/opencontainers/runc/issues/1596#issuecomment-330477399 makes sense (easy for spec validation, use interactively, easy to extend and etc.). In this PR I chose the structured JSON format.

The unstructured (opaque string) format also has some advantages: https://github.com/opencontainers/runc/issues/1596#issuecomment-330328687. It will be easier to conjunct L3 cache and memory bandwidth schema into a single string, and write once to schemata file. I'd like to hear other reviewers' opinions and suggestions.

For L3 cache schema, it is already used opaque string format (e.g., `"l3CacheSchema": "L3:0=7f0;1=1f"`). Anyway, I have plan to unify the formats for L3 cache and memory bandwidth. 

If we choose the structured JSON format, how about this mapping from the schemata file content to JSON config?
```
cat /sys/fs/resctrl/container_id/schemata
L3:0=7f0;1=1f
MB:0=20;1=70
```

```
"linux": {
    "intelRdt": {
        "l3Cache": [{"id": 0, "value": 7f0}, {"id": 1, "value": 1f}],
        "memBandwidth": [{"id": 0, "value": 20}, {"id": 1, "value": 70}]
    }
}
```

wking: > The unstructured (opaque string) format also has some advantages…

I agree that there are tradeoffs between structured data and isolating users from the underlying kernel API vs. unstructured data that provides more direct access to the underlying kernel API.  And I don't really care if we go with a structured approach or an unstructured approach.  What doesn't make sense to me is going with an unstructured approach for `l3CacheSchema` and then going for a structured approach in `memBandwidth`.  It seems like the “we should isolate config authors from the details of the kernel API” (or not) would be a decision that applied equally to everything under `intelRdt`.

> Anyway, I have plan to unify the formats for L3 cache and memory bandwidth.

I don't think the spec is a good place to play with the config format, because now that we've cut 1.0.0 with the existing `l3CacheSchema`, we need to continue to support it until this spec hits v2.  If we decided to go with [your unified, structured approach][1] (or [the unified, unstructured `schemata` I'd floated earlier][2], or some other approach), we'd need to continue to support the deprecated `l3CacheSchema` throughout the 1.x spec lifetime and define how the deprecated property interacted with the new properties.  But we certainly don't want to stamp a 1.1 with something that gets deprecated in 1.2, because then we have to support both the 1.0 `l3CacheSchema` and the 1.1 `otherDeprecatedThing` until 2.0.  So if you have plans for something unified, I think they belong in this PR, and not in follow-up work.

[1]: https://github.com/opencontainers/runtime-spec/pull/932#discussion_r145560769
[2]: https://github.com/opencontainers/runtime-spec/pull/932#discussion_r145528966

xiaochenshen: @wking 
@cyphar @crosbymichael 
/cc @vishh @mrunalp @vbatts @dqminh @philips @tianon @hqhq

> I don't think the spec is a good place to play with the config format, because now that we've cut 1.0.0 with the existing l3CacheSchema, we need to continue to support it until this spec hits v2. If we decided to go with your unified, structured approach (or the unified, unstructured schemata I'd floated earlier, or some other approach), we'd need to continue to support the deprecated l3CacheSchema throughout the 1.x spec lifetime and define how the deprecated property interacted with the new properties. But we certainly don't want to stamp a 1.1 with something that gets deprecated in 1.2, because then we have to support both the 1.0 l3CacheSchema and the 1.1 otherDeprecatedThing until 2.0. So if you have plans for something unified, I think they belong in this PR, and not in follow-up work.

I understood the consistency and compatibility requirement in runtime-spec. This is my plan:
1. Firstly, I will address "L3 cache" and "memory bandwidth" with `unified` formats in this single PR. 
2. To support existed "l3CacheSchema" throughout 1.x spec lifetime, and to avoid confusion of deprecated property, how about using similar "unified unstructured schemata" format for memory bandwidth in this PR and throughout 1.x runtime-spec and runc ?
3. If we have requirement to change all Intel RDT resources into "structured schemata" in spec 2.0 or newer version, I could open a new PR to slightly rework on appropriate time slot in the phase of spec 2.0.

Please help review and comment. Thank you!

wking: > This is my plan…

The plan sounds reasonable to me.  For the 1.x structure, we probably want at least SHOULDs for the `L3:` and `MB:` prefixes and internal newlines, and clear requirements for how we handle cases where those SHOULDs are violated.  MUSTs for those requirements would be nice, and we can do that for `memBandwidth`, but it may be too late to add new MUSTs to `l3CacheSchema`.

And there's also no reason we couldn't specify the 2.x structure now, as long as we document that the `*Schema` property overrides the analogous 2.x property when the 1.x property is set.  In that approach, the plan would be to have this PR deprecate `l3CacheSchema` and add `l3Cache` and `memBandwidth`.  We'd remove `l3CacheSchema` in 2.0.

config-linux.md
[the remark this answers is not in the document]
xiaochenshen: @wking 
@cyphar @crosbymichael 
/cc @vishh @mrunalp @vbatts @dqminh @philips @tianon @hqhq 

I have updated this PR according to @wking 's suggestion. Please help review at your convenience. Thank you.

In this PR, the spec for memory bandwidth (memBwSchema) keeps
the same format as existed spec for L3 cache (l3CacheSchema)
for consistency and compatibility in runtime-spec 1.x.

Example:
```
"linux": {
    "intelRdt": {
        "l3CacheSchema": "L3:0=7f0;1=1f",
        "memBwSchema": "MB:0=20;1=70"
    }
}
```

xiaochenshen:
@wking 
@cyphar @crosbymichael 
/cc @vishh @mrunalp @vbatts @dqminh @philips @tianon @hqhq 

I have updated this PR according to @wking 's suggestion. Please help review at your convenience. Thank you.

In this PR, the spec for memory bandwidth (memBwSchema) keeps
the same format as existed spec for L3 cache (l3CacheSchema)
for consistency and compatibility in runtime-spec 1.x.

Example:
```
"linux": {
    "intelRdt": {
        "l3CacheSchema": "L3:0=7f0;1=1f",
        "memBwSchema": "MB:0=20;1=70"
    }
}
```

config-linux.md
```diff
@@ -466,19 +466,26 @@ The following parameters can be specified to set up the controller:
 The following parameters can be specified for the container:
 
 * **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
-    If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
+* **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth percentage per L3 cache id.
```
wking: Can we get a line after this like:

> The value MUST start with `MB:` and MUST NOT contain newlines.

I'd like a similar line with MUST -> SHOULD for `l3CacheSchema` (or also with MUST if maintainers are comfortable with the [compat impact][1]).

Then down below, I'd like two paragraphs like:

> If both `l3CacheSchema` and `memBwSchema` are set, runtimes MUST write the combined value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.  If `l3CacheSchema` contains a line beginning with `MB:`, the value written to `schemata` MUST be the non-`MB:` line(s) from `l3CacheSchema` and the line from `memBWSchema`.
>
> If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
> 

[1]: https://github.com/opencontainers/runtime-spec/pull/932#discussion_r145822155

xiaochenshen: @wking 
> The value MUST start with MB: and MUST NOT contain newlines.

Yes, that is right. I will add this line for `memBwSchema`.

> I'd like a similar line with MUST -> SHOULD for l3CacheSchema (or also with MUST if maintainers are comfortable with the compat impact).

Make sense to me. I will add this line for `l3CacheSchema`.

> If l3CacheSchema contains a line beginning with MB:, the value written to schemata MUST be the non-MB: line(s) from l3CacheSchema and the line from memBWSchema.

This is not the case. `l3CacheSchema` will never contains a line beginning with `MB:`. It only contains a line beginning with `L3:` and `memBWSchema` only contains a line beginning with `MB:`. If both `l3CacheSchema` and `memBwSchema` are set, runc runtimes will handle this case in Set() to write the combined value: "`l3CacheSchema`"+"\n"+ "`memBwSchema`", e.g., "L3:0=7f0;1=1f\nMB:0=20;1=70" to `schemata` file.

The paragraphs may like:
```
* **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
    The value MUST start with `L3:` and MUST NOT contain newlines.
* **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth percentage per L3 cache id.
    The value MUST start with `MB:` and MUST NOT contain newlines.

    If both `l3CacheSchema` and `memBwSchema` are set, runtimes MUST write the combined value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
    If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
```

wking: > > If `l3CacheSchema` contains a line beginning with `MB:`, the value written to schemata MUST be the non-`MB:` line(s) from `l3CacheSchema` and the line from `memBWSchema`.
>
> This is not the case. `l3CacheSchema` will never contains a line beginning with `MB:`. It only contains a line beginning with `L3:`...

That would have been nice, but I see no such MUST in 1.0.0, which means it may be too late to add it now.  To add it before 2.0 (where we intend to drop `l3CacheSchema` anyway), you'd need to make a "nobody was actually doing that, despite it being technically legal" argument.  And withought an exhaustive list of config authors and a relatively straightforward compat spec, it may not be worth trying to build that case.

xiaochenshen: @wking 
> That would have been nice, but I see no such MUST in 1.0.0, which means it may be too late to add it now. 

How about removing these two lines?
`- The value MUST start with "L3:" and MUST NOT contain newlines.`
`- The value MUST start with "MB:" and MUST NOT contain newlines.`

In my opinion, it looks like runtime implementation detail. The validity of `l3CacheSchema` and `memBwSchema` will be checked in runc, and the validity of string writing to schemata file will be also checked by kernel.

wking: > In my opinion, it looks like runtime implementation detail. The validity of `l3CacheSchema` and `memBwSchema` will be checked in runc…

I think all configs not explicitly denied by the spec are allowed.  The 1.0.0 spec contains no `L3:` prefix requirement for `l3CacheSchema`, and it contains no anti-newline requirement for `l3CacheSchema`.  And runc [currently writes the opaque `l3CacheSchema` value to the `schemata` file][1].  So folks writing 1.0.0 configs can *already* set `l3CacheSchema` to `L3:0=7f0;1=1f\nMB:0=20;1=70` or `MB:0=20;1=70` and have a compliant config which is supported by runc.  I don't want to break those users, which means we need to either:

a. [Make a case for those users not existing][2], or
b. Add spec wording to [SHOULD against those configs, and MUST the runtime procedure for handling them if/when they occur][3].

b seems much more straightforward to me.

If we do not do either and those users happen to exist, having runc start erroring for `l3CacheSchema` like `L3:0=7f0;1=1f\nMB:0=20;1=70` is just pulling the rug out from under those users in order to satisfy an aesthetic concern about consistency between the existing `l3CacheSchema` and the new `memBwSchema`.  I think there is a benefit to consistency (it makes getting acquainted with the spec easier for newcomers), but it's not worth breaking existing users to achieve it. 

[1]: https://github.com/opencontainers/runc/blob/9a1186d128c4c4df4b2c6722976b49898cbe2e03/libcontainer/intelrdt/intelrdt.go#L510-L513
[2]: https://github.com/opencontainers/runtime-spec/pull/932#discussion_r146109506
[3]: https://github.com/opencontainers/runtime-spec/pull/932#discussion_r145983198

xiaochenshen: @wking 
> a. Make a case for those users not existing, or
b. Add spec wording to SHOULD against those configs, and MUST the runtime procedure for handling them if/when they occur.

> b seems much more straightforward to me.

I'd like to replace wording MUST with SHOULD for `l3CacheSchema` and new `memBwSchema`.

` The value SHOULD start with "L3:" and SHOULD NOT contain newlines.`
` The value SHOULD start with "MB:" and SHOULD NOT contain newlines.`

wking: > I'd like to replace wording MUST with SHOULD for `l3CacheSchema` and new `memBwSchema`.

I [prefer MUSTs for `memBwSchema`][1]; why would you SHOULD there?  It's a new property, so we do not have `l3CacheSchema`'s backwards-compat constraints.

And either way I think you need something like [the paragraphs I recommended earlier][1] to require runtimes to handle `l3CacheSchema` SHOULD violations in a consistent way.

[1]: https://github.com/opencontainers/runtime-spec/pull/932#discussion_r145983198

xiaochenshen: @wking 

Thank you for your scrupulous review and good suggestions.
I have updated this PR to address the comments.

The documentation for `IntelRdt` may look like:

## <a name="configLinuxIntelRdt" />IntelRdt

**`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
    If `intelRdt` is set, the runtime MUST write the container process ID to the `<container-id>/tasks` file in a mounted `resctrl` pseudo-filesystem, using the container ID from [`start`](runtime.md#start) and creating the `<container-id>` directory if necessary.
    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
    If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` pseudo-filesystems.

* **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
    The value SHOULD start with `L3:` and SHOULD NOT contain newlines.
* **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth percentage per L3 cache id.
    The value MUST start with `MB:` and MUST NOT contain newlines.

    If both `l3CacheSchema` and `memBwSchema` are set, runtimes MUST write the combined value to the `schemata` file in the <container-id> directory discussed in intelRdt. If `l3CacheSchema` contains a line beginning with `MB:`, the value written to `schemata` file MUST be the non-`MB:` line(s) from `l3CacheSchema` and the line from `memBWSchema`.

    If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.

    If neither `l3CacheSchema` nor `memBwSchema` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.

### Example

Consider a two-socket machine with two L3 caches where the default CBM is 0x7ff and the max CBM length is 11 bits,
and minimum memory bandwidth of 10% with a memory bandwidth granularity of 10%.

Tasks inside the container only have access to the "upper" 7/11 of L3 cache on socket 0 and the "lower" 5/11 L3 cache on socket 1,
and may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.

```json
"linux": {
    "intelRdt": {
        "l3CacheSchema": "L3:0=7f0;1=1f",
        "memBwSchema": "MB:0=20;1=70"
    }
}
```

xiaochenshen:
@wking 

Thank you for your scrupulous review and good suggestions.
I have updated this PR to address the comments.

The documentation for `IntelRdt` may look like:

## <a name="configLinuxIntelRdt" />IntelRdt

**`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
    If `intelRdt` is set, the runtime MUST write the container process ID to the `<container-id>/tasks` file in a mounted `resctrl` pseudo-filesystem, using the container ID from [`start`](runtime.md#start) and creating the `<container-id>` directory if necessary.
    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
    If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` pseudo-filesystems.

* **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
    The value SHOULD start with `L3:` and SHOULD NOT contain newlines.
* **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth percentage per L3 cache id.
    The value MUST start with `MB:` and MUST NOT contain newlines.

    If both `l3CacheSchema` and `memBwSchema` are set, runtimes MUST write the combined value to the `schemata` file in the <container-id> directory discussed in intelRdt. If `l3CacheSchema` contains a line beginning with `MB:`, the value written to `schemata` file MUST be the non-`MB:` line(s) from `l3CacheSchema` and the line from `memBWSchema`.

    If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.

    If neither `l3CacheSchema` nor `memBwSchema` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.

### Example

Consider a two-socket machine with two L3 caches where the default CBM is 0x7ff and the max CBM length is 11 bits,
and minimum memory bandwidth of 10% with a memory bandwidth granularity of 10%.

Tasks inside the container only have access to the "upper" 7/11 of L3 cache on socket 0 and the "lower" 5/11 L3 cache on socket 1,
and may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.

```json
"linux": {
    "intelRdt": {
        "l3CacheSchema": "L3:0=7f0;1=1f",
        "memBwSchema": "MB:0=20;1=70"
    }
}
```

config-linux.md
```diff
@@ -460,25 +460,34 @@ The following parameters can be specified to set up the controller:
 **`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
     If `intelRdt` is set, the runtime MUST write the container process ID to the `<container-id>/tasks` file in a mounted `resctrl` pseudo-filesystem, using the container ID from [`start`](runtime.md#start) and creating the `<container-id>` directory if necessary.
     If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
-
```
wking: nit: I liked having the not-set sentence in its own paragraph to make it even more obvious that the no-mounted line belonged to the if-set case.  But I don't care all that much, and am fine with this PR landing with this paragraph-splitting-newline removed.

config-linux.md
```diff
@@ -460,25 +460,34 @@ The following parameters can be specified to set up the controller:
...
 
 The following parameters can be specified for the container:
 
 * **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
-    If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
+    The value SHOULD start with `L3:` and SHOULD NOT contain newlines.
+* **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth percentage per L3 cache id.
+    The value MUST start with `MB:` and MUST NOT contain newlines.
+
+    If both `l3CacheSchema` and `memBwSchema` are set, runtimes MUST write the combined value to the `schemata` file in the <container-id> directory discussed in intelRdt. If `l3CacheSchema` contains a line beginning with `MB:`, the value written to `schemata` file MUST be the non-`MB:` line(s) from `l3CacheSchema` and the line from `memBWSchema`.
```
wking: I'd drop the indent here to make it a paragraph following the list (it's currently a paragraph inside the `memBwSchema` list entry).  This (and the following paragraphs) are about both `l3CacheSchema` and `memBwSchema`, and the `memBwSchema` list entry seems like too specific a location for that information.

You're also dropping the previous backticks around `<container-id>`, and we want to keep those because Markdown is a subset of HTML, so an unticked version will be invisible (as an unrecognized HTML tag).

And you have two sentences in one line, when [prefer one sentence per line][1].

schema/config-linux.json
```diff
@@ -273,7 +273,11 @@
                     "l3CacheSchema": {
                         "id": "https://opencontainers.org/schema/bundle/linux/intelRdt/l3CacheSchema",
                         "type": "string"
-                    }
+                    },
+                    "memBwSchema": {
+                        "id": "https://opencontainers.org/schema/bundle/linux/intelRdt/memBwSchema",
+                        "type": "string"
```
wking: Can we use a regexp for validation (like we do [here][1] and other places)?  That would be:

```json
"memBwSchema": {
  "id": "https://opencontainers.org/schema/bundle/linux/intelRdt/memBwSchema",
  "type": "string",
  "pattern": "^MB:[^\n]*"
}
```

And we can unit-test the schema on this point with an entry in [here][2].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0/schema/defs-linux.json#L121
[2]: https://github.com/opencontainers/runtime-spec/tree/v1.0.0/schema/test/config/bad

config-linux.md
xiaochenshen: @wking 
The reason why I delete this blank line is that: I found a nit in [original master branch code](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md): two or more spaces don't work at the end of "no-mounted" sentence for newline, and the "no-set" sentence is in code syntax highlighting style.

I could fix this issue to make "not-set" sentence in its own paragraph:
(without four spaces at the beginning of "not-set" sentence)
```
    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).

If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` pseudo-filesystems.
```

config-linux.md
xiaochenshen: @wking 
Thank you. I will fix these issues.

schema/config-linux.json
xiaochenshen: @wking 

Good idea. I'd like to add this.
We could use "./schema/validate" to validate the format of `memBwSchema` in config.json.

xiaochenshen:
@wking 

I have updated this PR to commit https://github.com/opencontainers/runtime-spec/pull/932/commits/a3b588e0901620d6e76f1d3935742380c6f69b13.

It addressed these comments:
1.  Make "not-set" sentence in its own paragraph.
2. Drop the indent of if conditions of `l3CacheSchema` and `memBwSchema`, to make it a paragraph following the list.
3.  Get back the missing backticks around `<container-id>`.
4. Make one sentence per line for paragraph "If both `l3CacheSchema` and `memBwSchema`...".
5. Add `pattern` constraint for `memBwSchema`

xiaochenshen:
@crosbymichael @vishh @mrunalp @vbatts @dqminh @philips @tianon @hqhq

Hi Maintainers,
This PR has been pending for a long time. Could you help code review at your convenience? Thank you!
I have rebased this PR to current master branch, commit https://github.com/opencontainers/runtime-spec/pull/932/commits/8b7085441adff343133bd70603d3aaec3ef582c0

crosbymichael:
LGTM

xiaochenshen:
@crosbymichael @vishh @mrunalp @vbatts @dqminh @philips @tianon @hqhq

I have rebased this PR to current master branch to address conflict issues with merged #988 .
Could you help code review at your convenience? Thank you!
commit https://github.com/opencontainers/runtime-spec/pull/932/commits/557b7f3ea4b5b0dda89a752ec7d9ab85343b376a

This PR is part of https://github.com/opencontainers/runc/issues/1596
The runc implementation: https://github.com/opencontainers/runc/pull/1632

vbatts:
non-breaking change, adding one new OPTIONAL field. A bit of semantics on the values when set.
LGTM

PR #934, opened 2017-11-01
add RFC requirement for mountLabel

Mashimiao:


config-linux.md
```diff
@@ -629,7 +629,8 @@ The following parameters can be specified to set up seccomp:
 
 ## <a name="configLinuxMountLabel" />Mount Label
 
-**`mountLabel`** (string, OPTIONAL) will set the Selinux context for the mounts in the container.
+**`mountLabel`** (string, OPTIONAL) specifies SELinux context for the mounts in the container.
+    If `mountLabel` is set, the runtime MUST set SELinux context of the mounts in the container to the given value.
```
wking: I think you want to keep “the” before “SELinux context”.

Mashimiao:

hqhq:
LGTM

config-linux.md
```diff
@@ -629,7 +629,8 @@ The following parameters can be specified to set up seccomp:
 
 ## <a name="configLinuxMountLabel" />Mount Label
 
-**`mountLabel`** (string, OPTIONAL) will set the Selinux context for the mounts in the container.
+**`mountLabel`** (string, OPTIONAL) specifies the SELinux context for the mounts in the container.
+    If `mountLabel` is set, the runtime MUST set the SELinux context of the mounts in the container to the given value.
```
wking: You want to drop this indent (I've made that mistake before, #936 ;).

Mashimiao: updated, it seems we should submit a PR to fix other places.

Mashimiao:

TomSweeneyRedHat:
LGTM

Mashimiao:
@opencontainers/runtime-spec-maintainers   PTAL

vbatts:
NACK
I'm :+1: with @wking https://github.com/opencontainers/runtime-spec/pull/934#pullrequestreview-73498295 here.
This would be a conditional `MUST` (the config generated should've accounted for whether selinux on the host is even supported). Which is needed for the trust model of handing down to the container runtime.
But to add a MUST here now, would be a revision in the spec and would be a compatibility issue for the community to need to update to using this newer version of the spec.

h-vetinari:
> @vbatts: But to add a MUST here now, would be a revision in the spec and would be a compatibility issue for the community to need to update to using this newer version of the spec.

Wouldn't that compatibility issue arise anyway due to deprecating the `prestart` hooks?

PR #936, opened 2017-11-09
config: Dedent root paragraphs, since they aren't a list entry

wking:
I'd broken this in 2022e3ec (#838), where I'd misread the initial bolding `**` as a list bullet or something :/.  The extra indents don't cause a problem for the first paragraph, but they [cause the “On all other platforms…” paragraph to be interpreted as a `<pre>` block][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#root


hqhq:
LGTM

crosbymichael:
LGTM

PR #937, opened 2017-12-01
remove unneeded indent

Mashimiao:

Some indents will cause erorr,  we should remove them.
Some places are similar with #936

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -46,6 +46,6 @@ Specifications have a variety of different timelines in their lifecycle.
     This means a major release like a v1.0.0 or v2.0.0 release will take 1 month at minimum: one week for rc1, one week for rc2, one week for rc3, and one week for the major release itself.
     Maintainers SHOULD strive to make zero breaking changes during this cycle of release candidates and SHOULD restart the three-candidate count when a breaking change is introduced.
     For example if a breaking change is introduced in v1.0.0-rc2 then the series would end with v1.0.0-rc4 and v1.0.0.
-- Minor and patch releases SHOULD be made on an as-needed basis.
+* Minor and patch releases SHOULD be made on an as-needed basis.
 
 [charter]: https://www.opencontainers.org/about/governance
--- a/config-linux.md
+++ b/config-linux.md
@@ -34,7 +34,6 @@ The following parameters can be specified to set up namespaces:
     * **`uts`** the container will be able to have its own hostname and domain name.
     * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container.
     * **`cgroup`** the container will have an isolated view of the cgroup hierarchy.
-
 * **`path`** *(string, OPTIONAL)* - namespace file.
     This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
     The runtime MUST place the container process in the namespace associated with that `path`.
@@ -48,31 +47,31 @@ If a `namespaces` field contains duplicated namespaces with same `type`, the run
 ### Example
 
 ```json
-    "namespaces": [
-        {
-            "type": "pid",
-            "path": "/proc/1234/ns/pid"
-        },
-        {
-            "type": "network",
-            "path": "/var/run/netns/neta"
-        },
-        {
-            "type": "mount"
-        },
-        {
-            "type": "ipc"
-        },
-        {
-            "type": "uts"
-        },
-        {
-            "type": "user"
-        },
-        {
-            "type": "cgroup"
-        }
-    ]
+"namespaces": [
+    {
+        "type": "pid",
+        "path": "/proc/1234/ns/pid"
+    },
+    {
+        "type": "network",
+        "path": "/var/run/netns/neta"
+    },
+    {
+        "type": "mount"
+    },
+    {
+        "type": "ipc"
+    },
+    {
+        "type": "uts"
+    },
+    {
+        "type": "user"
+    },
+    {
+        "type": "cgroup"
+    }
+]
 ```
 
 ## <a name="configLinuxUserNamespaceMappings" />User namespace mappings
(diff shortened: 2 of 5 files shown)


RELEASES.md
```diff
@@ -46,6 +46,6 @@ Specifications have a variety of different timelines in their lifecycle.
     This means a major release like a v1.0.0 or v2.0.0 release will take 1 month at minimum: one week for rc1, one week for rc2, one week for rc3, and one week for the major release itself.
     Maintainers SHOULD strive to make zero breaking changes during this cycle of release candidates and SHOULD restart the three-candidate count when a breaking change is introduced.
     For example if a breaking change is introduced in v1.0.0-rc2 then the series would end with v1.0.0-rc4 and v1.0.0.
-- Minor and patch releases SHOULD be made on an as-needed basis.
+* Minor and patch releases SHOULD be made on an as-needed basis.
```
wking: Nice.  I've submitted this upstream with opencontainers/project-template#40.

config-linux.md
```diff
@@ -509,10 +508,8 @@ The actions, architectures, and operators are strings that match the definitions
 The following parameters can be specified to set up seccomp:
 
 * **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as `syscalls[].action`.
-
 * **`architectures`** *(array of strings, OPTIONAL)* - the architecture used for system calls.
     A valid list of constants as of libseccomp v2.3.2 is shown below.
-
```
wking: Why are you removing this line?  A blank line before and after a Markdown list is traditional, and [it renders fine][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/75c847a4e6f4f2e5d13a2477fd57d5ceb98252dc/config-linux.md#seccomp

config-linux.md
```diff
@@ -531,35 +528,26 @@ The following parameters can be specified to set up seccomp:
...
     While this property is OPTIONAL, some values of `defaultAction` are not useful without `syscalls` entries.
     For example, if `defaultAction` is `SCMP_ACT_KILL` and `syscalls` is empty or unset, the kernel will kill the container process on its first syscall.
-
     Each entry has the following structure:
-
     * **`names`** *(array of strings, REQUIRED)* - the names of the syscalls.
         `names` MUST contain at least one entry.
     * **`action`** *(string, REQUIRED)* - the action for seccomp rules.
         A valid list of constants as of libseccomp v2.3.2 is shown below.
-
```
wking: I think we want to keep this pre-list (and the later post-list) blank lines too.  There are a number of other instances of this as well.

glossary.md
```diff
@@ -34,5 +34,4 @@ On Linux, the namespaces from which new [container namespaces](#container-namesp
 
 [JSON]: https://tools.ietf.org/html/rfc7159
 [UTF-8]: http://www.unicode.org/versions/Unicode8.0.0/ch03.pdf
-
```
wking: From the [style docs][1]:

> Referenced links should be kept in two alphabetically sorted sets, a general reference section followed by a man page section.

I'm fine dropping that and going to a single alphabetically sorted list, but if we want to do that we should also update the style docs.  I'm also fine keeping the separation and the blank line here.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/style.md#links

config-linux.md
Mashimiao: I removed them, because the blank line does not make sense.
But if it's traditional, I will keep them.

glossary.md
Mashimiao: If styles requries, by now I have no intention to change it. Let's keep where it was.

Mashimiao:
updated based on comments, PTAL

TomSweeneyRedHat:
LGTM, nice touch ups.

config-linux.md
```diff
@@ -509,10 +508,9 @@ The actions, architectures, and operators are strings that match the definitions
 The following parameters can be specified to set up seccomp:
 
 * **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as `syscalls[].action`.
-
 * **`architectures`** *(array of strings, OPTIONAL)* - the architecture used for system calls.
-    A valid list of constants as of libseccomp v2.3.2 is shown below.
 
+    A valid list of constants as of libseccomp v2.3.2 is shown below.
```
wking: I think you want to keep this line where it was (as part of the initial paragraph, with a blank line after it separating the sub-list).

config-linux.md
```diff
@@ -531,12 +529,9 @@ The following parameters can be specified to set up seccomp:
     * `SCMP_ARCH_S390X`
     * `SCMP_ARCH_PARISC`
     * `SCMP_ARCH_PARISC64`
-
 * **`syscalls`** *(array of objects, OPTIONAL)* - match a syscall in seccomp.
-
     While this property is OPTIONAL, some values of `defaultAction` are not useful without `syscalls` entries.
     For example, if `defaultAction` is `SCMP_ACT_KILL` and `syscalls` is empty or unset, the kernel will kill the container process on its first syscall.
-
```
wking: I think this “you'll probably want to set this” text is worth keeping in its own paragraph.  But I don't feel too strongly about it, and am ok with all of this getting squashed into a single paragraph like you're doing here.

config-linux.md
Mashimiao: updated.

vbatts:
rebase please. This LGTM.

vbatts:
_house keeping_: if someone else wants to pick up these commits and rebase, feel free to. It's been 1.5yrs needing only a rebase.

config-linux.md
```diff
@@ -450,25 +450,24 @@ The following parameters can be specified to set up the controller:
...
+If `intelRdt` is not set, the runtime MUST NOT manipulate any `resctrl` pseudo-filesystems.
 
 The following parameters can be specified for the container:
 
 * **`l3CacheSchema`** *(string, OPTIONAL)* - specifies the schema for L3 cache id and capacity bitmask (CBM).
-    If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
-
-    If `l3CacheSchema` is not set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
+    * If `l3CacheSchema` is set, runtimes MUST write the value to the `schemata` file in the `<container-id>` directory discussed in `intelRdt`.
+    * If `l3CacheSchema` is not set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
```
h-vetinari: Tagging this for reference in rebased PR

tianon:
Rebased/merged in #1031 :+1:

Thanks!

PR #938, opened 2017-12-01
config*: Dedent paragraphs that aren't a list entry

wking:
Like #936.  @Mashimiao pointed out that [there may be more instances of this][1], and this commit fixes all instances turned up with:

```console
$ git grep -B1 '^    [^ ]' | grep -A1 '[:-]\*\*'
```

[1]: https://github.com/opencontainers/runtime-spec/pull/934#discussion_r154246493


Mashimiao:
Uh....    I fixed them in #937

wking:
Ah good stuff.  I'll comment there.

PR #939, opened 2017-12-02
add definite limit for username mappings

Mashimiao:

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -87,6 +87,7 @@ Each entry has the following structure:
 * **`size`** *(uint32, REQUIRED)* - is the number of ids to be mapped.
 
 The runtime SHOULD NOT modify the ownership of referenced filesystems to realize the mapping.
+The runtime MUST generate an error when user namespace mappings is specified, but a new user namespace is not specified to create.
 Note that the number of mapping entries MAY be limited by the [kernel][user-namespaces].
 
 ### Example


vbatts:
I suppose this is fine. Any assumptions on providing that mapping would be left to the runtime implementation, so the MUST here makes sense.
LGTM

config-linux.md
```diff
@@ -87,6 +87,7 @@ Each entry has the following structure:
 * **`size`** *(uint32, REQUIRED)* - is the number of ids to be mapped.
 
 The runtime SHOULD NOT modify the ownership of referenced filesystems to realize the mapping.
+The runtime MUST generate an error when user namespace mappings is specified, but a new user namespace is not specified to create.
```
wking: Should “generate an error” be a link, like we generally use now:

```console
$ git describe
v1.0.1-20-g4ad8e74
$ git grep 'generate an error]' config-linux.md
config-linux.md:    The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
config-linux.md:If a `namespaces` field contains duplicated namespaces with same `type`, the runtime MUST [generate an error](runtime.md#errors).
config-linux.md:    If a [file][] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
config-linux.md:Runtimes MAY consider certain `cgroupsPath` values to be invalid, and MUST generate an error if this is the case.
config-linux.md:    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
```

wking: I'm also ok rewording this to be a config limitation, with something like:

> Configurations that set an ID mapping MUST have entry in `linux.namespaces` where `type` is `user` and `path` is unset.

That way runtimes can punt to the kernel, but folks who want to pre-check their config with runtime-tools (or other) validation can catch instances of this.

vbatts:
_house keeping:_ hrm. This is another conditional `MUST`, but it is a behavior I think is consistent with how runc and pals currently does. So rev'ing the spec version ought not be needed.
@opencontainers/runtime-spec-maintainers is this sufficient to add just this sentence? (i think it should be fine)

PR #941, opened 2017-12-06
config: add "umask" field to POSIX "user" section

cyphar:
Users may want to specify the umask(2) of the init process in a
container. This value is identical in semantics to POSIX. This is in
order to allow usage of an OCI container for a service which normally
only inherits the umask given to it.

See-also: https://github.com/opencontainers/runc/issues/1650
Requested-by: @leberknecht

### DIFF
--- a/config.md
+++ b/config.md
@@ -217,6 +217,7 @@ For POSIX platforms the `user` structure has the following fields:
 
 * **`uid`** (int, REQUIRED) specifies the user ID in the [container namespace](glossary.md#container-namespace).
 * **`gid`** (int, REQUIRED) specifies the group ID in the [container namespace](glossary.md#container-namespace).
+* **`umask`** (int, OPTIONAL) specifies the [umask][umask_2] of the user. If unspecified, the umask should not be changed from the calling process' umask.
 * **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs in the [container namespace](glossary.md#container-namespace) to be added to the process.
 
 _Note: symbolic name for uid and gid, such as uname and gname respectively, are left to upper levels to derive (i.e. `/etc/passwd` parsing, NSS, etc)_
@@ -233,6 +234,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
     "user": {
         "uid": 1,
         "gid": 1,
+        "umask": 63,
         "additionalGids": [5, 6]
     },
     "env": [
@@ -291,6 +293,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
     "user": {
         "uid": 1,
         "gid": 1,
+        "umask": 7,
         "additionalGids": [2, 8]
     },
     "env": [
@@ -843,6 +846,7 @@ Here is a full example `config.json` for reference.
 [selinux]:http://selinuxproject.org/page/Main_Page
 [no-new-privs]: https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt
 [proc_2]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt
+[umask.2]: http://pubs.opengroup.org/onlinepubs/009695399/functions/umask.html
 [semver-v2.0.0]: http://semver.org/spec/v2.0.0.html
 [ieee-1003.1-2008-xbd-c8.1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_01
 [ieee-1003.1-2008-functions-exec]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -60,6 +60,9 @@
         "GID": {
             "$ref": "#/definitions/uint32"
         },
+		"Umask": {
+            "$ref": "#/definitions/uint32"
+		},
         "ArrayOfGIDs": {
             "type": "array",
             "items": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -85,6 +85,8 @@ type User struct {
 	UID uint32 `json:"uid" platform:"linux,solaris"`
 	// GID is the group id.
 	GID uint32 `json:"gid" platform:"linux,solaris"`
+	// Umask is the umask for the init process.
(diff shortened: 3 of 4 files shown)


crosbymichael:
When do you plan to set this during init? Right before exec or when the rootfs is being setup like the existing umask settings?

config.md
```diff
@@ -217,6 +217,7 @@ For POSIX platforms the `user` structure has the following fields:
 
 * **`uid`** (int, REQUIRED) specifies the user ID in the [container namespace](glossary.md#container-namespace).
 * **`gid`** (int, REQUIRED) specifies the group ID in the [container namespace](glossary.md#container-namespace).
+* **`umask`** (int, OPTIONAL) specifies the [umask][umask_2] of the user. If unspecified, the umask should not be changed from the calling process' umask.
```
wking: nit: the possessive of “process” is “process's”, see us [here][1] and the Linux man pages [here][2].

nit: the “If unspecified…” sentence should [go on its own line][3].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config.md#L149
[2]: https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/commit/?id=a5a3e91bf97f08652849367eeef112813dceb809
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/style.md#one-sentence-per-line

schema/defs.json
```diff
@@ -60,6 +60,9 @@
         "GID": {
             "$ref": "#/definitions/uint32"
         },
+		"Umask": {
```
wking: The indent here is a bit strange.  You can automatically format these files by running `make fmt` in the `schema` directory.

wking:
> When do you plan to set this during init? Right before exec or when the rootfs is being setup…

The way [the spec reads now][1] (#700), runtimes will be able to chose either of those as they see fit.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L105

config.md
```diff
@@ -843,6 +846,7 @@ Here is a full example `config.json` for reference.
 [selinux]:http://selinuxproject.org/page/Main_Page
 [no-new-privs]: https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt
 [proc_2]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt
+[umask.2]: http://pubs.opengroup.org/onlinepubs/009695399/functions/umask.html
```
wking: nit: POSIX functions should go be in section 3 (or 3p), so `umask.2` should be `umask.3`.  `umask.2` would be [the Linux kernel docs for `umask`][1].

nit: you're linking to the 2004 edition of POSIX.  I'd rather stay consistent with our other links and use [the 2016 edition][2] (#858).

[1]: http://man7.org/linux/man-pages/man2/umask.2.html
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/umask.html

specs-go/config.go
```diff
@@ -85,6 +85,8 @@ type User struct {
 	UID uint32 `json:"uid" platform:"linux,solaris"`
 	// GID is the group id.
 	GID uint32 `json:"gid" platform:"linux,solaris"`
+	// Umask is the umask for the init process.
+	Umask uint32 `json:"umask,omitempty" platform:"linux,solaris"`
```
wking: Zero is a valid umask (it means “leave the permissions entirely up to the process itself”), so I think we need a pointer here.

cyphar:
@crosbymichael In `runc` I would want to do this right before exec because it refers to the container process, no? But maybe there's an argument for it to be done earlier -- I'm not sure?

vbatts:
@cyphar @crosbymichael I interpret this as being set right before exec. Perhaps that could be clearer.

wking:
On Wed, Apr 04, 2018 at 03:08:23PM -0700, Vincent Batts wrote:
> @cyphar @crosbymichael I interpret this as being set right before
> exec. Perhaps that could be clearer.

If you want that interpretation to be portable across compliant
runtimes, I think you'd want to land wording to that effect around
[1], with something like:

  Runtimes MUST NOT apply process.user.umask as part of creation; that
  property is only applied during [start](#start).

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/runtime.md#L103

stain:
As this is still pending approval since April - what needs to be done to progress this PR?

leberknecht:
Anything we can do here?

config.md
```diff
@@ -233,6 +234,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are
     "user": {
         "uid": 1,
         "gid": 1,
+        "umask": 63,
```
vbatts: i get that they're just ints, but i wish octals were a viewable in json, and 077 wouldn't be 77

vbatts:
> When do you plan to set this during init? Right before exec or when the rootfs is being setup like the existing umask settings?

I would see this as right before exec. What would you're hopes be?

cyphar:
I think right before exec would be most reasonable, since it's a setting about the container and not about configuring devices and others such things. Generally most `config.json` options aren't meant to impact internal runtime setup (except in cases where it can't be avoided like cgroups).

rhatdan:
This has come up again in a request for Podman.  Is there anything we could do to move this forward?

giuseppe:
> @crosbymichael In `runc` I would want to do this right before exec because it refers to the container process, no? But maybe there's an argument for it to be done earlier -- I'm not sure?

doing it just before exec seems like the correct thing to me.  Currently runc does it in `finalizeRootfs` (https://github.com/opencontainers/runc/blob/master/libcontainer/rootfs_linux.go#L151).  The umask inside the container is always set to `022`

crosbymichael:
LGTM

leberknecht:
@caniszczyk @dqminh @hqhq @mrunalp @philips @tianon @vbatts @vishh
sorry for spamming, but can someone please do the required second review here? :)

caniszczyk:
RFC @opencontainers/runtime-spec-maintainers

tianon:
I think @wking's comments look to be relevant and should be addressed (especially the inconsistent formatting).  It also seems like a good idea IMO to clarify when this is expected to be applied, per @crosbymichael's comments.

vbatts:
LGTM
And as this is `OPTIONAL` it is a non-breaking, forward compatible feature.

PR #942, opened 2017-12-07
Add definitions for RDMA controller/cgroup of Linux kernel 4.11

paravmellanox:
This request is for the inclusion of RDMA controller/cgroup which is added in Linux kernel 4.11.

Background:
RDMA Networking adapters provide low latency, direct access to hardware (HCA) to user space applications to send/receive messages and data in a clustered systems.
With latest HCAs, application to application level messages delivery latency can be as low as < 1usec.
RDMA kernel subsystem achieves this using RDMA resources such as QPs, MRs CQs objects.

RDMA cgroup allows limiting resources of the such RDMA HCA networking adapters.

RDMA resource creations such as hca_handle and hca_objects must be controlled so that certain set of processes doesn't take away all hardware resources.

RDMA controller allows limiting such HCA resource on per HCA basis.

Example implementation of this controller is already available below.
https://github.com/paravmellanox/cgroups/commit/469ac1c45b37c765411978b4417f21ff30a31ab0

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -169,7 +169,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 ## <a name="configLinuxControlGroups" />Control groups
 
 Also known as cgroups, they are used to restrict resource usage for a container and handle device access.
-cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
+cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids, network and RDMA resources for the container.
 For more information, see the [kernel cgroups documentation][cgroup-v1].
 
 ### <a name="configLinuxCgroupsPath" />Cgroups Path
@@ -455,6 +455,36 @@ The following parameters can be specified to set up the controller:
    }
 ```
 
+### <a name="configLinuxRDMA" />RDMA
+
+**`rdma`** (object, OPTIONAL) represents the cgroup subsystem `rdma`.
+For more information, see the kernel cgroups documentation about [rdma][cgroup-v1-rdma].
+
+The name of the device to limit is the entry key.
+Entry values are objects with the following properties:
+
+* **`hcaHandles`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_handles in the cgroup
+* **`hcaObjects`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_objects in the cgroup
+
+You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given entry, and MAY specify both.
+
+#### Example
+
+```json
+"rdma": {
+    "mlx5_1": {
+        "hcaHandles": 3,
+        "hcaObjects": 10000
+    },
+    "mlx4_0": {
+        "hcaObjects": 1000
+    },
+    "rxe3": {
+        "hcaObjects": 10000
+    }
+}
+```
+
 ## <a name="configLinuxIntelRdt" />IntelRdt
 
 **`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
@@ -647,6 +677,7 @@ The following parameters can be specified to set up seccomp:
 [cgroup-v1-net-cls]: https://www.kernel.org/doc/Documentation/cgroup-v1/net_cls.txt
 [cgroup-v1-net-prio]: https://www.kernel.org/doc/Documentation/cgroup-v1/net_prio.txt
 [cgroup-v1-pids]: https://www.kernel.org/doc/Documentation/cgroup-v1/pids.txt
+[cgroup-v1-rdma]: https://www.kernel.org/doc/Documentation/cgroup-v1/rdma.txt
 [cgroup-v2]: https://www.kernel.org/doc/Documentation/cgroup-v2.txt
 [devices]: https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
 [devpts]: https://www.kernel.org/doc/Documentation/filesystems/devpts.txt
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -320,6 +320,14 @@ type LinuxNetwork struct {
 	Priorities []LinuxInterfacePriority `json:"priorities,omitempty"`
 }
 
+// LinuxRdma for Linux cgroup 'rdma' resource management (Linux 4.11)
+type LinuxRdma struct {
+	// Maximum number of HCA handles that can be opened. Default is "no limit".
+	HcaHandles *uint32 `json:"hcaHandles,omitempty"`
+	// Maximum number of HCA objects that can be created. Default is "no limit".
+	HcaObjects *uint32 `json:"hcaObjects,omitempty"`
+}
+
 // LinuxResources has container runtime resource constraints
 type LinuxResources struct {
 	// Devices configures the device whitelist.
@@ -336,6 +344,10 @@ type LinuxResources struct {
 	HugepageLimits []LinuxHugepageLimit `json:"hugepageLimits,omitempty"`
 	// Network restriction configuration
 	Network *LinuxNetwork `json:"network,omitempty"`
+	// Rdma resource restriction configuration.
+	// Limits are a set of key value pairs that define RDMA resource limits,
+	// where the key is device name and value is resource limits.
+	Rdma map[string]LinuxRdma `json:"rdma,omitempty"`
 }
 
 // LinuxDevice represents the mknod information for a Linux special device file
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -175,6 +175,12 @@
                                 }
                             }
                         }
+                    },
+                    "rdma": {
+                        "type": "object",
+                        "additionalProperties": {
+                            "$ref": "defs-linux.json#/definitions/Rdma"
+                        }
                     }
                 }
             },
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -240,6 +240,17 @@
                 "priority"
             ]
         },
+        "Rdma": {
+            "type": "object",
+            "properties": {
+                "hcaHandles": {
+                    "$ref": "defs.json#/definitions/uint32"
+                },
+                "hcaObjects": {
+                    "$ref": "defs.json#/definitions/uint32"
+                }
+            }
+        },
         "NamespaceType": {
             "type": "string",
             "enum": [
--- a/schema/test/config/bad/linux-rdma.json
+++ b/schema/test/config/bad/linux-rdma.json
@@ -1,0 +1,15 @@
+{
+    "ociVersion": "1.0.0",
+    "root": {
+        "path": "rootfs"
+    },
+    "linux": {
+        "resources": {
+            "rdma": {
+                "mlx5_1": {
+                    "hcaHandles": "not a uint32"
+                }
+            }
+        }
+    }
+}
--- a/schema/test/config/good/linux-rdma.json
+++ b/schema/test/config/good/linux-rdma.json
@@ -1,0 +1,22 @@
+{
+    "ociVersion": "1.0.0",
+    "root": {
+        "path": "rootfs"
+    },
+    "linux": {
+        "resources": {
+            "rdma": {
+                "mlx5_1": {
+                    "hcaHandles": 3,
+                    "hcaObjects": 10000
+                },
+                "mlx4_0": {
+                    "hcaObjects": 1000
+                },
+                "rxe3": {
+                    "hcaObjects": 10000
+                }
+            }
+        }
+    }
+}


config-linux.md
```diff
@@ -169,7 +169,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 ## <a name="configLinuxControlGroups" />Control groups
 
 Also known as cgroups, they are used to restrict resource usage for a container and handle device access.
-cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids and network for the container.
+cgroups provide controls (through controllers) to restrict cpu, memory, IO, pids, network and rdma resources for the container.
```
wking: The kernel docs aren't particularly consistent about this, but since RDMA is [an acronym for remote direct memory access][1], I think we want to use RDMA here (and in other places, except for the literal `rdma`).

[1]: https://en.wikipedia.org/wiki/Remote_direct_memory_access

config-linux.md
```diff
@@ -455,6 +455,34 @@ The following parameters can be specified to set up the controller:
...
+### <a name="configLinuxRDMA" />RDMA
+
+**`rdma`** (object, OPTIONAL) represents the cgroup subsystem `rdma`.
+For more information, see the kernel cgroups documentation about [rdma][cgroup-v1-rdma].
+
+The following parameters can be specified to set up the controller:
+
+* **`hca_device`** *(string, REQUIRED)* - specifies the device name whose resources limit to be configured
+* **`hca_handles`** *(uint32, REQUIRED)* - specifies the maximum number of hca_objects in the cgroup for a specified device
+* **`hca_objects`** *(uint32, REQUIRED)* - specifies the maximum number of hca_handles in the cgroup for a specified device
```
wking: You have “Default is "no limit".” in the Go comments, but REQUIRED here.  Reading the kernel docs, I think `hca_handles` and `hca_objects` should both be OPTIONAL.  If you like, you can add a requirement like [this][1]

> You MUST specify at least one of `hca_handles` or `hca_objects` in a given entry, and MAY specify both.

The kernel docs also make a distinction between leaving those unset (no change from the previous limit) or setting them to `max` (clearing a possible previous limit).  Since the spec supports joining and tweaking existing cgroups, I think we need a way to distinguish between the two cases.  The current policy seems to be using signed integers and [using -1 for “explicitly remove any limits”][2].

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L338
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L262

specs-go/config.go
```diff
@@ -320,6 +320,16 @@ type LinuxNetwork struct {
...
 }
 
+// LinuxRdma for Linux cgroup 'rdma' resource management (Linux 4.11)
+type LinuxRdma struct {
+	// Hca device name whose resources to be restricted
+	HcaDevice string `json:"hca_device"`
+	// Maximum number of HCA handles that can be opened. Default is "no limit".
+	HcaHandles uint32 `json:"hca_handles"`
+	// Maximum number of HCA objects that can be created. Default is "no limit".
+	HcaObjects uint32 `json:"hca_objects"`
```
wking: Are zero values allowed (e.g. for “you can't use any HCA handles”)?  If not, we should say so in the Markdown spec.  If so, we should make these `omitempty` pointers so we can distinguish between “set to zero” and “unset”.

paravmellanox: yes. setting zero is allowed.

config-linux.md
paravmellanox: ok. yes, I will change rdma to RDMA.

config-linux.md
paravmellanox: You are right. I initially had it optional but than made it required to simplify the -1. But I think its fine as it already exist. 
I will make both of them as optional.
So when user doesn't want to configure any one of the value, it should set to -1. Correct?
Because at kernel level they are just optional KV pair, but at cgroups spec level it is a structure. So no_limit should be initialized as -1. Right?

wking: > So when user doesn't want to configure any one of the value, it should set to -1. Correct?

No, leave it unset to not change the limit (nothing written to the control file for this parameter).  Set to -1 to *clear* the limit (`max` written to the cgroup control file for this parameter).

paravmellanox: so when cgroups.New() is invokes, there are 4 valid values for hca_objects/hca_handles.
1. 0 - no allocation possible
2. +ve - valid positive value
3. -1 to clear the limits
4. do not touch the limits 

If we make these two parameters optional, what should user set in specs.LinuxRdma.HcaObjects to convey - 'do not touch the limits'.
Setting 0 will disable it (which is valid value).

wking: > ... what should user set in specs.LinuxRdma.HcaObjects to convey - 'do not touch the limits'.

`nil`, which is why [we want pointers][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155593626

paravmellanox: ok. Sounds good. I will revise the spec.

config-linux.md
```diff
@@ -455,6 +455,35 @@ The following parameters can be specified to set up the controller:
...
 ```
 
+### <a name="configLinuxRDMA" />RDMA
+
+**`rdma`** (object, OPTIONAL) represents the cgroup subsystem `rdma`.
+For more information, see the kernel cgroups documentation about [rdma][cgroup-v1-rdma].
+
+The following parameters can be specified to set up the controller:
+
+* **`hca_device`** *(string, OPTIONAL)* - specifies the device name whose resources limit to be configured
```
wking: I'd rather keep this REQUIRED, because there's no point in supporting `"rdma": [{}]`.  It also lets you drop the later line about “You must specify a valid `hca_device`…”.

specs-go/config.go
```diff
@@ -320,6 +320,16 @@ type LinuxNetwork struct {
 	Priorities []LinuxInterfacePriority `json:"priorities,omitempty"`
 }
 
+// LinuxRdma for Linux cgroup 'rdma' resource management (Linux 4.11)
+type LinuxRdma struct {
+	// Hca device name whose resources to be restricted
+	HcaDevice string `json:"hca_device,omitempty"`
```
wking: If you go back to REQUIRing `hca_device`, you'll want to drop `omitempty` here.

config-linux.md
paravmellanox: ok. so if I understand correctly, "rdma": [{}] can be treated as illegal config.
"rdma": [{"hca_device": "m1x5_1"}] is valid. And other two parameters are optional anyway.

So you are saying if end user passes hca_device only then we create the rdma controller?
Won't Docker or other engines create all empty cgroup controllers?

wking: > so if I understand correctly, `"rdma": [{}]` can be treated as illegal config

That's what you'd get if you REQUIRED `hca_device` (which I think is a good idea for consistency with [`rlimits[].type`][1] ).

> `"rdma": [{"hca_device": "m1x5_1"}]` is valid…

As you have it now.  But if you [requiring at least one of `hca_handles` or `hca_objects` to be set][2], then that would be invalid too (and we'd be consistent with the [current block I/O approach][3]).

> So you are saying if end user passes `hca_device` only then we create the rdma controller?  Won't Docker or other engines create all empty cgroup controllers as `"rdma": [{}]`?

If you have `rdma` unset or set to `[{}]` (or `[{"hca_device": "m1x5_1"}]`, etc.), you're not adding new limits.  The spec position on that [is][4]:

> Runtimes MAY attach the container process to additional cgroup controllers beyond those necessary to fulfill the `resources` settings.

which means that, for a config that does not adjust limits via `linux.resources.rdma`, the runtime can attach the container process to the rdma controller at `cgroupsPath` or not as it sees fit.  See #493, #834, and opencontainers/runc#861 for more background.

In case it helps clarify, I think this section of text should look like:

```

* **`hca_device`** *(string, REQUIRED)* - specifies the device name whose resources limit to be configured
* **`hca_handles`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_objects in the cgroup for a specified device
* **`hca_objects`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_handles in the cgroup for a specified device

You MUST specify at least one of `hca_handles` or `hca_objects` in a given entry, and MAY specify both.
```

To close another no-op loophole we could follow [this][5] and require `rdma` to, when set, contain at least one entry.  That would be new wording though, since the spec doesn't currently do that for any OPTIONAL property.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config.md#L167
[2]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155593193
[3]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155593193
[4]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L197
[5]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L543

paravmellanox: section of text that you described looks good to me. If we follow the text you described, it closes the no-op loophole too, isn't it for rdma controller?

wking: > If we follow the text you described, it closes the no-op loophole too, isn't it for rdma controller?

You could still have `"rdma": []` unless you REQUIRE the array to have at least one entry.  We don't close that loophole for any other OTIONAL properties at the moment (e.g. `"mounts": []` is ok), so we'd need maintainer input on whether it was worth setting a new precedent.

Another issue is the possibility of duplicate `hca_device` entries.  My preferred approach to tgat would be to use an object:

```
"rdma": {
  "m1x5_1": {
    "hca_handles": 3
  }
}
```

My attempt to do that for rlimits was rejected (#583), but I'm not clear on why.  It's possible that maintainers would support an object for this new property.

paravmellanox: I wouldn't worry about duplicate entries in case of rdma, because caller who is setting duplicate entries knows what is set. Last accessed entry would be applied, so that is fine.

I think keeping "rdma": [] in sync with "mounts":[] is ok.
This also allows to update the entry at later point without the need to new cgroup and moving all container processes to that new cgroup. Because in that process, some process might get left out without freeze-resume operation as mentioned in https://github.com/opencontainers/runc/issues/861.
so "rdma": [] seems better idea to me and I actually prefer that because it allows more flexibility in future for dynamically created RDMA devices which is on horizon currently and I am actively involved on kernel side of it, but not yet ready enough to discuss in this forum.

I don't have any strong preference or objection to schema proposed in this PR and below one.
Both schema are fine to me. Below one is little more intuitive where its evident that limits are of the specified device. I would +1 for below schema regardless of consideration for duplication.

```
"rdma": {
  "m1x5_1": {
    "hca_handles": 3
  }
}
```

wking: > I wouldn't worry about duplicate entries in case of rdma, because caller who is setting duplicate entries knows what is set. Last accessed entry would be applied, so that is fine.

We worry about them for `rlimits` (#607) and `namespaces` (#597), I don't see why this would be different.  One use-case would be if one person writes the template, and another uses [oci-runtime-tool][1] or `jq` to edit a particular entry.  That editing is easier if you're using a map, because accessing a particular entry is easier, you don't need to worry about finding the last matching entry, and you don't have to decide what to do about earlier matching entries.

> This also allows to update the entry at later point without the need to new cgroup and moving all container processes to that new cgroup.

If there are semantic differences between leaving `rdma` unset and setting it to `[]` (or, with the object approach, `{}`), I think we need to be very clear about what they are.  And we probably need to adjust the Go property to use a pointer to an array, and to not be `omitempty`.  I don't think that path is going to be particularly friendly :p.

[1]: https://github.com/opencontainers/runtime-tools

paravmellanox: There is no mention in runc spec on processing order of array or list of objects so finding last matching entry is possibly one way.
But I prefer less ambiguous schema as you have described, so I would go forward with it.

how about below summary.

1. If rdma is unset - no need to create rdma controller.
2. If "rdma": [] is provided, create rdma controller. It is ok to create empty controller.
Advantages: 
(a) No need to create new rdma cgroup when limits are updated
(b) As described at start of https://github.com/opencontainers/runc/issues/861, code is lot simpler which doesn't need to migrate existing processes to new cgroup via freeze-resume approach
3. "rdma":[ {} } ] accepts below zero or more entries.
```
  "mlx5_1": {
    "hca_handles": 3
    "hca_objects": 3
  },
  "mlx5_2": {
    "hca_handles": 3
  },
```
If this looks reasonable starting point, I will update the spec.

wking: > If this looks reasonable starting point...

(2) would be different from how the spec handles the current cgroups, at least to my reading.  And it doesn't avoid migration when the user leaves `rdma` unset at creation (your case (1)) but later sets it in an update (unless the runtime always joins all controllers).  And Go handling would need to distinguish between unset and an empty map.  But mitigating migration would be useful.  I'm now cautiously optimistic about this direction, although it would be good to have some maintainers weigh in.

paravmellanox: > (2) would be different from how the spec handles the current cgroups, at least to my reading. 

Below sample creates both the cgroups test1 and test2, which I think is correct behavior what I was expecting in sync with (2). or I misunderstood your comment?

```
package main

import (
        "cgroups"
        specs "github.com/opencontainers/runtime-spec/specs-go"
)

func main() {
        limits := new(specs.LinuxPids)

        _, _ = cgroups.New(cgroups.V1, cgroups.StaticPath("/test1"),
                           &specs.LinuxResources{
                           Pids : limits},)

        _, _ = cgroups.New(cgroups.V1, cgroups.StaticPath("/test2"),
                           &specs.LinuxResources{
                           Pids : nil},)
}
```
Regarding updating limits, cgroups go code does create new cgroup if it doesn't exist and apply the limits. I am not sure of process migration part using freeze-resume.

wking: > ```
> import (
>         "cgroups"
> ```

What is `cgroups`?  With your snippet in `test.go`:

```console
$ go run test.go 
test.go:4:2: cannot find package "cgroups" in any of:
        /usr/lib/go/src/cgroups (from $GOROOT)
        /home/wking/.local/lib/go/src/cgroups (from $GOPATH)
```

It looks like reading in the JSON would be ok, but I can't find a convenient approach to consistent serialization:

```
$ cat test.go
package main

import (
	"encoding/json"
	"fmt"
	"os"
)

type SomeType struct {
	SomeMap map[string]string `json:"someMap"`
}

func main() {
	for _, data := range []string{
		"{}",
		"{\"someMap\": {}}",
		"{\"someMap\": {\"a\": \"b\"}}",
	} {
		someType := SomeType{}
		err := json.Unmarshal([]byte(data), &someType)
		if err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
		out, err := json.Marshal(someType)
		if err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
		fmt.Printf("%s -> %v (nil? %t, serialized: %s)\n", data, someType, someType.SomeMap == nil, string(out))
	}
}
$ go run test.go
{} -> {map[]} (nil? true, serialized: {"someMap":null})
{"someMap": {}} -> {map[]} (nil? false, serialized: {"someMap":{}})
{"someMap": {"a": "b"}} -> {map[a:b]} (nil? false, serialized: {"someMap":{"a":"b"}})
```

So you can use `nil` comparison to distinguish between “unset” (`== nil`) and “set to an empty object” (`!= nil && len(…) == 0`).  But we don't want the literal `null` in the serialized output (#662) for the originally-unset case.  And if we use:

```
SomeMap map[string]string `json:"someMap,omitempty"`
```

both the “unset” and “set to an empty object” cases serialize to “unset”.  So what Go type should we use to round-trip the distinction between “unset” and “set to an empty object”?

paravmellanox: >What is cgroups? With your snippet in test.go:
I am sorry for not putting github path there.
It is
"github.com/containerd/cgroups"

wking: > Below sample creates both the cgroups test1 and test2, which I think is correct behavior what I was expecting in sync with (2)…
>
> ```
>        limits := new(specs.LinuxPids)
>
>        _, _ = cgroups.New(cgroups.V1, cgroups.StaticPath("/test1"),
>                           &specs.LinuxResources{
>                           Pids : limits},)
> ```

`pids` is not a great example, because [`limits` is REQUIRED][1].  That means you cannot access [case (2)][2], which would be `"pids": {}` (meaning “create the cgroup (if necessary) and attach the container process to the cgroup, but don't adjust any limits”).

> ```
>         _, _ = cgroups.New(cgroups.V1, cgroups.StaticPath("/test2"),
>                            &specs.LinuxResources{
>                            Pids : nil},)
> ```

This is [case (1)][2], where runtimes [MAY][3] create the cgroup (if necessary) and attach the container process to the cgroup.  Apparently	containerd/cgroups decides to do that.  But runtimes don't have to.  And not creating the cgroup (and not attaching the container process to the non-existent cgroup) is the lazy-cgroups approach discussed in opencontainers/runc#861.  If containerd/cgroups decided to implement the lazy-cgroups approach on `nil`, their [`Create`][4] would instead look like:

```go
func (p *pidsController) Create(path string, resources *specs.LinuxResources) error {
  resources.Pids == nil {
    return nil  // case (1), opt for lazy creation and user did not set "pids"
  }
  if err := os.MkdirAll(p.Path(path), defaultDirPerm); err != nil {
    return err
  }
  if resources.Pids.Limit == nil {
    return nil  // case (2), user set "pids": {}, and we're pretending limit is OPTIONAL and an *int64
  }
  // case (3), user set "pids": {"limit": …}
  return ioutil.WriteFile(
    filepath.Join(p.Path(path), "pids.max"),
    []byte(strconv.FormatInt(resources.Pids.Limit, 10)),
    defaultFilePerm,
  )
}
```

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L448
[2]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155914884
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L197
[4]: https://github.com/containerd/cgroups/blob/29da22c6171a4316169f9205ab6c49f59b5b852f/pids.go#L31-L43

paravmellanox: > This is case (1), where runtimes MAY create the cgroup (if necessary) and attach the container process to the cgroup. Apparently containerd/cgroups decides to do that. But runtimes don't have to. And not creating the cgroup (and not attaching the container process to the non-existent cgroup) is the lazy-cgroups approach discussed in opencontainers/runc#861. If containerd/cgroups decided to implement the lazy-cgroups approach on nil, their Create would instead look like:

I agree to what you described.
However current scheme (1) of runc-spec and its implementation sounds reasonble to me and fits the requirements of rdma too without #861.
When https://github.com/opencontainers/runc/issues/861 scheme is adopted, rdma could possibly follow that.

In order to make progress, I propose to move forward with currently defined scheme and implementation of (1), (2) and (3).
(2) and (3) are already matches generic runc-spec definition.

(1) is implemented by all cgroups (not just pids) in https://github.com/containerd/cgroups.
If (1) needs change in runc-spec definition or wording, it is going to be different PR which will affect other resources too.

wking: > If (1) needs change in runc-spec definition or wording, it is going to be different PR which will affect other resources too.

That makes sense to me.  I had been under the impression (based on [this comment][1]), that you intended to start requiring cgroup creation (if necessary) and joining in the `"rmda": {}` case, which is not a requirement that the current spec has for other controllers.  Currently the spec, as I read it, allows runtimes to decide whether they want to create/join or not for both the “unset” (e.g. `"resources": {}`) and “set empty” cases (e.g. `"resources": {"memory": {}}`) because of [this clause][2].  If you're ok using that same logic (for now) for `rmda`, then I'm back to preferring the wording I floated [here][3].

[1]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155914884
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L197
[3]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155908710

paravmellanox: >That makes sense to me. I had been under the impression (based on this comment), that you intended to start requiring cgroup creation (if necessary) and joining in the "rmda": {} case, which is not a requirement that the current spec has for other controllers. Currently the spec, as I read it, allows runtimes to decide whether they want to create/join or not for both the “unset” (e.g. "resources": {}) and “set empty” cases (e.g. "resources": {"memory": {}}) because of this clause. If you're ok using that same logic (for now) for rmda, then I'm back to preferring the wording I floated here.

Based on the clause you mention https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config-linux.md#L197 and based on implementation at https://github.com/containerd/cgroups, which follows the MAY part, -> "rdma": [{}] and "rdma": [{"hca_device": "m1x5_1"}] is also valid similar to priorities array of objects. For rdma instead of array of objects, as you suggested, we will have map as 
```
"rdma": {
  "m1x5_1": {
    "hca_handles": 3
  }
}
```
Are you ok with that?

wking: > For rdma instead of array of objects, as you suggested, we will have map as
>
> ```
> "rdma": {
>   "m1x5_1": {
>     "hca_handles": 3
>   }
> }
> ```
>
> Are you ok with that?

Ah, right, that would change the wording.  I'm in favor of the object/map approach, yes.

paravmellanox: ok. Great. So let me revise the spec to have treat them as optional and as map instead of array.

config-linux.md
```diff
@@ -455,6 +455,39 @@ The following parameters can be specified to set up the controller:
...
+        * **`hca_handles`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_objects in the cgroup
+        * **`hca_objects`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_handles in the cgroup
+        You MUST specify at least one of the `hca_handles` or `hca_objects` in a given entry, and MAY specify both.
+
+#### Example
+
+```json
+    "rdma": {
+        "limits": [
+            "mlx5_1": {
```
wking: This is not valid JSON.  I think you want to drop `limits` and just make [`rdma` an object with device-name keys and handles/objects-limit values][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155913558

specs-go/config.go
```diff
@@ -320,6 +320,20 @@ type LinuxNetwork struct {
 	Priorities []LinuxInterfacePriority `json:"priorities,omitempty"`
 }
 
+// LinuxRdma for Linux cgroup 'rdma' resource management (Linux 4.11)
+type LinuxRdma struct {
+	// Limits are a set of key value pairs that define RDMA resource limits,
+	// where the key is device name and value is resource limits.
+	Limits map[string]LinuxRdmaLimit `json:"limits,omitempty"`
```
wking: This is whatI think you want for `rdma`; no need for a `Limits`.  Just use:

```go
type LinuxRdma map[string]LinuxRdmaLimit
```

Or skip `LinuxRdma entirely and just inline `map[string]LinuxRdmaLimit` in `LinuxResources`.

paravmellanox: > Or skip LinuxRdma entirely and just inlinemap[string]LinuxRdmaLimitinLinuxResources`.
Simplification looks fine to me. Sending updates.

paravmellanox:
@wking any updates? Do you need any more information for this PR?

wking:
On Tue, Jan 23, 2018 at 02:16:29PM +0000, Parav Pandit wrote:
> @wking any updates?

I've filed paravmellanox/runtime-spec#1 with some suggested changes.
If those look good, feel free to squash them into your PR.

wking:
With paravmellanox/runtime-spec#1 merged, the new content looks good to me.  You might want to squash this PR down to a single commit.  And passing tests are blocked on #945.  Any maintainers want to look this over?

paravmellanox:
I am not sure that if I squash and force push the new squashed commit, will the discussions happened here will be lost or not.
I was hoping that If maintainer does squash merge, than discussion is preserved and final merged code also has single commit.
I little new in this area, so please guild me.

wking:
On Wed, Jan 24, 2018 at 11:28:54PM +0000, Parav Pandit wrote:
> I am not sure that if I squash and force push the new squashed
> commit, will the discussions happened here will be lost or not.

Whatever happens here, discussion in the PRs will be preserved
(possibly under “Show outdated” accordions like [1]).

> I was hoping that If maintainer does squash merge, than discussion
> is preserved and final merged code also has single commit.

I don't do much squash-merging via GitHub's UI, so I'm not sure.  But
one commit of mine that was squash-merged didn't end up with any
useful commit message [2] despite the original commits having commit
messages [3].  If you are comfortable squashing, I suggest you do so
and write up notes about alternatives you considered and why you
eventually went with what you did for the squashed commit's message.
That way you don't have to rely on the squash-merge to get something
sensible.

> I little new in this area, so please guild me.

And I'm not a maintainer, so this is all just my 2¢.  Do whatever
makes sense to you.

[1]: https://github.com/opencontainers/runtime-spec/pull/942#discussion_r155590497
[2]: https://github.com/opencontainers/runtime-tools/commit/7a4cb362793899d1dbf04cb5eb8bf5c6577cb55c
[3]: https://github.com/opencontainers/runtime-tools/pull/321/commits/a2725b42d7b312681ed5315cf7e663a13c487445

paravmellanox:
Hi @wking, @mrunalp, and other reviewers, 
I squashed all the commits and now have single commit for this PR.
Can you please review and merge it?
Code will follow soon sometime next week once this is merged.

paravmellanox:
@wking , thanks a lot for the inputs, fixup and additional JSON definitions.

paravmellanox:
Hi, 

@vishh @vbatts @philips @mrunalp @hqhq @dqminh @tianon @crosbymichael @caniszczyk 

Can you please review this PR? @wking and I have went through several iterations. Feel that this is good start now and like to merge this PR at earliest to provide this feature to users waiting for it.

config-linux.md
```diff
@@ -455,6 +455,36 @@ The following parameters can be specified to set up the controller:
...
 
+### <a name="configLinuxRDMA" />RDMA
+
+**`rdmaLimits`** (object, OPTIONAL) represents the cgroup subsystem `rdma`.
+For more information, see the kernel cgroups documentation about [rdma][cgroup-v1-rdma].
+
+The name of the device to limit is the entry key.
+Entry values are objects with the following properties:
+
+* **`hca_handles`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_objects in the cgroup
```
crosbymichael: There should be no `_` in the json properties

crosbymichael:
Overall, this looks good to me once the `_` are removed from the json.

paravmellanox:
Removed the underscore and updated the PR.

crosbymichael:
Why did you close?

paravmellanox:
Closed by mistake. I am sorry.

crosbymichael:
Can you please rebase on master to pick up the changes that fix the CI?

crosbymichael:
Sorry, did you merge or rebase?

crosbymichael:
```bash
git fetch origin && git reset --hard origin/master
git cherry-pick 5f778472aa6a2f168964e9105ecd797fe1426372
```

;)

paravmellanox:
Got closed during rebase process, must be something wrong I did from the GUI interface. Reopening it again.

paravmellanox:
After the rebase, travis-ci still seems to fail while executing:
make -C schema test

ERROR: invalid document-file path: stat test/config/good/invalid-json.json: no such file or directory

  testing test/config/bad/linux-rdma.json
The document is valid
    received unexpected validation success
make: *** [test] Error 1

@wking or @crosbymichael, how to check which out of the 2 check failed the build and where to see what caused the failure if it is due to error in linux-rdma.json

wking:
> After the rebase…

Looks like you haven't pushed [the cherry-pick][1]:

```console
$ git fetch origin
$ git log -1 --oneline --decorate origin/pr/942
5d9bbad (origin/pr/942, origin/pr/936) config: Dedent root paragraphs, since they aren't a list entry
```

[1]: https://github.com/opencontainers/runtime-spec/pull/942#issuecomment-366335180

paravmellanox:
I followed below steps, likely not optimal.
1. Reset the head of paravmellanox/runtime-spec master branch to an older version.
2. Rebase & merge from opencontainers/runtime-spec.
3. git am <rdma-cgroup.patch>
4. git push origin master

This updated the PR with 4 patches (patches after rebase and new patch).
I am not sure how to avoid these patches which came as rebase to not show up in old PR.

As a result of this push command, travis-ci build triggered.
It looked at test/config/good/invalid-json.json

schema/test/config/good directory contains only 3 files. minimal*.json and spec-example.json.
invalid-json.json file is not found. It is present in test/config/bad directory.

wking:
On Fri, Feb 16, 2018 at 12:56:45PM -0800, Parav Pandit wrote:
> 3. git am <rdma-cgroup.patch>

Without the content of the patch, it's hard to know what this is
doing.  It'd easier to tell with @crosbymichael's cherry-pick
suggestion [1].  Filling in a bit more detail:

  $ git remote -v
  origin  git://github.com/opencontainers/runtime-spec.git (fetch)
  origin  git://github.com/opencontainers/runtime-spec.git (push)
  paravmellanox  https://github.com/paravmellanox/runtime-spec.git (fetch)
  paravmellanox  https://github.com/paravmellanox/runtime-spec.git (push)

I'm guessing there, you may be using other remote names for the
opencontainers repo and/or your fork.  Just substitute your names
below.  The scheme (git:// vs. https:// vs. ssh://) doesn't matter
either, as long as you have something pushable for paravmellanox's
push entry.

  $ git fetch origin
  $ git checkout master
  $ git reset --hard origin/master
  $ git cherry-pick 5f778472aa6a2f168964e9105ecd797fe1426372
  $ git push -f paravmellanox/master

[1]: https://github.com/opencontainers/runtime-spec/pull/942#issuecomment-366335180

paravmellanox:
Hi @wking, I did follow the steps as described. However I continue to see the error:

ERROR: invalid document-file path: stat test/config/good/invalid-json.json: no such file or directory

As I mentioned, this file doesn't exist in https://github.com/opencontainers/runtime-spec/tree/master/schema/test/config/good directory where Travis-CI is looking for a file.
It is instead located in https://github.com/opencontainers/runtime-spec/tree/master/schema/test/config/bad

wking:
On Tue, Feb 20, 2018 at 10:40:56PM +0000, Parav Pandit wrote:
> Hi @wking, I did follow the steps as described.

The f25d7b463 you pushed today is a clean cherry-pick of 5f77847 onto
master, so that's good.

> ERROR: invalid document-file path: stat
> test/config/good/invalid-json.json: no such file or directory

This sounds like #947.  I've just filed a bump there.

paravmellanox:
Hi @wking, @crosbymichael, Is there anything I can do to remove the bump of #947? I do not have much know-how of Travis-CI though, but I can attempt.

wking:
On Thu, Feb 22, 2018 at 12:09:29AM +0000, Parav Pandit wrote:
> Hi @wking, @crosbymichael, Is there anything I can do to remove the
> bump of #947?

You can avoid the problem by dropping the schema/test/config/bad/…
additions.  But #947 is a four-character fix, I'd just wait for the
maintainers to get around to it.

paravmellanox:
Sure. @wking lets wait for another day to resolve #947.

paravmellanox:
In order to make progress to make use of this functionality, I updated PR to drop schema/test/config/bad/linux-rdma.json for now. @crosbymichael and others please continue to review and merge the PR.

config-linux.md
```diff
@@ -455,6 +455,36 @@ The following parameters can be specified to set up the controller:
...
+
+* **`hcaHandles`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_handles in the cgroup
+* **`hcaObjects`** *(uint32, OPTIONAL)* - specifies the maximum number of hca_objects in the cgroup
+
+You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given entry, and MAY specify both.
+
+#### Example
+
+```json
+"rdmaLimits": {
```
crosbymichael: is there a reason why you named this `rdmaLimits` instead of just `rdma`?

wking:
> In order to make progress to make use of this functionality, I updated PR to drop schema/test/config/bad/linux-rdma.json for now.

#947 landed today, so you can restorw the bad example.

config-linux.md
wking: > is there a reason why you named this `rdmaLimits` instead of just `rdma`?

I like `rdma` better, although I don't care much.  Previously, @paravmellanox [cited][1] `hugepageLimits` for precedent, although the spec obviously also contains limiting properties without a `Limits` suffix.

[1]: https://github.com/paravmellanox/runtime-spec/pull/1#pullrequestreview-91323208

paravmellanox:
both changes done. @wking , @crosbymichael 
rdmaLimits -> rdma.
Added back bad/linux-rdma.json file.
Travis-CI build is passing now.

crosbymichael:
LGTM

paravmellanox:
@crosbymichael, who else from the group should be reviewing these changes?

crosbymichael:

mrunalp:
LGTM

PR #943, opened 2017-12-07
meeting: Bump January meeting from the 3rd to the 10th

wking:
Catching up with:

On Thu, Dec 07, 2017 at 09:14:32AM -0600, @caniszczyk [wrote][1]:
> btw, I bumped the January meeting a week later due to the holidays, that should solve issues and updated the calendar invite

I've used [`RDATE`][2] to add the new date and [`EXDATE`][3] to remove the old date.  The ICS was validated [here][4].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/FzclHQsMH_c/QKJMybRkBAAJ
[2]: https://tools.ietf.org/html/rfc5545#section-3.8.5.2
[3]: https://tools.ietf.org/html/rfc5545#section-3.8.5.1


philips:
LGTM

wking:
There's not all that much point in merging the January bump now, although we may still want this PR for the “which are usually” wiggle.

vbatts:
LGTM

PR #945, opened 2018-01-05
schema: Completely drop our JSON Schema 'id' properties

wking:
We're using [JSON Schema draft-04][1], as [declared][1.5] by our [`$schema` properties][2].  In draft-04, the `id` keyword alters the resolution scope.  But our current `$ref` values use [JSON Pointers][4] with relative references like `defs-linux.json#/definitions/Device` that ignore the `id`s.

By draft-07, `id` has become `$id`, and [the docs say][5]:

> The root schema of a JSON Schema document SHOULD contain an "$id" keyword with a URI (containing a scheme).

But since xeipuuv/gojsonschema@83a7f6369d5, including any URI that cannot be retrieved generates an error:

```console
$ ./validate config-schema.json test/config/good/minimal.json
Could not read schema from HTTP, response status is 404 Not Found
```

While a root `id` entry would be nice, we don't currently host these anywhere with a useful URI.  We could use GitHub links like [this][7], but then testing pull requests would be difficult.

By draft-07, the purpose of internal `$id` entries is clearly [explained][5]:

> Providing a plain name fragment enables a subschema to be relocated within a schema without requiring that JSON Pointer references are updated.

We don't need that, because we control all the references.  In the infrequent event of a subschema move, we can update the consuming references in the same commit.

The draft-07 `$ref` docs also explain that `$ref` targets [may be URNs][8]:

> The URI is not a network locator, only an identifier.  A schema need not be downloadable from the address if it is a network-addressable URL, and implementations SHOULD NOT assume they should perform a network operation when they encounter a network-addressable URI.

I haven't found analogous wording for `$id`, but it's possible that gojsonschema is being overly agressive with its attempted retrievals.

This commit removes all of our `id` entries.  The resulting JSON Schema is valid (regardless of where you host it) and does not generate the 404s.

If/when we merge/release this it should address opencontainers/runc#1680.

[1]: https://tools.ietf.org/html/draft-zyp-json-schema-04#section-7.2
[1.5]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/schema/config-schema.json#L3
[2]: https://tools.ietf.org/html/draft-zyp-json-schema-04#section-6
[3]: https://tools.ietf.org/html/draft-ietf-appsawg-json-pointer-07
[4]: https://tools.ietf.org/html/rfc6901
[5]: https://tools.ietf.org/html/draft-handrews-json-schema-00#section-9.2
[6]: https://github.com/xeipuuv/gojsonschema/commit/83a7f6369d552cba07cdeea7dc622f1161a07deb
[7]: https://raw.githubusercontent.com/opencontainers/runtime-spec/v1.0.1/schema/config-schema.json
[8]: https://tools.ietf.org/html/draft-handrews-json-schema-00#section-8

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -1,32 +1,27 @@
 {
     "linux": {
         "description": "Linux platform-specific configurations",
-        "id": "https://opencontainers.org/schema/bundle/linux",
         "type": "object",
         "properties": {
             "devices": {
-                "id": "https://opencontainers.org/schema/bundle/linux/devices",
                 "type": "array",
                 "items": {
                     "$ref": "defs-linux.json#/definitions/Device"
                 }
             },
             "uidMappings": {
-                "id": "https://opencontainers.org/schema/bundle/linux/uidMappings",
                 "type": "array",
                 "items": {
                     "$ref": "defs.json#/definitions/IDMapping"
                 }
             },
             "gidMappings": {
-                "id": "https://opencontainers.org/schema/bundle/linux/gidMappings",
                 "type": "array",
                 "items": {
                     "$ref": "defs.json#/definitions/IDMapping"
                 }
             },
             "namespaces": {
-                "id": "https://opencontainers.org/schema/bundle/linux/namespaces",
                 "type": "array",
                 "items": {
                     "anyOf": [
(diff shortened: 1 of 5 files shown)


crosbymichael:
LGTM

tianon:
LGTM

PR #949, opened 2018-02-12
config: Add VM-based container configuration section

sameo:
This adds a section to describe VM based container configurations to be
used by OCI runtimes using hardware virtualization to provide another
layer of isolation.

As part of this section we define 3 entries:

-  A virtual machine root image opbject. This is the guest image that
   contains the virtual machine root filesystem. The container image will
   be mounted on top of that filesystem.

- A virtual machine kernel object. This is the kernel that will boot
  inside the virtual machine. The object describes the host kernel image
  path, additional parameters and an optional guest initrd for the
  kernel to use.

- A virtual machine hypervisor object. This is the hypervisor that will
  manage the container virtual machine from the host. The object
  describe a hypervisor binary path and some additional parameters.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>
Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>

Important note: This is a refresh of the very long standing PR #405.

### DIFF
--- a/config-vm.md
+++ b/config-vm.md
@@ -1,0 +1,68 @@
+# <a name="VirtualMachineSpecificContainerConfiguration" /> Virtual-machine-specific Container Configuration
+
+This section describes the schema for the [virtual-machine-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+The virtual-machine container specification provides additional configuration for the hypervisor, kernel, and image.
+
+## <a name="HypervisorObject" /> Hypervisor Object
+
+**`hypervisor`** (object, OPTIONAL) specifies details of the hypervisor that manages the container virtual machine.
+* **`path`** (string, REQUIRED) path to the hypervisor binary that manages the container virtual machine.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+* **`parameters`** (array of strings, OPTIONAL) specifies an array of parameters to pass to the hypervisor.
+
+### Example
+
+```json
+    "hypervisor": {
+        "path": "/path/to/vmm",
+        "parameters": ["opts1=foo", "opts2=bar"]
+    }
+```
+
+## <a name="KernelObject" /> Kernel Object
+
+**`kernel`** (object, REQUIRED) specifies details of the kernel to boot the container virtual machine with.
+* **`path`** (string, REQUIRED) path to the kernel used to boot the container virtual machine.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+* **`parameters`** (array of strings, OPTIONAL) specifies an array of parameters to pass to the kernel.
+* **`initrd`** (string, OPTIONAL) path to an initial ramdisk to be used by the container virtual machine.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+
+### Example
+
+```json
+    "kernel": {
+        "path": "/path/to/vmlinuz",
+        "parameters": ["foo=bar", "hello world"],
+        "initrd": "/path/to/initrd.img"
+    }
+```
+
+## <a name="ImageObject" /> Image Object
+
+**`image`** (object, OPTIONAL) specifies details of the image that contains the root filesystem for the container virtual machine.
+* **`path`** (string, REQUIRED) path to the container virtual machine root image.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+* **`format`** (string, REQUIRED) format of the container virtual machine root image. Commonly supported formats are:
+    * **`raw`** [raw disk image format][raw-image-format]. Unset values for `format` will default to that format.
+    * **`qcow2`** [QEMU image format][qcow2-image-format].
+    * **`vdi`** [VirtualBox 1.1 compatible image format][vdi-image-format].
+    * **`vmdk`** [VMware compatible image format][vmdk-image-format].
+    * **`vhd`** [Virtual Hard Disk image format][vhd-image-format].
+
+This image contains the root filesystem that the virtual machine **`kernel`** will boot into, not to be confused with the container root filesystem itself. The latter, as specified by **`path`** from the [Root Configuration](config.md#Root-Configuration) section, will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
+
+### Example
+
+```json
+    "image": {
+        "path": "/path/to/vm/rootfs.img",
+	"format": "raw"
+    }
+```
+
+[raw-image-format]: https://en.wikipedia.org/wiki/IMG_(file_format)
+[qcow2-image-format]: https://git.qemu.org/?p=qemu.git;a=blob_plain;f=docs/interop/qcow2.txt;hb=HEAD
+[vdi-image-format]: https://forensicswiki.org/wiki/Virtual_Disk_Image_(VDI)
+[vmdk-image-format]: http://www.vmware.com/app/vmdk/?src=vmdk
+[vhd-image-format]: https://github.com/libyal/libvhdi/blob/master/documentation/Virtual%20Hard%20Disk%20(VHD)%20image%20format.asciidoc
--- a/config.md
+++ b/config.md
@@ -349,6 +349,8 @@ For Windows based systems the user structure has the following fields:
     This MUST be set if the target platform of this spec is `windows`.
 * **`solaris`** (object, OPTIONAL) [Solaris-specific configuration](config-solaris.md).
     This MAY be set if the target platform of this spec is `solaris`.
+* **`vm`** (object, OPTIONAL) [Virtual-machine-specific configuration](config-vm.md).
+    This MAY be set if the target platform and architecture of this spec support hardware virtualization.
 
 ### Example (Linux)
 
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -163,6 +163,9 @@
         },
         "windows": {
             "$ref": "config-windows.json#/windows"
+        },
+        "vm": {
+            "$ref": "config-vm.json#/vm"
         }
     },
     "required": [
--- a/schema/config-vm.json
+++ b/schema/config-vm.json
@@ -1,0 +1,60 @@
+{
+    "vm": {
+        "description": "configuration for virtual-machine-based containers",
+        "type": "object",
+        "required": [
+            "kernel"
+        ],
+        "properties": {
+            "hypervisor": {
+                "description": "hypervisor config used by VM-based containers",
+                "type": "object",
+                "required": [
+                    "path"
+                ],
+                "properties": {
+                    "path": {
+                        "$ref": "defs.json#/definitions/FilePath"
+                    },
+                    "parameters": {
+                        "$ref": "defs.json#/definitions/ArrayOfStrings"
+                    }
+                }
+            },
+            "kernel": {
+                "description": "kernel config used by VM-based containers",
+                "type": "object",
+                "required": [
+                    "path"
+                ],
+                "properties": {
+                    "path": {
+                        "$ref": "defs.json#/definitions/FilePath"
+                    },
+                    "parameters": {
+                        "$ref": "defs.json#/definitions/ArrayOfStrings"
+                    },
+                    "initrd": {
+                        "$ref": "defs.json#/definitions/FilePath"
+                    }
+                }
+            },
+            "image": {
+                "description": "root image config used by VM-based containers",
+                "type": "object",
+                "required": [
+                    "path",
+                    "format"
+                ],
+                "properties": {
+                    "path": {
+                        "$ref": "defs.json#/definitions/FilePath"
+                    },
+                    "format": {
+                        "$ref": "defs-vm.json#/definitions/RootImageFormat"
+                    }
+                }
+            }
+        }
+    }
+}
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -25,6 +25,8 @@ type Spec struct {
 	Solaris *Solaris `json:"solaris,omitempty" platform:"solaris"`
 	// Windows is platform-specific configuration for Windows based containers.
 	Windows *Windows `json:"windows,omitempty" platform:"windows"`
+	// VM specifies configuration for virtual-machine-based containers.
+	VM *VM `json:"vm,omitempty" platform:"vm"`
 }
 
 // Process contains information to start a specific application inside the container.
@@ -497,6 +499,42 @@ type WindowsNetwork struct {
 type WindowsHyperV struct {
 	// UtilityVMPath is an optional path to the image used for the Utility VM.
 	UtilityVMPath string `json:"utilityVMPath,omitempty"`
+}
+
+// VM contains information for virtual-machine-based containers.
+type VM struct {
+	// Hypervisor specifies hypervisor-related configuration for virtual-machine-based containers.
+	Hypervisor VMHypervisor `json:"hypervisor,omitempty"`
+	// Kernel specifies kernel-related configuration for virtual-machine-based containers.
+	Kernel VMKernel `json:"kernel"`
+	// Image specifies guest image related configuration for virtual-machine-based containers.
+	Image VMImage `json:"image,omitempty"`
+}
+
+// VMHypervisor contains information about the hypervisor to use for a virtual machine.
+type VMHypervisor struct {
+	// Path is the host path to the hypervisor used to manage the virtual machine.
+	Path string `json:"path"`
+	// Parameters specifies parameters to pass to the hypervisor.
+	Parameters string `json:"parameters,omitempty"`
+}
+
+// VMKernel contains information about the kernel to use for a virtual machine.
+type VMKernel struct {
+	// Path is the host path to the kernel used to boot the virtual machine.
+	Path string `json:"path"`
+	// Parameters specifies parameters to pass to the kernel.
+	Parameters string `json:"parameters,omitempty"`
+	// InitRD is the host path to an initial ramdisk to be used by the kernel.
+	InitRD string `json:"initrd,omitempty"`
+}
+
+// VMImage contains information about the virtual machine root image.
+type VMImage struct {
+	// Path is the host path to the root image that the VM kernel would boot into.
+	Path string `json:"path"`
+	// Format is the root image format type (e.g. "qcow2", "raw", "vhd", etc).
+	Format string `json:"format"`
 }
 
 // LinuxSeccomp represents syscall restrictions
--- a/schema/defs-vm.json
+++ b/schema/defs-vm.json
@@ -1,0 +1,14 @@
+{
+    "definitions": {
+        "RootImageFormat": {
+            "type": "string",
+            "enum": [
+                "raw",
+                "qcow2",
+                "vdi",
+                "vmdk",
+                "vhd"
+            ]
+        }
+    }
+}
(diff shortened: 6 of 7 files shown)


sameo:
cc @vbatts @mrunalp @runcom @stevvooe

sameo:
@gnawux @bergwolf @laijs

config-vm.md
```diff
@@ -0,0 +1,49 @@
...
+* **`path`** (string, required) specifies the host path to the kernel used to boot the container virtual machine.
+* **`parameters`** (string, optional) specifies a space-separated list of parameters to pass to the kernel.
+* **`initrd`** (string, optional) specifies the host path to an initial ramdisk to be used by the container virtual machine.
+
+## <a name="HypervisorObject" /> Hypervisor Object
+
+Used by virtual-machine-based runtimes only.
+
+* **`path`** (string, required) specifies the host path to the hypervisor used to manage the container virtual machine.
+* **`parameters`** (string, optional) specifies a space-separated list of parameters to pass to the hypervisor.
```
wking: If this is going to be an array, can it be a JSON array (“array of strings”) instead of a single JSON string with internal space delimiters?

config-vm.md
```diff
@@ -0,0 +1,49 @@
+# <a name="VirtualMachineSpecificContainerConfiguration" /> Virtual-machine-specific Container Configuration
+
+Virtual-machine-based containers require additional configuration to that specified in the [default spec configuration](config.md).
```
wking: `config.md` isn't the “default” configuration, it's the (mostly) platform-agnostic configuration.  It's not entirely platform-agnostic (e.g. `hooks` is POSIX-only), but I think we can do better than “default”.  Maybe “base”?  “main”?

config-vm.md
```diff
@@ -0,0 +1,49 @@
...
+
+This **optional** configuration is specified in a "VM" object:
+
+* **`imagePath`** (string, required) host path to file that represents the root filesystem for the container virtual machine.
```
wking: “required” → “REQUIRED” (and similarly below, also “optional” → “OPTIONAL”.  The casing needs to match the [declarations here][1].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/spec.md#notational-conventions

config-vm.md
sameo: Done.

config-vm.md
sameo: Done.

config-vm.md
sameo: Given that this is supposed to be a parameter list that could be passed as is to the command, and would obviously be quite specific to that binary, what would be the advantage of using a JSON array?

wking: > Given that this is supposed to be a parameter list that could be passed as is to the command, and would obviously be quite specific to that binary...

So is [`execv`'s `argv`][1], and we use an array there.

> ... what would be the advantage of using a JSON array?

No need to quote/escape spaces within arguments and a simpler structure for removing/replacing arguments while composing the array.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html

config-vm.md
```diff
@@ -0,0 +1,49 @@
+# <a name="VirtualMachineSpecificContainerConfiguration" /> Virtual-machine-specific Container Configuration
+
+Virtual-machine-based containers require additional configuration to that specified in the [base spec configuration](config.md).
+
+This **optional** configuration is specified in a "VM" object:
+
+* **`imagePath`** (string, REQUIRED) host path to file that represents the root filesystem for the container virtual machine.
+* **`kernel`** (object, REQUIRED) specifies details of the kernel to boot the container virtual machine with.
+* **`hypervisor`** (object, REQUIRED) specifies details of the hypervisor that manages the container virtual machine.
```
wking: Do all VM-based runtimes support multiple hypervisors?  I'd expect this to be OPTIONAL, with an implementation-defined default.

config-vm.md
```diff
@@ -0,0 +1,49 @@
+# <a name="VirtualMachineSpecificContainerConfiguration" /> Virtual-machine-specific Container Configuration
+
+Virtual-machine-based containers require additional configuration to that specified in the [base spec configuration](config.md).
+
+This **optional** configuration is specified in a "VM" object:
+
+* **`imagePath`** (string, REQUIRED) host path to file that represents the root filesystem for the container virtual machine.
```
laijs: It (**`imagePath`**) should be optional.

**`imagePath`** and **`initrd`** are both booting media. We need to mark them both optional and document that at least one of them should be specified.

Moreover, since **`imagePath`** and **`initrd`** are both booting media, they are recommended to be put together. So we can move **`imagePath`** to the **`kernel`** section around the **`initrd`** field. After we do it, we may also need to rename **`kernel`** to **`boot`** and rename `kernel.path` to `boot.kernel`. Another reason why **`boot`** is better than **`kernel`** here is that **`bios`** or **`cbfs`** can be added to the **`boot`** section if any team strongly require them.

Finally, I like the name **`imagePath`**, but I didn't expect that it has ambiguous with the **`path`** field in the [Root Configuration](config.md#Root-Configuration) section. So I suggest to rename it to **`initImage`**, since it and **`initrd`** are almost the same functionality.

sameo:
@wking @laijs Updated with changes for your suggestions/comments.

laijs:
LGTM

config-vm.md
```diff
@@ -0,0 +1,51 @@
...
+* **`initrd`** (string, OPTIONAL) specifies the host path to an initial ramdisk to be used by the container virtual machine.
+
+## <a name="ImageObject" /> Image Object
+
+Used by virtual-machine-based runtimes only.
+
+* **`path`** (string, REQUIRED) specifies the absolute host path to the container virtual machine root image. This image contains the root filesystem that the virtual machine **`kernel`** will boot into, not to be confused with the container root filesystem itself. The latter, as specified by **`path`** from the [Root Configuration](config.md#Root-Configuration) section, will be mounted inside the virtual machine at a location chosen by the virtual-machine-based runtime.
+
+
+## <a name="FullyPopulatedVMExample" /> Example of a fully-populated `VM` object
```
wking: There's no precedent for “fully-populated”.  I'd rather have per-`vm`-child-property examples like the other platform-specific sections.  I have no preference between rooted examples (like [the Windows section uses][1]) or relative example (like the [Solaris][2] and [Linux][3] sections use).  A fully-populated example could go [in here][4].  And if you decide to keep a full `vm` example here, I'd just title it “Example” and rely on [section nesting][5] to define the example scope.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-windows.md#example
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-solaris.md#example
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config-linux.md#example
[4]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#configuration-schema-example
[5]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/style.md#example

config.md
```diff
@@ -462,6 +462,12 @@ Instead they MUST ignore unknown properties.
 Runtimes that are reading or processing this configuration file MUST generate an error when invalid or unsupported values are encountered.
 Unless support for a valid value is explicitly required, runtimes MAY choose which subset of the valid values it will support.
 
+## <a name="VirtualMachine" />VM
```
wking: The VM-specific section isn't quite platform-specific, but I think this needs to go into the platform-specific section because not all runtimes will be able to support these properties, and platforms are linked to the [compliance language][1].  This is one of the reasons [I prefer “protocol” to “platform”][2].

You should also define the **`vm`** property wherever the Virtual Machine entry ends up in `config.md`.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/spec.md#L40-L41
[2]: https://github.com/opencontainers/runtime-spec/pull/570#issue-85434395

sameo:
@wking I think I addressed all your comments, thanks a lot for the thorough review.

schema/config-schema.json
```diff
@@ -199,6 +199,9 @@
         },
         "windows": {
             "$ref": "config-windows.json#/windows"
+        },
+	"vm": {
```
wking: This looks like a non-standard indent.  You can fix it with:

```
schema $ make fmt
```

config.md
```diff
@@ -349,6 +349,8 @@ For Windows based systems the user structure has the following fields:
     This MUST be set if the target platform of this spec is `windows`.
 * **`solaris`** (object, OPTIONAL) [Solaris-specific configuration](config-solaris.md).
     This MAY be set if the target platform of this spec is `solaris`.
+* **`vm`** (object, OPTIONAL) [Virtual-machine-specific configuration](config-vm.md).
```
wking: You'll also want to add virtual-machine entries [here][1] and [here][2].

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/spec.md#platforms
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/spec.md#table-of-contents

specs-go/config.go
```diff
@@ -487,6 +489,40 @@ type WindowsHyperV struct {
 	UtilityVMPath string `json:"utilityVMPath,omitempty"`
 }
 
+// VM contains information for virtual-machine-based containers.
+type VM struct {
+	// Hypervisor specifies hypervisor-related configuration for virtual-machine-based containers.
+	Hypervisor VMHypervisor `json:"hypervisor"`
```
wking: `hypervisor` is optional, so you'll want:

```go
Hypervisor *VMHypervisor `json:"hypervisor,omitempty"`
```

And similarly for `Image` below.

schema/config-vm.json
```diff
@@ -0,0 +1,66 @@
+{
+    "vm": {
+        "description": "configuration for virtual-machine-based containers",
+        "id": "https://opencontainers.org/schema/bundle/vm",
```
wking: You should drop these `id` entries (#945).

schema/config-vm.json
```diff
@@ -0,0 +1,66 @@
...
+        "type": "object",
+        "required" : [
+            "kernel",
```
wking: This trailing comma [is invalid JSON][1].

[1]: https://travis-ci.org/opencontainers/runtime-spec/jobs/342293097#L584

sameo:
@wking All comments addressed, thanks.

wking:
3d8cce44a looks good to me.  It doesn't add any new runtime requirements though, so there would be no way to check for [runtime compliance][1].  That's true for other portions of the spec as well though (e.g. most of the Linux cgroups config, #746), so it may not be an issue for maintainers.  If it is an issue, we'll want additional language like:

```markdown
Runtimes MUST execute the `hypervisor` with [IEEE Std 1003.1-2008 `execvp`'s *argv*][ieee-1003.1-2008-functions-exec] set to the union of `path` and `parameters`.
```

for `hypervisor`, and similarly for the other properties.

Actually, for `hypervisor`, I'd encourage reuse of the existing [`process`][2] or [`hooks`][3] schema, which would make life easier for implementations (see [here][4] where I called for those to be unified).  If you use the `hooks` schema, you get very similar semantics to your existing `hypervisor` schema, except that you have `args` instead of `parameters`, and you gain `env` and `timeout`.  You also get to punt on runtime requirements as “not a problem for this PR, because we're leaning on current spec language”.  That seems like a win to me.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/spec.md#L41
[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#process
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#posix-platform-hooks
[4]: https://groups.google.com/a/opencontainers.org/d/msg/dev/Cn5L6g1prgA/AaUSpmJBEQAJ

specs-go/config.go
```diff
@@ -487,6 +489,40 @@ type WindowsHyperV struct {
...
+	// Parameters specifies parameters to pass to the kernel.
+	Parameters string `json:"parameters,omitempty"`
+	// InitRD is the host path to an initial ramdisk to be used by the kernel.
+	InitRD string `json:"initrd,omitempty"`
+}
+
+// VMImage contains information about the virtual machine root image.
+type VMImage struct {
+	// Path is the host path to the root image that the VM kernel would boot into.
+	Path string `json:"path"`
```
vbatts: Is this leaving the type of that path completely up to the runtime? (qcow, raw, directory, etc)

config.md
vbatts: done now

mrunalp:
LGTM

mrunalp:
@crosbymichael PTAL

wking:
> Actually, for `hypervisor`, I'd encourage reuse of the existing [`process`][2] or [`hooks`][3] schema…

I've filed a sub-PR for this at sameo/runtime-spec#1, with commits for both the “reference `process`” and “copy/paste `process` approaches.

[2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#process
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#posix-platform-hooks

specs-go/config.go
```diff
@@ -499,6 +501,42 @@ type WindowsHyperV struct {
...
+	// Format is the root image format type (e.g. "qcow2", "raw", "vhd", etc).
+	Formar string `json:"format"`
```
laijs: here is a typo.

config-vm.md
```diff
@@ -0,0 +1,63 @@
...
+    }
+```
+
+## <a name="ImageObject" /> Image Object
+
+**`image`** (object, OPTIONAL) specifies details of the image that contains the root filesystem for the container virtual machine.
+* **`path`** (string, REQUIRED) path to the container virtual machine root image.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+* **`format`** (string, REQUIRED) format of the container virtual machine root image. The following root image formats are supported:
+    * **`raw`** raw disk image format. Unset values for `type` will default to that format.
```
wking: `type` -> `format`.

config-vm.md
```diff
@@ -0,0 +1,63 @@
...
+    * **`qcow2`** QEMU image format.
+    * **`qcow`** Old QEMU image format.
+    * **`vdi`** VirtualBox 1.1 compatible image format.
+    * **`vmdk`** VMware compatible image format.
+    * **`vhd`** Virtual Hard Disk image format.
```
wking: Ideally these would have links to specs/docs, to make sure everyone agreed on what "Virtual Hard Disk", etc. meant.

specs-go/config.go
sameo: Since we're making it a required field for VM images, I don't think `omitempty` is needed there.

sameo:
@laijs @wking PR fixed now.

config-vm.md
```diff
@@ -0,0 +1,70 @@
...
+
+## <a name="ImageObject" /> Image Object
+
+**`image`** (object, OPTIONAL) specifies details of the image that contains the root filesystem for the container virtual machine.
+* **`path`** (string, REQUIRED) path to the container virtual machine root image.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+* **`format`** (string, REQUIRED) format of the container virtual machine root image. The following root image formats are supported:
+    * **`raw`** [raw disk image format][raw-image-format]. Unset values for `format` will default to that format.
+    * **`qcow2`** [QEMU image format][qcow2-image-format].
+    * **`qcow`** [Old QEMU image format][qcow-image-format].
```
wking: If this format is deprecated in QEMU, do we want to include it here?  Maybe leave it off until someone asks for it specifically?

sameo: Fair enough, `qcow` removed.

config-vm.md
```diff
@@ -0,0 +1,68 @@
...
+        "initrd": "/path/to/initrd.img"
+    }
+```
+
+## <a name="ImageObject" /> Image Object
+
+**`image`** (object, OPTIONAL) specifies details of the image that contains the root filesystem for the container virtual machine.
+* **`path`** (string, REQUIRED) path to the container virtual machine root image.
+    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
+* **`format`** (string, REQUIRED) format of the container virtual machine root image. The following root image formats are supported:
```
crosbymichael: Supported by what? Who? Support for formats seem like an implementation detail to me for various runtimes, not something that should be dictated by the spec.

sameo: @crosbymichael Supported by the current major hypervisors. So you'd prefer to see a free form string that would only be a hint to the hypervisor?

crosbymichael: I guess the only worry I have is that people would read the spec and the certification committee will say that these are MUSTs that all runtimes have to support if we have it in the spec.

crosbymichael: Could we change the wording to something like "commonly supported formats"?

sameo: That would be better indeed, thanks. Let me change that.

crosbymichael: Cool, LGTM after this change

crosbymichael:
LGTM

crosbymichael:
Thanks!

wking:
Now that two separate maintainers have (at various points) LGTMed this PR, I'm getting jumpy about it landing without sameo/runtime-spec#1 getting a look over.  I've just rebased that on top of is PR's current 74b670e, but I'd really like to avoid a third process-launching schema in `hypervisor`, so if there's anything I can do to make the sub-PR more appealing, let me know.

sameo:
@wking **If** maintainers want to re-use existing process schema, I'd much prefer to go with wking/opencontainer-runtime-spec@edd501a . The `Process` description is fairly container specific but we've contemplated containerizing the hypervisor in the past anyway. The only concern I have about it is `User` being a mandatory field. In most cases the hypervisor/VMM will be called from a runtime implementation (kata-runtime binary or a CRI implementation) and will likely want to run under the same user as the caller, under the same user namespace. Forcing caller to specify this is a little redundant and cumbersome, or is there a way to specify that implicitly?

wking:
> If maintainers want to re-use existing process schema, I'd much prefer to go with wking/opencontainer-runtime-spec@edd501a .

I agree.  The commit after that was because @crosbymichael specifically requested a copy/paste implementation and not edd501a's reference.  I'm hoping I misunderstood, or that he changes his mind after seeing both approaches.

> The only concern I have about it is `User` being a mandatory field.

I'm in favor of making it optional, with unset `user` properties meaning "don't change this".  Previous discussion in #417, #489, #618.  It's been required since 8af3eb64b

sameo:
> A possible compromise would be removing it from hypervisor

You mean by specifying that it `MUST NOT be configured` ?

wking:
> > A possible compromise would be removing it from `hypervisor`
>
> You mean by specifying that it `MUST NOT be configured`?

Yeah, at least as long as it's required for `process`.

crosbymichael:
ping @mrunalp can you review again?

sameo:
@crosbymichael @mrunalp @stevvooe Could you also please let us know if you want to use the process schema @wking is proposing or not?

crosbymichael:
@sameo I think it's fine as is.  The majority of the fields on Process do not apply here.

wking:
> I think it's fine as is. The majority of the fields on Process do not apply here.

So MUSTmore of them to be unset?  Or recycle the hook structure instead?  I'd still like to have this process-launching structure avoid being a third, incompatible way to set `argv`.

crosbymichael:

specs-go/config.go
```diff
@@ -499,6 +501,42 @@ type WindowsHyperV struct {
 	UtilityVMPath string `json:"utilityVMPath,omitempty"`
 }
 
+// VM contains information for virtual-machine-based containers.
+type VM struct {
+	// Hypervisor specifies hypervisor-related configuration for virtual-machine-based containers.
+	Hypervisor VMHypervisor `json:"hypervisor,omitempty"`
```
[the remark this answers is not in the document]
crosbymichael: Ahh, good point.  Shouldn't the hypervisor be exec'd with the correct path and args anyways from a runtime?

hqhq:
LGTM

specs-go/config.go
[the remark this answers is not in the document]
wking: This has been spun off into #964.

PR #951, opened 2018-02-20
glossary: Bump JSON spec to RFC 8259

wking:
The new RFC was released in 2017-12 and obsoletes 7159.  No real changes for us, although it does [fix some minor errata][1].

[1]: https://tools.ietf.org/html/rfc8259#appendix-A


vbatts:
LGTM

PR #954, opened 2018-03-06
config: Drop 'type' from bind-mount example

wking:
This example initially came in with b3918a25, but the `bind` value is not one of the types you can get out of `/proc/filesystems`.  Instead, bind mounts should leave the type unset and put either `bind` or `rbind` in options (although neither of those are documented either, #771).  Since documenting `(r)bind` seems to be too difficult (for reasons [I don't understand][1]), we should at least stop setting type in the example to stop [confusing][2] [users][3].

Runc still checks `.Type` instead of `.Options` for bind-ness in [a][4] [few][5] [places][6], but we can address all of those by setting `.Device` to `bind` depending on `.Options` [here][4].

[1]: https://github.com/opencontainers/runtime-spec/pull/771#issuecomment-314193121
[2]: https://github.com/opencontainers/runc/issues/1744
[3]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/2gia6t1Dnv0
[4]: https://github.com/opencontainers/runc/blob/v1.0.0-rc5/libcontainer/specconv/spec_linux.go#L272-L276
[5]: https://github.com/opencontainers/runc/blob/v1.0.0-rc5/libcontainer/rootfs_linux.go#L33
[6]: https://github.com/opencontainers/runc/blob/v1.0.0-rc5/libcontainer/rootfs_linux.go#L234

### DIFF
--- a/config.md
+++ b/config.md
@@ -115,7 +115,6 @@ For POSIX platforms the `mounts` structure has the following fields:
     },
     {
         "destination": "/data",
-        "type": "bind",
         "source": "/volumes/testing",
         "options": ["rbind","rw"]
     }


wking:
> Runc still checks .Type instead of .Options for bind-ness in a few places, but we can address all of those by setting .Device to bind depending on .Options here.

Filed as opencontainers/runc#1753.

config.md
```diff
@@ -115,7 +115,6 @@ For POSIX platforms the `mounts` structure has the following fields:
     },
     {
         "destination": "/data",
-        "type": "bind",
         "source": "/volumes/testing",
         "options": ["rbind","rw"]
```
Nabarun: Does it makes sense to have rbind in the options if type:bind is removed

cyphar: Yes, `rbind`/`bind` are options and are not meant to be the "type". The type is not used when performing bind-mounts.

Nabarun: @cyphar please help me understand this, if I remove rbind from my options it will then not do the binding right?

wking: > ... if I remove rbind from my options it will then not do the binding right?

You need either `bind` or `rbind` in `options`.  You do not need to set `type` for binds per the spec, and if/when opencontainers/runc#1753 lands, you won't need to set `type` for runc binds either.

vbatts:
i reject this. 'bind' is suitable type.

wking:
> i reject this. 'bind' is suitable type.

I'm fine with folks including `bind` as their type if they want, but I don't see a need to when we *also* require them to set `(r)bind` in `options`.  And `mount(2)` works fine without requiring `"bind"` for `filesystemtype`.  Can you give more details about why you'd like to keep it?  Especially since it diverges from the [specified values][1]?  If you want to keep it, would you be comfortable documenting it in the spec?

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.1/config.md#L103

vbatts:
that is just an e.g. not a mandated list ... "bind" is a type of mount

wking:
On Wed, Apr 04, 2018 at 10:13:08PM +0000, Vincent Batts wrote:
> that is just an e.g. not a mandated list ... "bind" is a type of
> mount

Right.  But the mount(2) API shows that you don't have to explicitly
pass it in when you are separately setting MS_BIND or similar.

And if you expect runtimes to *require* type set to ‘bind’ for bind
mounts, I think you really want to land wording to that effect in the
spec.  I certainly don't see it now, and portable syntax for bind
mounts seems like an important part of compliance testing.

wking:
Cross-linking #878 (which I'd forgotten was open).  That PR is in flight with docs that would support (but not require) the example change I'd proposed here.

wking:
Any chance of revisiting this now that opencontainers/runc#1753 has [landed][1]?

[1]: https://github.com/opencontainers/runc/pull/1753#event-1576491460

PR #956, opened 2018-03-08
uidMappings: change order of fields for clarity

alban:
"man 7 user_namespaces" explains the format of uid_map and gid_map:
```
    <containerID> <hostID> <mapSize>
```

The order of map entries in JSON does not matter. But for the clarity of
the spec, I find it easier to understand if the order of the JSON fields is
the same as the order of the fields in the underlying uid_map/gid_map
files.

I am about to file a PR in runtime-tools because the fields in
uid_map/gid_map were parsed in the wrong order.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -82,8 +82,8 @@ If a `namespaces` field contains duplicated namespaces with same `type`, the run
 
 Each entry has the following structure:
 
-* **`hostID`** *(uint32, REQUIRED)* - is the starting uid/gid on the host to be mapped to *containerID*.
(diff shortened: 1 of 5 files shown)


wking:
Maybe swap them [in the Go structure][1] too?  I think (but haven't tested) that Go's default JSON serializer uses the struct's property order for its output.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/specs-go/config.go#L197-L200

tianon:
LGTM

(purely cosmetic change)

mrunalp:
LGTM

Issue #957, opened 2018-03-12
Clarify Linux devices uid/gid ownership wrt uid/gid mapping


alban:
The spec describes [Devices](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#devices) and gives an example:

```
    "devices": [
        {
            "path": "/dev/fuse",
            "type": "c",
            "major": 10,
            "minor": 229,
            "fileMode": 438,
            "uid": 0,
            "gid": 0
        },
```

Are the uid & gid relative to the host, or to the container after conversion with the userns uid/gid mappings?

The spec should clarify that.

cyphar:
It would be relative to the container. All `uid`s and `gid`s specified should be considered to be from the container's point of view (aside from the mapping entries themselves -- though a container will see the same values "host-side" if it looks at `/proc/self/*id_map`).

Maybe there should be some wording making it clear that the mappings apply to everything.

wking:
On Mon, Mar 12, 2018 at 02:30:33PM +0000, Aleksa Sarai wrote:
> Maybe there should be some wording making it clear that the mappings
> apply to everything.

I recommend something like:

  … id of device owner in the [container user namespace](glossary.md#container-namespace)

to follow the approach we already use for paths:

  $ git describe
  v1.0.1-16-g6be516e
  $ git grep 'namespace](glossary'
  config-linux.md:    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
  config-linux.md:    If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
  config-linux.md:If a namespace type is not specified in the `namespaces` array, the container MUST inherit the [runtime namespace](glossary.md#runtime-namespace) of that type.
  config-linux.md:    * **`name`** *(string, REQUIRED)* - interface name in [runtime network namespace](glossary.md#runtime-namespace)
  config-linux.md:    If no mounted `resctrl` pseudo-filesystem is available in the [runtime mount namespace](glossary.md#runtime-namespace), the runtime MUST [generate an error](runtime.md#errors).
  config-linux.md:    The values MUST be absolute paths in the [container namespace](glossary.md#container_namespace).
  config-linux.md:    The values MUST be absolute paths in the [container namespace](glossary.md#container-namespace).
  config.md:* **`uid`** (int, REQUIRED) specifies the user ID in the [container namespace](glossary.md#container-namespace).
[...]

PR #959, opened 2018-03-13
Linux devices: uid/gid relative to container

alban:
It was not clear in the spec if the uid and gid owning a Linux device
were relative to the runtime namespace or to the container namespace.
This patch clarifies that.

Fixes https://github.com/opencontainers/runtime-spec/issues/957


vbatts:
LGTM

PR #962, opened 2018-04-02
config-vm: Recycle the hook-process schema

wking:
We've had two ways to specify a process to launch (for the container process and for hooks) [for a while now][0].  Last week, we landed #949 with a third process-launching structure.  This pull request replaced that third structure with a recycled hook-process schema for launcing the hypervisor.  I prefer using the more-fundamental container process schema, but @crosbymichael felt [it had too many properties that wouldn't apply to the hypervisor-launching case][1].  Versus master, this pull request adds:

* The ability to set `argv[0]` independently of `path` (like [`execv`][2] and similar).
* The ability to set `env` and `timeout` for the hypervisor process.
* The ability to lean on existing hook-launching code for launching hypervisor processes, with the difference being just whether you [pass the state in on stdin][3] or not.

And removes:

* 18 lines of source (says `git diff --diffstat origin/master`).
* A third take at [`argv` phrasing][4] which is much different from our [existing][5] [phrasing][6].

If we get this or a similar unification landed before we cut a release that includes #949, then I think this change is all positive with no downside.  If we punt on this until after 1.1.0, then this is probably just going to end up as one of those unfortunate historical quirks that are not worth the migration pain of fixing.

Ping @sameo and @crosbymichael, since they're the only two who have [expressed][7] [opinions][1] on this sort of thing so far.

[0]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/Cn5L6g1prgA
[1]: https://github.com/opencontainers/runtime-spec/pull/949#issuecomment-373730290
[2]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html
[3]: https://github.com/opencontainers/runtime-spec/blame/d362ed3abe646248f1f84a6e5871d05513f9167b/config.md#L389
[4]: https://github.com/opencontainers/runtime-spec/blame/d362ed3abe646248f1f84a6e5871d05513f9167b/config-vm.md#L11
[5]: https://github.com/opencontainers/runtime-spec/blame/d362ed3abe646248f1f84a6e5871d05513f9167b/config.md#L157-L158
[6]: https://github.com/opencontainers/runtime-spec/blame/d362ed3abe646248f1f84a6e5871d05513f9167b/config.md#L378
[7]: https://github.com/opencontainers/runtime-spec/pull/949#issuecomment-372299408

### DIFF
--- a/config-vm.md
+++ b/config-vm.md
@@ -5,17 +5,14 @@ The virtual-machine container specification provides additional configuration fo
 
 ## <a name="HypervisorObject" /> Hypervisor Object
 
-**`hypervisor`** (object, OPTIONAL) specifies details of the hypervisor that manages the container virtual machine.
-* **`path`** (string, REQUIRED) path to the hypervisor binary that manages the container virtual machine.
-    This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
-* **`parameters`** (array of strings, OPTIONAL) specifies an array of parameters to pass to the hypervisor.
+**`hypervisor`** (object, OPTIONAL) configures the hypervisor process using the [hook-process schema](config.md#hook-process).
 
 ### Example
 
 ```json
     "hypervisor": {
         "path": "/path/to/vmm",
-        "parameters": ["opts1=foo", "opts2=bar"]
+        "args": ["vmm", "opts1=foo", "opts2=bar"]
     }
 ```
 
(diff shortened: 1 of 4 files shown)


vbatts:
_house keeping:_ i'm inclined to close this for now. While I am all for re-using schema, this slight change imposes a change and hitching VM intricacies to the other hooks (which sounds fine, but may not align to ongoing reality)

vbatts:
(decided to close, since the neither of the two reviewers have chimed in on this in 1.5yrs...)

PR #968, opened 2018-05-09
.travis: Bump minimum Go version to 1.9

wking:
[Go 1.10 was released on 2018-02-16][1] and [Go only supports the last two major releases][2] (so currently 1.9 and 1.10).  Besides ensuring that we are compatible with the currently-maintained releases, this will avoid our current test failure now that [golint has dropped support for Go 1.6][3]:

```console
$ make install.tools
go get -u github.com/golang/lint/golint
../../../golang.org/x/tools/go/internal/gcimporter/iimport.go:77: undefined: io.SeekCurrent
../../../golang.org/x/tools/go/internal/gcimporter/iimport.go:80: undefined: io.SeekCurrent
../../../golang.org/x/tools/go/internal/gcimporter/iimport.go:156: undefined: io.SeekCurrent
../../../golang.org/x/tools/go/internal/gcimporter/iimport.go:187: r.declReader.Reset undefined (type bytes.Reader has no field or method Reset)
../../../golang.org/x/tools/go/internal/gcimporter/iimport.go:226: r.declReader.Reset undefined (type bytes.Reader has no field or method Reset)
make: *** [.install.golint] Error 2
```

[1]: https://golang.org/doc/devel/release.html#go1.10
[2]: https://golang.org/doc/devel/release.html#policy
[3]: https://travis-ci.org/opencontainers/runtime-spec/jobs/376692151#L469

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: go
 go:
-  - 1.7
-  - 1.6.3
-  - 1.5.4
+  - "1.10.x"
+  - "1.9.x"
 
 sudo: required
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ services:
 before_install:
   - make install.tools
   - docker pull vbatts/pandoc
+  - go get -d ./schema/...
 
 install: true
 
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ test: .govet .golint .gitvalidation
 
 # `go get github.com/golang/lint/golint`
 .golint:
-ifeq ($(call ALLOWED_GO_VERSION,1.6,$(HOST_GOLANG_VERSION)),true)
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,7 @@
 language: go
 go:
-  - 1.7
-  - 1.6.3
-  - 1.5.4
+  - 1.10
```
tianon: This gets interpreted as 1.1 (and thus fails) -- you probably want to use `1.10.x` anyhow (to get the latest minor release in the 1.10 series).

Same with `1.9.x`.

See also https://docs.travis-ci.com/user/languages/go/#Specifying-a-Go-version-to-use.

wking: > This gets interpreted as 1.1 (and thus fails)...

Thanks.  Fixed with 99c66a0 -> 692abcb.

tianon:
LGTM

crosbymichael:
LGTM

Issue #592, opened 2016-10-20
missing NIS domainname


justincormack:
We have had various complaints in Docker eg see https://github.com/docker/docker/issues/27067 that you cannot set the `domainname`, just the `hostname`.  As `runc` has no way to set the domain name, we cannot easily implement this. Can we add a new optional field for domain name? It would exactly match hostname, except on Linux it would use the `setdomainname(2)` syscall not `sethostname(2)`.

This is used by some applications eg NIS (where it originated as a field I believe).

Have not yet determined if Windows has a comparable field, it seems likely. Other Unixes certainly do.

Happy to add a patch (and in `runc`) if this is acceptable.

cyphar:
Heh, you beat me to this. :wink:

I reckon the config should look like:

```
{ "nis": { "hostname": "a", "domainname": "b.com" } }
```

Where you can replace "nis" with "uname" (or "uts") if more appropriate.

justincormack:
Actually I think this may be Linux specific, looking at FreeBSD https://www.freebsd.org/cgi/man.cgi?query=domainname&sektion=1&apropos=0&manpath=FreeBSD+10.3-RELEASE+and+Ports it is set in `/etc/rc.conf` not kernel mediated and read via `sysctl` so there would be nothing for `runc` to usefully do here.

cyphar:
We could shove it in config-linux then under `nisdomainname`.

justincormack:
yes I think that may be the best option.

justincormack:
It looks to me like Solaris is like FreeBSD, although I could not find a definitive answer. BSD says their version came from SunOS. It is not in Posix as part of `uname`.

cyphar:
Thinking about this more I think it's possible to just use `kernel.domainname` to set the domain name without needing any OCI support (especially because the idea would be to just make it Linux-specific anyway).

cyphar:
I've figured out this can be entirely done with opencontainers/runc#1827 and moby/moby#37302.

AkihiroSuda:
Completed in:
- https://github.com/opencontainers/runtime-spec/pull/1156

PR #974, opened 2018-06-14
spec: add field for NIS domain name

cyphar:
A common complaint is that there is no way to explicitly set the NIS
domain name of a container (through setdomainname(2) and
kernel.domainname). Some legacy software uses the domain name of the
host for license verification and other similar functions, so having
this is a very useful feature for people who use such software.

Part of moby/moby#27067.

### DIFF
--- a/config.md
+++ b/config.md
@@ -341,6 +341,19 @@ For Windows based systems the user structure has the following fields:
 "hostname": "mrsdalloway"
 ```
 
+## <a name="configDomainName" />Domain Name
+
+* **`domainname`** (string, OPTIONAL) specifies the container's NIS domain name as seen by processes running inside the container.
+    On Linux, for example, this will [change the domain name][setdomainname.2] in the [container](glossary.md#container-namespace) [UTS namespace][uts-namespace.7].
+    Depending on your [namespace configuration](config-linux.md#namespaces), the container UTS namespace may be the [runtime](glossary.md#runtime-namespace) [UTS namespace][uts-namespace.7].
+
+### Example
+
+```json
+"hostname": "mrsdalloway",
+"domainname": "opencontainers.org"
+```
+
 ## <a name="configPlatformSpecificConfiguration" />Platform-specific configuration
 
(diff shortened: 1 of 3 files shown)


wking:
Do we know how non-Linux platforms could support this?  For example, see discussion in #592 about this being set only via `/etc/rc.conf` (in userspace, not the kernel) for [FreeBSD][1] and potentially [Solaris][2].  Nobody has weighed in on whether this is a kernel or userspace property on Windows.  Having runtimes configure userspace properties is not a problem in itself, but it's not clear to me what would happen if, say, Solaris set this only in `/etc/rc.conf`, and a user called a Solaris runtime with `domainname` set *and* and `/etc/rc.conf` in their [`root.path`][3].  Should the runtime clobber the supplied `rc.conf`?  Should the runtime ignore the configured `domainname`?  Should the runtime parse the supplied `rc.conf` and error if and only if the domain name configured there differed from the `domainname` in the OCI configuration?

If we don't have time to get to the bottom of this on other platforms (at least Windows and Solaris), it may make sense to make this part of the Linux-specific config (which [you'd floated][4] back in 2016 ;).

[1]: https://github.com/opencontainers/runtime-spec/issues/592#issuecomment-255098941
[2]: https://github.com/opencontainers/runtime-spec/issues/592#issuecomment-255128118
[3]: https://github.com/opencontainers/runtime-spec/blob/v1.0.1/config.md#root
[4]: https://github.com/opencontainers/runtime-spec/issues/592#issuecomment-255099751

cyphar:
> (which you'd floated back in 2016 ;)

Oh god. I guess I really am getting old. :wink:

But (agreeing with myself) I think that putting it under the Linux section makes more sense given the weirdness with Solaris/FreeBSD that I forgot about.

cyphar:
Actually I just noticed that we could actually already implement this through the `kernel.domainname` sysctl. So maybe it's not necessary to have a separate field for this. I'll investigate further.

cyphar:
I've figured out this can be entirely done with opencontainers/runc#1827 and moby/moby#37302.

PR #976, opened 2018-06-20
config: Add Windows Devices to Schema

cwilhit:
Signed-off-by: Craig Wilhite <crwilhit@microsoft.com>

This adds Windows Device support to the OCI runtime-spec: code, JSON schema and markdown documentation. This is a first step in our approach to lighting up access to host devices for Windows Server containers.

Devices in Windows implement any number of interface class GUIDs. By configuring a container with a device interface class GUID, we will plug all devices on the host that implement the interface class GUID
into the device namespace of the container.

While this approach does not allow for the speciication of a single, particular device, our schema definition is defined to be flexible; we interpret the 'id' of the device passed in based upon the 'id_type'.

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -26,7 +26,7 @@ The Windows container specification uses APIs provided by the Windows Host Compu
 Each entry has the following structure:
 
 * **`id`** *(string, REQUIRED)* - specifies the device which the runtime MUST make available in the container.
-* **`id_type`** *(string, REQUIRED)* - tells the runtime how to interpret `id`. Today, Windows only supports a value of `class`, which identifies `id` as a [device interface class GUID][interfaceGUID].
+* **`idType`** *(string, REQUIRED)* - tells the runtime how to interpret `id`. Today, Windows only supports a value of `class`, which identifies `id` as a [device interface class GUID][interfaceGUID].
 
 [interfaceGUID]: https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-interface-classes
 
@@ -37,11 +37,11 @@ Each entry has the following structure:
         "devices": [
             {
                 "id": "24E552D7-6523-47F7-A647-D3465BF1F5CA",
-                "id_type": "class"
+                "idType": "class"
             },
             {
                 "id": "5175d334-c371-4806-b3ba-71fd53c9258d",
-                "id_type": "class"
+                "idType": "class"
             }
         ]
     }
--- a/schema/defs-windows.json
+++ b/schema/defs-windows.json
@@ -6,7 +6,7 @@
                 "id": {
                     "type": "string"
                 },
-                "id_type": {
+                "idType": {
                     "type": "string",
                     "enum": [
                         "class"
@@ -15,7 +15,7 @@
             },
             "required": [
                 "id",
-                "id_type"
+                "idType"
             ]
         }
     }
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -454,7 +454,7 @@ type WindowsDevice struct {
 	// Device identifier: interface class GUID, etc.
 	ID string `json:"id"`
 	// Device identifier type: "class", etc.
-	IDType string `json:"id_type"`
+	IDType string `json:"idType"`
 }
 
 // WindowsResources has container runtime resource constraints for containers running on Windows.


config-windows.md
```diff
@@ -19,6 +19,32 @@ The Windows container specification uses APIs provided by the Windows Host Compu
     }
 ```
 
+## <a name="configWindowsDevices" />Devices
+
+**`devices`** (array of objects, OPTIONAL) lists devices that MUST be available in the container.
+
+
+* **`id`** *(string, REQUIRED)* - specifies the device for which the runtime must make available in the container.
```
wking: "for which" -> "which"

config-windows.md
```diff
@@ -19,6 +19,32 @@ The Windows container specification uses APIs provided by the Windows Host Compu
...
+* **`id_type`** *(string, REQUIRED)* - tells the runtime how to interpret `id`. Today, Windows only supports a value of `class`, which identifies `id` as a device class GUID.
```
wking: Is there a doc link with more background on device class GUIDs?  Maybe [this][1]?  It would be nice to give readers a way to unpack that jargon.

[1]: https://docs.microsoft.com/en-us/windows-hardware/drivers/install/system-defined-device-setup-classes-available-to-vendors

config-windows.md
```diff
@@ -19,6 +19,32 @@ The Windows container specification uses APIs provided by the Windows Host Compu
...
+            "id_type": "class"
+        },
+        {
+            "id": "5175d334-c371-4806-b3ba-71fd53c9258d",
+            "id_type": "class"
+        }
+    ]
+```
+
+All devices on the host that ascribe to the declared device class GUID interface will be made available in the container.
```
wking: This line seems redundant with the earlier "MUST be available in the container".  Can we drop this line?

schema/defs-windows.json
```diff
@@ -0,0 +1,11 @@
+{
+    "definitions": {
+        "Device": {
+            "type": "object",
+            "required": [
+                "id",
+                "id_type"
```
wking: I think you should define these as strings as well as marking them required.

config-windows.md
cwilhit: Yes, that sounds reasonable.

config-windows.md
```diff
@@ -19,6 +19,32 @@ The Windows container specification uses APIs provided by the Windows Host Compu
...
+
+* **`id`** *(string, REQUIRED)* - specifies the device which the runtime MUST make available in the container.
+* **`id_type`** *(string, REQUIRED)* - tells the runtime how to interpret `id`. Today, Windows only supports a value of `class`, which identifies `id` as a [device interface class GUID][interfaceGUID].
+
+[interfaceGUID]: https://docs.microsoft.com/en-us/windows-hardware/drivers/install/overview-of-device-interface-classes
+
+### Example
+
+```json
+    "devices": [
```
wking: nit: The other examples in `config-windows.md` include the `"windows"` object to which this is an entry.  What you have here, on the other hand, is closer to the example approach taken in `config-linux.md`, where you only show the key being defined in this subsection.  I don't know if the maintainers have a preference, but personally I think consistency within this file is more important.  And ideally we'd have consistency across the whole spec, so we wouldn't have to make these decisions for each new example ;).

cwilhit:
@wking  - fixed from suggestions. 
ping @jhowardmsft if you'd like to give it another look over.

cwilhit:
Bumping this thread to see status on getting this PR pulled in.

config-windows.md
```diff
@@ -19,6 +19,34 @@ The Windows container specification uses APIs provided by the Windows Host Compu
...
 ```
 
+## <a name="configWindowsDevices" />Devices
+
+**`devices`** (array of objects, OPTIONAL) lists devices that MUST be available in the container.
+
+
+* **`id`** *(string, REQUIRED)* - specifies the device which the runtime MUST make available in the container.
+* **`id_type`** *(string, REQUIRED)* - tells the runtime how to interpret `id`. Today, Windows only supports a value of `class`, which identifies `id` as a [device interface class GUID][interfaceGUID].
```
crosbymichael: Are there any other types than `class`? 

Also this should be `idType` not `id_type`

vbatts:
after @crosbymichael feedback, LGTM

config-windows.md
tianon: I also wonder if this is really flexible enough for future needs -- will all future resources that might count as devices have some kind of ID? (I guess it's Windows we're talking about, so everything has a GUID of some kind?)

Being able to attach all sound devices to a container is neat, but if you want something like raw hard drive or USB drive access you'd have to attach _all_ drives, which isn't great, right? (I'm guessing that's why `idType` is included here -- so `id` could instead point to the GUID of an individual device at some point in the future?)

cwilhit: @crosbymichael Thanks for catching the typo, let me commit a fix. In the future, we'll certainly have more types. 

That's correct, @tianon. Each device in Windows is guaranteed to have a unique device instance path ID. In the future it would be easy to extend so that id=<some dev instance path ID> and idType="instance" or some other label. Given this, I think what we have is extensible to give finer granularity in defining a device for a Windows container in the future.

crosbymichael:
LGTM

PR #977, opened 2018-06-28
meeting: Bump July meeting from the 4th to the 11th

wking:
Catching up with:

On Thu, Jun 28, 2018 at 11:37:02AM -0500, @caniszczyk [wrote][1]:
> Due to holidays in the US, we will be moving the OCI dev meeting ahead a week.

I've used [`RDATE`][2] to add the new date and [`EXDATE`][3] to remove the old date.  The ICS was validated [here][4].

The change is very similar to the one that landed earlier in #943 for the January meeting.

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/FzclHQsMH_c/QKJMybRkBAAJ
[2]: https://tools.ietf.org/html/rfc5545#section-3.8.5.2
[3]: https://tools.ietf.org/html/rfc5545#section-3.8.5.1


wking:
The currently-live calendar still claims the meeting is tomorrow.  Any chance we can get this PR landed before then so folks importing this calendar can see the bumped day?

mrunalp:
LGTM

vbatts:
LGTM

PR #982, opened 2018-08-02
maskedPaths and readonlyPaths: ignore unexistent paths

alban:
runc ignores unexistent paths in maskedPaths and readonlyPaths. That's useful for blocking /proc/latency_stats (default in buildah) because this path is not existing on all kernels.

In this case, no error should be generated. Other errors should be generated. For example, using readonlyPaths on a unbindable path fails and this error must be generated, otherwise the path would silently stay read-write.


cyphar:
While I understand why `readonlyPaths` might be okay to ignore (though I'd be worried about the security implications -- really I think an empty file or empty directory should be made and then re-binded as `ro`), `maskedPaths` **MUST NOT** be ignored. They should always be made masked and inaccessible. If currently `runc` doesn't do this, it should be fixed rather than the incorrect behaviour codified in the specification.

crosbymichael:
@cyphar runc is not ignoring any of them.  I tries to setup the ro or mask, but if the destination does not exist, then there is nothing for it mask or ro so it returns nil in this case.  I think the runc functionality is correct but the spec change with the wording `ignore` could be a little dangerous.

cyphar:
> It tries to setup the ro or mask, but if the destination does not exist, then there is nothing for it mask or ro so it returns nil in this case.

I think that this isn't right though -- I think that if we're told to mask a non-existent path then we should still mask it (to be fair, we won't be able to figure out if we should put a file or a directory over the path). The concern would be that the path isn't immediately available, but we want to make sure that the container process can never access that path.

cyphar:
I'm not sure that that this is correct though -- it depends what a user expects and what people want to use masking for. If we're told to mask a non-existent path then we arguably should still mask it (to be fair, we won't be able to figure out if we should put a file or a directory over the path). The concern would be that the path isn't immediately available, but we want to make sure that the container process can never access that path. Of course, on the other hand, runc does not traditionally modify `rootfs` and so doing so could cause confusion...

crosbymichael:
I'm not sure that would work, we mask an non-existent path and then someone mounts over the top later, it's not masked.

alban:
We cannot always create an empty file or directory when the destination is non-existent: in the example of `/proc/latency_stats`, we cannot `touch` or `mkdir` in procfs.

I think the current `runc` behaviour is fine.

runc can only masking paths it knows about at the time the container is set up. Subsequent mounts cannot be protected with `maskedPaths`. Users should just not give `CAP_SYS_ADMIN` for subsequent mounts.

> Unexistent paths MUST be ignored without generating an error.

Any other suggestion for the word `ignore`?

crosbymichael:
should this be a MUST or SHOULD?

cyphar:
If anything it should be SHOULD, but I'm still not a fan of this...

config-linux.md
```diff
@@ -635,6 +635,7 @@ The following parameters can be specified to set up seccomp:
 
 **`maskedPaths`** (array of strings, OPTIONAL) will mask over the provided paths inside the container so that they cannot be read.
     The values MUST be absolute paths in the [container namespace](glossary.md#container_namespace).
+    Unexistent paths MUST be skipped without generating an error.
```
vbatts: "Unexistent" is not a word. Either "inexistent" or "nonexistent".
And this is best a `SHOULD`, not a `MUST`.

config-linux.md
```diff
@@ -648,6 +649,7 @@ The following parameters can be specified to set up seccomp:
 
 **`readonlyPaths`** (array of strings, OPTIONAL) will set the provided paths as readonly inside the container.
     The values MUST be absolute paths in the [container namespace](glossary.md#container-namespace).
+    Unexistent paths MUST be skipped without generating an error.
```
vbatts: same

Issue #979, opened 2018-07-25
effective and permitted capability sets cannot be taken into account


alban:
https://github.com/opencontainers/runtime-spec/pull/675 introduced the ability to specify capabilities for each of the Linux capability sets:
- bounding
- permitted
- inheritable
- effective
- ambient

But this is very generic and not possible to implement meaningfully in a container runtime because:
- `effective` and `permitted` cannot be passed to the container because the kernel will reset them during the exec of the pid1 process in the container (see details in [Transformation of capabilities during execve()](http://man7.org/linux/man-pages/man7/capabilities.7.html)), so that's not really useful.
- all combinations are not possible, e.g. each `ambient` cap has to be in both `permitted` and `effective`.

To test this in practice, I use `runc run --bundle=$PWD foo | grep Cap` with the command `cat /proc/self/status` in `config.json`:

```
# runc run --bundle=$PWD foo | grep Cap
CapInh:	00000000a80425fb
CapPrm:	00000000a80425fb
CapEff:	00000000a80425fb
CapBnd:	00000000a80425fb
CapAmb:	00000000a80425fa
# capsh --decode=00000000a80425fb
0x00000000a80425fb=cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap
# capsh --decode=00000000a80425fa
0x00000000a80425fa=cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap
```

In this test, I had a `config.json` with `CAP_CHOWN` in the `bounding`, `inheritable` and `ambient` sets but not in the `effective` and `permitted` sets. But confusingly, the cap appears in the `effective` and `permitted` sets but not in the `ambient` set. With the rules in [Transformation of capabilities during execve()](http://man7.org/linux/man-pages/man7/capabilities.7.html)), that's the expected behaviour.

|             | Specified in config.json | Tested in the container |
|-------------|:------------------------:|:-----------------------:|
| bounding    |             ✔            |            ✔            |
| effective   |                          |            ✔            |
| inheritable |             ✔            |            ✔            |
| permitted   |                          |            ✔            |
[...]

PR #987, opened 2018-08-23
[RFC] Remove effective and permitted capabilities

dongsupark:
Both `effective` and `permitted` capabilities cannot be passed to a container because the Linux kernel will reset them during the exec of the pid1 process in the container. In practice, no matter how these two capabilities are set in a runtime config file, those can be set in the container to a totally different values.

So let's remove these capabilities from the runtime spec, because it's not possible to make them work inside containers.

NOTE: I'm aware that it would take much time for removing anything from the spec, because then all relevant runtimes and tools need to be also updated. However, I would like to get feedback from maintainers, to know if this removal makes sense at all. I waited for a reply since the last week, but got no answer. So I'm creating this PR. Any comment would be welcome.

Fixes https://github.com/opencontainers/runtime-spec/issues/979

Signed-off-by: Dongsu Park <dongsu@kinvolk.io>

### DIFF
--- a/config.md
+++ b/config.md
@@ -191,10 +191,8 @@ For Linux-based systems, the `process` object supports the following process-spe
     Any value which cannot be mapped to a relevant kernel interface MUST cause an error.
     `capabilities` contains the following properties:
 
-    * **`effective`** (array of strings, OPTIONAL) the `effective` field is an array of effective capabilities that are kept for the process.
     * **`bounding`** (array of strings, OPTIONAL) the `bounding` field is an array of bounding capabilities that are kept for the process.
     * **`inheritable`** (array of strings, OPTIONAL) the `inheritable` field is an array of inheritable capabilities that are kept for the process.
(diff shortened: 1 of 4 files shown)


crosbymichael:
I'm not sure. We have this logic in Docker for non-root users. 

https://github.com/moby/moby/commit/15ff09395c001bcb0f284461abbc404a1d8bab4d#diff-ea22bb26655e758dc94c291dbaee0e76R267

justincormack:
I don't think that is true in the case of `exec` a file with filesystem capabilities, so I don't think that removing them does make sense, even if the use cases are fairly niche.

crosbymichael:
I'm -1 on this change as well

vbatts:
_house keeping:_ I don't disagree with @dongsupark or @alban about the behavior they're seeing. The trouble is that folks may have already worked around the behavior, and by removing this would cause an effectively nop-version-bump.
Unfortunately, I think this would better be a PR adding a statement about the behavior that is implied to work around the niche case you've found.
I'm closing this PR for now.

PR #998, opened 2019-01-23
Windows:Have native CommandLine in Process

jhowardmsft:

Windows natively doesn't launch processes through an array of arguments. Instead, it uses a command line single string, which the startup code (eg https://docs.microsoft.com/en-us/previous-versions/ms880421(v=msdn.10)) interprets into an argv set. Golang does identical parsing of the command-line on Windows in commandLineToArgv: https://github.com/golang/go/blob/ff7b245a31394b700a252fd547cf16ad0ad838b6/src/os/exec_windows.go#L100-L174

See the MSDN `CreateProcess` documentation for the actual API into Windows: https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-createprocessa

Using args alone on Windows has possible loss of fidelity causing some commands to not operate as expected. (This has long been the case on Windows in docker for example). A contrived (but real) example in a test case in the moby repo where fidelity is lost is the following command-line:

`cmd /S /C mkdir "c:/foo"`

When parsed into a JSON array (such as the args), this becomes 5 elements as follows:

 - `cmd`
 - `/S`
 - `/C`
 - `mkdir`
 - `c:/foo`

Note specifically that the lost information are the double-quotes around `c:/foo`. When using the required contatenation/space separation, and argument escaping required on Windows (https://github.com/golang/sys/blob/c4afb3effaa53fd9a06ca61262dc7ce8df4c081b/windows/exec_windows.go#L9-L18), this becomes the following command line:

`cmd /S /C mkdir c:/foo`

When the double-quotes are missing, mkdir here fails, but with the double-quotes succeeds as expected, as shown in the following screenshot.

image

The addition of a full `commandLine` in Process for use on Windows alleviates issues where fidelity can be lost. 

If commandLine is omitted, the Windows runtime will fall-back to existing behaviour of contatenation/escaping described previously.

@crosbymichael PTAL

@dmcgowan & @jterry75 FYI.

### DIFF
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -50,12 +50,14 @@
         "process": {
             "type": "object",
             "required": [
-                "cwd",
-                "args"
+                "cwd"
             ],
             "properties": {
                 "args": {
                     "$ref": "defs.json#/definitions/ArrayOfStrings"
+                },
+                "commandLine": {
+                    "type": "string"
                 },
                 "consoleSize": {
                     "type": "object",
--- a/config.md
+++ b/config.md
@@ -155,8 +155,11 @@ For POSIX platforms the `mounts` structure has the following fields:
 * **`cwd`** (string, REQUIRED) is the working directory that will be set for the executable.
     This value MUST be an absolute path.
 * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
-* **`args`** (array of strings, REQUIRED) with similar semantics to [IEEE Std 1003.1-2008 `execvp`'s *argv*][ieee-1003.1-2008-functions-exec].
-    This specification extends the IEEE standard in that at least one entry is REQUIRED, and that entry is used with the same semantics as `execvp`'s *file*.
+* **`args`** (array of strings, OPTIONAL) with similar semantics to [IEEE Std 1003.1-2008 `execvp`'s *argv*][ieee-1003.1-2008-functions-exec].
+    This specification extends the IEEE standard in that at least one entry is REQUIRED (non-Windows), and that entry is used with the same semantics as `execvp`'s *file*. This field is OPTIONAL on Windows, and `commandLine` is REQUIRED if this field is omitted.
+* **`commandLine`** (string, OPTIONAL) specifies the full command line to be executed on Windows.
+    This is the preferred means of supplying the command line on Windows. If omitted, the runtime will fall back to escaping and concatenating fields from `args` before making the system call into Windows.
+
 
 ### <a name="configPOSIXProcess" />POSIX process
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -38,7 +38,9 @@ type Process struct {
 	// User specifies user information for the process.
 	User User `json:"user"`
 	// Args specifies the binary and arguments for the application to execute.
-	Args []string `json:"args"`
+	Args []string `json:"args,omitempty"`
+	// CommandLine specifies the full command line for the application to execute on Windows.
+	CommandLine string `json:"commandLine,omitempty" platform:"windows"`
 	// Env populates the process environment for the process.
 	Env []string `json:"env,omitempty"`
 	// Cwd is the current working directory for the process and must be


jhowardmsft:
@jterry75 Updated

jhowardmsft:
Fixed CI issues. Ready for review.

crosbymichael:
LGTM

jterry75:
@jhowardmsft - Might be worth adding a few unit tests to https://github.com/opencontainers/runtime-tools once this merges with the different Windows `Cmdline vs Args` and to update the default generator to `Cmdline` and not `Args`

schema/config-schema.json
```diff
@@ -50,13 +50,15 @@
         "process": {
             "type": "object",
             "required": [
-                "cwd",
-                "args"
+                "cwd"
```
jterry75: @crosbymichael - Is there a way to do `required: oneOf{ "args", "commandLine" }` or something here?

jhowardmsft:
Hold fire merging this until I have verified the whole end-to-end scenario

jhowardmsft:
OK, end-to-end has been validated. This is good to go if another maintainer can review please. Thanks.

jhowardmsft:
@tianon Any chance you could look at this quickly?

tianon:
This really makes no sense to me -- don't Windows applications still have `argc` and `argv`?  Don't they have explicit argument lists?

In rough Go pseudocode, I would imagine something like `RUN mkdir "foo"` to turn into ``exec.Command(`cmd`, `/S`, `/C`, `mkdir "foo"`)`` (using backtick-style strings to illustrate simpler), which should pass the entire `mkdir "foo"` string as-is to `cmd.exe` for further quote parsing, right?  Why doesn't Windows have a way to execute a process directly without Go having to do insane reverse quoting just to get it to do the "right" thing and launch a process with a given arbitrary `argv`?

jhowardmsft:
> Why doesn't Windows have a way to execute a process directly without

Don't shoot the messenger. I'm just saying what we have to work with here :innocent:

tianon:
I do come from a Windows background, but it's been a very long time now
since I read the "CreateProcess" docs, sorry 😅

So, I guess to put this another way, why does it make sense to embed this
legacy design quirk into the runtime specification? It seems to me it'd
make more sense to link to docs about how to quote an argv properly for use
in a Windows command line (like Go does) than to complicate the spec by
having two incompatible and platform specific ways to specify the same
thing?  Isn't consistency and standardization what we're going for here?

jhowardmsft:
I forgot. Sorry 😇 

Unfortunately legacy design quirk is the reality is current/how Windows still works today. Perhaps if you see this commit https://github.com/moby/moby/pull/38541/commits/3d1c46d8e9ad3c8b8b2595faab0f5b9b72d05ec4, it will start to make more sense. It's absolutely impossible to get moby to use containerd for its runtime without moving to what is natural on Windows for launching process which is a full command line, not a set of argv which need escaping prior to building a command line. Consistency and standardisation are all well and good when there is parity between platforms. This is not the case though here.

jhowardmsft:
@tianon See my message on slack too.

jhowardmsft:
Perhaps a real example where fidelity is lost might help.

```
package main

import (
	"fmt"
	"strings"

	"golang.org/x/sys/windows"
)

// Demonstrate why converting a command line into an array of
// args, then trying to rebuild the command line loses fidelity.
// In other words, showing why Windows needs a full command line
// without any other conversion.

func main() {
	cl := `cmd /s /c mkdir "c:/foo"`
	
	fmt.Printf("Original: '%s'\n", cl)

	args := commandLineToArgv(cl)
	for i, v := range args {
		fmt.Printf(" - %d: %s\n", i, v)
	}

	convertedBackToCommandLine := ""
	for _, v := range args {
		convertedBackToCommandLine += windows.EscapeArg(v) + " "
	}
	convertedBackToCommandLine = strings.TrimSpace(convertedBackToCommandLine)
	fmt.Printf("converted back: '%s'\n", convertedBackToCommandLine)

	if convertedBackToCommandLine != cl {
		fmt.Println("**** FIDELITY HAS BEEN LOST ****")
	}
	
}

// Following are copied directly from golang in the startup code.
// Same as for C/C++/Other languages runtime startup to convert
// command line to argv.

// appendBSBytes appends n '\\' bytes to b and returns the resulting slice.
func appendBSBytes(b []byte, n int) []byte {
	for ; n > 0; n-- {
		b = append(b, '\\')
	}
	return b
}

// readNextArg splits command line string cmd into next
// argument and command line remainder.
func readNextArg(cmd string) (arg []byte, rest string) {
	var b []byte
	var inquote bool
	var nslash int
	for ; len(cmd) > 0; cmd = cmd[1:] {
		c := cmd[0]
		switch c {
		case ' ', '\t':
			if !inquote {
				return appendBSBytes(b, nslash), cmd[1:]
			}
		case '"':
			b = appendBSBytes(b, nslash/2)
			if nslash%2 == 0 {
				// use "Prior to 2008" rule from
				// http://daviddeley.com/autohotkey/parameters/parameters.htm
				// section 5.2 to deal with double double quotes
				if inquote && len(cmd) > 1 && cmd[1] == '"' {
					b = append(b, c)
					cmd = cmd[1:]
				}
				inquote = !inquote
			} else {
				b = append(b, c)
			}
			nslash = 0
			continue
		case '\\':
			nslash++
			continue
		}
		b = appendBSBytes(b, nslash)
		nslash = 0
		b = append(b, c)
	}
	return appendBSBytes(b, nslash), ""
}

// commandLineToArgv splits a command line into individual argument
// strings, following the Windows conventions documented
// at http://daviddeley.com/autohotkey/parameters/parameters.htm#WINARGV
func commandLineToArgv(cmd string) []string {
	var args []string
	for len(cmd) > 0 {
		if cmd[0] == ' ' || cmd[0] == '\t' {
			cmd = cmd[1:]
			continue
		}
		var arg []byte
		arg, cmd = readNextArg(cmd)
		args = append(args, string(arg))
	}
	return args
}
```


Running this:

```
S E:\docker\test\cltoargv> go build; .\cltoargv.exe
Original: 'cmd /s /c mkdir "c:/foo"'
 - 0: cmd
 - 1: /s
 - 2: /c
 - 3: mkdir
 - 4: c:/foo
converted back: 'cmd /s /c mkdir c:/foo'
**** FIDELITY HAS BEEN LOST ****
PS E:\docker\test\cltoargv>
```

And showing how the double quotes are required in the real world:

image

tianon:
Ah, so if I'm understanding correctly, `cmd.exe` is a Windows application which doesn't use `int main(...)`, but instead uses `WinMain` (https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-winmain), and parses `lpCmdLine` directly? (that makes a lot more sense and makes me more positive to this change)

tianon:
LGTM

crosbymichael:
@opencontainers/runtime-spec-maintainers Any other comments or questions before I merge this?

vbatts:
I get why the argv array can be lossy, it's a shame we don't have a generic fix rather than a platform specific thing. But so it goes

jhowardmsft:
I said this offline to @tianon, but for completeness, adding it here too:

You essentially have it right, but there's a tiny bit of clarification. Indeed what you say is the case for a GUI app using standard Windows entrypoints. However, for all command line apps using the standard C or C++ runtime, even go programs, they too get invoked by Windows as a command line. In fact, ALL programs on Windows gets invoked by a command line. Not by an array of arguments.

So you would still have a C-style `int main(int argc, char* argv)` from a coders perspective writing an application on Windows. But the C/C++/Go/<insert other language> startup code will do additional processing prior to the applications main function being called. That is to convert from a command line to the argv style.

In go, that's the `commandLineToArgv` function which is called in the init() function of the os package https://github.com/golang/go/blob/ff7b245a31394b700a252fd547cf16ad0ad838b6/src/os/exec_windows.go#L100, so that in the actual programs `func main()` it can use `os.Args` to query the golang equivalent of argv[].

In the Microsoft C++ startup code, it does exactly the same processing as documented here prior to invoking main itself: https://docs.microsoft.com/en-us/cpp/cpp/parsing-cpp-command-line-arguments?view=vs-2017. The processing it describes is documented https://msdn.microsoft.com/en-us/library/windows/desktop/ms683156(v=vs.85).aspx and https://docs.microsoft.com/en-us/windows/desktop/api/shellapi/nf-shellapi-commandlinetoargvw

thaJeztah:
I see a lot of background is captured in the discussion here on GitHub; should a summary be created and added as commit message?

jhowardmsft:
> I see a lot of background is captured in the discussion here on GitHub; should a summary be created and added as commit message?

Sure, I'll add that.

jhowardmsft:
@thaJeztah @tianon @crosbymichael 

Have updated the commit message to capture the discussion above. Unfortunately, that means that the approvals have been lost due to an updated commit ID.

vbatts:
LGTM

caniszczyk:
LGTM

(test to see if this works now)

caniszczyk:
ok, we are unblocked

thaJeztah:
Thanks for the 😍 commit message, @jhowardmsft

PR #1008, opened 2019-04-24
Add create-host, create-container, start-container hooks

RenaudWasTaken:
Hello!

Since we discussed this during the last OCI weekly meeting, I wanted to continue the conversation on a more concrete change.
Related:
- https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/WTHcf-b3wXA
- https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/_dmqMkvX_iE

I've used the same names as LXC for the hooks. I'm happy to change them if we feel there are more appropriate names.
My reasoning while drafting this was to answer some of these questions:
- Definition: How can we define the `mount` hook in a way that would clearly define the ability to mount "stuff" inside the container
- Ordering: Is there an ordering that is required?
  Here it would be helpful to define the ordering such that the pre-start hook detects that it doesn't need to be executed in a setting where we specify both pre-start hook and mount hook for backwards compat reasons.

Questions still open:
- Should we define "outside" and "inside" (in relation to the VM debate)

What do you think about this first draft?

### DIFF
--- a/config.md
+++ b/config.md
@@ -378,39 +378,118 @@ For Windows based systems the user structure has the following fields:
 For POSIX platforms, the configuration structure supports `hooks` for configuring custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
 
 * **`hooks`** (object, OPTIONAL) MAY contain any of the following properties:
-    * **`prestart`** (array of objects, OPTIONAL) is an array of [pre-start hooks](#prestart).
-        Entries in the array contain the following properties:
-        * **`path`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2008 `execv`'s *path*][ieee-1003.1-2008-functions-exec].
-            This specification extends the IEEE standard in that **`path`** MUST be absolute.
-            Runtimes MUST resolve this value in the [runtime namespace](glossary.md#runtime-namespace).
-        * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008 `execv`'s *argv*][ieee-1003.1-2008-functions-exec].
-        * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
-        * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
-            If set, `timeout` MUST be greater than zero.
-    * **`poststart`** (array of objects, OPTIONAL) is an array of [post-start hooks](#poststart).
-        Entries in the array have the same schema as pre-start entries.
-    * **`poststop`** (array of objects, OPTIONAL) is an array of [post-stop hooks](#poststop).
-        Entries in the array have the same schema as pre-start entries.
+    * **`prestart`** (array of objects, OPTIONAL, **DEPRECATED**) is an array of [`prestart` hooks](#prestart).
+        * Entries in the array contain the following properties:
+            * **`path`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2008 `execv`'s *path*][ieee-1003.1-2008-functions-exec].
+                This specification extends the IEEE standard in that **`path`** MUST be absolute.
+            * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008 `execv`'s *argv*][ieee-1003.1-2008-functions-exec].
+            * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
+            * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
+                If set, `timeout` MUST be greater than zero.
+        * The value of `path` MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+        * The `prestart` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+    * **`createRuntime`** (array of objects, OPTIONAL) is an array of [`createRuntime` hooks](#createRuntime-hooks).
+        * Entries in the array contain the following properties (the entries are identical to the entries in the deprecated `prestart` hooks):
+            * **`path`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2008 `execv`'s *path*][ieee-1003.1-2008-functions-exec].
+                This specification extends the IEEE standard in that **`path`** MUST be absolute.
+            * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008 `execv`'s *argv*][ieee-1003.1-2008-functions-exec].
+            * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
+            * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
+                If set, `timeout` MUST be greater than zero.
+        * The value of `path` MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+        * The `createRuntime` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+    * **`createContainer`** (array of objects, OPTIONAL) is an array of [`createContainer` hooks](#createContainer-hooks).
+        * Entries in the array have the same schema as `createRuntime` entries.
+        * The value of `path` MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+        * The `createContainer` hooks MUST be executed in the [container namespace](glossary.md#container-namespace).
+    * **`startContainer`** (array of objects, OPTIONAL) is an array of [`startContainer` hooks](#startContainer-hooks).
+        * Entries in the array have the same schema as `createRuntime` entries.
+        * The value of `path` MUST resolve in the [container namespace](glossary.md#container-namespace).
+        * The `startContainer` hooks MUST be executed in the [container namespace](glossary.md#container-namespace).
+    * **`poststart`** (array of objects, OPTIONAL) is an array of [`poststart` hooks](#poststart).
+        * Entries in the array have the same schema as `createRuntime` entries.
+        * The value of `path` MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+        * The `poststart` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+    * **`poststop`** (array of objects, OPTIONAL) is an array of [`poststop` hooks](#poststop).
+        * Entries in the array have the same schema as `createRuntime` entries.
+        * The value of `path` MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+        * The `poststop` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
 
 Hooks allow users to specify programs to run before or after various lifecycle events.
 Hooks MUST be called in the listed order.
-Hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
 The [state](runtime.md#state) of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.
 
 ### <a name="configHooksPrestart" />Prestart
 
-The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+The `prestart` hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
 On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+Note: `prestart` hooks were deprecated in favor of `createRuntime`, `createContainer` and `startContainer` hooks, which allow more granular hook control during the create and start phase.
+
+The `prestart` hooks' path MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+The `prestart` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+
+### <a name="configHooksCreateRuntime" />CreateRuntime Hooks
+
+The `createRuntime` hooks MUST be called as part of the [`create`](runtime.md#create) operation after the runtime environment has been created (according to the configuration in config.json) but before the `pivot_root` or any equivalent operation has been executed.
+
+The `createRuntime` hooks' path MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+The `createRuntime` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+
+On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+The definition of `createRuntime` hooks is currently underspecified and hooks authors, should only expect from the runtime that the mount namespace have been created and the mount operations performed. Other operations such as cgroups and SELinux/AppArmor labels might not have been performed by the runtime.
+
+Note: `runc` originally implemented `prestart` hooks contrary to the spec, namely as part of the `create` operation (instead of during the `start` operation). This incorrect implementation actually corresponds to `createRuntime` hooks. For runtimes that implement the deprecated `prestart` hooks as `createRuntime` hooks, `createRuntime` hooks MUST be called after the `prestart` hooks.
+
+### <a name="configHooksCreateContainer" />CreateContainer Hooks
+
+The `createContainer` hooks MUST be called as part of the [`create`](runtime.md#create) operation after the runtime environment has been created (according to the configuration in config.json) but before the `pivot_root` or any equivalent operation has been executed.
+The `createContainer` hooks MUST be called after the `createRuntime` hooks.
+
+The `createContainer` hooks' path MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+The `createContainer` hooks MUST be executed in the [container namespace](glossary.md#container-namespace).
+
+For example, on Linux this would happen before the `pivot_root` operation is executed but after the mount namespace was created and setup.
+
+The definition of `createContainer` hooks is currently underspecified and hooks authors, should only expect from the runtime that the mount namespace and different mounts will be setup. Other operations such as cgroups and SELinux/AppArmor labels might not have been performed by the runtime.
+
+### <a name="configHooksStartContainer" />StartContainer Hooks
+
+The `startContainer` hooks MUST be called [before the user-specified process is executed](runtime.md#lifecycle) as part of the [`start`](runtime.md#start) operation.
+This hook can be used to execute some operations in the container, for example running the `ldconfig` binary on linux before the container process is spawned.
+
+The `startContainer` hooks' path MUST resolve in the [container namespace](glossary.md#container-namespace).
+The `startContainer` hooks MUST be executed in the [container namespace](glossary.md#container-namespace).
 
 ### <a name="configHooksPoststart" />Poststart
 
-The post-start hooks MUST be called [after the user-specified process is executed](runtime.md#lifecycle) but before the [`start`](runtime.md#start) operation returns.
+The `poststart` hooks MUST be called [after the user-specified process is executed](runtime.md#lifecycle) but before the [`start`](runtime.md#start) operation returns.
 For example, this hook can notify the user that the container process is spawned.
+
+The `poststart` hooks' path MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+The `poststart` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
 
 ### <a name="configHooksPoststop" />Poststop
 
-The post-stop hooks MUST be called [after the container is deleted](runtime.md#lifecycle) but before the [`delete`](runtime.md#delete) operation returns.
+The `poststop` hooks MUST be called [after the container is deleted](runtime.md#lifecycle) but before the [`delete`](runtime.md#delete) operation returns.
 Cleanup or debugging functions are examples of such a hook.
+
+The `poststop` hooks' path MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
+The `poststop` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+
+### Summary
+
+See the below table for a summary of hooks and when they are called:
+
+|           Name          | Namespace |                                                            When                                                                    |
+| ----------------------- | --------- | -----------------------------------------------------------------------------------------------------------------------------------|
+| `prestart` (Deprecated) | runtime   | After the start  operation is called but before the user-specified program command is executed.                                    |
+| `createRuntime`         | runtime   | During the create operation, after the runtime environment has been created and before the pivot root or any equivalent operation. |
+| `createContainer`       | container | During the create operation, after the runtime environment has been created and before the pivot root or any equivalent operation. |
+| `startContainer`        | container | After the start operation is called but before the user-specified program command is executed.                                     |
+| `poststart`             | runtime   | After the user-specified process is executed but before the start operation returns.                                               |
+| `poststop`              | runtime   | After the container is deleted but before the delete operation returns.                                                            |
 
 ### Example
 
@@ -424,6 +503,28 @@ Cleanup or debugging functions are examples of such a hook.
             },
             {
                 "path": "/usr/bin/setup-network"
+            }
+        ],
+        "createRuntime": [
+            {
+                "path": "/usr/bin/fix-mounts",
+                "args": ["fix-mounts", "arg1", "arg2"],
+                "env":  [ "key1=value1"]
+            },
+            {
+                "path": "/usr/bin/setup-network"
+            }
+        ],
+        "createContainer": [
+            {
+                "path": "/usr/bin/mount-hook",
+                "args": ["-mount", "arg1", "arg2"],
+                "env":  [ "key1=value1"]
+            }
+        ],
+        "startContainer": [
+            {
+                "path": "/usr/bin/refresh-ldcache"
             }
         ],
         "poststart": [
--- a/runtime.md
+++ b/runtime.md
@@ -55,18 +55,24 @@ The lifecycle describes the timeline of events that happen from when a container
     If the runtime is unable to create the environment specified in the [`config.json`](config.md), it MUST [generate an error](#errors).
     While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified program (from [`process`](config.md#process)) MUST NOT be run at this time.
     Any updates to [`config.json`](config.md) after this step MUST NOT affect the container.
-3. Runtime's [`start`](runtime.md#start) command is invoked with the unique identifier of the container.
-4. The [prestart hooks](config.md#prestart) MUST be invoked by the runtime.
-    If any prestart hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 9.
-5. The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
-6. The [poststart hooks](config.md#poststart) MUST be invoked by the runtime.
-    If any poststart hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
-7. The container process exits.
+3. The [`prestart` hooks](config.md#prestart) MUST be invoked by the runtime.
+    If any `prestart` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+4. The [`createRuntime` hooks](config.md#createRuntime-hooks) MUST be invoked by the runtime.
+    If any `createRuntime` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+5. The [`createContainer` hooks](config.md#createContainer-hooks) MUST be invoked by the runtime.
+    If any `createContainer` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+6. Runtime's [`start`](runtime.md#start) command is invoked with the unique identifier of the container.
+7. The [`startContainer` hooks](config.md#startContainer-hooks) MUST be invoked by the runtime.
+    If any `startContainer` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+8. The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
+9. The [`poststart` hooks](config.md#poststart) MUST be invoked by the runtime.
+    If any `poststart` hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
+10. The container process exits.
     This MAY happen due to erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
-8. Runtime's [`delete`](runtime.md#delete) command is invoked with the unique identifier of the container.
-9. The container MUST be destroyed by undoing the steps performed during create phase (step 2).
-10. The [poststop hooks](config.md#poststop) MUST be invoked by the runtime.
-    If any poststop hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
+11. Runtime's [`delete`](runtime.md#delete) command is invoked with the unique identifier of the container.
+12. The container MUST be destroyed by undoing the steps performed during create phase (step 2).
+13. The [`poststop` hooks](config.md#poststop) MUST be invoked by the runtime.
+    If any `poststop` hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
 
 ## <a name="runtimeErrors" />Errors
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -125,13 +125,26 @@ type Hook struct {
 	Timeout *int     `json:"timeout,omitempty"`
 }
 
+// Hooks specifies a command that is run in the container at a particular event in the lifecycle of a container
 // Hooks for container setup and teardown
 type Hooks struct {
-	// Prestart is a list of hooks to be run before the container process is executed.
+	// Prestart is Deprecated. Prestart is a list of hooks to be run before the container process is executed.
+	// It is called in the Runtime Namespace
 	Prestart []Hook `json:"prestart,omitempty"`
+	// CreateRuntime is a list of hooks to be run after the container has been created but before pivot_root or any equivalent operation has been called
+	// It is called in the Runtime Namespace
+	CreateRuntime []Hook `json:"createRuntime,omitempty"`
+	// CreateContainer is a list of hooks to be run after the container has been created but before pivot_root or any equivalent operation has been called
+	// It is called in the Container Namespace
+	CreateContainer []Hook `json:"createContainer,omitempty"`
+	// StartContainer is a list of hooks to be run after the start operation is called but before the container process is started
+	// It is called in the Container Namespace
+	StartContainer []Hook `json:"startContainer,omitempty"`
 	// Poststart is a list of hooks to be run after the container process is started.
+	// It is called in the Runtime Namespace
 	Poststart []Hook `json:"poststart,omitempty"`
 	// Poststop is a list of hooks to be run after the container process exits.
+	// It is called in the Runtime Namespace
 	Poststop []Hook `json:"poststop,omitempty"`
 }
 
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -12,6 +12,15 @@
                 "prestart": {
                     "$ref": "defs.json#/definitions/ArrayOfHooks"
                 },
+                "createRuntime": {
+                    "$ref": "defs.json#/definitions/ArrayOfHooks"
+                },
+                "createContainer": {
+                    "$ref": "defs.json#/definitions/ArrayOfHooks"
+                },
+                "startContainer": {
+                    "$ref": "defs.json#/definitions/ArrayOfHooks"
+                },
                 "poststart": {
                     "$ref": "defs.json#/definitions/ArrayOfHooks"
                 },
--- a/schema/test/config/good/spec-example.json
+++ b/schema/test/config/good/spec-example.json
@@ -155,6 +155,28 @@
                 "path": "/usr/bin/setup-network"
             }
         ],
+        "createRuntime": [
+            {
+                "path": "/usr/bin/fix-mounts",
+                "args": ["fix-mounts", "arg1", "arg2"],
+                "env":  [ "key1=value1"]
+            },
+            {
+                "path": "/usr/bin/setup-network"
+            }
+        ],
+        "createContainer": [
+            {
+                "path": "/usr/bin/mount-hook",
+                "args": ["-mount", "arg1", "arg2"],
+                "env":  [ "key1=value1"]
+            }
+        ],
+        "startContainer": [
+            {
+                "path": "/usr/bin/refresh-ldcache"
+            }
+        ],
         "poststart": [
             {
                 "path": "/usr/bin/notify-start",


RenaudWasTaken:
cc @vbatts @crosbymichael

vbatts:
cc @cyphar

config.md
```diff
@@ -375,7 +375,7 @@ For Windows based systems the user structure has the following fields:
 For POSIX platforms, the configuration structure supports `hooks` for configuring custom actions related to the [lifecycle](runtime.md#lifecycle) of the container.
 
 * **`hooks`** (object, OPTIONAL) MAY contain any of the following properties:
-    * **`prestart`** (array of objects, OPTIONAL) is an array of [pre-start hooks](#prestart).
+    * **`prestart`** (array of objects, OPTIONAL, **DEPRECATED**) is an array of [pre-start hooks](#prestart).
```
alban: In the section `### <a name="configHooksPrestart" />Prestart` below, you could explain that it's deprecated in favor of the two other hooks.

config.md
```diff
@@ -399,6 +403,18 @@ The [state](runtime.md#state) of the container MUST be passed to hooks over stdi
 The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
 On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
 
+### <a name="configHooksMount" />Mount Hooks
```
alban: Could you fix the example line 432 below?

config.md
```diff
@@ -399,6 +403,18 @@ The [state](runtime.md#state) of the container MUST be passed to hooks over stdi
...
+
+The mount hooks MUST be called as part of the create operation at a time that would allow the hook to execute operations both inside and outside the container.
+For example, on linux this would happen before the `pivot_root` operation is executed but after the mount namespace was created and setup.
+
+For runtimes that implement the deprecated pre-start hooks as mount hooks, mount hooks MUST be called after the prestart hooks.
```
alban: I'm not sure we should specify the order. Since the definition of the prestart hook is not changed, this order makes it impossible to implement a container runtime with support for the two hooks following the spec. And for compatibility, hooks installed on both prestart and mount could both check if the mount task is already done, so the order should not matter.

RenaudWasTaken: > And for compatibility, hooks installed on both prestart and mount could both check if the mount task is already done, so the order should not matter.

 The goal here was that you can rely on ordering instead of checking for state you might have persisted in the mount or prestart hook.

> this order makes it impossible to implement a container runtime with support for the two hooks following the spec

Do you mind detailling a bit more what you mean here :) ?

RenaudWasTaken:
Thanks for your review @alban, will get to them tomorrow!

config.md
```diff
@@ -399,6 +403,18 @@ The [state](runtime.md#state) of the container MUST be passed to hooks over stdi
 The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
 On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
 
+### <a name="configHooksMount" />Mount Hooks
+
+The mount hooks MUST be called as part of the create operation at a time that would allow the hook to execute operations both inside and outside the container.
+For example, on linux this would happen before the `pivot_root` operation is executed but after the mount namespace was created and setup.
```
cyphar: linux -> Linux.

RenaudWasTaken:
Comments addressed :)

RenaudWasTaken:
Ping @cyphar @vbatts @alban :)

config.md
alban: > Do you mind detailing a bit more what you mean here :) ?

Sorry, I take it back.

So now, the spec mandates steps to happen in this order:
1. the container namespaces are created
2. Prestart hooks
3. Mount hooks
4. pivot_root
5. Start hooks
6. The user-specified process is executed 
7. Poststart hooks

I thought the spec was saying that the Prestart hooks must happen after the namespaces are fully configured (including the pivot_root), which would make things impossible. But the wording is more vague:

> [The pre-start hooks] are called after the container namespaces are created

So I think it's ok :)

h-vetinari:
Any update on this?

RenaudWasTaken:
@cyphar can you look into this :) ?
We would be happy to have your signoff!

RenaudWasTaken:
Ping @cyphar

RenaudWasTaken:
Ping @cyphar @vbatts

h-vetinari:
Can anyone provide an update/review here? Pings based on [recent merges](https://github.com/opencontainers/runtime-spec/commits/master)
@tianon @crosbymichael @hqhq @vbatts @caniszczyk

RenaudWasTaken:
Ping @cyphar @vbatts

RenaudWasTaken:
@cyphar @vbatts can you guys take a stab at reviewing this :) ?

Thanks!

config.md
```diff
@@ -384,6 +384,10 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
         * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
         * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
             If set, `timeout` MUST be greater than zero.
+    * **`mount`** (array of objects, OPTIONAL) is an array of [mount hooks](#mount-hooks).
```
kad: "mount" is a really bad naming selection. It is applicable to single vendor at the moment (NVidia) and from naming denotes other usages. Currently prestart hook is used by others not only to do mounts, but also to do various devices initialization. If we want to be explicit in defining execution timings of the hooks, let's define that, but please don't imply in the names single use case of the hook.

config.md
bart0sh: I'd propose to not deprecate prestart hooks. They may be used for other purposes than mount and start hooks. For example our [FPGA device plugin](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/fpga_plugin/README.md) verifies and programs FPGA devices in a prestart hook.

RenaudWasTaken:
> I'd propose to not deprecate prestart hooks. They may be used for other purposes than mount and start hooks. For example our FPGA device plugin verifies and programs FPGA devices in a prestart hook.
> "mount" is a really bad naming selection. It is applicable to single vendor at the moment (NVidia) and from naming denotes other usages. Currently prestart hook is used by others not only to do mounts, but also to do various devices initialization. If we want to be explicit in defining execution timings of the hooks, let's define that, but please don't imply in the names single use case of the hook.

Let's take a step back :)

Pre-start hooks are not getting deprecated. In the spec, pre-start hooks are defined as follows:
> The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed.

After discussing this in https://github.com/opencontainers/runc/pull/1811 we realized that this meant that this meant that pre-start hooks needed to be called after the `pivot_root` operation.
However, runc implemented pre-start hooks such that they were called before the `pivot_root` operation.

After discussion, we decided to clarify the meaning of pre-start hooks and allow to have hooks both "before" and "after" `pivot_root`.

Hence, we came up with the two new hooks:
- `mount` hooks, which would be called at the exact same time as the current pre-start hook as currently implemented in runc
- `start` hooks which would be called at the exact same time as the one defined in the spec.

So pre-start hooks aren't being deprecated "per-se", they are being renamed. The names were chosen based on LXC's current hooks but if you feel strongly about another name, I'll happily change the PR.

kad:
@RenaudWasTaken the problem is that "mount" implies a lot of meaning and does more confusion than clarification in comparison with current `prestart` name.

 I would rather do the names that will explicitly refer to environment where it is executed (before/after pivot_root, to be explicit). Something like: keep current `prestart` name as it is, with adding clarification to the spec that it is executed in the host namespace and something like `containerprestart` to explicitly indicate that it will be executed in the container namespace.
Or something like you mentioned `inside` / `outside` combined with  operations `pre`/`post` `start`/`stop`.  Names should be descriptive enough for sysadmin who will be configuring those to understand without going and reading deeply OCI spec.

RenaudWasTaken:
> Something like: keep current prestart name as it is, with adding clarification to the spec that it is executed in the host namespace

That's unfortunately not an option, clarifying it either way is a breaking change.

> `containerprestart` to explicitly indicate that it will be executed in the container namespace.

I have no strong feelings on the naming, `containerprestart` seems fine. The naming used here is the same as the one chosen by LXC with the intention to cause less confusion.

If we think there is more value in using a different name that is clearer than I'm happy to change this PR.

config.md
```diff
@@ -399,6 +403,20 @@ The [state](runtime.md#state) of the container MUST be passed to hooks over stdi
 The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
 On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
 
+Note: Prestart hooks were deprecated in favor of mount and start hooks, which have a clearer definition.
```
cyphar: I would explicitly mention that `runc` incorrectly implemented them, to ensure users don't depend on them.

cyphar:
@kad runc will continue to support `prestart` indefinitely as-is (internally mapping them to `mount` hooks). This is precisely the reason why we need to have new hook names, to stop breaking existing users. But `prestart` is wrong as-is and so we need to deprecate it (to stop users from using it given the mismatch between runc and runtime-spec).

@RenaudWasTaken I think copying the LXC hooks is the only reasonable choice. I don't want to get caught up in bike-shedding over what hook names we should use.

From my side, this LGTM -- as long as the NVIDIA folks are happy with it (sorry for the long time to review, I've been swamped with `openat2` and related work).

RenaudWasTaken:
I've updated the PR :)

config.md
bart0sh: Are we going to use -host and -container or -inside and -outside suffixes to the hook names? It's still not clear to me if proposed hooks will be executed in the host or container namespace. "mount" looks quite vague to me in this regard. It requires some guesswork to understand its purpose. Would it be better to have prestart-host and prestart-container instead?

h-vetinari:
> @bart0sh: Would it be better to have prestart-host and prestart-container instead?

I like this idea. It keeps the original name "prestart", and clarifies how it has been split up between before/after the pivot_root.

> @RenaudWasTaken: I've updated the PR :)

Did you forget to push? ;-)

> @cyphar: sorry for the long time to review, I've been swamped with `openat2` and related work

Glad to see this moving forward again in any case! I'm guessing you're thinking about solving this issue already for `runc-1.0.0rc9`, which I'm also guessing should go out quickly due to the new CVE? But IMO that shouldn't rush the decision of the naming here, which the community will be stuck with for a long time.

If my surmise of the situation is correct, how about releasing rc9 first and then solving the hooks under less time pressure?

config.md
kad: @cyphar @Renaudwastaken  Can you clarify the statement about "`runc` incorrectly implemented them" ?  From what I see in https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks it says "Hooks MUST be executed in the runtime namespace.", and  prestart is executed accordingly to this specification.

RenaudWasTaken: > The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed. 

This points to having the hook after the `pivot_root` is called :)

https://github.com/opencontainers/runtime-spec/blob/master/config.md#prestart

RenaudWasTaken: @bart0sh, is the naming something that can be resolved by better detailling the hooks in the spec?

Today they are worded as follows:
*mount hooks*
> The mount hooks MUST be called as part of the create operation at a time that would allow the hook to execute operations both inside and outside the container.
> For example, on Linux this would happen before the `pivot_root` operation is executed but after the mount namespace was created and setup.

*start hooks*
> The start hooks MUST be called [before the user-specified process is executed](runtime.md#lifecycle) as part of the [`start`](runtime.md#start) operation.
> This hook can be used to execute some operations in the container, for example running the ldconfig binary on linux before the container process is spawned.

Do you think we could reword the definition in a different way that would reduce the confusion?

RenaudWasTaken:
> Did you forget to push? ;-)

Thanks! Updated 

>> @bart0sh: Would it be better to have prestart-host and prestart-container instead?
> I like this idea. It keeps the original name "prestart", and clarifies how it has been split up.

The issue that I faced with using the `host` word, is that it becomes pretty confusing as to what it means for a vm based runtime (is host the hypervisor or the guest?).

It seemed to me that keeping existing terminology (i.e: used in LXC) was a lot less confusing  since a `hook` author is likely to be writing hooks for both OCI compliant runtimes (runc, ...) and non OCI compliant runtimes (LXC).

config.md
kad: > > The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed.
> 
> This points to having the hook after the `pivot_root` is called :)
> https://github.com/opencontainers/runtime-spec/blob/master/config.md#prestart

Well, not really. It specifies the timing. But where hook itself is called is specified by line that I quoted from spec above: "Hooks MUST be executed in the runtime namespace", and runtime namespace defined as  https://github.com/opencontainers/runtime-spec/blob/master/glossary.md#runtime-namespace

"After `pivot_root`" will be already not a runtime namespace.

bart0sh:
@RenaudWasTaken 

> The issue that I faced with using the host word, is that it becomes pretty confusing as to what it means for a vm based runtime (is host the hypervisor or the guest?).

That makes sense to me. What about -runtime and -container suffixes? They'd match namespaces  mentioned in the spec.

> It seemed to me that keeping existing terminology (i.e: used in LXC) was a lot less confusing since a hook author is likely to be writing hooks for both OCI compliant runtimes (runc, ...) and non OCI compliant runtimes (LXC).

Not my case, I'm afraid. 'mount' seems very confusing to me for the following reasons:
- it's not generic as it suggests particular use case
- it's not clear in which namespace it runs
- I've never written LXC hooks and not going to do it in foreseeable future. Our use cases are fully covered by OCI compliant runtimes.

cyphar:
> @h-vetinari: If my surmise about the situation is correct, how about releasing rc9 first and then solving the hooks under less time pressure?

I already sent out the vote for 1.0.0-rc9 -- this is a blocker for runc 1.0 not any of the rcs. The main reason I responded to this is because @dongsupark pinged me in person about it.

> @bart0sh What about -runtime and -container suffixes? They'd match namespaces mentioned in the spec.

My problem with the suffixes is that they are not accurate -- both of the hooks will run inside the container's namespace (it's just that one of them is before `pivot_root`). 

> @bart0sh: it's not generic as it suggests particular use case

In my mind "mount" means "call this hook during the setup of the container's mountpoints" (though to be fair, LXC probably named them "mount" specifically because that was their purpose -- they have many other "pre-start" hooks with very different names as well).

By using a more specific name, in the future we can have more hooks at different points during OCI runtime execution -- but if we limit ourselves to `prestart-runtime` then it will be more confusing if we wish to add more hooks in the future (`pre-prestart-runtime` if you want a slightly earlier hook?).

Not to mention that reusing the name `prestart` is just begging for people to get confused.

> @bart0sh: it's not clear in which namespace it runs

See above point about the suffixes -- but additionally, the solution is people should read the documentation and not rely on what the name seems to imply (since `prestart-runtime` might lead you to conclude you can modify the host's mounts -- which you cannot (easily) do).

> @bart0sh: I've never written LXC hooks and not going to do it in foreseeable future. Our use cases are fully covered by OCI compliant runtimes.

But you're not the only user of OCI's runtime hooks. If you won't ever use LXC, that's totally fine -- but it also means that if there is an argument to using them (consistency between runtimes, making it easier to implement tools like NVIDIA's hooks that run under both LXC and runc, and so on) then you saying "I don't care about that" isn't really an argument against doing it.

kad:
> > @bart0sh What about -runtime and -container suffixes? They'd match namespaces mentioned in the spec.
> 
> My problem with the suffixes is that they are not accurate -- both of the hooks will run inside the container's namespace (it's just that one of them is before `pivot_root`).

Wait. That significantly breaks the existing functionality. See my previous comment to @RenaudWasTaken: https://github.com/opencontainers/runtime-spec/pull/1008#discussion_r330697376 
If `mount` you expect to execute in container namespace, then `prestart` can't be deprecated as it explicitly stated to be executed in runtime namespace. See https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks

cyphar:
@kad Sorry, you're completely right -- I got what `runc` (incorrectly) does mixed up with what the spec says. In that case `prestart-runtime` is an accurate name for the hook, though I'm still of the mind that we should use the LXC names (for the other reasons I outlined above).

bart0sh:
@cyphar 

> In that case prestart-runtime is an accurate name for the hook

Looks like we're slowly making progress on this :)

> though my gut instinct is still to use the LXC hook names

ok, let's look at which hooks currently exist in LXC and OCI runtime. Hopefully it'll help us to come up with the right solution.

LXC:
```
Name                Namespace     When
lxc.hook.pre-start  runtime       Before the container ttys, consoles, or mounts are up.
lxc.hook.pre-mount  container     Before the rootfs has been set up.
lxc.hook.mount      container     After mounting has been done, but before the pivot_root.
lxc.hook.autodev    container     After mounting has been done and after any mount hooks have run, but before the pivot_root.
lxc.hook.start-host runtime       After the container has been setup, and immediately before starting the container init..
lxc.hook.start      container     Immediately before executing the container's init
lxc.hook.stop       runtime       After the container has been shut down with references to the container's namespaces.
lxc.hook.post-stop  runtime       After the container has been shut down.
lxc.hook.clone      ???           When the container is cloned to a new one.
lxc.hook.destroy    ???           when the container is destroyed.
```

OCI runtime:
```
Name                Namespace     When
prestart            runtime       After the start operation is called but before the user-specified program command is executed.
poststart           runtime       After the user-specified process is executed but before the start operation returns.
poststop            runtime       After the container is deleted but before the delete operation returns.
```

OCI runtime hooks are similar, but not quite the same as lxc.hook.pre-start, lxc.hook.start-host and lxc.hook.post-stop. prestart and poststop have slightly different names, poststart name is different from  lxc.hook.start-host.

Looks like the names are not fully matched already. However if we want to match them then introducing mount and start hooks would help with this. Introducing -runtime and -container suffixes would make the name convention more understandable though.

Thoughts?

h-vetinari:
A ping for `prestart-runtime` and `prestart-container` hooks. :)

RenaudWasTaken:
I took another good look at this today (and experimented with lxc), I agree on the naming though I'm wondering if it would make sense to have the following:

```
Name                   Namespace     When
prestart-runtime   runtime            After the create operation is called but before the pivot_root.
mount                  container          After the create operation is called but before the pivot_root (as part of the create operation at a time that would allow the hook to execute operations both inside and outside the container)
start                     container          Before the user-specified process is executed (should we specify after start? Or better in the same way as lxc defines it `Immediately before executing the container's init`)
```

So that would mean the current `prestart` becomes `pre-start-runtime` and we add two hooks that are the same as the LXC ones?

kad:
> ```
> Name                   Namespace     When
> prestart-runtime       runtime       After the create operation is called but before the pivot_root.
> ```
Functionality-wise, this will be same as current `prestart`, isn't it?
Why not to keep existing name, or just add alias?

RenaudWasTaken:
Oh right sorry I forgot to answer you on the thread :)

Yes you are right, prestart hooks are defined to be called from the runtime namespace. However they are also defined to be called after the start operation.
I think everyone's view on this issue was that we shouldn't be changing the definition of something that's already part of the spec (as this would probably create a mess in terms of implementation).

Functionality wise this is indeed the prestart hook implemented by runc. I'm not sure what you mean by alias though :)

kad:
> Yes you are right, prestart hooks are defined to be called from the runtime namespace. However they are also defined to be called after the start operation.

Well, according to spec, it must be called in the implementation of the "start" operation of the container, before executing user process. This is not really "after the start":

```
Prestart
The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed.
On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
```

> I think everyone's view on this issue was that we shouldn't be changing the definition of something that's already part of the spec (as this would probably create a mess in terms of implementation).

Indeed. Trying to rename/deprecate something without changing functionality will create even more for users who are using that on clusters with different versions of runc.
 
> Functionality wise this is indeed the prestart hook implemented by runc. I'm not sure what you mean by alias though :)

I mean, if current `prestart` and proposed `prestart-runtime` is exactly same from functionality point of view, what is the the point of renaming? If somebody really want to introduce new name, we can do it without deprecating existing `prestart` to do not break compatibility with older versions.

RenaudWasTaken:
> Well, according to spec, it must be called in the implementation of the "start" operation of the container, before executing user process. This is not really "after the start":

>> Prestart
>> The pre-start hooks MUST be called after the start operation is called but before the user-specified program command is executed.
>> On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).

Agreed, we're still stuck with a definition that doesn't really match the reality :D !

> I mean, if current prestart and proposed prestart-runtime is exactly same from functionality point of view, what is the the point of renaming? If somebody really want to introduce new name, we can do it without deprecating existing prestart to do not break compatibility with older versions.

Because `pre-start` as defined by the OCI is not `pre-start` as implemented in runc. The goal here is to have the `pre-start` hook as implemented in runc be part of the spec.

Because we can't change the definition of `pre-start` as defined in runc, our thought process is:
1. Deprecated `pre-start` (which doesn't mean that it should be removed).
2. Introduce `pre-start-runtime` which is `pre-start` as implemented in runc.

Additionally I'm also suggesting that we should have `mount` and `start` hooks as they would be pretty useful, e.g: running a container binary (ldconfig) in the container namespace rather than the host namespace.


What that means is that for existing consumers of the `pre-start` hook in runc, is that they will be able to continue using that hook until the runc maintainers decided they want to remove it (which might very well be never).

They can also use the new `pre-start-runtime` hook that is going to be called at the same time as `prestart`

kad:
> Agreed, though we're still stuck with a definition that doesn't really match the reality :D !

Hm. Can you point me to the code in runc which doesn't match reality?
What I see, is that in libcontainer/process_linux.go, in initProc:start() it setups cgroup, executes `prestart` and then sends `run` to the user specified container process.
What I'm missing?

> > I mean, if current prestart and proposed prestart-runtime is exactly same from functionality point of view, what is the the point of renaming? If somebody really want to introduce new name, we can do it without deprecating existing prestart to do not break compatibility with older versions.
> 
> Because `pre-start` as defined by the OCI is not `pre-start` as implemented in runc. The goal here is to have the `pre-start` hook as implemented in runc be part of the spec.

So far, I don't see contradictions between spec and implementation. What I'm missing?

> Because we can't change the definition of `pre-start` as defined in runc, our thought process is:
> 
> 1. Deprecated `pre-start` (which doesn't mean that it should be removed).
> 2. Introduce `pre-start-runtime` which is `pre-start` as implemented in runc.

So, what will be difference in those two in terms of implementation in runc?


> Additionally I'm also suggesting that we should have `mount` and `start` hooks as they would be pretty useful, e.g: running a container binary (ldconfig) in the container namespace rather than the host namespace.

Even were I consider those hooks as something that helps to violate one of the main principles of the container architecture ("immutable container image in all environments"), I don't mind to add `start`/`mount`. My main questions right now are about `prestart`.

 > What that means is that for existing consumers of the `pre-start` hook in runc, is that they will be able to continue using that hook until the runc maintainers decided they want to remove it (which might very well be never).
> 
> They can also use the new `pre-start-runtime` hook that is going to be called at the same time as `prestart`

Ok, so, again, what is the difference between existing `prestart` and `pre-start-runtime`?

h-vetinari:
@kad, context is here: opencontainers/runc#1710 opencontainers/runc#1811. Since the latter PR would break already-deployed NVidia code, `runc 1.0.0` has been stuck for more than a year.

AFAICT there's no way around deprecating the wrongly-implemented `prestart`-hook, and replacing it with new and correct hooks that allow all existing functionality to be reflected. The largest open question seems to be the colour of the bikeshed (the names of the new hooks) - I'm sure the implementation would follow quickly, since there are many people interested in getting `runc 1.0.0` finalised.

kad:
> @kad, context is here: [opencontainers/runc#1710](https://github.com/opencontainers/runc/issues/1710) [opencontainers/runc#1811](https://github.com/opencontainers/runc/pull/1811). Since the latter PR would break already-deployed NVidia code, `runc 1.0.0` has been stuck for more than a year.

Thanks @h-vetinari . I'll have a look at those lengthly discussions. I don't mind new hooks that NVidia needs, I'm more concerned about things that supposedly working now, so was trying to understand how/why it might break nvidia and will it break our usages. I'll try to dig out that info from issues above.

RenaudWasTaken:
> So far, I don't see contradictions between spec and implementation. What I'm missing?

Thanks @h-vetinari for linking to the correct issues :)
The implementation is different in runc vs the definition of the OCI (Prestart, Poststart hooks are called from the context of create).

> So, what will be difference in those two in terms of implementation in runc?
> Ok, so, again, what is the difference between existing prestart and pre-start-runtime?

As mentioned higher ("Introduce pre-start-runtime which is pre-start as implemented in runc.") from runc's perspective these would be the same.

Hope this helps move the conversation forward!

RenaudWasTaken:
I updated the PR as follows:
- Added "prestart-runtime"
- Rephrased the following definition:
    - From: "The mount hooks MUST be called as part of the create operation at a time that would allow the hook to execute operations both inside and outside the container."
    - To "The mount hooks MUST be called as part of the create operation after the runtime environment has been created (according to the configuration in config.json) but before the pivot_root or any equivalent operation has been executed."
- Added for each hook the namespace in which it is expected to be executed
- Updated specs-go/config.go
- Updated the schema
- Updated the runtime flow
- Added a summary table

config.md
```diff
@@ -384,31 +384,84 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
         * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
         * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
             If set, `timeout` MUST be greater than zero.
+    * **`prestart-runtime`** (array of objects, OPTIONAL) is an array of [prestart-runtime hooks](#prestart-runtime-hooks).
```
bart0sh: I'm not sure I follow this. prestart-runtime  hooks look the same as prestart, so why not to keep using prestart hooks?

bart0sh:
I'd propose these:
- leave prestart hook as it is as its new name seems to cause confusions here.
- Don't add prestart-runtime
- remove mention of incorrect runc implementation or give more details

The rest looks good to me.

RenaudWasTaken:
The following script highlight the behavior that runc has (i.e: prestart is part of the create operation):
```
$ mkdir alpine && cd alpine
$ mkdir rootfs
$ docker export $(docker create alpine) | tar -C rootfs -xf -
$ runc spec
$ cat config.json | jq -r '.process.terminal = false' | jq '. + {"hooks": {"prestart": [{"path": "/bin/sh", "args": ["/bin/sh", "-c", "echo prestart is in create > /tmp/foo"]}]}}' &> new_config.json
$ mv new_config.json config.json
$ sudo runc create --bundle . foo
$ cat /tmp/foo
prestart is in create
$ runc --version
runc version 1.0.0-rc8
commit: 425e105d5a03fabd737a126ad93d62a9eeede87f
spec: 1.0.1-dev
```
So to drive down the point that's been discussed at length here and in  opencontainers/runc#1710 opencontainers/runc#1811:
- OCI defines prestart hooks to be executed during the START operation: [here](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#lifecycle) and [here](https://github.com/opencontainers/runtime-spec/blob/master/config.md#prestart)
- runc implements prestart hooks as part of the CREATE operation

This PR makes the following changes:
   - Add prestart-runtime which is prestart as implemented in runc
   - Deprecate `prestart`
   - Add mount and start hooks

I'm happy to add more details on the runc incorrect implementation by linking to the correct issues :)

kad:
I have a proposal on how to move forward with smaller steps: let's split two things:
1. mount/start hooks in container namespace. -- let's do it in this PR. Reasons and implementation are understood and clear. Let's get it merged to spec and allow `runc` to implement those two new hooks.
2. For second change in this PR, about `prestart*` - let's do separate PR for the spec once we figure out what we should do with `prestart`. 

Reason for second item is because as I mentioned in my comment in https://github.com/opencontainers/runc/pull/1811#issuecomment-545620464 - I tried to apply patch from opencontainers/runc#1811 and run some of my tests: I don't see any significant change in behaviour after applying patch. So, I'm really curious what will break on nvidia side after applying opencontainers/runc#1811 ?

One more thing: spec document should have specification, and should not talk about bugs in one of the implementations of that specification, thus IMHO all notes/refs to runc issues should not be in the official specification document.

RenaudWasTaken:
@kad I don't think merging `mount`/`start` now changes anything, this PR is blocking runc 1.0.0

We all agree that `prestart` implemented in runc is different than `prestart` described in the OCI spec.
Can you guys detail what is the resistance of making the `prestart` hook that is implemented in runc officially part of the spec? Do you have alternatives that you think make more sense?

> Reason for second item is because as I mentioned in my comment in opencontainers/runc#1811 (comment) - I tried to apply patch from opencontainers/runc#1811 and run some of my tests: I don't see any significant change in behaviour after applying patch. So, I'm really curious what will break on nvidia side after applying opencontainers/runc#1811 ?

The side effects are fairly well described in the PR, see this comment: https://github.com/opencontainers/runc/pull/1811#issuecomment-420441651

config.md
```diff
@@ -384,31 +384,84 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
 ### <a name="configHooksPrestart" />Prestart
 
-The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+The prestart hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+Note: Prestart hooks were deprecated in favor of prestart-runtime, mount and start hooks, which have a clearer definition.
+Additionally note that `runc` incorrectly implemented the prestart hooks, users should be warry of relying on a particular behavior (see [https://github.com/opencontainers/runc/issues/1710](https://github.com/opencontainers/runc/issues/1710) and [https://github.com/opencontainers/runc/pull/1811](https://github.com/opencontainers/runc/pull/1811)).
+
+Prestart hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
```
h-vetinari: This should now be specified to refer either specifically to the deprecated `prestart` hook, or it is contradictory to the table below (where mount/start hooks are specified to be in the container namespace).

config.md
```diff
@@ -384,31 +384,84 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+This hook can be used to execute some operations in the container, for example running the ldconfig binary on linux before the container process is spawned.
+
+Start hooks MUST be executed in the [container namespace](glossary.md#container-namespace).
+
 ### <a name="configHooksPoststart" />Poststart
 
 The post-start hooks MUST be called [after the user-specified process is executed](runtime.md#lifecycle) but before the [`start`](runtime.md#start) operation returns.
 For example, this hook can notify the user that the container process is spawned.
 
+Postart hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
```
h-vetinari: Postart -> Poststart

config.md
RenaudWasTaken: Sorry I'm not sure I understand your comment :)

This statement should refer to the deprecated `prestart` hook because it is in the `prestart` section.

config.md
RenaudWasTaken: Thanks!

config.md
```diff
@@ -384,31 +384,84 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
-Hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
 The [state](runtime.md#state) of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.
 
 ### <a name="configHooksPrestart" />Prestart
 
-The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+The prestart hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+Note: Prestart hooks were deprecated in favor of prestart-runtime, mount and start hooks, which have a clearer definition.
```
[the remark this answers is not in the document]
RenaudWasTaken: Thanks!

config.md
```diff
@@ -384,31 +384,84 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+Additionally note that `runc` incorrectly implemented the prestart hooks, users should be warry of relying on a particular behavior (see [https://github.com/opencontainers/runc/issues/1710](https://github.com/opencontainers/runc/issues/1710) and [https://github.com/opencontainers/runc/pull/1811](https://github.com/opencontainers/runc/pull/1811)).
```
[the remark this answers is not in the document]
RenaudWasTaken: Thanks!

config.md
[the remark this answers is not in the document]
RenaudWasTaken: Thanks! Fixed :)

config.md
h-vetinari: I understood that only after re-reading, which is why I commented. I think the confusing part is that it refers to the plural "hooks" even though it talks about the single kind of `prestart` hook (of which there can be several), and *not* the several prestart hooks (as in collectively about the hooks available during the start phase) that are being introduced here.

I think one way to make this clearer would be to use the name as "`prestart` hooks" (and correspondingly for the other hooks).

h-vetinari:
@RenaudWasTaken 
Thanks for the response!

> @RenaudWasTaken: I'm not attached to the name but as a hook author having hooks that do the same thing be named consistently across runtimes is a nice thing to have (here LXC and OCI runtimes).

This can be a goal of course, but it's not a goal that applies to everyone. First and foremost, the OCI spec should IMO be looking to be consistent within itself, before looking to be consistent with other specs. As such, I think your option `b)` would be completely feasible.

> @RenaudWasTaken: I also believe that hook authors will be reading the definition of the hook

It was explicitly brought up [above](https://github.com/opencontainers/runtime-spec/pull/1008#issuecomment-529539881) that this does not cover all relevant cases, and - along the same lines - names should IMO be as self-descriptive as possible:
> @kad: Names should be descriptive enough for sysadmin who will be configuring those to understand without going and reading deeply OCI spec.
----------------

> @RenaudWasTaken: *Note: `create-runtime` does sound better.*

Glad to hear it! :)

RenaudWasTaken:
>> It was explicitly brought up above that this does not cover all relevant cases, and - along the same lines - names should IMO be as self-descriptive as possible:
>> @kad: Names should be descriptive enough for sysadmin who will be configuring those to understand without going and reading deeply OCI spec.

To be honest, I view looking at the hooks definition as running `man` on a command you are not familiar with (whether it's called `mount` or `create-runtime`).
Additionally note that not all hooks will be manually configured by a sysadmin.

> This can be a goal of course, but it's not a goal that applies to everyone. First and foremost, the OCI spec should IMO be looking to be consistent within itself, before looking to be consistent with other specs. As such, I think your option b) would be completely feasible.

My thought process is that hooks author will be interacting more with that part of the spec than sysadmins hence prioritizing a) above option b).

Should we bring this discussion to the next OCI meeting and see if we can have a decision on it?

cyphar:
@amshinde I have no problem with adding `*-guest-*` hooks as well, though I think that:

 * We should put that in a separate PR to not further block this one.
 * We will need to figure out what the semantics for non-VM runtimes should be. We could make `create-guest-* == create-*` in that case, but I'm not sure how much I like that. I'd also want to put a the word "vm" in the hook name to clarify that those hooks are only useful for VM-based runtimes (`create-vmguest-*` or something).

RenaudWasTaken:
Following today's meeting I'll be updating the PR with the following names:
- prestart-runtime -> create-runtime
- mount -> create-container
- start -> start-container

@cyphar @h-vetinari @bart0sh @alban are there any comments you think need to be addressed in this PR?

bart0sh:
@RenaudWasTaken new names look good to me.

> Note: Prestart hooks were deprecated in favor of create-runtime, create-container and start-container hooks, which allow more granular hook control during the create and start phase.

I'm still not sure how prestart hooks that run in runtime namespace can be deprecated in favor of create-container and start-container that run in a container namespace.
I'd suggest to only mention create-runtime, i.e. Prestart hooks were deprecated in favor of create-runtime hooks.

> Additionally note that `runc` incorrectly implemented the prestart hooks,

After giving it some thought I'd suggest to remove this note. Mentioning implementation bugs in the specification doesn't look like a good practice to me.

amshinde:
> @amshinde I have no problem with adding `*-guest-*` hooks as well, though I think that:
> 
>     * We should put that in a separate PR to not further block this one.

@cyphar I am good with addressing this in a separate PR.

RenaudWasTaken:
>> Additionally note that runc incorrectly implemented the prestart hooks,

> After giving it some thought I'd suggest to remove this note. Mentioning implementation bugs in the specification doesn't look like a good practice to me.

Done :)

>> Note: Prestart hooks were deprecated in favor of create-runtime, create-container and start-container hooks, which allow more granular hook control during the create and start phase.

> I'm still not sure I understand how prestart hooks that run in runtime namespace can be deprecated in favor of create-container and start-container that run in a container namespace.
I'd suggest to only mention create-runtime, i.e. Prestart hooks were deprecated in favor of create-runtime hooks.


Can we agree to disagree on this one :D ?
To me this says that we deprecated a specific hook in favor of other more granular and explicit hooks. They don't map exactly but fit several use case that were intended to be addressed by the first one :).


@cyphar @h-vetinari @bart0sh @alban do you see any other comments you want addressed in this PR? If not I think we have a pretty good consensus and only need some official stamp of approval from you!

config.md
```diff
@@ -384,31 +384,84 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
 
 ### <a name="configHooksPrestart" />Prestart
 
-The pre-start hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+The prestart hooks MUST be called after the [`start`](runtime.md#start) operation is called but [before the user-specified program command is executed](runtime.md#lifecycle).
+On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+Note: Prestart hooks were deprecated in favor of mount and start hooks, which have a clearer definition.
+Additionally note that `runc` incorrectly implemented the prestart hooks, users should be warry of relying on a particular behavior.
+
```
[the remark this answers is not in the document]
RenaudWasTaken: I think we've agreed that we didn't want to link to the specific runc issue that was detailing the implementation issues. I'll mark this resolved.

bart0sh:
@RenaudWasTaken 

> Can we agree to disagree on this one :D ?
To me this says that we deprecated a specific hook in favor of other more granular and explicit hooks. They don't map exactly but fit several use case that were intended to be addressed by the first one :).

That's minor thing I believe. However, the deprecation of the existing hook is not. The main reason of the deprecation as far as I understand it is that runc incorrectly implements prestart hook. If this is true then this reason is not enough to deprecate it from my point of view. Bugs in one of the implementations shouldn't influence spec change in my opinion.

h-vetinari:
> @bart0sh: The main reason of the deprecation as far as I understand it is that runc incorrectly implements prestart hook. If this is true then this reason is not enough to deprecate it from my point of view. Bugs in one of the implementations shouldn't influence spec change in my opinion.

Well, `runc` is the reference implementation of the spec, which gives this more weight than just "one of the implementations". Furthermore, it would be a pretty invasive change to existing behaviour of `runc` (which itself needs at least a roll-out of the hooks proposed in this PR, to replace those usages that require the current incorrect implementation in the runtime namespace), which would probably mean an extra 1-2 rc's and possibly 6-12 months of transition (just guesstimating here...).

Assuming that, it would then be possible to keep the `prestart` hooks *as originally spec'd* (i.e. during the `start` phase) by going to the following order:
1. `create-runtime`
1. `create-container`
1. `prestart` (= `start-container`; assuming correct(ed) implementation)

But even then, already deployed software (depending on the incorrect implementation) would start failing in strange ways if someone were to upgrade `runc` without knowing about this (and all downstream application authors would need to add complicated version-guards around their hooks).

I honestly think that deprecation is preferable; it's just a cleaner solution all around. And it's not like the spec needs to be cast in stone. **Just bump the version, that's what it's there for.**

cyphar:
@bart0sh The old hooks must be deprecated (that's ignoring the fact that the names weren't very good in the first place -- I'm of half a mind to suggest `org.opencontainers.runtime.hook.*` as the hook names). Sorry, but that's non-negotiable -- the current hooks are never going to be usable because some project won't correctly update their handling or documentation.

Every other OCI runtime I'm aware of has (incorrectly, though understandably) copied `runc`'s incorrect semantics because the widely-used NVIDIA and CNI hooks wouldn't work as-is with the semantics defined in the current spec. As much as I wish this wasn't the case, people blindly copy what `runc` does because the runtime-spec doesn't correctly capture the semantics needed to get an OCI container runtime to be fully functional (don't even get me started on `--console-socket` and how inadequate `"terminal": true` is).

I do wish the situation was better, and that we could just ignore a `runc` bug as being "just a bug in one implementation" (I would've loved to release `runc 1.0` two years ago) but that simply isn't the state of things.

h-vetinari:
@cyphar 
Thanks for the response! Can you give an idea of what's still necessary from your POV to get this PR over the line (re: getting towards `runc 1.0`)?

config.md
```diff
@@ -384,31 +384,82 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+### Summary
+
+See the below table for a summary of hooks and when they are called:
+
+|          Name         | Namespace |                                                            When                                                                   |
+| --------------------- | --------- | ----------------------------------------------------------------------------------------------------------------------------------|
+| prestart (Deprecated) | runtime   | After the start  operation is called but before the user-specified program command is executed                                    |
+| create-runtime        | runtime   | During the create operation, after the runtime environment has been created and before the pivot root or any equivalent operation |
+| create-container      | container | During the create operation after the runtime environment has been created and before the pivot root or any equivalent operation  |
```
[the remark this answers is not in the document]
cyphar: And a missing fullstop for the first three hooks.

config.md
```diff
@@ -384,31 +384,82 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
             If set, `timeout` MUST be greater than zero.
+    * **`create-runtime`** (array of objects, OPTIONAL) is an array of [create-runtime hooks](#create-runtime-hooks).
+        Entries in the array have the same schema as prestart entries.
+    * **`create-container`** (array of objects, OPTIONAL) is an array of [create-container hooks](#create-container-hooks).
+        Entries in the array have the same schema as prestart entries.
+    * **`start-container`** (array of objects, OPTIONAL) is an array of [start-container hooks](#start-container-hooks).
+        Entries in the array have the same schema as prestart entries.
     * **`poststart`** (array of objects, OPTIONAL) is an array of [post-start hooks](#poststart).
-        Entries in the array have the same schema as pre-start entries.
+        Entries in the array have the same schema as prestart entries.
```
cyphar: nit: "prestart" isn't really a word -- write it as `prestart` (the same applies throughout your changes).

config.md
```diff
@@ -384,31 +384,82 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+
+### <a name="configHooksCreateRuntime" />Create Runtime Hooks
+
+The create-runtime hooks MUST be called as part of the [`create`](runtime.md#create) operation after the runtime environment has been created (according to the configuration in config.json) but before the pivot_root or any equivalent operation has been executed.
+The create-runtime hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+
+On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+Note: the incorrect `runc` implementation of the prestart hooks, maps to a create-runtime hook.
+For runtimes that implement the deprecated prestart hooks as create-runtime hooks, create-runtime hooks MUST be called after the prestart hooks.
```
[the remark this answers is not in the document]
cyphar: I agree with @h-vetinari -- please re-add the `runc` reference.

cyphar:
@h-vetinari Aside from the few nits that I just commented, I'm fine with the bulk of the PR. We just need to get the @opencontainers/runtime-spec-maintainers to review it.

config.md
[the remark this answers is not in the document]
RenaudWasTaken: Done, thanks!

config.md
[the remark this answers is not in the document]
RenaudWasTaken: Done, thanks!

config.md
RenaudWasTaken: Done! I ensure the following:
- All references to post-start and post-stop have been renamed to `poststart` and `poststop`
- All references of prestart, create-runtime, ... have been surrounded with backticks.

h-vetinari:
> @cyphar: We just need to get the @opencontainers/runtime-spec-maintainers to review it.

I'm wondering if there's a better way for this than pinging on Github. This PR's been open for half a year, and none of them have yet responded (except an early CC comment from @vbatts). Would you be able to reach out to them through a different channel?

config.md
```diff
@@ -384,31 +384,80 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+The `create-runtime` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
+
+On Linux, for example, they are called after the container namespaces are created, so they provide an opportunity to customize the container (e.g. the network namespace could be specified in this hook).
+
+Note: `runc` originally implemented `prestart` hooks contrary to the spec, namely as part of the `create` operation (instead of during the `start` operation). This incorrect implementation actually corresponds to `create-runtime` hooks. For runtimes that implement the deprecated `prestart` hooks as `create-runtime` hooks, `create-runtime` hooks MUST be called after the `prestart` hooks.
+
+### <a name="configHooksCreateContainer" />Create Container Hooks
+
+The `create-container` hooks MUST be called as part of the [`create`](runtime.md#create) operation after the runtime environment has been created (according to the configuration in config.json) but before the `pivot_root` or any equivalent operation has been executed.
+The `create-container` hooks MUST be executed in the [container namespace](glossary.md#container-namespace).
```
mrunalp: Who ensures this?

mrunalp: If the hooks are provided access information such as the linux namespaces for the container, they can join the namespaces and do their work. However at this point, `create-container` and `create-runtime` could be the same.

config.md
```diff
@@ -384,31 +384,80 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+The `create-container` hooks MUST be called after the `create-runtime` hooks.
+
+For example, on Linux this would happen before the `pivot_root` operation is executed but after the mount namespace was created and setup.
+
+### <a name="configHooksStartContainer" />Start Container Hooks
+
+The `start-container` hooks MUST be called [before the user-specified process is executed](runtime.md#lifecycle) as part of the [`start`](runtime.md#start) operation.
```
mrunalp: From the description, this sounds like a pre exec hook. I understand the use case but wonder if we can name it something else.

config.md
cyphar: The runtime -- the point is that the runtime should spawn the hook in the container's mount (and other) namespaces *before* performing `pivot_root`.

h-vetinari:
>  @RenaudWasTaken changed the title ~**Add mount and start hooks**~ **Add precreate-host, precreate-container, prestart-container hooks**

Another naming change? At a minimum, the title does not reflect the content of the diff (or vice versa).

Also, thanks @mrunalp for taking a look.

config.md
giuseppe: I still don't get `create-container`.

Is the difference between `create-runtime` and `create-container` just that in the latter case all the user namespaces are joined?

Are the `create-container` hooks supposed to be spawned from the container init process (just before `pivot_root`)?  
I guess no, as at this point the container is not yet fully configured and different runtimes may perform the initialization in a different order.  Conversely, if it is called from the runtime process, it just seem like a special case of `create-runtime` where the runtime also joins the namespaces before exec'ing the hook executable.

cyphar: I don't really see why it's better to punt doing `nsenter` off to the hook authors. Given how trivial it is to implement as part of the runtime, and the fact that LXC has incredibly similar hooks, I don't think adding more annoyances to hook authors is a good idea. Especially if your hooks are incredibly trivial bash scripts, having to mess with nsenter just to do something as simple as create a mount seems excessive.

giuseppe: > It's not required, but that is one way of doing it (and is probably how we'd implement it in `runc`).

At this point the container is not fully configured, what is left to configure and what is already configured is not known to the hook.  Since there is no requirement on what the OCI runtime must do before the `pivot_root` (beside of course setting the mounts), the hook might run from different contexts depending on the OCI runtime used and the container configuration.  It will easily turn into issues like "hook X works with the runtime Y but doesn't work with the runtime Z".

e.g. should cgroups apply to the hook process?  Seccomp?  Resource limits?

config.md
h-vetinari: IIUC, this is indeed what the `prestart` hooks were originally specced to be (but which never got implemented in this way). Due to the deprecation, `prestart` cannot be reused as a name, while the suggested names follow a `<operation>-<namespace>` pattern.

The easiest adaptation might be to re-add the "pre"-prefix, i.e. `pre<operation>-<namespace>`, yielding `precreate-runtime`, `precreate-container` & `prestart-container`, but I'm not sure that's better (since the hook is executed _during_  rather than _before_ `<operation>`). Of course, there's no fundamental reason why this hook could not be named completely differently, just that the bikeshedding has gone on for a long time already. ;-)

config.md
cyphar: Hmm, you have a point. How about we make it `preexec-runtime` and you spawn the hook right before the container's `execve` time (but after the `start` socket has been hit) -- this is kind of the intended use-case for the whole `create`+`start` split but no higher-level runtimes expose that as a feature that you can use for your own extensions (in fact the lack of usability when using higher-level runtimes of the `create`+` start` split is the entire reason for this PR to extend hooks).

But in that case we'd have to ask the hook to deal with joining all of the relevant namespaces -- not to mention that dealing with a container after `pivot_root` isn't really nice if you want to do any operations without touching (**extremely untrusted and dangerous**) container binaries.

RenaudWasTaken:
My bad I didn't intend to rename the hooks, just change the title. Let me re-edit

config.md
h-vetinari: @RenaudWasTaken, thoughts on this? How is it the equivalent hook used in LXC?

vbatts:
@RenaudWasTaken you and I chatted at KubeCon last week. You indicated there was a little cleanup needed, but this conversation is ready to move. What all is left?

config.md
RenaudWasTaken: > It will easily turn into issues like "hook X works with the runtime Y but doesn't work with the runtime Z".

I expect this to happen over time to be honest, runc, gvisor and kata are already pretty different and need different configs to happen.

> Since there is no requirement on what the OCI runtime must do before the pivot_root (beside of course setting the mounts)

This is a problem that is not just specific to the `create-container` but also to the `create-runtime` hook, though the way it is addressed in spec today is in the `runtime.md` file, step 2 reads as follows:
> container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).  While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified program (from [`process`](config.md#process)) MUST NOT be run at this time. Any updates to [`config.json`](config.md) after this step MUST NOT affect the container.

> @RenaudWasTaken, thoughts on this? How is it the equivalent hook used in LXC?

This would be the `mount hook` equivalent of lxc (before pivot_root and in the container namespace), which is the behavior we are trying to mirror here.

h-vetinari: @mrunalp @giuseppe @cyphar @RenaudWasTaken
This thread seems unresolved.

1. Should there be a distinction between `create-runtime` and `create-container` at all?
1. If so, how is it this specified?
1. Should hooks be expected to be compatible across different runtimes at all?

To me, the answer to the last question should be yes (that's why there's a runtime-spec), but maybe this goal is at odds with the kind of granular hook control that some applications need...?

h-vetinari:
Regarding the [thread](https://github.com/opencontainers/runtime-spec/pull/1008#discussion_r352361390) that's still open - maybe this needs a pragmatic way forward, in the sense that this PR is ~95% of what's necessary to resolve the hook ordering problem, and the last 5% could be fine-tuned in future spec-runc commits (or versions).

I say this because in [58 days](https://daycalc.appspot.com/02/02/2020), the [original bug](https://github.com/opencontainers/runc/issues/1710) about the hook ordering turns two years old (the only blocker for `runc 1.0.0`), and it seems everyone involved is burnt out from the infinite slog towards `1.0.0` (or from other work).

So maybe perfect is the enemy of good here, and this should just go forward, get tried out by a runc-rc (and maybe other runtimes; kata, crun, ...), and then refined from there?

config.md
cyphar: The answer to the last question most definitely is yes. I think that @giuseppe has a valid point -- this is arguably why we created the whole `create`+`start` split to begin with (it allows you to do hook-like operations without significantly constraining the runtime implementation). But no higher-level runtime exposes the `create`+`start` split properly.

And so we need to use hooks to achieve the same thing. I think that both approaches have downsides:

* Making the hook effectively act as a hook for "post-create" means that the container will have already `pivot_root`ed, making it harder to implement hooks reasonably. There's almost certainly many possible security bugs you could spawn by asking third-party hooks to safely interact with a fully-set-up container rootfs. To be fair, those worries also exist with this PR's hooks -- but by running them before `pivot_root` iti sn't as bad.

* The approach in this PR (as @giuseppe says) results in more and more problems where "runc is the standard". While I would like to avoid this, at this point it's a bit naive to pretend that runc's semantics have not already been implemented by all other runtimes. Maybe the solution here is to just define these semantics in the spec, and move on.

Anyway, I think that (given the security worries I have about "post-create" hooks) we should just go with this PR and see whether the other runtimes have any strong complaints. If nobody screams, then we can just stick with this.

RenaudWasTaken: Sorry, the past two weeks have been hard for getting some work done on this topic. I'll be more available during December :)

> Should there be a distinction between create-runtime and create-container at all?

This is a nice to have as a hooks author, because that means not having to deal with joining the different namespaces (namely the mount one). As pointed out by @cyphar above it would also help avoid security issues if we mess up our implementation.

> If so, how is it this specified?

I was hoping that by leaving the definition vague and defining an example in the runtime.md file it would be enough.
e.g:

`config.md`:
> The `create-runtime` hooks MUST be called as part of the [`create`](runtime.md#create) operation after the runtime environment has been created (according to the configuration in config.json) but before the `pivot_root` or any equivalent operation has been executed.

`runtime.md`
> 1. OCI compliant runtime's [`create`](runtime.md#create) command is invoked with a reference to the location of the bundle and a unique identifier.
> 2. The container's runtime environment MUST be created according to the configuration in [`config.json`](config.md).
> While the resources requested in the [`config.json`](config.md) MUST be created, the user-specified program (from [`process`](config.md#process)) MUST NOT be run at this time.
> ...
> The [`create-runtime` hooks](config.md#create-runtime-hooks) MUST be invoked by the runtime.


> Should hooks be expected to be compatible across different runtimes at all?

Can you detail what you mean by compatible?

As a hooks developer, I have pretty low expectation of my hook working on multiple runtimes without me having to handle different scenarios (for example we need to pass-through the device in Kata before hand if I want to have access to it in the hook). 
I'm more hoping to have a uniformity of the time it is expected to be called.

To be honest, I don't have a strong opinion of either scenario, I can work with both. If `create-container` is blocking this PR, I'm happy to drop it. It's a nice to have, will likely avoid headaches to many people but we can work around it not being here.

h-vetinari: > @RenaudWasTaken: Can you detail what you mean by compatible?

I'm really not the best person to argue this case, but the clear goal of a spec (as reaffirmed by @cyphar) above is that different implementations should be compatible, including the hooks. Of course, the different hooks may depend on each other (in the order that they are guaranteed to be executed; c.f. your device example), but the point is that this should work equally across implementations.

A starting point for open questions might be [this comment](https://github.com/opencontainers/runtime-spec/pull/1008#discussion_r347354422) by @giuseppe:

> e.g. should cgroups apply to the hook process? Seccomp? Resource limits?

giuseppe: could we perhaps just document that there should not be expectations on the status of the container initialization (such as cgroups/seccomp might be already in place) except that the specified mounts are created?

At least hook developers will be aware of such potential issues so that if a specific configuration is needed another kind of hook can be used?

h-vetinari: > @giuseppe: could we perhaps just document that there should not be expectations on the status of the container initialization (such as cgroups/seccomp might be already in place) except that the specified mounts are created?

I think this is a good idea - in fact, I'm wondering if this couldn't/shouldn't be strengthened to be mandated by the spec?

I'll admit my understanding is limited, but to me, the question that @giuseppe raised is basically where `create-container` lives relative to the non-mount container setup:

1. `create` command is invoked
1. runtime environment created according to `config.json`
1. `create-runtime` hook is executed
1. mount points are set up, change into container namespace
1. ??? unspecified order of
    - `create-container` hook is executed
    - various bits of container setup: seccomp, cgroups, etc.
1. `pivot_root`
1. `start` command is invoked
1. `start-container` hook is executed
1. ...

I think that we could hit two birds with one stone by prescribing:
1. `create` command is invoked
1. runtime environment created according to `config.json`
1. `create-runtime` hook is executed
1. mount points are set up, change into container namespace
1. `create-container` hook is executed
1. all other container setup: seccomp, cgroups, etc.
1. `pivot_root`
1. `start` command is invoked
1. `start-container` hook is executed
1. ...

The two birds in question are:
* there's a clear specification that's unambiguous (i.e. only the mount and namespace operations happen between `create-runtime` and `create-container`) - which means the answer should be clear (if something doesn't work), whether it's the hook or the runtime that's not spec-conformant. If a hook relies on some of the other bits of container setup, it must be shifted to `start-container`.
* The spec is minimal in the way that `create-container` hooks written for this spec are not at risk of breaking if a later spec version adds something more between the two hooks, because correct hooks (as per the old spec) would/could not be relying on the additional steps.

config.md
```diff
@@ -384,31 +384,80 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
         * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
         * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
             If set, `timeout` MUST be greater than zero.
-    * **`poststart`** (array of objects, OPTIONAL) is an array of [post-start hooks](#poststart).
-        Entries in the array have the same schema as pre-start entries.
-    * **`poststop`** (array of objects, OPTIONAL) is an array of [post-stop hooks](#poststop).
-        Entries in the array have the same schema as pre-start entries.
+    * **`create-runtime`** (array of objects, OPTIONAL) is an array of [`create-runtime` hooks](#create-runtime-hooks).
+        Entries in the array have the same schema as `prestart` entries.
```
h-vetinari: I realized while reading the rendered [`config.json`](https://github.com/opencontainers/runtime-spec/blob/master/config.md) that:
* the spec for the new hooks shouldn't build on the structure of a deprecated hook
* the spec for the new hooks _cannot_ reuse the structure verbatim because the definition for **`path`** explicitly says:
    >  Runtimes MUST resolve this value in the [runtime namespace](glossary.md#runtime-namespace).
  
  which - AFAICT - is contradictory to the requirement of `create-container` and `start-container` to be run in the container namespace

runtime.md
```diff
@@ -55,18 +55,24 @@ The lifecycle describes the timeline of events that happen from when a container
...
-6. The [poststart hooks](config.md#poststart) MUST be invoked by the runtime.
-    If any poststart hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
-7. The container process exits.
+3. The [`prestart` hooks](config.md#prestart) MUST be invoked by the runtime.
+    If any `prestart` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+4. The [`create-runtime` hooks](config.md#create-runtime-hooks) MUST be invoked by the runtime.
+    If any `create-runtime` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+5. The [`create-container` hooks](config.md#create-container-hooks) MUST be invoked by the runtime.
+    If any `create-container` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+6. Runtime's [`start`](runtime.md#start) command is invoked with the unique identifier of the container.
```
h-vetinari: Since `pivot_root` features prominently in the specification of the `create-*` hooks, I think it should be listed in this life-cycle presentation as well.

config.md
RenaudWasTaken: Good catch!
Though, I think for **`path`**, we could lookup the binary in the runtime namespace (and should?) for the `create-container` hook.

runtime.md
RenaudWasTaken: Will do!

config.md
RenaudWasTaken: The motivation for the create-runtime hook was to specify the behavior that is currently implemented in runc. That behavior [fully sets up the container (including cgroups, though I haven't looked into selinux)](https://github.com/opencontainers/runc/blob/c8337777b666dec6f4f602849b81096983cdac6a/libcontainer/process_linux.go#L375-L385) before invoking the prestart hooks.

I had in my mind the following flow:
1. `create` command is invoked
2. runtime environment created according to config.json
    mount points are set up, as well as other container setup operations: seccomp, cgroups, etc.
3. `create-runtime` hook is invoked
4. `create-container` hook is invoked 
5. `pivot_root` or equivalent operations
6. `start` command is invoked
7. `start-container` hook is executed

As an example, the `mount` hooks in LXC have joined all the namespaces of the container. We could either follow that model or drop the create-runtime hook and leave the work of joining the different namespace to the hook.

Side note, I'm a bit confused as to why we can't specify completing all container setup operations "except pivot_root or an equivalent operation" before the create-container / create-runtime operation.

cyphar: > Side note, I'm a bit confused as to why we can't specify completing all container setup operations "except pivot_root or an equivalent operation" before the create-container / create-runtime operation.

There are some operations which have to be done as late as possible -- much later than `pivot_root` (such as seccomp application or setting SELinux/AppArmor labels). It's better to not permit containers to `pivot_root`.

h-vetinari: > @giuseppe: could we perhaps just document that there should not be expectations on the status of the container initialization (such as cgroups/seccomp might be already in place) except that the specified mounts are created?

@cyphar @mrunalp @vbatts 
What do you think about @giuseppe's suggestion? That would probably be the fastest way forward. Or should the spec be more specific?

cyphar: I think to solve this particular blocking problem we should just go with @giuseppe's suggestion to make hook authors aware that this is an underspecified area of the spec and the only guarantees they can be sure of are that the mounts will be set up and `pivot_root` will not have been called yet.

In the future we can work on making the spec more specific.

runtime.md
RenaudWasTaken: I realized this is something we probably don't want to specify. In case runtimes don't execute `pivot_root` as part of the `create` command, we are better off not indicating when runtimes should execute specific operations (or the equivalent).

RenaudWasTaken:
Summary of the latest update, which I believe resolves the last comments:
1. Updated the hooks argument definition as follows:
    - create-* start-*, poststart and poststop no longer rely on the same entries as the prestart entries [discussion happened here](https://github.com/opencontainers/runtime-spec/pull/1008#discussion_r355720003)
    - create-runtime was updated to include a copy of the entries of the `prestart` hooks.
    - create-container, poststart and poststop now use the sentence `Entries in the array have the same schema as create-runtime entries.`
    - create-container contains a note: `Note that even if the hook will be executed in the container namespace, the `path` should resolve in the runtime namespace.`
    - start-container now uses a slightly different definition (with the only difference being the entry `path` changing to `path-container`). The json definition was updated accordingly:
```
        * **`path-container`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2008 `execv`'s *path*][ieee-1003.1-2008-functions-exec].
            This specification extends the IEEE standard in that **`path-container`** MUST be absolute.
            Runtimes MUST resolve this value in the [container namespace](glossary.md#container-namespace).
...
```

I considered updating `create-container`, `poststart` and `poststop` to have a definition where `path` would become `path-runtime`, but this would force us to deprecate the `path` field of `poststart` and `poststop` and create a mess of arguments accepted the different hooks.
e.g: `prestart` accepts a `path`, `poststart` and `poststop` accept both `path` and `path-runtime` and `create-*` only accept `path-runtime`.

We could also keep the definition the same and not change that entry, this would probably simplify the definition, but make things a bit more error prone.

2. Update to the definition of `create-container` and `create-runtime`, I've added the following definition under both hooks:
```
The definition of `create-runtime` hooks is currently underspecified and hooks authors, should only expect from the runtime that the mount namespace and different mounts will be setup. Other operations such as cgroups and SELinux/AppArmor labels might not have been performed by the runtime.
```

runtime.md
RenaudWasTaken: @h-vetinari I'm keeping this conversation open, If you have a specific idea it's probably worth discussing !

config.md
RenaudWasTaken: Updated the PR to echo these thoughts :)

h-vetinari:
@RenaudWasTaken 

Something is upsetting Travis:
```
[...]
  testing test/config/good/spec-example.json
- hooks.startContainer.0: path is required
    received unexpected validation failure
Makefile:22: recipe for target 'test' failed
make: *** [test] Error 1
make: Leaving directory '/home/travis/gopath/src/github.com/opencontainers/runtime-spec/schema'
The command "make -C schema test" exited with 2.
```

runtime.md
h-vetinari: It's OK if `pivot_root` can not be reflected accurately in this very linear lifecycle description (as I naively thought); feel free to resolve this comment.

RenaudWasTaken:
@h-vetinari thanks for reviewing! I fixed the last issue but it looks like master is failing now!

giuseppe:
> it looks like master is failing now!

it is fixed by https://github.com/opencontainers/runtime-spec/pull/1024

vbatts:
which is merged now. You can rebase.

RenaudWasTaken:
CI is green!

h-vetinari:
Time for the next (final...?) round of reviews?
@alban @amshinde @crosbymichael @cyphar @giuseppe @mrunalp @vbatts 

Maybe we can start off 2020 with a new runtime-spec version? =)

config.md
```diff
@@ -387,31 +387,102 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
...
+        * **`args`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008 `execv`'s *argv*][ieee-1003.1-2008-functions-exec].
+        * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
+        * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
+            If set, `timeout` MUST be greater than zero.
+    * **`create-container`** (array of objects, OPTIONAL) is an array of [`create-container` hooks](#create-container-hooks).
+        Entries in the array have the same schema as `create-runtime` entries.
+        Note that even if the hook will be executed in the container namespace, the `path` should resolve in the runtime namespace.
+    * **`start-container`** (array of objects, OPTIONAL) is an array of [`start-container` hooks](#start-container-hooks).
+        Entries in the array contain the following properties:
+        * **`pathContainer`** (string, REQUIRED) with similar semantics to [IEEE Std 1003.1-2008 `execv`'s *path*][ieee-1003.1-2008-functions-exec].
```
giuseppe: is there any advantage in calling it `pathContainer` instead of `path`?  It would avoid a new type as everything else seems the same as with the other hooks

RenaudWasTaken: It makes it clearer in the API where the path is expected to be resolved. Since this is the only hook that is expected to be resolved in the container, I thought it made sense to make an explicit separation.

giuseppe: I'd still prefer to simplify it and reuse the existing types where possible.  IMO they have the same semantic, the only difference is the mount namespace used

RenaudWasTaken: Addressed in the latest update

RenaudWasTaken:
In the latest update, the following changes were made:
- `create-container` has now the same structure as the other hooks (before it had a field called `pathContainer` rather than `path`)
- I specified for each hook in what namespace the `path` value should be resolved
- I specified for each hook in what namespace it should be executed

runtime.md
```diff
@@ -55,18 +55,24 @@ The lifecycle describes the timeline of events that happen from when a container
...
-4. The [prestart hooks](config.md#prestart) MUST be invoked by the runtime.
-    If any prestart hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 9.
-5. The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
-6. The [poststart hooks](config.md#poststart) MUST be invoked by the runtime.
-    If any poststart hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
-7. The container process exits.
+3. The [`prestart` hooks](config.md#prestart) MUST be invoked by the runtime.
+    If any `prestart` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
+4. The [`create-runtime` hooks](config.md#create-runtime-hooks) MUST be invoked by the runtime.
+    If any `create-runtime` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
```
giuseppe: should `create-runtime` and `create-container` be specified in camel case as well as in `config.md`?  At least the schema and the Go code seems to use the camel case version.

schema/defs.json
```diff
@@ -108,6 +108,12 @@
                 "$ref": "#/definitions/Hook"
             }
         },
+        "ArrayOfContainerHooks": {
```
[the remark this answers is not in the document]
RenaudWasTaken: Addressed, thanks for noticing!

specs-go/config.go
```diff
@@ -125,13 +125,26 @@ type Hook struct {
 	Timeout *int     `json:"timeout,omitempty"`
 }
 
+// Hook specifies a command that is run in the container at a particular event in the lifecycle of a container
```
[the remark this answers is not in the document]
RenaudWasTaken: Addressed, thanks for noticing!

runtime.md
RenaudWasTaken: Addressed!

cyphar:
LGTM from my side.

crosbymichael:
LGTM

h-vetinari:
I'm quite sure no-one wants to receive a ping from this thread ever again, but since some parts of the new hooks were left unspecified, I opened a follow-up issue to track things: #1039.

PR #1011, opened 2019-06-03
config-linux: add more info about hugetlb page size

odinuge:
Currently the docs don't say anything about what the "pageSize" is other
than the fact that it is a string. This makes it easier for developers
to understand how it works, and may help avoiding mistakes which are
hard to spot.

This can help other doing the same mistake as `runc` and k8s when using hugepages, used `<size>kB` instead of the correct `<size>KB`. (https://github.com/opencontainers/runc/pull/2065)

Signed-off-by: Odin Ugedal <odin@ugedal.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -394,6 +394,9 @@ For more information, see the kernel cgroups documentation about [HugeTLB][cgrou
 Each entry has the following structure:
 
 * **`pageSize`** *(string, REQUIRED)* - hugepage size
+    The value has the format "<size><unit-prefix>B" (64KB, 2MB, 1GB), and must match the <hugepagesize> of the
+    corresponding control file found in "/sys/fs/cgroup/hugetlb/hugetlb.<hugepagesize>.limit_in_bytes".
+    Values of <unit-prefix> are intended to be parsed using base 1024 ("1KB" = 1024, "1MB" = 1048576, etc).
 * **`limit`** *(uint64, REQUIRED)* - limit in bytes of *hugepagesize* HugeTLB usage
 
 #### Example
@@ -403,6 +406,10 @@ Each entry has the following structure:
         {
             "pageSize": "2MB",
             "limit": 209715200
+        },
+        {
+            "pageSize": "64KB",
+            "limit": 1000000
         }
    ]
 ```
--- a/config.md
+++ b/config.md
@@ -710,6 +710,10 @@ Here is a full example `config.json` for reference.
                 {
                     "pageSize": "2MB",
                     "limit": 9223372036854772000
+                },
+                {
+                    "pageSize": "64KB",
+                    "limit": 1000000
                 }
             ],
             "memory": {
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -124,7 +124,8 @@
                             "type": "object",
                             "properties": {
                                 "pageSize": {
-                                    "type": "string"
+                                    "type": "string",
+                                    "pattern": "^[1-9][0-9]*[KMG]B$"
                                 },
                                 "limit": {
                                     "$ref": "defs.json#/definitions/uint64"
--- a/schema/test/config/bad/linux-hugepage.json
+++ b/schema/test/config/bad/linux-hugepage.json
@@ -1,0 +1,16 @@
+{
+    "ociVersion": "1.0.0",
+    "root": {
+        "path": "rootfs"
+    },
+    "linux": {
+        "resources": {
+            "hugepageLimits": [
+                {
+                    "limit": 1234123,
+                    "pageSize": "64kB"
+                }
+            ]
+        }
+    }
+}
--- a/schema/test/config/good/spec-example.json
+++ b/schema/test/config/good/spec-example.json
@@ -232,6 +232,10 @@
                 {
                     "pageSize": "2MB",
                     "limit": 9223372036854772000
+                },
+                {
+                    "pageSize": "64KB",
+                    "limit": 1000000
                 }
             ],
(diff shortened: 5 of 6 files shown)


crosbymichael:
LGTM

config-linux.md
```diff
@@ -394,6 +394,8 @@ For more information, see the kernel cgroups documentation about [HugeTLB][cgrou
 Each entry has the following structure:
 
 * **`pageSize`** *(string, REQUIRED)* - hugepage size
+    The value has the format "<size><unit-prefix>B" (64KB, 2MB, 1GB), and must match the <hugepagesize> of the
+    corresponding control file found in "/sys/fs/cgroup/hugetlb/hugetlb.<hugepagesize>.limit_in_bytes"
```
tianon: Should we also specify explicitly that these are intended to be 1024-based units?

I think https://github.com/opencontainers/runc/blob/e4fa8a457544ca646e02e60d124aebb0bb7f52ad/libcontainer/cgroups/utils.go#L411-L429 is also related -- the kernel appears to use a lowercase `k` but uppercase for the rest?

odinuge: Hi @tianon! :smile: 

> Should we also specify explicitly that these are intended to be 1024-based units?

Smart! Any suggestion about how to write it?

> I think https://github.com/opencontainers/runc/blob/e4fa8a457544ca646e02e60d124aebb0bb7f52ad/libcontainer/cgroups/utils.go#L411-L429 is also related -- the kernel appears to use a lowercase k but uppercase for the rest?

The kernel does indeed use "KB" (https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/mm/hugetlb_cgroup.c?h=v5.0#n349), and that has been the case from day 1. To avoid the mistake, I posted a patch adding a note about it in the kernel docs too. https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/cgroup-v1/hugetlb.txt?h=for-5.2-fixes It was applied to the cgroup-tree, and will most likely be a part of the `5.2` release :smile: 

I have a PR in runc to fix it here: https://github.com/opencontainers/runc/pull/2065, together with this PR https://github.com/kubernetes/kubernetes/pull/78495 in k8s, blocked by runc. :smile: Both PRs have some background info about the issue. Even `containerd` ~use `kB`~, patched here: https://github.com/containerd/cgroups/pull/85

odinuge: Also, I think the lowercase `k` is simply a result of copy-paste from [docker/go-units](https://github.com/docker/go-units/blob/master/size.go#L37) into runc, and further into k8s etc.

tianon: I don't think it is -- when I do an `ls` on the path that bit of code is looking at, I get lowercase `k` values:

```console
$ ls /sys/kernel/mm/hugepages/
hugepages-1048576kB  hugepages-2048kB
```

> Smart! Any suggestion about how to write it?

Technically, what we're talking about is [officially "MiB", etc](https://en.wikipedia.org/wiki/Megabyte), but still commonly used both ways (which is why I figure it's probably worth calling out, since `MB` is still considered ambiguous). 
 Here's the best idea I've got, maybe someone else can do / adapt this better:

> Values of `<unit-prefix>` are intended to be parsed using base 1024 ("1KB" = 1024, "1MB" = 1048576, etc).

odinuge: > I don't think it is -- when I do an ls on the path that bit of code is looking at, I get lowercase k values:

You are looking at the normal kernel memory section (idk. what they are calling it). There, all the sizes are rendered with `kB` (yes, lowercase). The `cgroup` control files (v1) we are talking about, are  located in `/sys/fs/cgroup/hugetlb/`. When setting the huge page limits we use the cgroup control files for the given container (further down in the `/sys/fs/cgroup/hugetlb` dir), and not the files inside `/sys/kernel/mm/hugepages/`.

I totally agree that this is strange, and that it would be easier to just use the same format both places. Maybe they does something else when they implement huge pages in cgroup v2 (I dont think they have yet)..

Screenshot from 2019-06-03 21-45-09

> Technically, what we're talking about is officially "MiB", etc, but still commonly used both ways (which is why I figure it's probably worth calling out, since MB is still considered ambiguous).
Here's the best idea I've got, maybe someone else can do / adapt this better:

Yee. MiB is the correct. When dealing with memory, the kernel people (and more or less all of us) always use `1 << 10`, `1 << 20` etc. when talking about sizes, and in my opinion they (the kernel people) do a lot of strange and inconsistent stuff when naming these things. They never (almost never) break userspace, so I guess that is the reason stuff like this is kept the way it is.

> Values of <unit-prefix> are intended to be parsed using base 1024 ("1KB" = 1024, "1MB" = 1048576, etc).

Nice! :smile:

tianon: Interesting, thanks for the additional background!

odinuge: This is where the `pageSize` is used to write to the cgroup control file: https://github.com/opencontainers/runc/blob/e4fa8a457544ca646e02e60d124aebb0bb7f52ad/libcontainer/cgroups/fs/hugetlb.go#L31

odinuge: Updated the description here according to the discussion now @tianon! 😄

tianon:
LGTM

:+1:

odinuge:
CC: @crosbymichael

crosbymichael:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/runtime-spec/pull-request/1011/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/runtime-spec)

PR #1012, opened 2019-06-03
Add Linux personality support

justincormack:
A lot of people use the Linux `personality` support to allow a 64 bit machine
to emulate a 32 bit machine. In particular if you just run 32 bit binaries, many
build processes will fail as `uname` will still return a value appropriate for
a 64 bit system. Including the personality syscall wil change this to reflect the
value from a 32 bit system, such as `i686` rather than `x86_64`.

Note that this patch only supports the base 32 bit/64 bit calls. The other options
are largely obsolete and rarely used. I left flexibility to add other base domains
and to add flags in future, but I am not sure there is any demand for them. The
only use case I found in the recent past was the `ADDR_NO_RANDOMIZE` option that
disables ASLR, which older versions of Emacs required, but generally they set this
themselves, so it is not needed as a Runc option, and it is a serious security
reduction. The 32 bit option is different as if you are running 32 bit containers
for build, they generally do not know they are "supposed" to run 32 bit, and so
this option allows you do do the equivalent of running a `chroot` with `linux32`
as is often done on non containerised build systems.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -684,6 +684,23 @@ The following parameters can be specified to set up seccomp:
     "mountLabel": "system_u:object_r:svirt_sandbox_file_t:s0:c715,c811"
 ```
 
+## <a name="configLinuxPersonality" />Personality
+
+**`personality`** (object, OPTIONAL) sets the Linux execution personality. For more information
+see the [personality](personality.2) syscall documentation. As most of the options are
+obsolete and rarely used, and some reduce security, the currently supported set is a small
+subset of the available options.
+
+* **`domain`** *(string, REQUIRED)* - the execution domain.
+    The valid list of constants is shown below. `LINUX32` will set the `uname` system call to show
+    a 32 bit CPU type, such as `i686`.
+
+    * `LINUX`
+    * `LINUX32`
+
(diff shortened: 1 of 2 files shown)


justincormack:
cc @mattspencer-arm

m5p3nc3r:
@justincormack I have reviewed this with our internal team, and we agree that this should be the correct solution to the root problem.  Thanks for pulling this together.

mattspencer-arm:
hey @justincormack , I have reviewed this with the internal team here and we agree this should be the correct solution to the problem.  Thanks for pulling this together.

crosbymichael:
LGTM

vbatts:
LGTM

As this introduces an `OPTIONAL` struct it is forward compatible.

giuseppe:
the schema file is not updated

vbatts:
Ah you're right. I'm driving home now. Would you open an issue?

giuseppe:
opened here: https://github.com/opencontainers/runtime-spec/issues/1023

giuseppe:
and a PR: https://github.com/opencontainers/runtime-spec/pull/1024

vbatts:
Thank you!

PR #1018, opened 2019-09-09
config-linux: support seccomp flags

giuseppe:
allow to specify what flags must be passed to seccomp(2) when
installing the filter.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -586,6 +586,14 @@ The following parameters can be specified to set up seccomp:
     * `SCMP_ARCH_PARISC`
     * `SCMP_ARCH_PARISC64`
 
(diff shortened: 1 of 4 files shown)


giuseppe:
CC @mrunalp @vbatts

rhatdan:
@caniszczyk @crosbymichael @dqminh @hqhq @mrunalp @philips @tianon @vbatts @vishh PTAL

A couple of reasons for these changes:

We are working on generating seccomp rules per container, and would like to be able to see missing syscalls in the audit.log. 

People are not using SECCOMP separation, because it hurts performance by turning on the spectre/meltdown mitigations by default.  People who want performance could still use seccomp if we did not turn on the mitigation by default.

crosbymichael:
LGTM

hqhq:
LGTM

PR #1038, opened 2020-03-30
Add the specification of seccomp userspace notification

KentaTada:
This commit adds the specification of seccomp userspace notification.

The runtime will pass the file descriptor of seccomp notification via the UNIX domain socket using SCM_RIGHT to **"listenerPath"**.
The reference link is below.
https://www.kernel.org/doc/html/v5.0/userspace-api/seccomp_filter.html#userspace-notification

I want to share our background. As Sony, we are building our rootless embedded container system. We need to hook some syscalls like mount(2) or mknod(2) inside our container and handle those syscalls in userspace in accordance with our use case.
I think this specification is useful for all container users, especially rootless container users.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -624,6 +624,9 @@ The following parameters can be specified to set up seccomp:
     * `SECCOMP_FILTER_FLAG_TSYNC`
     * `SECCOMP_FILTER_FLAG_LOG`
     * `SECCOMP_FILTER_FLAG_SPEC_ALLOW`
+    * `SECCOMP_FILTER_FLAG_NEW_LISTENER`
+
+* **`listenerPath`** *(string, OPTIONAL)* - specifies the path of UNIX domain socket which the runtime will pass the file descriptor of seccomp notification using SCM_RIGHT to.
 
 * **`syscalls`** *(array of objects, OPTIONAL)* - match a syscall in seccomp.
     While this property is OPTIONAL, some values of `defaultAction` are not useful without `syscalls` entries.
@@ -633,7 +636,7 @@ The following parameters can be specified to set up seccomp:
     * **`names`** *(array of strings, REQUIRED)* - the names of the syscalls.
         `names` MUST contain at least one entry.
     * **`action`** *(string, REQUIRED)* - the action for seccomp rules.
-        A valid list of constants as of libseccomp v2.4.0 is shown below.
+        A valid list of constants is shown below.
 
         * `SCMP_ACT_KILL`
         * `SCMP_ACT_KILL_PROCESS`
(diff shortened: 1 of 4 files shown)


KentaTada:
PTAL
@crosbymichael @mrunalp @vbatts @dqminh @tianon @hqhq

config-linux.md
```diff
@@ -598,6 +598,9 @@ The following parameters can be specified to set up seccomp:
     * `SECCOMP_FILTER_FLAG_TSYNC`
     * `SECCOMP_FILTER_FLAG_LOG`
     * `SECCOMP_FILTER_FLAG_SPEC_ALLOW`
+    * `SECCOMP_FILTER_FLAG_NEW_LISTENER`
+
+* **`listenerPath`** *(string, OPTIONAL)* - specifies the path of UNIX domain socket which the runtime will pass the file descriptor of seccomp notification using SCM_RIGHT to.
```
AkihiroSuda, line 603: We may want to prioritize the DLL interface rather than socket. https://github.com/containers/crun/pull/438

cc @giuseppe

giuseppe: do you think that should be part of the specs?  I considered that more of a development tool.  How would the .so handled?  Through cgo?

alban:
Hi @KentaTada. I started a similar [proposal](https://docs.google.com/document/d/1xHw5GQjMj6ZKR-40aKmTWZRkvlPuzMGQRu-YpOFQc30/edit) and we will discuss this topic at the [OCI Weekly Discussion](https://hackmd.io/El8Dd2xrTlCaCG59ns5cwg) tomorrow. Maybe you can join? Sorry I hadn't noticed this PR before.

KentaTada:
> Sorry I hadn't noticed this PR before.

No problem. Thank you for your contact.
I'm sorry but I can't attend the meeting due to personal reasons.
In addition to that, I cannot explain the detailed mechanism of our proprietary container runtime currently.
To put it simply, I implemented it in our container runtime as same as crun.
I think we just only need "listenerPath" like crun as runtime-spec.

BTW, I'll take paternity leaves sometimes and my reply may be late.
@ManaSugi takes over our container runtime and he will reply to the question when I am not available.

config-linux.md
KentaTada: I'm sorry for the very late reply.
DLL interface is useful but I think we don' t need it as runtime-spec as @giuseppe said.

KentaTada:
Just rebased

rata:
@KentaTada I've just opened this PR: https://github.com/opencontainers/runtime-spec/pull/1074 that also overlaps.

It is quite similar to this, but it specifies: how the UNIX socket should be used, and more importantly for our use case a seccomp state to be sent over the socket, so the agent can have information about the container (like the pod it is, etc.).

IIUC, it should allow for your use case too. What do you think? Please feel free to review it :)

rata:
PR #1074  added support for this (is already merged). I think it is safe to close this one :)

Issue #1002, opened 2019-03-05
Tracker for cgroups v2


vbatts:
Cgroups unified hierarchy has been on the horizon for a while, and is now reaching critical mass to begin switching over. Most all the controllers or equivalent are there, but there are mappings needed.

This issue should serve as the tracker for remaining issues to close this gap.

- [x] ensure ebpf for devices maps to existing configs (device cgroup had a minor incompatibility (blacklisting isn't really possible))
- [x] freezer controller (expected in linux 5.2)
- [x] cgroupfs and/or systemd API
- [ ] There's no longer separate limits for kmem and kmem.tcp
- [ ] There's no longer a file to set "swappiness" for a single cgroup hierarchy
- [ ] There's no longer a way to disable OOMs through the memcg

Related https://github.com/opencontainers/runtime-spec/issues/948

vbatts:
cc @filbranden for several of these are his findings

AkihiroSuda:
@giuseppe are you willing to propose the crun conversion spec to OCI Runtime Spec?
https://github.com/giuseppe/crun/blob/master/crun.1.md#cgroup-v2

giuseppe:
> @giuseppe are you willing to propose the crun conversion spec to OCI Runtime Spec?
> https://github.com/giuseppe/crun/blob/master/crun.1.md#cgroup-v2

I can.  I was already discussing them with @filbranden, @vbatts and @mrunalp.

I wonder if someone else beside me played with it though :-)

h-vetinari:
@giuseppe, do you mind if I send a PR with the cgroups conversion specs from crun? Might still have a chance to squeeze in such non-invasive changes before the 1.0.2 release...

giuseppe:
> @giuseppe, do you mind if I send a PR with the cgroups conversion specs from crun? Might still have a chance to squeeze in such non-invasive changes before the 1.0.2 release...

sure, I've no problems with that

h-vetinari:
> @h-vetinari: @giuseppe, do you mind if I send a PR with the cgroups conversion specs from crun? Might still have a chance to squeeze in such non-invasive changes before the 1.0.2 release...

For the record, I had less time than expected in February, but I tried once, and found the undertaking more complex than expected (hence no PR).

[...]

PR #1040, opened 2020-04-13
cgroup: add cgroup v2 support

giuseppe:
allow users to specify cgroup v2 resources.

Each element in the map refers to a file in the cgroup v2 hierarchy and the element value has its content.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -494,6 +494,31 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
 }
 ```
 
+## <a name="configLinuxUnified" />Unified
+
+**`unified`** (object, OPTIONAL) allows cgroup v2 parameters to be to be set and modified for the container.
+
+Each key in the map refers to a file in the cgroup unified hierarchy.
+
+The OCI runtime MUST ensure that the needed cgroup controllers are enabled for the cgroup.
+
+Configuration unknown to the runtime MUST still be written to the relevant file.
+
+The runtime MUST generate an error when the configuration refers to a cgroup controller that is not present or that cannot be enabled.
+
+### Example
+
+```json
+"unified": {
+    "io.max": "259:0 rbps=2097152 wiops=120\n253:0 rbps=2097152 wiops=120",
+    "hugetlb.1GB.max": "1073741824"
+}
+```
+
+If a controller is enabled on the cgroup v2 hierarchy but the configuration is provided for the cgroup v1 equivalent controller, the runtime MAY attempt a conversion.
+
+If the conversion is not possible the runtime MUST generate an error.
+
 ## <a name="configLinuxIntelRdt" />IntelRdt
 
 **`intelRdt`** (object, OPTIONAL) represents the [Intel Resource Director Technology][intel-rdt-cat-kernel-interface].
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -34,6 +34,9 @@
             "resources": {
                 "type": "object",
                 "properties": {
+                    "unified": {
+                        "$ref": "defs.json#/definitions/mapStringString"
+                    },
                     "devices": {
                         "type": "array",
                         "items": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -372,6 +372,8 @@ type LinuxResources struct {
 	// Limits are a set of key value pairs that define RDMA resource limits,
 	// where the key is device name and value is resource limits.
 	Rdma map[string]LinuxRdma `json:"rdma,omitempty"`
+	// Unified resources.
+	Unified map[string]string `json:"unified,omitempty"`
 }
 
 // LinuxDevice represents the mknod information for a Linux special device file


giuseppe:
@AkihiroSuda @mrunalp @kolyshkin @vbatts @rhatdan

config-linux.md
```diff
@@ -299,7 +302,8 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
 
 The following parameters can be specified to set up the controller:
 
-* **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup
+* **`weight`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup.  The weight is in the range [1, 10000].
+* **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup.  Its value is in the range [2, 262144].
 * **`quota`** *(int64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
 * **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only)
 * **`realtimeRuntime`** *(int64, OPTIONAL)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources
```
AkihiroSuda, line 309: v1 only?

config-linux.md
```diff
@@ -261,9 +261,12 @@ For more information, see the kernel cgroups documentation about [memory][cgroup
...
 
-* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage
-* **`reservation`** *(int64, OPTIONAL)* - sets soft limit of memory usage
+* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage.  It maps to `memory.max` on cgroup v2
+* **`protection`** *(int64, OPTIONAL)* - sets hard memory protection.  It maps to `memory.min` on cgroup v2
+* **`reservation`** *(int64, OPTIONAL)* - sets soft limit of memory usage.  It maps to `memory.low` on cgroup v2
+* **`softLimit`** *(int64, OPTIONAL)* - sets soft memory limit.  It maps to `memory.high` on cgroup v2
 * **`swap`** *(int64, OPTIONAL)* - sets limit of memory+Swap usage
+* **`swapOnly`** *(int64, OPTIONAL)* - sets limit of swap usage.  It can be used only on cgroup v2.
 * **`kernel`** *(int64, OPTIONAL)* - sets hard limit for kernel memory
```
AkihiroSuda, line 270: v1 only?

config-linux.md
```diff
@@ -261,9 +261,12 @@ For more information, see the kernel cgroups documentation about [memory][cgroup
 
 Values for memory specify the limit in bytes, or `-1` for unlimited memory.
 
-* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage
-* **`reservation`** *(int64, OPTIONAL)* - sets soft limit of memory usage
+* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage.  It maps to `memory.max` on cgroup v2
```
AkihiroSuda, line 264: Let's show the v1 mappings explicitly as well

AkihiroSuda: (Maybe in tables)

giuseppe: should I copy the table from https://github.com/containers/crun/blob/master/crun.1.md#cgroup-v2 (and leave only the ones that have a conversion to a different range)?

giuseppe: I've specified the file name on cgroup v1 as well.

config-linux.md
giuseppe: sure, added

config-linux.md
giuseppe: added

giuseppe:
@filbranden

mrunalp:
I think we need a way to tell the container runtime whether to use v1 or v2. It could be an optional field `cgroupsVersion`. One possible way to interpret:
* If not set, assume v1 and convert to v2 values if running on a v2 system.
* If set to v2, then don't perform any conversions and directly set the v2 values.

h-vetinari:
> @mrunalp: I think we need a way to tell the container runtime whether to use v1 or v2. It could be an optional field cgroupsVersion. One possible way to interpret:

I think that's a good idea (see similar [thoughts](https://github.com/opencontainers/runtime-spec/issues/1002#issuecomment-612907538) in #1002)

giuseppe:
> I think we need a way to tell the container runtime whether to use v1 or v2. It could be an optional field `cgroupsVersion`. One possible way to interpret:
> 
>     * If not set, assume v1 and convert to v2 values when running on a v2 system.
> 
>     * If set to v2, then don't perform any conversions and directly set the v2 values.

I was convinced about having a way to tell whether it is cgroup v1 or cgroup v2, but since also runc has adopted implicit conversions now, there wouldn't be any difference in what the OCI runtime does.

The conversions are performed for properties that are only on cgroup v1, e.g. `shares`.  On cgroup v2 that should not be used, but instead both the OCI runtime and the container engine should use `cpu.weight`.

h-vetinari:
> @giuseppe: I was convinced about having a way to tell whether it is cgroup v1 or cgroup v2, but since also runc has adopted implicit conversions now, there wouldn't be any difference in what the OCI runtime does.

Isn't that putting the cart before the horse? Yes, runc has already implemented cgroups v2 in the absence of more details from the specification, but it seems like the runtime-spec should make the "right" choice independently from what runc has done (and of course: runc would be able to follow what the spec decides to mandate).

giuseppe:
> Isn't that putting the cart before the horse? Yes, runc has already implemented cgroups v2 in the absence of more details from the specification, but it seems like the runtime-spec should make the "right" choice independently from what runc has done (and of course: runc would be able to follow what the spec decides to mandate).

yes, I am for the "right" choice.  Just what would you do differently once you know what mode to use?  It is like duplicating the information we already have, if you use `share` on cgroup v2, then it is better to do the conversion when running on cgroup v2, differently if `weight` is used then that value is used without any conversion

h-vetinari:
> @giuseppe: Just what would you do differently once you know what mode to use?

Well, for one thing, runtimes (other than runc, where the established behaviour creates compatibility expectations) could choose to warn/error if the specified fields are not compatible with the given `cgroupsVersion` (in other words, your `share`-vs-`cpu.weight` example, IIUC).

Something very roughly like:
> possible spec-formulation: *[...] runtimes MAY choose to raise an error if any of the given fields doesn't match the specified `cgroupsVersion`. Alternatively, runtimes MAY warn only and proceed by converting the values as appropriate according to the following table [...]*

Runtimes might even choose to warn/error if using a v2-specified container on a v1-host (or vice versa) - however, this approach would likely be too harsh, as then containers with prespecified limits wouldn't be compatible across hosts with different cgroup versions anymore...

In any case, I think `cgroupsVersion` would be a good piece of information to have in a specification (whatever the runtime chooses to do).

crosbymichael:
Since v2 will be the future, should we map fields better to the v2 naming more than what we are doing here?  "low|high|max" for memory, etc?

mrunalp:
> Since v2 will be the future, should we map fields better to the v2 naming more than what we are doing here? "low|high|max" for memory, etc?
Yeah, another option is creating new fields that map better to v2 and the either root the cgroup heirarchy at v1 or v2.

giuseppe:
> In any case, I think `cgroupsVersion` would be a good piece of information to have in a specification (whatever the runtime chooses to do).

one thing to keep in mind is that v1 vs v2 is really about each controller.  The kernel allows to use different controllers (what is known as the hybrid mode: https://systemd.io/CGROUP_DELEGATION/#three-different-tree-setups-).  While both runc and crun decided to not support it, so it is either cgroup v1 or cgroup v2, I don't think the OCI specs should deny this configuration.

So for OCI conformance, I think it should be fine to choose either `cpu.shares` or `cpu.weight` (but not both), then it is left to the OCI runtime what to do in this case.  Perhaps we can expect that OCI runtimes MUST support when all controllers are used on the same cgroup v1 version and MAY support controllers on different versions?

giuseppe:
> Since v2 will be the future, should we map fields better to the v2 naming more than what we are doing here? "low|high|max" for memory, etc?

sure, should we also replace existing fields (e.g. the existing `limit` has the same semantic as `max`)?

AkihiroSuda:
Will the "unified" field contain devices?

kolyshkin:
> Instead of replacing fields, could we maybe make a new struct in the resources field, maybe name it "unified"?
> What do you all think?

I like it. Perhaps `"v2"` would be a better name.

crosbymichael:
idk, `v2` seems out of place within the spec as a type name

mrunalp:
> This makes it super simple and backwards compatible. I feel like if we didn't do something that maps well to v2, in a year we will just be frustrated.

What do you all think? >

I like this :+1:

mrunalp:
> Will the "unified" field contain devices?

Yes, I think it should, so unified is self-contained.

AkihiroSuda:
> ```
>  "unified": {
>    "memory.high": 234,
> ...
> ```

I think we should avoid using `.` symbol in the JSON key names. We could use `_` instead, but I feel it is better to keep using separate `memory{}` `cpu{}` objects.

> Yes, I think it should, so unified is self-contained.

Then we will probably end up duplicating huge `devices` objects in the both places for compatibility 🤔

giuseppe:
should we try to avoid using an extra level?  Since cgroup v2 is going to be the default, we will end up with something like:

```
"resources": {
    "unified": {
        "memory.high": 234,
        "memory.max": 555
        "cpu.max": 1
    }
},
```
or:
```
"resources": {
    "unified": {
        "memory" : {
            "high": 234,
            "max": 555
        }
    }
}
```

Could it be a new field at the same level?  Perhaps define a new field `cgroups`?
```
"resources": {
...
},
"cgroups": {
...
},
```

@AkihiroSuda could we just add in the specs that if `cgroups.devices` is missing to look into `resources.devices`?  That should be fine for new implementations

crosbymichael:
I'm leaning towards a flat structure since it is unified on disk as well.

config-linux.md
```diff
@@ -261,11 +261,14 @@ For more information, see the kernel cgroups documentation about [memory][cgroup
 
 Values for memory specify the limit in bytes, or `-1` for unlimited memory.
 
-* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage
-* **`reservation`** *(int64, OPTIONAL)* - sets soft limit of memory usage
+* **`limit`** *(int64, OPTIONAL)* - sets limit of memory usage.  It maps to `memory.limit_in_bytes` on cgroup v1 and to `memory.max` on cgroup v2
+* **`protection`** *(int64, OPTIONAL)* - sets hard memory protection.  It maps to `memory.min` on cgroup v2.  It can be used only on cgroup v2.
+* **`reservation`** *(int64, OPTIONAL)* - sets soft limit of memory usage.  It maps to `memory.soft_limit_in_bytes` on cgroup v1 and to `memory.low` on cgroup v2
+* **`softLimit`** *(int64, OPTIONAL)* - sets soft memory limit.  It maps to `memory.high` on cgroup v2.  It can be used only on cgroup v2.
```
kolyshkin, line 267: It might be beneficial to quote kernel docs, saying _this is the main mechanism to control memory usage. If usage goes over the high boundary, the processes of the container are throttled and put under heavy reclaim pressure_ or something like this.

Otherwise people will still use memory.limit which is not the best mechanism for v2.

kolyshkin:
What if we describe cgroups v1 and v2 in a separate sections? Otherwise it's kinda hard to read, since a reader has to always bear in mind which setting applies to which cgroup version. This will also help to retire cgroupv1 in the [hopefully not-so-distant] future.

In cgroup v1 section we can say that in case v2 is used on the host, the runtime SHOULD make effort to transform the settings to v2.

giuseppe:
what about my last comment?  Should we move the cgroup v2 configuration to a separate section like:

```
"resources": {
...
},
"cgroups": {
...
},
```

I'd prefer the cgroup v2 configuration is not a child of cgroup v1.

@cyphar do you have any suggestions?

h-vetinari:
> @crosbymichael: Instead of replacing fields, could we maybe make a new struct in the resources field, maybe name it "unified"? [...] This makes it super simple and backwards compatible. I feel like if we didn't do something that maps well to v2, in a year we will just be frustrated.

> @giuseppe: should we try to avoid using an extra level? [...]
> Could it be a new field at the same level? Perhaps define a new field cgroups?

> @crosbymichael: I'm leaning towards a flat structure since it is unified on disk as well.

I have to say I like @giuseppe's suggestion. This would make it backwards compatible (if not present, assume v1), and easy to switch (even a v2 specced-container would be runnable on a v1 host, with the appropriate reverse transformations, as long no v2-only features are used). And of course, the spec can mandate warnings/errors if the resource specs don't match the declared `cgroupsVersion`.

And regarding "I feel like if we didn't do something that maps well to v2, in a year we will just be frustrated.": IIUC, if something doesn't get mapped to v2 now, that conversion could then be added later. This would be painless as long as we don't enable "wrong" mappings in either directions that then need to be deprecated.

As an aside, I find "unified" to just be a crutch for the transition, and that, ideally,  it shouldn't be baked into the spec forever.

cyphar:
I think flat `unified` as proposed by @crosbymichael is a reasonable solution, because it's honestly what we should've done for most controllers right from the beginning. Playing catch-up with the kernel (when the kernel API had pretty big mistakes in it -- such as the entire `memory` controller) has been one of the most frustrating things when it comes to runtime-spec development. LXC's config format is also fairly similar in that it represents all of the cgroup keys using the same names as the kernel API -- this avoids adding an extra level of frustration to an existing API. The devices cgroup is a bit of an ugly duckling here (it's the only cgroup which completely changed its entire API model), but the current implicit conversion I think is entirely reasonable -- we never *really* exposed the textual representation of devices cgroupv1.

Also, a flat-structured `unified` actually allows a runtime to just use any new cgroup extensions without the need for us to define them in the specification (and now that all of the weird configuration of cgroupv1 is basically gone in cgroupv2, this would be quite nice). I do see the argument for a structured setup but to be honest it just makes me nervous (especially given the rest of our history with such things).

I will say though that the (necessary -- especially for the devices cgroup) conversions for cgroupv1 <-> cgroupv2 is going to be quite painful. We need to allow specifying both cgroupv1 and cgroupv2 (`memory.low` requires us to do this because you can specify `memory.max` in both versions, but in cgroupv2 you probably only want to specify `memory.low` -- so `memory.max` in cgroupv2 would be `max` which annoyingly users would have to manually specify :shrug:). I guarantee tools will end up breaking this somehow, so I look forward to those bug reports. Also specifying the implicit conversions out the wazoo is going to be at least a little bit of a pain.

I am not a fan of `cgroupVersion` because the cgroup version used is a property of the host system, and not something the container runtime can change or configure. Yes, we can just error out if the version is wrong but this would be the first configuration field that has semantics like this ("error out if the host system isn't using X"). Another issue is "what does it mean to have cgroupv2 enabled" -- does it mean that all the controllers you request have to be enabled in cgroupv2 (if so, what about controllers not yet migrated to cgroupv2)? Or does it just mean that cgroupv2 should be mounted? Or even does it just mean that only kernel support is needed? I think all of those options have their own issues and I don't think it'd be a good idea to deal with that whole shit-show in the specification. 

> I'd prefer the cgroup v2 configuration is not a child of cgroup v1.

Well, `"resources"` was technically never meant to mean just cgroupv1. The initial goal was that it would be agnostic (yeah, looking back now this was a complete pipe dream -- but oh well). I think having it as a child is "better" because it matches the hierarchy of `/sys/fs/cgroup` -- at least under systemd. And it also indicates that you can controllers in both cgroupv1 and cgroupv2 on a system (as frustrating as that is).

h-vetinari:
Thanks for the detailed write up @cyphar.

> I am not a fan of `cgroupVersion` because the cgroup version used is a property of the host system, and not something the container runtime can change or configure.

True. From the spec-PoV, it still makes sense IMO, because it's like a switch that toggles the interpretation of the resource limits that are being passed. And of course, as much as reasonably possible, a v2-specced container should still run on a v1 host with everything transformed correctly. Only where some v2-specific functionality is demanded, the runtime SHOULD raise a warning on a v1 host (possibly continuing on a best-effort basis), or error out.

I get what you mean that the devil is in the details on some of the edge cases, but it would be easy to start with conservative mappings (for those people who feel like they must specify resources in v2 already) back to v1. The rest could error out initially (with no harm done to existing setups in v1) and could be fleshed out in due course.

giuseppe:
thanks for the detailed explanation.

I agree with you on `cgroupVersion` as it doesn't make sense unless we specify a separate version for each controller.  @h-vetinari what more information would it give in addition to look where the cgroup controller is mounted  (e.g. if a memory limit is specified and the memory controller is on cgroup v1, then the OCI runtime has already enough information to apply it)?

I agree on the devices controller as well, in the end we can expose only a subset of the eBPF functionalities.

Do we agree on adding something like the following for cgroup v2?
```
"resources": {
    "unified"" {
        "memory.min": "100000",
        "some.other.file": "100000"
    }
}
```
Then we can document that devices MUST be configured from resources/devices and follow the same semantics.

h-vetinari:
> @giuseppe: what more information would it give in addition to look where the cgroup controller is mounted [...]?

It's possible that I misunderstood your original proposal. My point was about bridging v1<->v2 host vs spec (to the degree that the spec is portable). However (assuming I understand correctly now), if you specify the path to the cgroup controller, then the point becomes moot because then you know whether it's v1 or v2, and can do all the transforms accordingly.

giuseppe:
> It's possible that I misunderstood your original proposal. My point was about bridging v1<->v2 host vs spec (to the degree that the spec is portable). However (assuming I understand correctly now), if you specify the path to the cgroup controller, then the point becomes moot because then you know whether it's v1 or v2, and can do all the transforms accordingly.

yes exactly.

Also, I don't think we can add a new version field to modify the behaviour of existing fields.  OCI runtimes that don't understand the new version field will still apply the configuration as v1.

New features or settings that behave differently (e.g. cpu.shares vs cpu.weight), should either be specified as v1 or have new fields defined.

giuseppe:
can we move forward on this proposal?

@crosbymichael @AkihiroSuda @cyphar what do you think about the format in https://github.com/opencontainers/runtime-spec/pull/1040#issuecomment-621191609 ?

giuseppe:
updated to just use a `map[string]string`

config-linux.md
```diff
@@ -494,6 +494,21 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
 }
 ```
 
+## <a name="configLinuxUnified" />Unified
+
+**`unified`** (object, OPTIONAL) allows cgroup v2 parameters to be modified at runtime for the container.
+
+Each key in the map refers to a file in the cgroup unified hierarchy.  The OCI runtime must ensure that the needed cgroup controllers are enabled for the cgroup.
```
cyphar, line 501: Missing line break at the end of the sentence (I don't like this rule either, but :shrug:).

cyphar: Also, MUST should be capitalised. This should also have some language about how to deal with controllers that are unsupported by the system (or the runtime) -- but also that unknown-to-the-runtime configuration options for known-to-the-runtime controllers SHOULD (or maybe even MUST?) be written to the relevant cgroup file.

And I guess this means we're punting on the whole cgroupv1 conversion issue? Or do you want to tackle that in a separate PR?

AkihiroSuda, line 501:
```markdown
Restrictions:
* A key MUST NOT contain "/".
* A key MUST NOT start with "cgroup.".
```

AkihiroSuda, line 501: > The OCI runtime must ensure that the needed cgroup controllers are enabled for the cgroup.

s/must/SHOULD/ ?
What is the expected behavior if the runtime cannot enable the controller?

giuseppe: I agree on the first one, but why the second?  Wouldn't be useful to set `cgroup.type`?

config-linux.md
```diff
@@ -494,6 +494,21 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
 }
 ```
 
+## <a name="configLinuxUnified" />Unified
+
+**`unified`** (object, OPTIONAL) allows cgroup v2 parameters to be modified at runtime for the container.
```
AkihiroSuda, line 499:
```markdown
On cgroup v2 hosts, if both unified and legacy fields are set, the OCI runtime SHOULD apply both. 
In case of a conflict across the unified field and the legacy fields, the OCI runtime SHOULD error out.
```

config-linux.md
AkihiroSuda: Usecase?

giuseppe: I guess it should error out?

cyphar: It should error out because it cannot fulfill the configuration that was requested.

AkihiroSuda: "MUST error out" sounds too strict for rootless containers

giuseppe: rootless containers should not get a cgroup configuration if they are not able to satisfy it.  In what cases would the cgroup configuration be set if it is not usable?

giuseppe: just have the freedom of doing it :-)

AkihiroSuda: :-)

At least we should deny setting `cgroup.subtree_control`? And also `cgroup.procs`, `cgroup.threads`, `cgroup.freeze` probably.

cyphar: If rootless containers under `runc` today aren't giving errors if you pass them a cgroup configuration they cannot fulfil, then that's a bug in `runc`. It's a hard requirement of the runtime-spec that runtimes must not silently ignore any aspect of the configuration.

giuseppe: There are at least two: `cgroup.max.descendants` and `cgroup.max.depth` that makes sense to override.

Should the block list be specified in the OCI specs or is it an implementation detail and left to the OCI runtime?  @cyphar what do you think?

giuseppe: I can address it here.

Do we want the conversion to be mandatory?

I think it could be suggested, but I'd not put too much requirements on this area, something like:
```
If a controller is enabled on cgroup v2 and the configuration is provided
for cgroup v1, the runtime COULD attempt a conversion.
If the conversion is not possible the runtime MUST generate an error.

/* Some examples */
```

Also to keep in mind, an OCI runtime could support hybrid mode, so the cgroup v1 configuration still makes sense.

config-linux.md
giuseppe: I've added a new section for the "conversion" part

config-linux.md
```diff
@@ -494,6 +494,49 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
...
+    "io.max": "259:0 rbps=2097152 wiops=120\n253:0 rbps=2097152 wiops=120",
+    "hugetlb.1GB.max": "1073741824"
+}
+```
+
+If a controller is enabled on the cgroup v2 hierarchy but the configuration is provided for the cgroup v1 equivalent controller, the runtime MAY attempt a conversion.
+
+If the conversion is not possible the runtime MUST generate an error.
+
+### Example
```
AkihiroSuda, line 522: Any chance to add the complete conversion convention (either as an example or as a part of the spec)?

giuseppe: should I copy the table from https://github.com/containers/crun/blob/master/crun.1.md#cgroup-v2 ?

AkihiroSuda: yes please

giuseppe: added

config-linux.md
giuseppe: thanks, fixed

config-linux.md
```diff
@@ -494,6 +494,98 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
...
+"unified": {
+    "cpu.weight": "1"
+}
+```
+
+When converting from cgroup v1 to cgroup v2, the runtime SHOULD convert in the following way:
+
+## Memory controller
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
```
h-vetinari, line 544: Shouldn't this be `cgroup v1 value (x)` instead of `OCI (x)`? (also in the tables below)

giuseppe: yes, that is better.  Fixed

giuseppe:
@crosbymichael @mrunalp PTAL

giuseppe:
@dqminh @hqhq  @philips @tianon @vbatts @vishh PTAL

config-linux.md
h-vetinari: Correspondingly, `cgroup 2 value (y)` --> `cgroup v2 value (y)` would make a lot of sense.

giuseppe: thanks, fixed.

giuseppe: I am still not fully convinced the OCI runtime specs should mandate how the conversion is performed.  There are other valid ways to convert from one range to the other.

I'd prefer if it is just a suggestion.  @AkihiroSuda @cyphar what do you think?

h-vetinari: Personally, I think that linear transform of the respective ranges is the only way that satisfies the principle of least surprise, and would therefore be fine with mandating that.

giuseppe: the linear transform doesn't keep the default value unchanged though.

`cpu.shares=1024` will be converted to 39 (or 40), instead of the default `100` for `cpu.weight`.

So there are better ways of converting it, if keeping the default value is important.

I still think the OCI runtime specs should not mandate how the conversion is performed, it looks more like an implementation detail left to the OCI runtime.

giuseppe:
ping

config-linux.md
mrunalp: `to be set and modified for the`

config-linux.md
```diff
@@ -494,6 +494,98 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
...
 
+## <a name="configLinuxUnified" />Unified
+
+**`unified`** (object, OPTIONAL) allows cgroup v2 parameters to be modified at runtime for the container.
+
+Each key in the map refers to a file in the cgroup unified hierarchy.
+
+The OCI runtime MUST ensure that the needed cgroup controllers are enabled for the cgroup.
+
+Configuration unknown to the runtime MUST still be written to the relevant file.
```
mrunalp, line 505: I am not sure why we need this. Did you have a use case in mind?

config-linux.md
giuseppe: thanks, fixed

config-linux.md
giuseppe: it was pointed out here: https://github.com/opencontainers/runtime-spec/pull/1040#discussion_r435832813

When the runtime finds a configuration like `"$CONTROLLER.$NAME": "$VALUE"`, it must always try to write it to the `$CONTROLLER.$NAME` file in the hierarchy.  This is mostly needed to prevent that the OCI runtime ignores any of the cgroup v2 settings.

config-linux.md
```diff
@@ -494,6 +494,31 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
...
+"unified": {
+    "io.max": "259:0 rbps=2097152 wiops=120\n253:0 rbps=2097152 wiops=120",
+    "hugetlb.1GB.max": "1073741824"
+}
+```
+
+If a controller is enabled on the cgroup v2 hierarchy but the configuration is provided for the cgroup v1 equivalent controller, the runtime MAY attempt a conversion.
+
+If the conversion is not possible the runtime MUST generate an error.
+
```
AkihiroSuda, line 521: Can we have a link to https://github.com/containers/crun/blob/0.14.1/crun.1.md#cgroup-v2 as an Implementers' Note

giuseppe: I don't see other such cases in the runtime specs.  If that is ok for the maintainers, I can add it.

config-linux.md
```diff
@@ -494,6 +494,31 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
 }
 ```
 
+## <a name="configLinuxUnified" />Unified
+
+**`unified`** (object, OPTIONAL) allows cgroup v2 parameters to be to be set and modified for the container.
+
+Each key in the map refers to a file in the cgroup unified hierarchy.
```
vbatts, line 501: I understand not binding the to kernel API,  but should we link to docs and note that these names will change/add per the kernel versions. Also denote any error handling for unknown seettings?

config-linux.md
AkihiroSuda: ping

config-linux.md
giuseppe: sure I think we can add a note saying these files can change per kernel version.

The rationale for having a map here is that the OCI runtime doesn't have to know about what the settings mean as long as it writes the value to the specified file.  Should the error handling be anything more than propagating the errno from `open` or `write`?

dangowrt: > sure I think we can add a note saying these files can change per kernel version.
> 
> The rationale for having a map here is that the OCI runtime doesn't have to know about what the settings mean as long as it writes the value to the specified file. Should the error handling be anything more than propagating the errno from `open` or `write`?

The implementation also needs to make sure the corresponding controller is available and enabled (ie. recursively write to `cgroup.subtree_control`). The controller can be unknown to the implementation, not be available (due to being used in cgroup1 or not being complied into the kernel) or lack features (eg. `io.bfq.*`).

giuseppe: isn't that implicit with `The runtime MUST generate an error when the configuration refers to a cgroup controller that is not present or that cannot be enabled.`.

Even if the controller is not present because not compiled in the kernel, the runtime MUST fail and return an error if the configuration specified a setting that cannot be honored.

AkihiroSuda, line 501: I still think we should restrict key strings
https://github.com/opencontainers/runtime-spec/pull/1040#discussion_r465093114

giuseppe: is it something that we must enforce in the specs or that is up to the OCI runtime?  I am still not sure we need to enforce everything in the specs since the runtime can have more restrictions.

Are you fine just with `cgroup.subtree_control`, `cgroup.procs`, `cgroup.threads` and `cgroup.freeze` ?  Can it be `SHOULD`? :-)

AkihiroSuda: At least `/` MUST NOT be included in the key.  I'm not sure about `cgroup.*` stuff. Can we have a POC implementation in runc/crun repo for experiments?

config-linux.md
```diff
@@ -494,6 +494,31 @@ You MUST specify at least one of the `hcaHandles` or `hcaObjects` in a given ent
...
+### Example
+
+```json
+"unified": {
+    "io.max": "259:0 rbps=2097152 wiops=120\n253:0 rbps=2097152 wiops=120",
+    "hugetlb.1GB.max": "1073741824"
+}
+```
+
+If a controller is enabled on the cgroup v2 hierarchy but the configuration is provided for the cgroup v1 equivalent controller, the runtime MAY attempt a conversion.
```
AkihiroSuda, line 518: If both v1 and v2 config are provided and they conflict, what should the runtime do? My suggestion is `the runtime SHOULD use v2 value` (with a warning log)

giuseppe: that looks like an invalid configuration, isn't better to just error out?

config-linux.md
giuseppe: sure, that is why I'd prefer we are not too strict at the moment for what is not permitted :-)  It may stop valid use cases, and we can always amend the specs and be more specific in future.

config-linux.md
AkihiroSuda: Sometimes it might be unclear what does conflict and what does not. E.g. if CPU RT limit or Kmem limit is specified in v1 config, does it "conflict" with v2 config?

crosbymichael: I would call this invalid and you should error.  Lets not make things more complicated then it needs to be.  Producers of the runtime spec are expected to have a good understanding on how things work and they should create a correct spec or else we error.  

Lets keep this simple.

crosbymichael:
Overall I like this approach and it seems to be very good for additional subsystems and fields in the future.

I'm all for runtimes throwing an error if both v1 and v2 are populated, they should use the very basic checks to fill in the correct object, this is not hard at all.  I say we move forward with this.

vbatts:
LGTM

crosbymichael:
LGTM

crosbymichael:
ping @mrunalp you wanna check it over one last time and do the honors of merging it?

mrunalp:
@crosbymichael @vbatts Thanks for your reviews :) I think this is fine to merge to make progress. Once we start implementing and find issues then we could update the spec accordingly.

mrunalp:
LGTM

PR #1042, opened 2020-05-12
seccomp: fix go-specs for errnoRet

giuseppe:
commit 3bfcde28eab5d7e107ddd3771fc92399332932a8 introduced errnoRet
for seccomp syscalls but the Go specs were not implemented correctly.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -669,7 +669,7 @@ type LinuxSeccompArg struct {
 type LinuxSyscall struct {
 	Names    []string           `json:"names"`
 	Action   LinuxSeccompAction `json:"action"`
-	ErrnoRet uint               `json:"errno"`
+	ErrnoRet *uint              `json:"errnoRet,omitempty"`
 	Args     []LinuxSeccompArg  `json:"args,omitempty"`
 }
 


rhatdan:
@vbatts PTAL

giuseppe:
@mrunalp @vbatts  PTAL

specs-go/config.go
```diff
@@ -669,7 +669,7 @@ type LinuxSeccompArg struct {
 type LinuxSyscall struct {
 	Names    []string           `json:"names"`
 	Action   LinuxSeccompAction `json:"action"`
-	ErrnoRet uint               `json:"errno"`
+	ErrnoRet *uint              `json:"errnoRet,omitempty"`
```
thaJeztah, line 672: Wondering; Was the pointer needed? (wondering because IIRC, `omitempty` also omits `0`, in which case the default for an `uint` would be a meaningful value (no override)

giuseppe: I think we need the pointer here as specifying `ErrnoRet = 0` has a different meaning (report success), while the default with `ErrnoRet` unspecified is `EPERM`.

I've done a test for `omitempty` and it seems pointers are omitted only if they are `nil`: https://play.golang.org/p/wnzK9EQNCOS

thaJeztah: Wouldn't `ErrnoRet = 0` conflict with the purpose of `LinuxSeccompAction` ?

giuseppe: I think the combination `SCMP_ACT_ERRNO` and `ErrnoRet = 0` is useful and it is different than `SCMP_ACT_ALLOW`, since the syscall is not performed but it is reported as successful.

thaJeztah: Ah, gotcha, yes, that makes sense

mrunalp:
LGTM

specs-go/config.go
cyphar: Yeah, it's critical that it be clear whether `errnoRet` was unset (`-EPERM`) or `0` (a "success" but the syscall isn't run). As an aside, this trick is used for the relatively-new seccomp syscall emulation.

cyphar:
LGTM.

specs-go/config.go
thaJeztah: I assumed it would only be used with `> 0` values (`0` meaning; don't change the default); thx!

cyphar: All good, it's definitely not the intended use for seccomp. ;)

giuseppe:
@caniszczyk @crosbymichael @dqminh @hqhq @philips @tianon @vishh PTAL

tianon:
LGTM

PR #1046, opened 2020-05-20
Add State status constants to spec-go

RenaudWasTaken:
Hello!

The [State definition ](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state) defines 4 constants for the status field that are not present in the specs-go

This PR adds them to the state.go file to make it available to runc.
/cc @cyphar @mrunalp @mikebrow

### DIFF
--- a/specs-go/state.go
+++ b/specs-go/state.go
@@ -1,18 +1,21 @@
 package specs
 
+// ContainerState represents the state of a container.
+type ContainerState string
+
 const (
 	// StateCreating indicates that the container is being created
-	StateCreating = "creating"
+	StateCreating ContainerState  = "creating"
 
 	// StateCreated indicates that the runtime has finished the create operation
-	StateCreated = "created"
+	StateCreated ContainerState  = "created"
 
 	// StateRunning indicates that the container process has executed the
 	// user-specified program but has not exited
-	StateRunning = "running"
+	StateRunning ContainerState  = "running"
 
 	// StateStopped indicates that the container process has exited
-	StateStopped = "stopped"
+	StateStopped ContainerState  = "stopped"
 )
 
 // State holds information about the runtime state of the container.
--- a/specs-go/state.go
+++ b/specs-go/state.go
@@ -1,5 +1,20 @@
 package specs
 
+const (
+	// StateCreating indicates that the container is being created
+	StateCreating = "creating"
+
+	// StateCreated indicates that the runtime has finished the create operation
+	StateCreated = "created"
+
+	// StateRunning indicates that the container process has executed the
+	// user-specified program but has not exited
+	StateRunning = "running"
+
+	// StateStopped indicates that the container process has exited
+	StateStopped = "stopped"
+)
+
 // State holds information about the runtime state of the container.
 type State struct {
 	// Version is the version of the specification that is supported.


specs-go/state.go
```diff
@@ -1,5 +1,20 @@
 package specs
 
+const (
+	// Creating indicates that the container is being created
+	Creating = "creating"
+
+	// Created indicates that the runtime has finished the create operation
+	Created  =  "created"
```
tianon, line 8: I think this file probably needs a `gofmt`. :sweat_smile:

tianon: Also, given these constants are all within the `specs` namespace, shouldn't they be prefixed with something like `State` or `StateStatus` so it's clear what they're for (as in, `StateStatusCreated`, etc)?

RenaudWasTaken: Fixed thanks!

specs-go/state.go
```diff
@@ -1,5 +1,20 @@
...
+	// StateCreated indicates that the runtime has finished the create operation
+	StateCreated = "created"
+
+	// StateRunning indicates that the container process has executed the
+	// user-specified program but has not exited
+	StateRunning = "running"
+
+	// StateStopped indicates that the container process has exited
+	StateStopped = "stopped"
+)
```
mikebrow, line 16: can we add some comments about the delta between these and runc status states:
```
func (s Status) String() string {
	switch s {
	case Created:
		return "created"
	case Running:
		return "running"
	case Pausing:
		return "pausing"
	case Paused:
		return "paused"
	case Stopped:
		return "stopped"
	default:
		return "unknown"
	}
}
```
Creating being new.. (and not yet discussed in the spec?)  It needs a bit of an intro maybe? 

The missing ones caught my eye.. but I get that this is the runtime-spec and the other states are not yet covered.. maybe a TODO for pulling them in?

cyphar: The pause-related ones aren't in the runtime-spec because we don't have a pause operation in the spec (among other missing things, such as the lack of an exec operation). I'd say that we should only define the ones which are actually listed in the spec and in `runc` we can just add the ones we have.

mikebrow: agreed...shouldn't introduce new state values that are not in the spec. Might be better just to leave them all out..

RenaudWasTaken: So I suppose this is resolved?

mikebrow: need to remove ```StateCreating = "creating"```

mrunalp:
LGTM

specs-go/state.go
RenaudWasTaken: I thought I had left a comment here, it looks like it probably didn't get sent :D ?
Anyways, I just double checked, this enum has all the states defined in the [spec here](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md)

mikebrow: oh man.. I entirely missed that a new quasi state was added for between the create request and actual created.

thaJeztah, line 16: Do we want these to have the same type (similar to https://godoc.org/github.com/docker/docker/api/types/mount#Consistency) ?

```suggestion
// ContainerState represents the state of a container.
type ContainerState string

const (
	// StateCreating indicates that the container is being created
	StateCreating ContainerState  = "creating"

	// StateCreated indicates that the runtime has finished the create operation
	StateCreated ContainerState  = "created"

	// StateRunning indicates that the container process has executed the
	// user-specified program but has not exited
	StateRunning ContainerState  = "running"

	// StateStopped indicates that the container process has exited
	StateStopped ContainerState  = "stopped"
)
```

cyphar: Yeah I agree, there should be a dedicated type for it.

thaJeztah:
ping @arnaudwastaken could you have a look at https://github.com/opencontainers/runtime-spec/pull/1046#discussion_r448865229 ?

RenaudWasTaken:
Sorry looking at it now!

RenaudWasTaken:
@thaJeztah applied!

thaJeztah:
@mrunalp @tianon @cyphar PTAL

specs-go/state.go
kolyshkin: Now, should the `State.Status` field should also use this type, ie

```diff
// State holds information about the runtime state of the container.
type State struct {
...
         // Status is the runtime status of the container.
-        Status string `json:"status"`
+        Status ContainerState `json:"status"`
...
```

I am asking because of this https://github.com/opencontainers/runc/pull/2520/commits/34ac72f87cd41ea116828760bb70ed70960c8d8e#r456629563

tianon: IMO that makes sense

RenaudWasTaken: I think it also makes sense, existing hooks shouldn't be broken by this too so it would just be a matter of updating the existing runtimes.

I'll open a PR later today.

RenaudWasTaken: It's here: https://github.com/opencontainers/runtime-spec/pull/1056

cyphar: Agreed.

PR #1047, opened 2020-05-28
config: Add AllowSpeculation

KentaTada:
AllowSpeculation disables spectre mitigations for container.
For more information about that, please refer to:
https://github.com/opencontainers/runc/issues/2430


specs-go/config.go
```diff
@@ -58,6 +58,8 @@ type Process struct {
 	OOMScoreAdj *int `json:"oomScoreAdj,omitempty" platform:"linux"`
 	// SelinuxLabel specifies the selinux context that the container process is run as.
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
+	// AllowSpeculation disables spectre mitigations
```
giuseppe, line 61: could we just make it more generic like using `speculative execution` instead of `spectre`?

KentaTada: I agree with you. I changed the above and config.md and the commit message.

giuseppe: thanks!

config.md
```diff
@@ -208,6 +208,7 @@ For Linux-based systems, the `process` object supports the following process-spe
     For more information on how these two settings work together, see [the memory cgroup documentation section 10. OOM Contol][cgroup-v1-memory_2].
 * **`selinuxLabel`** (string, OPTIONAL) specifies the SELinux label for the process.
     For more information about SELinux, see  [SELinux documentation][selinux].
+* **`allowSpeculation`** (bool, OPTIONAL) setting `allowSpeculation` to true disable speculative execution mitigations to improve the performance.
```
mrunalp, line 211: I think we should name this `disableSpeculationMitigations` instead.

KentaTada: >I think we should name this disableSpeculationMitigations instead.

`disableSpeculationMitigations` is easy to understand although I wanted to use the positive boolean variable name. I don't mind whichever it is.
@giuseppe, Could you give me your opinion on the variable name?

giuseppe: I think the name @mrunalp suggested is clearer, since it is about allowing or disabling mitigations not the speculative execution itself.

KentaTada: Thanks! I'll modify it.

KentaTada: @mrunalp @giuseppe I changed the name.

config.md
```diff
@@ -208,6 +208,7 @@ For Linux-based systems, the `process` object supports the following process-spe
     For more information on how these two settings work together, see [the memory cgroup documentation section 10. OOM Contol][cgroup-v1-memory_2].
 * **`selinuxLabel`** (string, OPTIONAL) specifies the SELinux label for the process.
     For more information about SELinux, see  [SELinux documentation][selinux].
+* **`disableSpeculationMitigations`** (bool, OPTIONAL) setting `disableSpeculationMitigations` to true disable speculative execution mitigations to improve the performance.
```
cyphar, line 211: I think this needs more explanation as an option (and should be reworded to not repeat the name of the setting), and preferably a link to the kernel documentation for speculation mitigations.

```suggestion
* **`disableSpeculationMitigations`** (bool, OPTIONAL) specifies whether CPU speculative execution mitigations should be disabled for the process. Several mitigations are auto-enabled under Linux, and can cause a noticeable performance impact (depending on your workload). Note that enabling this option may reduce the security properties of containers created with this configuration. See [the kernel documentation][speculative-control] for more information.

[speculative-control]: https://www.kernel.org/doc/html/latest/userspace-api/spec_ctrl.html
```

However, looking at [the kernel documentation](https://www.kernel.org/doc/html/latest/userspace-api/spec_ctrl.html), i notice that this new configuration option doesn't allow users to specify `PR_SPEC_FORCE_DISABLE` -- which forces a container to have speculation mitigations *enabled*. There's also an argument that users should be able to specify which mitigations to disable... Maybe we should instead have a `mitigations` object that has more than one option:

```"speculationMitigations": {
	"defaultRule": "force-enable",
	"exceptions": [
		{
			"mitigation": "store-bypass",
			"rule": "disable",
		},
	],
}
```

The above configuration would result in `PR_SPEC_STORE_BYPASS` being set to `PR_SPEC_ENABLE`, and `PR_SPEC_INDIR_BRANCH` being set to `PR_SPEC_FORCE_DISABLE`.

KentaTada: >I think this needs more explanation as an option (and should be reworded to not repeat the name of the setting), and preferably a link to the kernel documentation for speculation mitigations.

I completely agree with you. I should prepare for more explanation. Thanks!

>Maybe we should instead have a mitigations object that has more than one option:

I agree to use object but please let me explain the background of prctl and seccomp.

The kernel automatically forces a container to have all speculation mitigations enabled when seccomp is used without `SECCOMP_FILTER_FLAG_SPEC_ALLOW` on x86_64.
https://github.com/torvalds/linux/blob/v5.7-rc7/arch/x86/kernel/cpu/bugs.c#L1201
It affects the result of prctl.

In addition to that, I found the kernel bug that the kernel only forced to enable Speculative Store Bypass mitigations and not to enable Indirect Branch Speculation mitigations even if seccomp was used. I may create a kernel patch after more investigation.

Anyway, I thought it was difficult for users to set up complex settings at first.
But I reconsider and it is no problem if container runtime handles it appropriately.

giuseppe: I've just realized we already have `SECCOMP_FILTER_FLAG_SPEC_ALLOW` in the seccomp flags (and I've added it): https://github.com/opencontainers/runtime-spec/pull/1018

KentaTada: @giuseppe @cyphar

>I've just realized we already have SECCOMP_FILTER_FLAG_SPEC_ALLOW in the seccomp flags (and I've added it): #1018

`SECCOMP_FILTER_FLAG_SPEC_ALLOW` is also needed to disable mitigations when secccomp is used as mentioned previously.
I think we will implement the container runtime using secomp with SECCOMP_FILTER_FLAG_SPEC_ALLOW + prctl like below:
https://chromium.googlesource.com/chromium/src/sandbox/+/be48c498815b50c9585332d35bb5f6f4920c41de

But 
>However, looking at the kernel documentation, i notice that this new configuration option doesn't allow users to specify PR_SPEC_FORCE_DISABLE -- which forces a container to have speculation mitigations enabled. There's also an argument that users should be able to specify which mitigations to disable... Maybe we should instead have a mitigations object that has more than one option:

I don't know the reason but https://www.kernel.org/doc/html/latest/userspace-api/spec_ctrl.html is different from the actual behavior.

- When PR_SPEC_FORCE_DISABLE is set for PR_SPEC_STORE_BYPASS, a subsequent prctl(…, PR_SPEC_ENABLE) will fail.  This is as same as the kernel document.
- When PR_SPEC_FORCE_DISABLE is set for PR_SPEC_INDIRECT_BRANCH, a subsequent prctl(…, PR_SPEC_ENABLE) **will not fail as below comments from kernel community.  This is not as same as the kernel document.**
https://lore.kernel.org/patchwork/patch/1251849/

Actually, my below modified runc can disable IBPB/STIBP mitigation even if seccomp without `SECCOMP_FILTER_FLAG_SPEC_ALLOW` is enabled. 
https://github.com/opencontainers/runc/pull/2433

I'm sorry but please give me some time to reconsider this interface because it is complicated. And please give me feedback.
Thanks!

KentaTada: FYI:
>When PR_SPEC_FORCE_DISABLE is set for PR_SPEC_INDIRECT_BRANCH, a subsequent prctl(…, PR_SPEC_ENABLE) will not fail as below comments from kernel community. This is not as same as the kernel document.
https://lore.kernel.org/patchwork/patch/1251849/

After my patch, another person modified this issue as same as my patch.
https://lore.kernel.org/patchwork/patch/1251849/

So, the complicated specification or bug is fixed. I'll reconsider the interface.

KentaTada: @cyphar  I apologize for the late reply. I updated to use object.
Could you review the latest commit?

KentaTada:
@cyphar 
This feature is also needed  in https://github.com/moby/moby/issues/42619
Could you review my update?

config.md
```diff
@@ -208,6 +208,23 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+        Each entry has the following structure:
+        * **`mitigation`** *(string, REQUIRED)* - the name of specific mitigation.
+            A valid list of mitigations.
+            * `store-bypass` - Speculative Store Bypass
+            * `indirect-branch` - Indirect Branch Speculation in User Processes
+        * **`rule`** *(string, REQUIRED)* - enables or disables the specific mitigation.
+            * `enable` - The mitigation of this particular speculation is disabled.
+            * `disable` - The mitigation of this particular speculation is enabled.
+            * `force-disable` - Same as disable, but it cannot be undone.
+            * `disable-noexec` - Same as disable, but the state will be cleared on execve(2).
```
kailun-qin, line 227: `disable-noexec` is not applicable/supported for `indirect-branch`. Maybe we need a note here?

specs-go/config.go
```diff
@@ -75,6 +77,18 @@ type LinuxCapabilities struct {
...
 }
 
+// LinuxDisableSpeculationMitigations sets up the rule of speculative execution mitigations.
+type LinuxDisableSpeculationMitigations struct {
+	DefaultRule string `json:"defaultRule"`
+	Exceptions []SpecExceptions `json:"exceptions,omitempty"`
+}
+
+// SpecExceptions is used to specify the setting of speculative execution mitigations.
+type SpecExceptions struct {
```
kailun-qin, line 87: `SpecException` instead of `SpecExceptions`, as each defines one exception?

config.md
```diff
@@ -208,6 +208,23 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+        * `force-disable` - Same as disable, but it cannot be undone.
+        * `disable-noexec` - Same as disable, but the state will be cleared on execve(2).
+    * **`exceptions`** *(array of objects, OPTIONAL)* - the configuration of specific mitigations.
+        Each entry has the following structure:
+        * **`mitigation`** *(string, REQUIRED)* - the name of specific mitigation.
+            A valid list of mitigations.
+            * `store-bypass` - Speculative Store Bypass
+            * `indirect-branch` - Indirect Branch Speculation in User Processes
+        * **`rule`** *(string, REQUIRED)* - enables or disables the specific mitigation.
+            * `enable` - The mitigation of this particular speculation is disabled.
```
kailun-qin, line 224: Should the `rules` under `exceptions` keep the same semantics with *disable*, or it should reflect/align with the corresponding *mitigation*'s?

config.md
```diff
@@ -208,6 +208,23 @@ For Linux-based systems, the `process` object supports the following process-spe
...
 * **`selinuxLabel`** (string, OPTIONAL) specifies the SELinux label for the process.
     For more information about SELinux, see  [SELinux documentation][selinux].
+* **`disableSpeculationMitigations`** (object, OPTIONAL) specifies whether CPU speculative execution mitigations should be disabled for the process. Several mitigations are auto-enabled under Linux, and can cause a noticeable performance impact (depending on your workload). Note that enabling this option may reduce the security properties of containers created with this configuration. See [the kernel documentation][speculative-control] for more information.
+    * **`defaultRule`** *(string, REQUIRED)*  sets up the default rule to enable or disable the mitigations.
+        * `enable` - The mitigation of speculations without `exceptions` is disabled.
+        * `disable` - The mitigation of speculations without `exceptions` is enabled.
+        * `force-disable` - Same as disable, but it cannot be undone.
+        * `disable-noexec` - Same as disable, but the state will be cleared on execve(2).
+    * **`exceptions`** *(array of objects, OPTIONAL)* - the configuration of specific mitigations.
```
kailun-qin, line 217: Is there any possibility that we need multiple exceptions with the same `mitigation` type? If not, what about using dedicate structs with rules for `store-bypass` and `indirect-branch`?

PR #1054, opened 2020-07-07
Change all references from whitelist to allowlist

rhatdan:
We want to move to more enclusive names/terms in our code, and remove problematic language
from code and comments.

We want to change reference for whitelist/blacklist to allowlist/denylist.

https://www.redhat.com/en/blog/making-open-source-more-inclusive-eradicating-problematic-language

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -120,7 +120,7 @@ Each entry has the following structure:
     If a [file][] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
 * **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - [major, minor numbers][devices] for the device.
 * **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
-    You can also control access to devices [with cgroups](#device-whitelist).
+    You can also control access to devices [with cgroups](#configLinuxDeviceAllowedlist).
 * **`uid`** *(uint32, OPTIONAL)* - id of device owner in the [container namespace](glossary.md#container-namespace).
 * **`gid`** *(uint32, OPTIONAL)* - id of device group in the [container namespace](glossary.md#container-namespace).
 
@@ -161,7 +161,7 @@ In addition to any devices configured with this setting, the runtime MUST also s
 * [`/dev/random`][random.4]
 * [`/dev/urandom`][random.4]
 * [`/dev/tty`][tty.4]
-* `/dev/console` is set up if [`terminal`](config.md#process) is enabled in the config by bind mounting the pseudoterminal slave to `/dev/console`.
+* `/dev/console` is set up if [`terminal`](config.md#process) is enabled in the config by bind mounting the pseudoterminal pty to `/dev/console`.
 * [`/dev/ptmx`][pts.4].
   A [bind-mount or symlink of the container's `/dev/pts/ptmx`][devpts].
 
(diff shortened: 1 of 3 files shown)


config-linux.md
```diff
@@ -214,9 +214,9 @@ Runtimes MAY attach the container process to additional cgroup controllers beyon
 }
 ```
 
-### <a name="configLinuxDeviceWhitelist" />Device whitelist
```
cyphar, line 217: Since "whitelist" implies a deny-by-default configuration (with "blacklist" implying the reverse), we should add some text to clarify that this is the case. At least to my ear, "allowlist" doesn't have that implication (at least not as apparently as "whitelist").

cyphar:
@giuseppe Indeed. In particular, the text on pseudoterminals should also be revised to use "ptmx" and "pty" or something like that. There was [a twitter thread about this a few days ago](https://twitter.com/hugelgupf/status/1279686813670498305).

config-linux.md
thaJeztah: block/unblock?

rhatdan:
Well allowlist and denylist was the name referenced in blog above, which is why I went with it.  I am fine with picking some other name or just adding clarification data.

giuseppe:
> @giuseppe Indeed. In particular, the text on pseudoterminals should also be revised to use "ptmx" and "pty" or something like that. There was [a twitter thread about this a few days ago](https://twitter.com/hugelgupf/status/1279686813670498305).

I've performed the same change in crun after seeing that discussion :-) my first thought was for using `pty/tty` as I was also wrongly assuming that `-mx` stands for master.  Since it is not the case, I think ptmx/pty is much clearer unless the kernel is going to use a different naming.

giuseppe:
> Well allowlist and denylist was the name referenced in blog above, which is why I went with it. I am fine with picking some other name or just adding clarification data.

thanks.  LGTM

rhatdan:
@cyphar @thaJeztah PTAL

config-linux.md
```diff
@@ -120,7 +120,7 @@ Each entry has the following structure:
     If a [file][] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
 * **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - [major, minor numbers][devices] for the device.
 * **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
-    You can also control access to devices [with cgroups](#device-whitelist).
+    You can also control access to devices [with cgroups](#device-allowedlist).
```
thaJeztah, line 123: This looks broken; should either be

```suggestion
    You can also control access to devices [with cgroups](#configLinuxDeviceAllowedlist).
```

Or

```suggestion
    You can also control access to devices [with cgroups](#allowed-device-list).
```

config-linux.md
cyphar: I should've better clarified my point here -- the title "device allowlist" is totally fine (and I prefer it to "allowed device list"). What I was suggesting is to add a single sentence which says something like:

> The allow-list MUST have a default deny-all policy, meaning that if it is omitted no device access is permitted.

This would not be a change in behaviour (since that is how all implementations have worked and is technically already required by the term "whitelist") but instead just a clarification.

specs-go/config.go
```diff
@@ -354,7 +354,8 @@ type LinuxRdma struct {
 
 // LinuxResources has container runtime resource constraints
 type LinuxResources struct {
-	// Devices configures the device whitelist.
+	// Devices configures the only devices allowed to be used within
+	// the container.
```
cyphar, line 358: s/whitelist/allowlist/ reads better IMHO.

rhatdan: Switched back.

rhatdan:
@caniszczyk @crosbymichael @cyphar @dqminh @hqhq @mrunalp @tianon @vbatts @vishh PTAL

hqhq:
LGTM

vbatts:
LGTM

PR #1056, opened 2020-07-21
Update State structure to use the new ContainerState type

RenaudWasTaken:
Following up on @kolyshkin 's [comment](https://github.com/opencontainers/runtime-spec/pull/1046#discussion_r457674394), we realized I should have updated the `State` structure to use that new type.

Thanks for picking up on this mistake!

/cc @kolyshkin @titanous 

### DIFF
--- a/specs-go/state.go
+++ b/specs-go/state.go
@@ -25,7 +25,7 @@ type State struct {
 	// ID is the container ID
 	ID string `json:"id"`
 	// Status is the runtime status of the container.
-	Status string `json:"status"`
+	Status ContainerState `json:"status"`
 	// Pid is the process ID for the container process.
 	Pid int `json:"pid,omitempty"`
 	// Bundle is the path to the container's bundle directory.


thaJeztah:
I wondered if this would be a breaking change (from a "go" perspective), but thinking about that, I guess the tags in this repository reflect the "spec", not the go API; should this be mentioned somewhere? (or should the "specs-go" get it's own versioning/tags? (`specs-go/vX.Y.Z`)

kolyshkin:
LGTM

thaJeztah:
> I think we've had the "Go breaking change" conversation previously, and came to the conclusion that the version is for the spec itself, and the Go code here is merely supporting rather than the focus (so "Go breaking changes" are not critical).

Absolutely makes sense. I went looking if this was mentioned anywhere, and couldn't find it in the readme or [RELEASES.md](https://github.com/opencontainers/runtime-spec/blob/master/RELEASES.md).

I'm wondering if (given that the Go code and spec don't follow the same "stability" from a SemVer perspective) if versioning should be decoupled. Doing so would also allow more flexible handling of changes desired in the Go code, without requiring a new release of the specification itself. Given that the go code lives in a subdirectory/package, the project is in a "fortunate" position that this would be possible (using the `module-name/v<version>` tagging scheme), and it would help with those that switched to Go modules (to not get caught by surprise).

I can open a ticket for discussion

AkihiroSuda:
ready to merge?

PR #1062, opened 2020-08-13
Add support for time namespace

KentaTada:
The time namespace is a new kernel feature available in 5.6+ to
isolate the system monotonic and boot-time clocks.


tianon:
I think it's worth linking to https://github.com/opencontainers/runc/issues/2345, which has some related discussion.

KentaTada:
@tianon Thanks!
I want to support time namespace both runc and our container runtime.
In any case, I think runtime-spec should be updated at first.

giuseppe:
I think this is not enough, it is necessary also to specify a value for `timens_offsets`.

KentaTada:
> I think this is not enough, it is necessary also to specify a value for `timens_offsets`.

You are right. I added the offset.

m-yosefpor:
Currently running `uptime` command in a container shows host uptime, instead of container uptime. Will time namespace solve this issue?

utam0k:
Hi, everyone. What is the status of this PR? [lxd has supported for Time namespace](https://discuss.linuxcontainers.org/t/lxc-5-0-lts-has-been-released/14381). I think some runtimes can support it.
https://github.com/opencontainers/runc/issues/2345

KentaTada:
I'm sorry I mistook and closed this request.
Please see https://github.com/opencontainers/runtime-spec/pull/1151

PR #1074, opened 2020-11-05
Add Seccomp Notify support using UNIX sockets and container metadata

rata:
Hi!

This is a follow-up of PR #1073 using the alternative [discussed here](https://github.com/opencontainers/runtime-spec/pull/1073#issuecomment-719465555). The reasoning to go this way over the others [is here](https://github.com/opencontainers/runtime-spec/pull/1073#issuecomment-720524314)

I’m a co-worker of @alban and the commit has his sign-off too :)

This adds the specification for Seccomp Userspace Notification and the
Golang bindings. This contains:
- New fields in the seccomp section to use with seccomp userspace
  notification.
- Additional SeccompState struct containing the container state and file
  descriptors passed for seccomp.

This was discussed in the OCI Weekly Discussion on September 16th,
2020. After review on github, this implementation was changed to the
"Proposal with listenerPath and listenerExtraMetadata". For more
information see:
- https://github.com/opencontainers/runtime-spec/pull/1073#issuecomment-719465555

Docs presented on the community meeting (for the old implementation
using hooks):
- https://hackmd.io/El8Dd2xrTlCaCG59ns5cwg#September-16-2020
- https://docs.google.com/document/d/1xHw5GQjMj6ZKR-40aKmTWZRkvlPuzMGQRu-YpOFQc30/edit

Documentation for this feature:
- https://www.kernel.org/doc/html/v5.0/userspace-api/seccomp_filter.html#userspace-notification
- man pages: seccomp_user_notif.2 at
  https://git.kernel.org/pub/scm/docs/man-pages/man-pages.git/log/?h=seccomp_user_notif
- brauner's blog:
  https://brauner.github.io/2020/07/23/seccomp-notify.html

This PR is an alternative proposal to PR #1038 . While similar in nature,
the main difference is that this PR adds optional metadata to be sent to
the seccomp agent and specifies how the UNIX socket MUST be used.

Signed-off-by: Rodrigo Campos <rodrigo@kinvolk.io>

---
We will make a PR for runc implementing this shortly.

Please also note that this change doesn’t imply a fork+exec, as @giuseppe  [wanted to avoid](https://github.com/opencontainers/runtime-spec/pull/1073#pullrequestreview-520449705)

Thanks again and don't hesitate to ask any questions! (I'll answer the ones in Alban's PR shortly :))

cc @KentaTada @ManaSugi @giuseppe @AkihiroSuda @brauner @alban  @mauriciovasquezbernal

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -629,6 +629,21 @@ The following parameters can be specified to set up seccomp:
     * `SECCOMP_FILTER_FLAG_LOG`
     * `SECCOMP_FILTER_FLAG_SPEC_ALLOW`
 
+* **`listenerPath`** *(string, OPTIONAL)* - specifies the path of UNIX domain socket over which the runtime will send the [container process state](#containerprocessstate) data structure when the `SCMP_ACT_NOTIFY` action is used.
+    This socket MUST use `AF_UNIX` domain and `SOCK_STREAM` type.
+    The runtime MUST send exactly one [container process state](#containerprocessstate) per connection.
+    The connection MUST NOT be reused and it MUST be closed after sending a seccomp state.
+    If sending to this socket fails, the runtime MUST [generate an error](runtime.md#errors).
+    If the `SCMP_ACT_NOTIFY` action is not used this value is ignored.
+
+    The runtime sends the following file descriptors using `SCM_RIGHTS` and set their names in the `fds` array of the [container process state](#containerprocessstate):
+
+    * **`seccompFd`** (string, REQUIRED) is the seccomp file descriptor returned by the seccomp syscall.
+
+* **`listenerMetadata`** *(string, OPTIONAL)* - specifies an opaque data to pass to the seccomp agent.
+    This string will be sent as the `metadata` field in the [container process state](#containerprocessstate).
+    This field MUST NOT be set if `listenerPath` is not set.
+
 * **`syscalls`** *(array of objects, OPTIONAL)* - match a syscall in seccomp.
     While this property is OPTIONAL, some values of `defaultAction` are not useful without `syscalls` entries.
     For example, if `defaultAction` is `SCMP_ACT_KILL` and `syscalls` is empty or unset, the kernel will kill the container process on its first syscall.
@@ -637,7 +652,7 @@ The following parameters can be specified to set up seccomp:
     * **`names`** *(array of strings, REQUIRED)* - the names of the syscalls.
         `names` MUST contain at least one entry.
     * **`action`** *(string, REQUIRED)* - the action for seccomp rules.
-        A valid list of constants as of libseccomp v2.4.0 is shown below.
+        A valid list of constants as of libseccomp v2.5.0 is shown below.
 
         * `SCMP_ACT_KILL`
         * `SCMP_ACT_KILL_PROCESS`
@@ -647,6 +662,7 @@ The following parameters can be specified to set up seccomp:
         * `SCMP_ACT_TRACE`
         * `SCMP_ACT_ALLOW`
         * `SCMP_ACT_LOG`
+        * `SCMP_ACT_NOTIFY`
 
     * **`errnoRet`** *(uint, OPTIONAL)* - the errno return code to use.
         Some actions like `SCMP_ACT_ERRNO` and `SCMP_ACT_TRACE` allow to specify the errno code to return.
@@ -688,6 +704,45 @@ The following parameters can be specified to set up seccomp:
             "action": "SCMP_ACT_ERRNO"
         }
     ]
+}
+```
+
+### <a name="containerprocessstate" />The Container Process State
+
+The container process state is a data structure passed via a UNIX socket.
+The container runtime MUST send the container process state over the UNIX socket as regular payload serialized in JSON and file descriptors MUST be sent using `SCM_RIGHTS`.
+The container runtime MAY use several `sendmsg(2)` calls to send the aforementioned data.
+If more than one `sendmsg(2)` is used, the file descriptors MUST be sent only in the first call.
+
+The container process state includes the following properties:
+
+* **`ociVersion`** (string, REQUIRED) is version of the Open Container Initiative Runtime Specification with which the container process state complies.
+* **`fds`** (array, OPTIONAL) is a string array containing the names of the file descriptors passed.
+    The index of the name in this array corresponds to index of the file descriptors in the `SCM_RIGHTS` array.
+* **`pid`** (int, REQUIRED) is the container process ID, as seen by the runtime.
+* **`metadata`** (string, OPTIONAL) opaque metadata.
+* **`state`** ([state](runtime.md#state), REQUIRED) is the state of the container.
+
+Example sending a single `seccompFD` file descriptor in the `SCM_RIGHTS` array:
+
+```json
+{
+    "ociVersion": "0.2.0",
+    "fds": [
+        "seccompFd"
+    ],
+    "pid": 4422,
+    "metadata": "MKNOD=/dev/null,/dev/net/tun;BPF_MAP_TYPES=hash,array",
+    "state": {
+        "ociVersion": "0.2.0",
+        "id": "oci-container1",
+        "status": "creating",
+        "pid": 4422,
+        "bundle": "/containers/redis",
+        "annotations": {
+            "myKey": "myValue"
+        }
+    }
 }
 ```
 
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -212,6 +212,12 @@
                             "$ref": "defs-linux.json#/definitions/SeccompFlag"
                         }
                     },
+                    "listenerPath": {
+                        "type": "string"
+                    },
+                    "listenerMetadata": {
+                        "type": "string"
+                    },
                     "architectures": {
                         "type": "array",
                         "items": {
--- a/specs-go/state.go
+++ b/specs-go/state.go
@@ -5,17 +5,17 @@ type ContainerState string
 
 const (
 	// StateCreating indicates that the container is being created
-	StateCreating ContainerState  = "creating"
+	StateCreating ContainerState = "creating"
 
 	// StateCreated indicates that the runtime has finished the create operation
-	StateCreated ContainerState  = "created"
+	StateCreated ContainerState = "created"
 
 	// StateRunning indicates that the container process has executed the
 	// user-specified program but has not exited
-	StateRunning ContainerState  = "running"
+	StateRunning ContainerState = "running"
 
 	// StateStopped indicates that the container process has exited
-	StateStopped ContainerState  = "stopped"
+	StateStopped ContainerState = "stopped"
 )
 
 // State holds information about the runtime state of the container.
@@ -32,4 +32,25 @@ type State struct {
 	Bundle string `json:"bundle"`
 	// Annotations are key values associated with the container.
 	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
+const (
+	// SeccompFdName is the name of the seccomp notify file descriptor.
+	SeccompFdName string = "seccompFd"
+)
+
+// ContainerProcessState holds information about the state of a container process.
+type ContainerProcessState struct {
+	// Version is the version of the specification that is supported.
+	Version string `json:"ociVersion"`
+	// Fds is a string array containing the names of the file descriptors passed.
+	// The index of the name in this array corresponds to index of the file
+	// descriptor in the `SCM_RIGHTS` array.
+	Fds []string `json:"fds"`
+	// Pid is the process ID as seen by the runtime.
+	Pid int `json:"pid"`
+	// Opaque metadata.
+	Metadata string `json:"metadata,omitempty"`
+	// State of the container.
+	State State `json:"state"`
 }
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -61,7 +61,8 @@
                 "SCMP_ACT_ERRNO",
                 "SCMP_ACT_TRACE",
                 "SCMP_ACT_ALLOW",
-                "SCMP_ACT_LOG"
+                "SCMP_ACT_LOG",
+                "SCMP_ACT_NOTIFY"
             ]
         },
         "SeccompFlag": {
@@ -69,7 +70,8 @@
             "enum": [
                 "SECCOMP_FILTER_FLAG_TSYNC",
                 "SECCOMP_FILTER_FLAG_LOG",
-                "SECCOMP_FILTER_FLAG_SPEC_ALLOW"
+                "SECCOMP_FILTER_FLAG_SPEC_ALLOW",
+                "SECCOMP_FILTER_FLAG_NEW_LISTENER"
             ]
         },
         "SeccompOperators": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -598,11 +598,13 @@ type VMImage struct {
 
 // LinuxSeccomp represents syscall restrictions
 type LinuxSeccomp struct {
-	DefaultAction   LinuxSeccompAction `json:"defaultAction"`
-	DefaultErrnoRet *uint              `json:"defaultErrnoRet,omitempty"`
-	Architectures   []Arch             `json:"architectures,omitempty"`
-	Flags           []LinuxSeccompFlag `json:"flags,omitempty"`
-	Syscalls        []LinuxSyscall     `json:"syscalls,omitempty"`
+	DefaultAction    LinuxSeccompAction `json:"defaultAction"`
+	DefaultErrnoRet  *uint              `json:"defaultErrnoRet,omitempty"`
+	Architectures    []Arch             `json:"architectures,omitempty"`
+	Flags            []LinuxSeccompFlag `json:"flags,omitempty"`
+	ListenerPath     string             `json:"listenerPath,omitempty"`
+	ListenerMetadata string             `json:"listenerMetadata,omitempty"`
+	Syscalls         []LinuxSyscall     `json:"syscalls,omitempty"`
 }
 
 // Arch used for additional architectures


config-linux.md
```diff
@@ -685,6 +699,48 @@ The following parameters can be specified to set up seccomp:
...
+* **`state`** (map, REQUIRED) is the [state](runtime.md#state) of the container.
+
+Example:
+
+```json
+{
+    "ociVersion": "0.2.0",
+    "seccompFd": 0,
+    "pid": 4422,
+    "pidFd": 1,
```
giuseppe, line 727: I am not sure about implementing `pidFd` as part of seccomp notifications.

There are cases where pidfd is useful also without seccomp notifications, should we have a separate way for passing the pidfd?

config-linux.md
```diff
@@ -685,6 +699,48 @@ The following parameters can be specified to set up seccomp:
 }
 ```
 
+### <a name="seccompstate" />The Seccomp State
```
mauriciovasquezbernal, line 702: nit: Missing empty line.

config-linux.md
```diff
@@ -685,6 +699,48 @@ The following parameters can be specified to set up seccomp:
...
+The seccomp state is a data structure passed via a UNIX socket.
+The container runtime MUST send the seccomp state over the UNIX socket as regular payload.
```
alban, line 704:
```suggestion
The container runtime MUST send the seccomp state over the UNIX socket as regular payload serialized in JSON.
```

config-linux.md
```diff
@@ -685,6 +699,48 @@ The following parameters can be specified to set up seccomp:
...
+The container runtime MUST also send the file descriptor(s) via `SCM_RIGHTS`: the seccomp file descriptor returned by the seccomp syscall and, optionally, the process file descriptor (e.g as returned by `pidfd_open(2)` or by `clone(2)` with the `CLONE_PID` flag).
+The container runtime MAY use several `sendmsg(2)` calls to send the aforementioned data.
+If more than one `sendmsg(2)` is used, the file descriptors MUST be sent only in the first call.
+
+The seccomp state includes the following properties:
+
+* **`ociVersion`** (string, REQUIRED) is version of the Open Container Initiative Runtime Specification with which the seccomp state complies.
+* **`seccompFd`** (int, REQUIRED) is the index of the file descriptor in the `SCM_RIGHTS` array refering to the seccomp notify file descriptor.
```
alban, line 712:
```suggestion
* **`seccompFd`** (int, REQUIRED) is the index of the file descriptor in the `SCM_RIGHTS` array referring to the seccomp notify file descriptor.
```

config-linux.md
```diff
@@ -685,6 +699,48 @@ The following parameters can be specified to set up seccomp:
...
+    The value MUST be 0.
+* **`pid`** (int, REQUIRED) is the process ID, as seen by the runtime, on which the seccomp filter is applied (target process).
+* **`pidFd`** (int, OPTIONAL) is the index of the file descriptor in the `SCM_RIGHTS` array refering to the target process file descriptor.
```
alban, line 715:
```suggestion
* **`pidFd`** (int, OPTIONAL) is the index of the file descriptor in the `SCM_RIGHTS` array referring to the target process file descriptor.
```

config-linux.md
```diff
@@ -685,6 +699,48 @@ The following parameters can be specified to set up seccomp:
...
+    If present, this value MUST NOT be zero.
+* **`metadata`** (string, OPTIONAL) is the string set in `listenerMetadata`.
```
alban, line 717:
```suggestion
* **`metadata`** (string, OPTIONAL) is the string set in `listenerMetadata`. If the `listenerMetadata` is set, then the runtime MUST set this field too.
```

config-linux.md
alban: I would be fine if the spec define another separate way for passing pidfds.

But I feel it is useful here as well. Note that this `pidFd` is not necessarily referring to the "pid 1" of the container, but in the case of "runc exec", it is the pid that entered in the container. I don't see how we can have a separate way for passing the pidfd in that case.

specs-go/state.go
```diff
@@ -33,3 +33,19 @@ type State struct {
 	// Annotations are key values associated with the container.
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
+
+type SeccompState struct {
+	// Version is the version of the specification that is supported.
+	Version string `json:"ociVersion"`
+	// SeccompFd is the file descriptor for Seccomp User Notification
```
alban, line 40:
```suggestion
	// SeccompFd is the index of the file descriptor in the
	// `SCM_RIGHTS` array referring to the seccomp notify file
	// descriptor. It is always zero (the first entry in the array).
```

specs-go/state.go
```diff
@@ -33,3 +33,19 @@ type State struct {
...
+	SeccompFd int `json:"seccompFd"`
+	// Pid is the process ID on which the seccomp filter is applied
+	Pid int `json:"pid"`
+	// PidFd is a pidfd for the process on which the seccomp filter is
+	// applied
```
alban, line 45:
```suggestion
	// PidFd is is the index of the file descriptor in the
	// `SCM_RIGHTS` array referring to the target process file
	// descriptor (e.g as returned by `pidfd_open(2)` or by
	// `clone(2)` with the `CLONE_PID` flag).
```

specs-go/state.go
```diff
@@ -33,3 +33,19 @@ type State struct {
...
+	PidFd int `json:"pidFd,omitempty"`
+	// Opaque metadata
```
alban, line 47:
```suggestion
	// Opaque metadata copied from the listenerMetadata field.
```

alban (commented):
I am adding some suggestions that I didn't notice before.

config-linux.md
giuseppe: what do you think if we had something not tied to seccomp like?

```
{
    "ociVersion": "0.2.0",
    "fds" : {
       "seccompFd" : 0,
       "pidFd": 1,
       "futureFeatureFd" : 2
    }
    "pid": 4422,
    "state": {
        "ociVersion": "0.2.0",
        "id": "oci-container1",
        "status": "creating",
        "pid": 4422,
        "bundle": "/containers/redis",
        "annotations": {
            "myKey": "myValue"
        }
    }
}
```
Would that work as well?

I am thinking we already have a way to pass the tty fd (with `--console-socket`) and we could perhaps cover it as part of the OCI configuration instead of having it as a CLI option

config-linux.md
rata: Fixed, thanks!

config-linux.md
rata: Fixed, thanks!

config-linux.md
rata: Fixed, thanks!

config-linux.md
rata: Thanks, fixed!

specs-go/state.go
rata: Oh, nice catch. I forgot to update these :see_no_evil:.

Fixed, thanks!

specs-go/state.go
rata: Fixed, thanks! I also updated the description of the `pid` file to include is the pid as seen by the runtime (as config-linux.md says)

specs-go/state.go
rata: Fixed, thanks!

config-linux.md
rata: Fixed, thanks!

config-linux.md
rata: Hm, I've added this, but the hooks, that also receive a state in JSON, don't say it is serialized in JSON. From https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-hooks:

> The state of the container MUST be passed to hooks over stdin so that they may do work appropriate to the current state of the container.

Then, [the state](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#state) says:
> When serialized in JSON, the format MUST adhere to the following pattern

So, I'm unsure if the spec says in some general way (like the way it says with unknown fields should be ignored) it should be serialized as JSON and I'm not finding it now.

Do you know?

I added it just in case, but please let me know if this is covered by some section already

config-linux.md
rata: @giuseppe maybe this is a silly question, I think I'm a bit lost. How will the seccomp agent know about the new containers? Do you mean to keep the listenerPath and listenerMatadata in the seccomp section but send this more generic state instead of a seccomp state? In that case, how will be useful for other (non-seccomp) uses? Or do you mean to add fields to the state (not as your example json, as they are not part of the state) and use something like [this](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#query-state)?

I'm not sure I follow what you mean, or how we can use it from a seccomp agent that doesn't know when runc creates a container or "runc exec" is run in a container, so we have two (or more) pidfds to report: the fd for pid 1 of the container and the fd for the process that run "runc exec".

I'll try to wake up and read it again, just in case :)

giuseppe: my suggestion was just about using a map for holding the file descriptors and pass the same metadata.  In this way we could extend it in future, if we'll need any new kind of fd.

alban: It sounds good to me to use a `map[string]int` instead of individual fields for seccompFd and pidFd.

If you want to make it more generic, I guess you want to rename `type SeccompState` in `specs-go/state.go`... Do you have an idea of new name?

rata: @giuseppe ohh, sorry, I see now! SGTM too, seems like a nice idea. I'll adapt to create a "fds" section in the seccomp state.

Let me know if you want to rename the seccomp state to some other (more generic) thing as alban suggested :)

rata: @giuseppe I've moved to the map as you suggested. I couldn't find a good name for [the struct](https://github.com/opencontainers/runtime-spec/pull/1074/files#diff-a1246a93f692debc1727ae5655763e78b2f1662f8cded2473d1403ae2b24d035R37-R48).

What name do you suggest? :)

giuseppe: can't really think of anything good to suggest.  Maybe `ContainerState` or `ContainerProcessState`?

mauriciovasquezbernal: @giuseppe I just updated the PR to call it `ContainerProcessState`. Please let us know what you think. Thanks!

alban:
runc has a draft PR (https://github.com/opencontainers/runc/pull/2682) implementing this.

config-linux.md
```diff
@@ -625,6 +625,22 @@ The following parameters can be specified to set up seccomp:
     * `SECCOMP_FILTER_FLAG_LOG`
     * `SECCOMP_FILTER_FLAG_SPEC_ALLOW`
 
+* **`listenerPath`** *(string, OPTIONAL)* - specifies the path of UNIX domain socket over which the runtime will send the [container process state](#containerprocessstate) data structure when the `SCMP_ACT_NOTIFY` action is used.
+    This socket MUST use `AF_UNIX` domain and `SOCK_STREAM` type.
+    The runtime MUST send exactly one [container process state](#containerprocessstate) per connection.
+    The connection MUST NOT be reused and it MUST be closed after sending a seccomp state.
+    If sending to this socket fails, the runtime MUST [generate an error](runtime.md#errors).
```
sargun, line 632: Do we want to allow for "fail-open" or "fail-closed behaviour?

config-linux.md
```diff
@@ -625,6 +625,22 @@ The following parameters can be specified to set up seccomp:
...
+    If the `SCMP_ACT_NOTIFY` action is not used this value is ignored.
```
sargun, line 633: Shouldn't we send it no matter what? This can also be used as a "poor man's" mechanism of tracking container lifetime.

config-linux.md
```diff
@@ -625,6 +625,22 @@ The following parameters can be specified to set up seccomp:
...
+
+    The runtime sends the following file descriptors using `SCM_RIGHTS` and set their corresponding indexes in the `fdIndexes` map of the [container process state](#containerprocessstate):
+
+    * **`seccompFd`** (int, REQUIRED) is the seccomp file descriptor returned by the seccomp syscall.
+    * **`pidFd`** (int, OPTIONAL) is the process file descriptor (e.g as returned by `pidfd_open(2)` or by `clone(2)` with the `CLONE_PID` flag).
```
sargun, line 638: This is the pidfd of the container's pid 1?

config-linux.md
sargun: There should be a version or something, given that container process state appears to be extensible. The receiving end has to know how many FDs to receive *before* it receives them. Unless we have some kind of negotiation mechanism which first sends the version to enable the receiving end to know what to expect.

specs-go/state.go
```diff
@@ -33,3 +33,27 @@ type State struct {
...
+	// PidFdIndexKey is the index of the target process file descriptor.
+	PidFdIndexKey FdIndexKey = "pidFd"
+)
+
+// ContainerProcessState holds information about the state of a container process.
+type ContainerProcessState struct {
+	// Version is the version of the specification that is supported.
+	Version string `json:"ociVersion"`
+	// FdIndexes is a map containing the indexes of the file descriptors in the `SCM_RIGHTS` array.
+	FdIndexes map[FdIndexKey]int `json:"fdIndexes"`
```
sargun, line 52: why isn't this just a normal array?

sargun (commented):
One question about the version negotiation / how many FDs to allocate on the receiving side prior to doing the receive, but otherwise lgtm.

config-linux.md
```diff
@@ -625,6 +625,22 @@ The following parameters can be specified to set up seccomp:
     * `SECCOMP_FILTER_FLAG_LOG`
     * `SECCOMP_FILTER_FLAG_SPEC_ALLOW`
 
+* **`listenerPath`** *(string, OPTIONAL)* - specifies the path of UNIX domain socket over which the runtime will send the [container process state](#containerprocessstate) data structure when the `SCMP_ACT_NOTIFY` action is used.
+    This socket MUST use `AF_UNIX` domain and `SOCK_STREAM` type.
```
sargun, line 629: Wait, why not SOCK_SEQPACKET, or DGRAM?

config-linux.md
alban: > Do we want to allow for "fail-open" or "fail-closed behaviour?

IMO I think it should fail if sending the message to the seccomp agent fails for some reason (e.g. it is not running, or permission denied to connect to the socket). In that way, we don't make debugging difficult by accidentally starting containers that would return ENOSYS for some system calls.

I want to be able to run a seccomp agent in the following setups:
- as a socket activatable systemd service. If it is not started yet, it would be started on demand
- as a DaemonSet on Kubernetes. If it is not started yet, other pods that need the seccomp agent would fail. But pods that don't need it (if they use a seccomp policy without any `SCMP_ACT_NOTIFY` rule), they would start fine. So in that case, the seccomp agent DaemonSet would not have a chicken-egg problem.

config-linux.md
alban: I think it should not:
- so containers that don't use `SCMP_ACT_NOTIFY` policies would continue to work if the seccomp agent is not running or not installed.
- libseccomp automatically enables or disables `SECCOMP_FILTER_FLAG_NEW_LISTENER`  depending on whether we have at least one rule in the seccomp policy with `SCMP_ACT_NOTIFY`. So if container runtime implementations use libseccomp, they would not have a seccomp-fd to send without `SCMP_ACT_NOTIFY` rules. We make the implementation easier by following this (= no need to patch libseccomp).

config-linux.md
alban: Either the container's pid 1, or the pid that entered the container with `exec` (`runc-exec`/`docker-exec`/`kubectl-exec`).

specs-go/state.go
alban: Do you mean like this?

```suggestion
	Fds []FdKey `json:"fds"`
```

So it would look like this?
```json
{
    "ociVersion": "0.2.0",
    "fds": [
        "seccompFd",
        "pidFd"
    ],
    "pid": 4422,
    ...
}
```

I think that should work too.

The map structure comes from the previous implementation as an OCI hook. But I agree it's no longer necessary here.

config-linux.md
rata: Yeah, I don't see any reason to allow it. Also, why not use the hooks for poor man's container lifecycle tracking?

specs-go/state.go
rata: No, I don't think that will work, @alban . If tomorrow another optional fd is added (let's call it newFD), you don't have any way to know if the second element in the array is the pidFd or newFD.

@sargun for some context, in the previous iteration of this patch we had a top level field for each fd. Giuseppe asked to make this a map here (https://github.com/opencontainers/runtime-spec/pull/1074#discussion_r518651960).

config-linux.md
rata: Regarding not failing on sending to the unix socket, what is the point of making it not fail? I see only downsides, it makes no sense to start containers and just throw the seccomp fds it needs to work to `/dev/null`. Am I missing something?

specs-go/state.go
rata: Ohh, sorry, I missed this was an array of strings. So, for the example I used it would be something like: ["seccompfd", "newFD"].

That can work too. Not sure about the type, though :)

config-linux.md
rata: Maybe we can add that to the description? Something like: `This refers to the container's pid 1 if the container is being created or to the process that entered the container with exec`.

Would that make it clearer?

specs-go/state.go
mauriciovasquezbernal: I updated the PR to use an array instead of a map.

config-linux.md
```diff
@@ -686,6 +703,45 @@ The following parameters can be specified to set up seccomp:
...
+The container runtime MAY use several `sendmsg(2)` calls to send the aforementioned data.
+If more than one `sendmsg(2)` is used, the file descriptors MUST be sent only in the first call.
+
+The container processs state includes the following properties:
+
+* **`ociVersion`** (string, REQUIRED) is version of the Open Container Initiative Runtime Specification with which the container processs state complies.
+* **`fds`** (array, OPTIONAL) is a string array containing the names of the file descriptors passed. The index of the name in this array corresponds to index of the file descriptor the `SCM_RIGHTS` array.
+* **`pid`** (int, REQUIRED) is the container process ID, as seen by the runtime.
+* **`metadata`** (string, OPTIONAL) opaque metadata.
+* **`state`** (map, REQUIRED) is the [state](runtime.md#state) of the container.
```
vbatts, line 719: maybe where you have `map` as the type, put the `state` as a link to that struct definition

config-linux.md
```diff
@@ -686,6 +703,45 @@ The following parameters can be specified to set up seccomp:
...
+### <a name="containerprocessstate" />The Container Process State
+
+The container processs state is a data structure passed via a UNIX socket.
+The container runtime MUST send the container processs state over the UNIX socket as regular payload serialized in JSON and file descriptors MUST be send using `SCM_RIGHTS`.
+The container runtime MAY use several `sendmsg(2)` calls to send the aforementioned data.
+If more than one `sendmsg(2)` is used, the file descriptors MUST be sent only in the first call.
+
+The container processs state includes the following properties:
+
+* **`ociVersion`** (string, REQUIRED) is version of the Open Container Initiative Runtime Specification with which the container processs state complies.
```
vbatts, line 715: nit: bullet list need not have periods at the end

vbatts:
On the whole, this seems fine and developed enough. (it helps to have worked through the implementation on runc, and related work already done in crun).
By being in the spec, we can have consistent behavior across runtimes. That's the point.

This is not a breaking change, as the usage and fields are optional.
:+1:

config-linux.md
mauriciovasquezbernal: Isn't it the other way around? I see end periods everywhere :thinking:

config-linux.md
mauriciovasquezbernal: Good idea. Done.

mauriciovasquezbernal:
@giuseppe  @AkihiroSuda @KentaTada @ManaSugi I discussed this very briefly at the last [OCI meeting](https://www.youtube.com/watch?v=Qzwa6R0uZXQ&feature=youtu.be&t=1470). 

Since AFAIK you're the ones more interested on this, I'd like to get your feedback and if there are any concerns to move this forward. Thanks!

config-linux.md
```diff
@@ -625,6 +625,22 @@ The following parameters can be specified to set up seccomp:
...
+* **`listenerPath`** *(string, OPTIONAL)* - specifies the path of UNIX domain socket over which the runtime will send the [container process state](#containerprocessstate) data structure when the `SCMP_ACT_NOTIFY` action is used.
+    This socket MUST use `AF_UNIX` domain and `SOCK_STREAM` type.
+    The runtime MUST send exactly one [container process state](#containerprocessstate) per connection.
+    The connection MUST NOT be reused and it MUST be closed after sending a seccomp state.
+    If sending to this socket fails, the runtime MUST [generate an error](runtime.md#errors).
+    If the `SCMP_ACT_NOTIFY` action is not used this value is ignored.
+
+    The runtime sends the following file descriptors using `SCM_RIGHTS` and set their names in the `fds` array of the [container process state](#containerprocessstate):
+
+    * **`seccompFd`** (int, REQUIRED) is the seccomp file descriptor returned by the seccomp syscall.
```
giuseppe, line 637: should *seccompFd* be optional as well?

config-linux.md
```diff
@@ -686,6 +703,45 @@ The following parameters can be specified to set up seccomp:
 }
 ```
 
+### <a name="containerprocessstate" />The Container Process State
+
+The container processs state is a data structure passed via a UNIX socket.
```
giuseppe, line 708: typo in `processs`

config-linux.md
```diff
@@ -686,6 +703,45 @@ The following parameters can be specified to set up seccomp:
...
+The container runtime MUST send the container processs state over the UNIX socket as regular payload serialized in JSON and file descriptors MUST be send using `SCM_RIGHTS`.
+The container runtime MAY use several `sendmsg(2)` calls to send the aforementioned data.
+If more than one `sendmsg(2)` is used, the file descriptors MUST be sent only in the first call.
+
+The container processs state includes the following properties:
```
giuseppe, line 713: same typo here

config-linux.md
alban: I don't mind if seccompFd is optional in the spec, but how would it work if the runtime does not pass the seccompFd? Do you want to make it optional for runtimes that don't want to implement seccomp-notify, or are you thinking of an alternative way to obtain the seccompFd?

For runtimes that use seccomp-notify but implements the seccomp agent internally, they can just not set the `listenerPath` field.

The "Container Process State" section is separate and does not mandate the seccompFd, but this section is in the description of "listenerPath".

config-linux.md
alban: It seems good to me to add this description.

config-linux.md
giuseppe: could it be possible to use it for retrieving `pidFd` (or anything more that is added in future) but not `seccompFd`?  e.g. what happens if the profile doesn't set any rule with `SECCOMP_ACTION_NOTIFY`?  Does the OCI runtime still need to create a  `seccompFd`?

config-linux.md
rata: nice catch, fixed locally. Will push soon!

config-linux.md
rata: nice catch, fixed locally. Will push soon!

config-linux.md
rata: If the profile doesn't set any notify rule, we propose this field should be ignored by the runtime. See here: https://github.com/opencontainers/runtime-spec/pull/1074/files/d87dd02ab1435993552f72fd1d98c7779f67228a#diff-048d23d864e15683f516d2c1768965d546e87f8a59b2606cf2f2d52500ba5a32R633.

Our idea is to use this UNIX socket to send to a seccomp agent. Do you want to make this field more "generic"? Be aware that this is in the seccomp section, we can reuse the container state process in other parts of the spec, but this should be about seccomp.

specs-go/state.go
rata: Thanks. Marking this as solved, then

config-linux.md
rata: Added it!

config-linux.md
giuseppe: so do you think we should move the `listenerPath` outside the `seccomp` section?  I think it should be possible to use `pidFd` without seccomp notify, or seccomp in general

alban: > so do you think we should move the listenerPath outside the seccomp section?

I think we should not; it would make it more difficult to use in Kubernetes (see https://github.com/opencontainers/runtime-spec/pull/1073#issuecomment-720524314).

> I think it should be possible to use pidFd without seccomp notify, or seccomp in general

If you want a mechanism to get the pidFd without seccomp, I'd rather doing that separately in a separate field than `seccomp.listenerPath`, even if it reuses this new "Container Process State" type.

There can be only one seccomp agent receiving the seccompFd (by design from https://github.com/opencontainers/runtime-spec/pull/1073#discussion_r514382166). But I see use cases where several processes could receive the pidFd, so maybe it could be an array of listeners. Could this be a follow-up PR for this? In your use cases, would you want to receive the pidFd at container-create time or at container-start time?

rata: @giuseppe friendly ping? :)

rata:
@AkihiroSuda @KentaTada @ManaSugi Friendly ping? It would be super helpful for us if we can get a review for this :)

rhatdan:
@runtime-spec PTAL

rhatdan:
@mrunalp PTAL

config-linux.md
```diff
@@ -625,6 +625,23 @@ The following parameters can be specified to set up seccomp:
...
+    This socket MUST use `AF_UNIX` domain and `SOCK_STREAM` type.
+    The runtime MUST send exactly one [container process state](#containerprocessstate) per connection.
+    The connection MUST NOT be reused and it MUST be closed after sending a seccomp state.
+    If sending to this socket fails, the runtime MUST [generate an error](runtime.md#errors).
+    If the `SCMP_ACT_NOTIFY` action is not used this value is ignored.
+
+    The runtime sends the following file descriptors using `SCM_RIGHTS` and set their names in the `fds` array of the [container process state](#containerprocessstate):
+
+    * **`seccompFd`** (string, REQUIRED) is the seccomp file descriptor returned by the seccomp syscall.
+    * **`pidFd`** (string, OPTIONAL) is the process file descriptor (e.g as returned by `pidfd_open(2)` or by `clone(2)` with the `CLONE_PID` flag).
```
sargun, line 638: So, originally, you needed to support pidfd, because seccompfd left you with no way to determine if the container had been terminated or not. Now that there is tracking for that, it might be required. Do you think we should keep it?

sargun:
LGTM. Previously, there was the issue that you needed the PIDFD in order to track the container's lifetime. Now that we no longer need to do that because the epoll call will return EPOLLHUP (https://github.com/torvalds/linux/blob/master/kernel/seccomp.c#L1697-L1698) on the fd if there are no more containers, do we still need that?

In addition to that, I fear that if we embed the pid into the metadata, it will be problematic if the notifier is running in its own pid ns (for example a daemonset?). Do we want to ensure that the process that is sending the is the one that matches the pid, therefore enabling the use of SCM_CREDENTIALS to find out the pid of the other side? Or at least being explicit and saying which process will send this?

config-linux.md
giuseppe: I still think the `listenerPath` should be outside the seccomp section so it can be used for multiple kind of fds

schema/config-linux.json
```diff
@@ -209,6 +209,12 @@
                             "$ref": "defs-linux.json#/definitions/SeccompFlag"
                         }
                     },
+                    "listenerPath": {
+                        "type": "string"
+                    },
+                    "listenerMetadata": {
+                        "type": "string"
+                    },
```
giuseppe, line 217: same here, I'd move this outside the seccomp section so it can be used also without seccomp

rata:
@sargun
> LGTM.

Thanks for the review!

> Previously, there was the issue that you needed the PIDFD in order to track the container's lifetime. Now that we no longer need to do that because the epoll call will return EPOLLHUP (https://github.com/torvalds/linux/blob/master/kernel/seccomp.c#L1697-L1698) on the fd if there are no more containers, do we still need that?

Good point. I think it can be useful anyways, though. The pidfd refers to the container 1 process if a container is being created, or to the process doing “kubectl exec”. I think in some cases (specially the pid of the process connected via an “exec” might be tricker to get) is useful, and if the agent will not use it can just close it.

> In addition to that, I fear that if we embed the pid into the metadata, it will be problematic if the notifier is running in its own pid ns (for example a daemonset?). Do we want to ensure that the process that is sending the is the one that matches the pid, therefore enabling the use of SCM_CREDENTIALS to find out the pid of the other side? Or at least being explicit and saying which process will send this?

Well, the same thing happens with the hooks. Just running the agent in the host pid will do the trick. In a daemonset, just using `hostPID: true` in the daemon set will do it, we do that [here](https://github.com/kinvolk/seccompagent/blob/b2d423695d6dfc976d2456769acb19765a9d7524/deploy/seccompagent.yaml#L37) for example.

Also, I don’t think we want to ensure that the sender is the pid. Runc implementation is not doing that. And that's not doable today if sendmsg() is a syscall blocked by the seccomp policy, for example. In addition, it can also create issues for rootless in the future (in runc, it is sent by the runc parent and you need to have CAP_SYS_ADMIN to send a different pid in SCM_CREDENTIALS, which will complicate rootless, etc.).

Furthermore, I don’t think the the spec needs to be explicit about which process is sending the message, it just needs to say that the pid is from the container runtime namespace. The PR is already explicit about that, so IMHO it is fine: https://github.com/opencontainers/runtime-spec/pull/1074/files#diff-048d23d864e15683f516d2c1768965d546e87f8a59b2606cf2f2d52500ba5a32R719

What do you think?

config-linux.md
rata: @sargun marking as solved as you recently reviwed with a LGTM. Please feel free to reopen if you think this isn't addressed :)

config-linux.md
rata: @sargun marking as solved as you recently reviwed with a LGTM. Please feel free to reopen if you think this isn't addressed :)

config-linux.md
rata: @sargun marking as solved as you recently reviwed with a LGTM. Please feel free to reopen if you think this isn't addressed :)

sargun:
> @sargun
> 
> > LGTM.
> 
> Thanks for the review!
> 
> > Previously, there was the issue that you needed the PIDFD in order to track the container's lifetime. Now that we no longer need to do that because the epoll call will return EPOLLHUP (https://github.com/torvalds/linux/blob/master/kernel/seccomp.c#L1697-L1698) on the fd if there are no more containers, do we still need that?
> 
> Good point. I think it can be useful anyways, though. The pidfd refers to the container 1 process if a container is being created, or to the process doing “kubectl exec”. I think in some cases (specially the pid of the process connected via an “exec” might be tricker to get) is useful, and if the agent will not use it can just close it.
> 
That's why I suggested the approach / description below, where the receiving process can use SCM_CREDENTIALS to determine who sent the FD. Also, reducing it down to one FD simplifies things.

> > In addition to that, I fear that if we embed the pid into the metadata, it will be problematic if the notifier is running in its own pid ns (for example a daemonset?). Do we want to ensure that the process that is sending the is the one that matches the pid, therefore enabling the use of SCM_CREDENTIALS to find out the pid of the other side? Or at least being explicit and saying which process will send this?
> 
> Well, the same thing happens with the hooks. Just running the agent in the host pid will do the trick. In a daemonset, just using `hostPID: true` in the daemon set will do it, we do that [here](https://github.com/kinvolk/seccompagent/blob/b2d423695d6dfc976d2456769acb19765a9d7524/deploy/seccompagent.yaml#L37) for example.
> 
> Also, I don’t think we want to ensure that the sender is the pid. Runc implementation is not doing that. And that's not doable today if sendmsg() is a syscall blocked by the seccomp policy, for example. In addition, it can also create issues for rootless in the future (in runc, it is sent by the runc parent and you need to have CAP_SYS_ADMIN to send a different pid in SCM_CREDENTIALS, which will complicate rootless, etc.).

We need to solve this problem either way. The seccomp policy is installed at the time the listenerfd is created. Either you need to do clone3 and not unshare the fd table, or sendmsg the listener FD back (or pidfd getfd shenanigans). 
> 
> Furthermore, I don’t think the the spec needs to be explicit about which process is sending the message, it just needs to say that the pid is from the container runtime namespace. The PR is already explicit about that, so IMHO it is fine: https://github.com/opencontainers/runtime-spec/pull/1074/files#diff-048d23d864e15683f516d2c1768965d546e87f8a59b2606cf2f2d52500ba5a32R719
> 
> What do you think?
Doesn't it have to be the pid1 of the container? Or the process which will fork to become the pid1 of the container because otherwise you cannot ensure the seccomp policy will be inherited by all processes in the container?

----

I imagine the listener flow would work something like this:
1. Receive notification
2. Recvmsg + ancillary data (SCM_RIGHTS + SCM_CREDENTIALS)
3. Take the notifier FD from SCM_RIGHTS, and the PID from SCM_CREDENTIALS
4. Call pidfd_open on the pid retrieved from SCM_CREDENTIALS
5. epoll on the pidfd + listener fd.
6. When the listener FD generates a SIGHUP, tear down.

giuseppe:
I had a private chat with @rata.  My only concern is about having the `listenerPath` under the seccomp section as I think it is a more generic functionality to pass FDs to the upper layer.  For example, I think it should be possible to use `pidfd` without requiring seccomp at all.

We already have a way of passing the tty FD through `--console-socket`.   This could be supported as well through a more generic `listenerPath`.

On the other hand, having `listenerPath` in the seccomp block makes the integration with with Kubernetes easier as the entire configuration can live on its own.

What do you think if we add (with a separate PR) a `listenerPath` field to the outer scope as well that can be used with the same semantic without requiring seccomp?  @opencontainers/runtime-spec-maintainers what do you think?

cyphar:
I think having a more generic (**and part of the spec**) mechanism for passing file descriptors back to the parent would be an improvement over the current situation. However it should be noted that this problem mostly exists because we don't have a container manager (like `conmon` or LXC's monitor process) as part of runc's implementation -- and VM-based runtimes are going to struggle to emulate this behaviour as well..

It would also be great if the `sendmsg` data actually had useful information (such as the container name and what type of data is being sent) so that you can have a single socket ingest many file descriptors rather than having to set up a new one for each container (my first draft of runc's `--console-socket` implemented this but we dropped it because we weren't sure if it would be useful).

rata:
@sargun 
> I imagine the listener flow would work something like this:
> 
>     1. Receive notification
> 
>     2. Recvmsg + ancillary data (SCM_RIGHTS + SCM_CREDENTIALS)
> 
>     3. Take the notifier FD from SCM_RIGHTS, and the PID from SCM_CREDENTIALS
> 
>     4. Call pidfd_open on the pid retrieved from SCM_CREDENTIALS
> 
>     5. epoll on the pidfd + listener fd.
> 
>     6. When the listener FD generates a SIGHUP, tear down.

Ohh, I see what you mean now. This won’t work in the [runc draft implementation](https://github.com/opencontainers/runc/pull/2682) as the process sending the seccomp fd is not the one that did the seccomp() syscall. Currently the process that sends on the unix socket [gets the seccompFd with pidfd_getfd](https://github.com/opencontainers/runc/pull/2682/files#diff-f0214a0f16408fc7f168c6fc9837d189590025cc1813ebf7c1d751136936dfbfR172) and then sends it over the unix socket with the ociState using SCM_RIGHTS.

So, if the receiver enables SCM_CREDENTIALS on the socket (SO_PASSCRED), it will

 see the pid of this other process in the current draft implementation. To illustrate, I happen to have an image handy of the current runc diagram, just in case: runc-seccomp-notify

To do what you describe (using the jargon of the attached image now), we need the runc child to send over the unix socket for SCM_CREDENTIALs to give what you expect. But that is not possible, as we run the seccomp syscall as late as possible (as filters apply since we run it) and send over the unix socket afterwards (we need the fd seccomp() gives us). The issue that arise from that is that as late as possible is already after pivot_root (so, we can’t write to the unix path), already running as a different user (so maybe don’t have permissions to write to the unix socket) and after SELinux context is applied (that might block that), etc. And of course, we can’t run the seccomp syscall earlier, as it can block the pivot_root and the rest of the container setup.

None of these issues happen if we send from the runc parent. The seccomp agent can use SO_PASSCRED in the receiving unix socket (and validate the UID/GID, not pid), if the agent wants to do that, though.

I’d avoid forcing the runc parent to write the proper pid in SCM_CREDENTIALS too, as that would force it to run with CAP_SYS_ADMIN and will be just impossible to sort out for rootless mode.

@sargun What do you think?

rata:
@cyphar 

> I think having a more generic (**and part of the spec**) mechanism for passing file descriptors back to the parent would be an improvement over the current situation. 

You mean the listenerPath in the seccomp section (as this PR adds) and another in the Linux seccion as @giuseppe suggested?

> It would also be great if the `sendmsg` data actually had useful information (such as the container name and what type of data is being sent) so that you can have a single socket ingest many file descriptors rather than having to set up a new one for each container (my first draft of runc's `--console-socket` implemented this but we dropped it because we weren't sure if it would be useful).

This is the case for the seccomp section in this PR. We add that info, so a single seccomp agent can handle several containers :)

rata:
@giuseppe @cyphar  Hi!

I lack some context about the linux-scope listenerPath. But if for the seccomp case we just send the seccompFd (not the pidfd), would it solve the concerns here? 

Just to be clear I mean this:
 * We keep a listenerPath in seccomp section
 * The container state only sends the seccompfd. In other words, we remove [this line](https://github.com/opencontainers/runtime-spec/pull/1074/files#diff-048d23d864e15683f516d2c1768965d546e87f8a59b2606cf2f2d52500ba5a32R638).

Would that solve the issues? As [Sargun pointed out](https://github.com/opencontainers/runtime-spec/pull/1074#issuecomment-749241338), we can be without the pidFd.

If the listenerPath is in the linux section instead of seccomp, for the seccomp use case some problems arise. I can expand on this, but don’t want to be repetitive (reasoning is linked in the PR description, etc. :).

I’d like to do the best spec change here, but I lack the context of what the use cases of the listenerPath at the linux section are, to know what is better (maybe having them separate is a good separation of concerns, maybe not, depends on what we want in the linux section :)).

Can someone please give me more details on that so I can propose something that addresses (most) concerns? :)


Thanks a lot,
Rodrigo

rata:
@giuseppe @cyphar friendly ping?

I'd love to move forward with this. Is there anything I can do to help? I'm open to do a meeting (or add it to the OCI weekly meeting whenever you can attend) if that is better than feedback on the PR for you.

Please let me know if I can help in any way to move this forward :)

Thanks!

giuseppe:
> @giuseppe @cyphar Hi!
> 
> I lack some context about the linux-scope listenerPath. But if for the seccomp case we just send the seccompFd (not the pidfd), would it solve the concerns here?

but once we want to support pidfd we will need to add another listenerPath.

I agree with what @cyphar said here:



> I think having a more generic (**and part of the spec**) mechanism for passing file descriptors back to the parent would be an improvement over the current situation. However it should be noted that this problem mostly exists because we don't have a container manager (like `conmon` or LXC's monitor process) as part of runc's implementation -- and VM-based runtimes are going to struggle to emulate this behaviour as well..
> 
> It would also be great if the `sendmsg` data actually had useful information (such as the container name and what type of data is being sent) so that you can have a single socket ingest many file descriptors rather than having to set up a new one for each container (my first draft of runc's `--console-socket` implemented this but we dropped it because we weren't sure if it would be useful).


If we solve the issue once, then we won't even need to change the specs to support new FD types, such as the pidfd.

Is there any other way we can solve the issue with Kubernetes without adding the listenerPath to the seccomp section?

alban:
@giuseppe @cyphar Hi!

> but once we want to support pidfd we will need to add another listenerPath.

For the seccomp agent, we wouldn't need to add another listenerPath: a container runtime that supports sending the pidfd could just add it in the "fds" array, and the seccomp agent would receive it along with the seccomp-fd.

That would work just for the seccomp agent though, and not for other generic parents.

> Is there any other way we can solve the issue with Kubernetes without adding the listenerPath to the seccomp section?

That would not be ideal to me, because it would require more changes:

* In the Kubernetes PodSpec: instead of just a field in the securityContext to reference a file in `/var/lib/kubelet/seccomp/*.json`, it would need another field (or annotation) to reference the configuration for the listenerPath such as (`/var/lib/kubelet/seccomp-listenerpath/`)... That does not look clean to me.

* In containerd: instead of just revendoring the runtime-spec without any changes in the Go code, it would need to add new code to pick up the second configuration file into config.json. For reference, currently I only need [this commit](https://github.com/kinvolk/containerd/commit/4ea2052b9df5d65224c6be8cb8b443aa98b777dd) without other changes in Go code. The actual copy in [seccomp.go:WithProfile](https://github.com/containerd/containerd/blob/8dbe53a2a930af3631229e4d92cf839b64ee5a38/contrib/seccomp/seccomp.go#L36-L40) does not need changes

* Same in cri-o and any other runtimes.

Alternatively, you could consider that the listenerPath is part of the containerd configuration (`/etc/containerd/config.toml`), configured statically without possibility for Kubernetes to configure it. I would not like that because I would not be able to deploy a seccomp agent as a simple Kubernetes DaemonSet (the daemon set would need to patch containerd's config.toml and restart containerd -> not feasible on a live cluster).

> It would also be great if the sendmsg data actually had useful information (such as the container name and what type of data is being sent) so that you can have a single socket ingest many file descriptors rather than having to set up a new one for each container

I don't understand that part. This PR is already doing that: we can have a single seccomp agent process and it receives the sendmsg data that includes the container id, with other metadata (annotations, the kind of fds attached etc).

giuseppe:
@cyphar what do you think?

rata:
We talked with @giuseppe over slack after the New Kernel Features for Containers talk in Container Plumbing Days. 

We were talking that if the idea to have add the pidFd here and have the `listenerPath` in the `linux` section is for sending signals reliably (or other stuff like manage the container lifecycle), we can move forward with this PR as it is.

Having the listenerPath on the seccomp section, as it is now, seems nice because: seccomp can only have one supervisor, so no point in sending the seccomp-fd to another process that will pass it to one seccomp agent in the host (specially problematic if there are more than one agent, as different agents may implement handling for different syscalls) and we can later add a new `listenerPath` in the linux section too if needed. Both listenerPath can even point to the same path if we want, the message currently in the spec has all the metadata needed to know what it is about (annotations, which fds are sent, etc.).

Would that make sense for anyone else? If not, please let me know your concerns so I can do another proposal addressing them :)

@giuseppe please correct me if I got something wrong :)

giuseppe:
> Having the listenerPath on the seccomp section, as it is now, seems nice because: seccomp can only have one supervisor, so no point in sending the seccomp-fd to another process that will pass it to one seccomp agent in the host (specially problematic if there are more than one agent, as different agents may implement handling for different syscalls) and we can later add a new `listenerPath` in the linux section too if needed. Both listenerPath can even point to the same path if we want, the message currently in the spec has all the metadata needed to know what it is about (annotations, which fds are sent, etc.).

yes I think this makes sense since we send also all the needed metadata and the same listenerPath can be used both for "seccomp" and future "linux" FDs.

LGTM

rata:
I squashed the commits and fixed the conflict!

giuseppe (approved):
LGTM

vbatts:
Glad to see this evolve and that there is a corresponding implementation for it (https://github.com/opencontainers/runc/pull/2682).

LGTM

PR #1078, opened 2020-12-07
pandoc: point to a joint OCI org image

vbatts:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ OUTPUT_DIRNAME		?= output
 DOC_FILENAME		?= oci-runtime-spec
 DOCKER			?= $(shell command -v docker 2>/dev/null)
 PANDOC			?= $(shell command -v pandoc 2>/dev/null)
+
+PANDOC_CONTAINER ?= quay.io/oci/pandoc:1.17.0.3-2.fc25.x86_64
 ifeq "$(strip $(PANDOC))" ''
 	ifneq "$(strip $(DOCKER))" ''
 		PANDOC = $(DOCKER) run \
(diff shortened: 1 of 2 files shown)


alban:
LGTM. Thanks @vbatts

For the record, before this PR, Travis tests was sometimes failing with:
```
$ docker pull vbatts/pandoc
Using default tag: latest
Error response from daemon: received unexpected HTTP status: 500 Server Error
The command "docker pull vbatts/pandoc" failed and exited with 1 during .
```

Makefile
```diff
@@ -4,6 +4,8 @@ OUTPUT_DIRNAME		?= output
 DOC_FILENAME		?= oci-runtime-spec
 DOCKER			?= $(shell command -v docker 2>/dev/null)
 PANDOC			?= $(shell command -v pandoc 2>/dev/null)
+
+PANDOC_CONTAINER ?= quay.io/oci/pandoc:1.17.0.3-2.fc25.x86_64
```
tianon, line 8:
```suggestion
PANDOC_IMAGE ?= quay.io/oci/pandoc:1.17.0.3-2.fc25.x86_64
```

:innocent: 

Given this version string is now much more complex than trusting `vbatts/pandoc:latest`, would it also make sense to either have a `docker pull $(PANDOC_IMAGE)` target or perhaps have `install.tools` do that?

tianon: (To avoid the `.travis.yml` duplication, which will inevitably get out of sync or trip folks up if they need to bump this for whatever reason)

giuseppe: `quay.io/oci/pandoc:latest` seems to work fine.  Is there any reason for using an older version?

vbatts:
stale PR

PR #1083, opened 2021-02-08
schema: update README.md

Iceber:
Using a fixed version number is not very convenient 


vbatts (approved):
I prefer having examples that are directly copy-paste'able, but this is fine

PR #1084, opened 2021-02-08
schema: make with golang 1.16

Iceber:
Overriding `GOPATH` may cause some problems


giuseppe (approved):
LGTM

vbatts (approved):

Issue #1052, opened 2020-07-02
Prepare / Tag v1.1.0 release


thaJeztah:
(related to https://github.com/containerd/containerd/pull/4357/files#r448837520)

Runc tagged a new release (v1.0.0-rc91), and is currently depending on a non-tagged version of the runtime-spec. With various distribution packagers requiring tagged releases, and with go modules encouraging consumers to packages to use tagged releases, I'm wondering if a new release should be tagged.

runc is currently consuming 237cc4f519e2e8f9b235bacccfa8ef5a84df2875, (changes since v1.0.2: https://github.com/opencontainers/runtime-spec/compare/v1.0.2...237cc4f519e2e8f9b235bacccfa8ef5a84df2875), but there's a couple more changes in master (https://github.com/opencontainers/runtime-spec/compare/237cc4f519e2e8f9b235bacccfa8ef5a84df2875...master)

Are there pending pull requests that should be merged before tagging v1.0.3? (perhaps https://github.com/opencontainers/runtime-spec/pull/1046 would be nice to have). tags should be "cheap" so, perhaps not a blocker for a v1.0.3, but suggestions welcome 🤗 

@cyphar @crosbymichael @tianon @mrunalp @vbatts

h-vetinari:
Not to further hold up an already long process, but #1040 should also be very close to finished (just waiting for final review).

AkihiroSuda:
#1040 should be probably v1.0.4 or later

thaJeztah:
Given that it's adding new features, #1040 should probably bump the _minor_ version to stick with SemVer (so `v1.1.0`)

thaJeztah:
Thinking of that, possibly https://github.com/opencontainers/runtime-spec/pull/1046 warrants a "minor" version bump as well, as it's adding new functionality.

cyphar:
Yeah, we should tag a `1.0.3`. I would like to see #1046 merged (I don't think it makes sense to refer to it as a new feature -- it's just defining Go constants that were already defined by the spec -- though arguably the new hooks we added should've made the last release a `1.1.0`). And yeah, #1040 should definitely wait -- I don't want to rush adding more logic to cgroup configuration until we're sure it makes sense.

mskarbek:
@cyphar what more do you expect from croupv2 work?

cyphar:
[...]

PR #1087, opened 2021-02-19
seccomp: allow to override default errno return code

giuseppe:
the specs already support overriding the errno code for the syscalls
but the default value is hardcoded to EPERM.

Add a new attribute to override the default value.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -645,8 +645,9 @@ The following parameters can be specified to set up seccomp:
         * `SCMP_ACT_LOG`
 
     * **`errnoRet`** *(uint, OPTIONAL)* - the errno return code to use.
-        Some actions like `SCMP_ACT_ERRNO` and `SCMP_ACT_TRACE` allow to specify the errno
-        code to return. If not specified its default value is `EPERM`.
+        Some actions like `SCMP_ACT_ERRNO` and `SCMP_ACT_TRACE` allow to specify the errno code to return.
+        When the action doesn't support an errno, the runtime MUST print and error and fail.
+        If not specified its default value is `EPERM`.
 
     * **`args`** *(array of objects, OPTIONAL)* - the specific syscall in seccomp.
         Each entry has the following structure:
--- a/config-linux.md
+++ b/config-linux.md
@@ -594,6 +594,10 @@ The actions, architectures, and operators are strings that match the definitions
 The following parameters can be specified to set up seccomp:
 
 * **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as `syscalls[].action`.
+* **`defaultErrnoRet`** *(uint, OPTIONAL)* - the errno return code to use.
+    Some actions like `SCMP_ACT_ERRNO` and `SCMP_ACT_TRACE` allow to specify the errno code to return.
+    When the action doesn't support an errno, the runtime MUST print and error and fail.
+    If not specified then its default value is `EPERM`.
 * **`architectures`** *(array of strings, OPTIONAL)* - the architecture used for system calls.
     A valid list of constants as of libseccomp v2.5.0 is shown below.
 
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -203,6 +203,9 @@
                     "defaultAction": {
                         "$ref": "defs-linux.json#/definitions/SeccompAction"
                     },
+                    "defaultErrnoRet": {
+                        "$ref": "defs.json#/definitions/uint32"
+                    },
                     "flags": {
                         "type": "array",
                         "items": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -598,10 +598,11 @@ type VMImage struct {
 
 // LinuxSeccomp represents syscall restrictions
 type LinuxSeccomp struct {
-	DefaultAction LinuxSeccompAction `json:"defaultAction"`
-	Architectures []Arch             `json:"architectures,omitempty"`
-	Flags         []LinuxSeccompFlag `json:"flags,omitempty"`
-	Syscalls      []LinuxSyscall     `json:"syscalls,omitempty"`
+	DefaultAction   LinuxSeccompAction `json:"defaultAction"`
+	DefaultErrnoRet *uint              `json:"defaultErrnoRet,omitempty"`
+	Architectures   []Arch             `json:"architectures,omitempty"`
+	Flags           []LinuxSeccompFlag `json:"flags,omitempty"`
+	Syscalls        []LinuxSyscall     `json:"syscalls,omitempty"`
 }
 
 // Arch used for additional architectures


rhatdan:
LGTM
@mrunalp @cyphar PTAL

mrunalp:
LGTM

fweimer-rh:
This makes the `runc` workaround explicitly non-compliant, doesn't it? It sometimes returns `EPERM`, and sometimes `ENOSYS`, and not always `EPERM`.

As previously discussed, it is not entirely clear *when* that default should be applied. To any system call not listed in the policy?

Is `EPERM` really the right default here?

richfelker:
I have always believed (1) `ENOSYS` should be the default and (2) the system should not have any method for causing as-yet-unknown syscalls (at the time of profile authorship) to fail with any error code other than `ENOSYS`. Of these, (2) is the most important, and I understand there's a backwards-compatibility argument for returning `EPERM` as default for syscalls that were in the known range when the filter was written. I'm indifferent to adding the ability to change the default for this range as long as the unknown range is always forcibly `ENOSYS`.

giuseppe:
> Is `EPERM` really the right default here?

Probably not, I've picked it just for backward compatibility.

> I have always believed (1) `ENOSYS` should be the default and (2) the system should not have any method for causing as-yet-unknown syscalls (at the time of profile authorship) to fail with any error code other than `ENOSYS`. Of these, (2) is the most important, and I understand there's a backwards-compatibility argument for returning `EPERM` as default for syscalls that were in the known range when the filter was written. I'm indifferent to adding the ability to change the default for this range as long as the unknown range is always forcibly `ENOSYS`.

I personally think the default for other actions should still be `EPERM` as they are known at the time the profile was written but it is just a matter of taste.  If a container engine uses the new `defaultErrnoRet` then it knows about its meaning and we can pick the behavior we prefer and a seccomp profile can already override the errno value for each specified syscall.

thaJeztah (approved):
LGTM, thanks!

Actually had a branch with this, but didn't find the time yet to open 😂

giuseppe:
thanks anyone for the reviews.  Before merging let's wait for @cyphar to weight in as he was looking the most at this issue.

vrothberg (commented):
LGTM

config-linux.md
```diff
@@ -594,6 +594,9 @@ The actions, architectures, and operators are strings that match the definitions
 The following parameters can be specified to set up seccomp:
 
 * **`defaultAction`** *(string, REQUIRED)* - the default action for seccomp. Allowed values are the same as `syscalls[].action`.
+* **`defaultErrnoRet`** *(uint, OPTIONAL)* - the errno return code to use.
```
cyphar, line 597: It should probably be illegal to specify this for an action that doesn't have an errno value.

cyphar:
I'm fine with this PR, but as I've discussed before -- just to make it completely explicit -- **this does not solve the `-ENOSYS` problem**.  It's good that we're doing this for completeness's sake, but you would need all profiles to switch to `ENOSYS` and even then you're going to end up in scenarios where `socket(2)` returns `-ENOSYS` spuriously.

I think it would've been nicer to do this as one big changeset -- I think it would be a bad idea to release a new runtime-spec which adds this and doesn't address the larger `-ENOSYS` problem.

> (2) the system should not have any method for causing as-yet-unknown syscalls (at the time of profile authorship) to fail with any error code other than ENOSYS

I agree, and the reason I wanted to wait before we did this change is because I wanted us to include all of the other changes we need to solve this problem properly in the spec. But we can take this change now and do the other necessary changes later.

h-vetinari:
> @cyphar: I think it would be a bad idea to release a new runtime-spec which adds this and doesn't address the larger `-ENOSYS` problem.

By that argument, merging this PR would make the spec "unreleaseable". Given how thorny this problem seems to be, and how there have been [calls](https://github.com/opencontainers/runtime-spec/issues/1052) for a new version for ~8 months, would it be feasible to release `1.0.3` before merging this?

richfelker:
Why can't the spec just be fixed to specify what runc is doing to fix the problem?

giuseppe:
>  It's good that we're doing this for completeness's sake, but you would need all profiles to switch to `ENOSYS` and even then you're going to end up in scenarios where `socket(2)` returns `-ENOSYS` spuriously.

would this happen when a syscall is not fully specified in the profile and the defaultErrnoRet is used or are there other cases?

cyphar:
@giuseppe 

> would this happen when a syscall is not fully specified in the profile and the defaultErrnoRet is used or are there other cases?

That is the main issue -- but given that libseccomp today doesn't let you create fully-specified rules (depending on how complicated the allow rule is), that's sufficient to say that `defaultErrnoRet` cannot alone be used to solve the `-ENOSYS` problem.

In addition, in my view we should not be relying on profile authors to solve the `-ENOSYS` problem at all -- this is such a footgun that runtimes should handle it automatically (similarly to what runc is doing today). So in my view the complete solution for this problem is that we specify that the runtime should (or must) resolve this problem in a particular way. (Maybe there should be an opt-out from this behaviour, I'd have to think about exactly how that would work.)

To answer @richfelker, I tihnk that the spec-level solution should be a little bit less janky than the current runc solution. (Ideally we would do this through minimum kernel version information.)

@h-vetinari 

> Given how thorny this problem seems to be, and how there have been calls for a new version for ~8 months, would it be feasible to release 1.0.3 before merging this?

Yeah let's do that first.

giuseppe:
> @giuseppe
> 
> > would this happen when a syscall is not fully specified in the profile and the defaultErrnoRet is used or are there other cases?
> 
> That is the main issue -- but given that libseccomp today doesn't let you create fully-specified rules (depending on how complicated the allow rule is), that's sufficient to say that `defaultErrnoRet` cannot alone be used to solve the `-ENOSYS` problem.

true that won't fully solve the issue with `MASKED_EQ`, that must be addressed in libseccomp first (and it seems it is likely going to be added soon (2.6.0?))

> In addition, in my view we should not be relying on profile authors to solve the `-ENOSYS` problem at all -- this is such a footgun that runtimes should handle it automatically (similarly to what runc is doing today). So in my view the complete solution for this problem is that we specify that the runtime should (or must) resolve this problem in a particular way. (Maybe there should be an opt-out from this behaviour, I'd have to think about exactly how that would work.)
> 
> To answer @richfelker, I tihnk that the spec-level solution should be a little bit less janky than the current runc solution. (Ideally we would do this through minimum kernel version information.)

do you see the kernel version information as a replacement for `defaultErrnoRet` or the two features can coexist together?

cyphar:
@giuseppe 

> true that won't fully solve the issue with MASKED_EQ

`MASKED_EQ` is *a* problem (though it is one you can work around in a fairly dodgy manner by generating `2^n` `MASKED_EQ` rules). But the biggest issue is that you cannot construct multiple AND rules for the same argument with today's libseccomp (and there is no roadmap for fixing this). This means you cannot generate the inverse of `(arg0 == 1 || arg1 == 2)` because `(arg0 != 1 && arg1 != 2)` is an invalid rule.

> Do you see the kernel version information as a replacement for defaultErrnoRet or the two features can coexist together?

They can both co-exist -- the lack of `defaultErrnoRet` should be fixed purely for completeness's sake (and I can imagine a few cases where someone might want to use a different default errno).

giuseppe:
> `MASKED_EQ` is _a_ problem (though it is one you can work around in a fairly dodgy manner by generating `2^n` `MASKED_EQ` rules). But the biggest issue is that you cannot construct multiple AND rules for the same argument with today's libseccomp (and there is no roadmap for fixing this). This means you cannot generate the inverse of `(arg0 == 1 || arg1 == 2)` because `(arg0 != 1 && arg1 != 2)` is an invalid rule -- and as far as I can tell there is no clever way to work around this without having to generate an exhaustive set of `2^64` rules in the pathological case (and that's only if one argument has `||` rules -- it'll get even more complicated with multiple arguments).

Did you mean `(arg0 == 1 || arg0 == 2)`?  Yeah, that is another possible problem with libseccomp :(

In the long term I really hope we can bypass the seccomp generation and add to the specs something like https://github.com/containers/crun/pull/578 (in the seccomp section).  Alternatively, at least agree on something nicer than JSON and that users are not expected to deal directly with the libseccomp limitations.

config-linux.md
giuseppe: Thanks, I've added a sentence to clarify it.

giuseppe:
> I agree, though I am on the fence on whether we should still have `-ENOSYS` stubs be auto-added by runtimes even with custom seccomp profiles. Maybe it makes sense to allow users to opt-out of it, but I still think seccomp is far too footgun-friendly for us to leave it all up to profile generators.

I still think it is useful to allow the raw BPF.  I have already considered the pre-compilation of BPF in the past, since most containers use the same BPF profile, to avoid the cost of looking up syscalls in libseccomp.  The lookup function in 2.4.0 had `O(n)` complexity and it is used for each syscall specified in the profile (but then I preferred to go with https://github.com/seccomp/libseccomp/pull/223 and address it in libseccomp itself).

The raw BPF feature should not replace the existing mechanism, neither be the default; but it would open up the possibility to experiment with different ways of generating the BPF.  IMO, users should be able to shoot themselves in the foot if they wish so.  

> I'm fine with this PR, but as I've discussed before -- just to make it completely explicit -- **this does not solve the `-ENOSYS` problem**. It's good that we're doing this for completeness's sake, but you would need all profiles to switch to `ENOSYS` and even then you're going to end up in scenarios where `socket(2)` returns `-ENOSYS` spuriously

Could we move forward with it as it is just for completeness and we can address the `ENOSYS` issue separately?  I'd expect the discussion to take some time, while this feature will at least help us to temporarily address the issue with most seccomp profiles.

cyphar:
> Could we move forward with it as it is just for completeness and we can address the ENOSYS issue separately? I'd expect the discussion to take some time, while this feature will at least help us to temporarily address the issue with most seccomp profiles.

Yeah we can merge this first.

cyphar (approved):
LGTM.

mrunalp:
LGTM

tianon:
Arg, we've run into some fun with this downstream in moby/moby thanks to errno values being a per-architecture thing: https://github.com/moby/moby/pull/42836#issuecomment-963429850 / https://github.com/moby/moby/pull/42681#issuecomment-963433969

```console
$ git grep -inE 'ENOSYS[[:space:]]+=' | grep linux | column -t
unix/zerrors_linux_386.go:588:       ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_amd64.go:588:     ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_arm.go:594:       ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_arm64.go:585:     ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_mips.go:591:      ENOSYS  =  syscall.Errno(0x59)
unix/zerrors_linux_mips64.go:591:    ENOSYS  =  syscall.Errno(0x59)
unix/zerrors_linux_mips64le.go:591:  ENOSYS  =  syscall.Errno(0x59)
unix/zerrors_linux_mipsle.go:591:    ENOSYS  =  syscall.Errno(0x59)
unix/zerrors_linux_ppc.go:646:       ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_ppc64.go:650:     ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_ppc64le.go:650:   ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_riscv64.go:575:   ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_s390x.go:650:     ENOSYS  =  syscall.Errno(0x26)
unix/zerrors_linux_sparc64.go:639:   ENOSYS  =  syscall.Errno(0x5a)
```

Perhaps this should've been a string? :disappointed:

utam0k:
Hi! Is there any plan to add a check for this to the runtime-tools tests?

giuseppe:
> Arg, we've run into some fun with this downstream in moby/moby thanks to errno values being a per-architecture thing: [moby/moby#42836 (comment)](https://github.com/moby/moby/pull/42836#issuecomment-963429850) / [moby/moby#42681 (comment)](https://github.com/moby/moby/pull/42681#issuecomment-963433969)
> 
> ```
> $ git grep -inE 'ENOSYS[[:space:]]+=' | grep linux | column -t
> unix/zerrors_linux_386.go:588:       ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_amd64.go:588:     ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_arm.go:594:       ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_arm64.go:585:     ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_mips.go:591:      ENOSYS  =  syscall.Errno(0x59)
> unix/zerrors_linux_mips64.go:591:    ENOSYS  =  syscall.Errno(0x59)
> unix/zerrors_linux_mips64le.go:591:  ENOSYS  =  syscall.Errno(0x59)
> unix/zerrors_linux_mipsle.go:591:    ENOSYS  =  syscall.Errno(0x59)
> unix/zerrors_linux_ppc.go:646:       ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_ppc64.go:650:     ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_ppc64le.go:650:   ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_riscv64.go:575:   ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_s390x.go:650:     ENOSYS  =  syscall.Errno(0x26)
> unix/zerrors_linux_sparc64.go:639:   ENOSYS  =  syscall.Errno(0x5a)
> ```
> 
> Perhaps this should've been a string? disappointed

indeed, that was a mistake :/

Any suggestions on how we can fix it?  Do we need a new field and obsolete the current one?

thaJeztah:
The portable way would probably to have a string representation, so that the runtime that sets the actual profile can do the conversion 🤔

FWIW, I _think_ this would only be a problem if a profile is read from a JSON, and the JSON is created on  / with a non-matching architecture in mind, but I can see a string representation easier to use (when writing a profile).

thaJeztah:
> Do we need a new field and obsolete the current one?

If we decide to add a new field, we could make the new filed take precedence (if new field is set, old field must be ignored)

giuseppe:
> If we decide to add a new field, we could make the new filed take precedence (if new field is set, old field must be ignored)

yes, I agree with it.  I've used the same technique for containers/common seccomp profiles: https://github.com/containers/common/pull/821

thaJeztah:
For the meantime, for moby, I opened https://github.com/moby/moby/pull/43005, which uses conditional rules for MIPS / non-MIPS, with hard-coded values for `ENOSYS`. It's a bit dirty, but should do the trick for now (to allow users to save the profile in JSON).

tianon:
From the perspective of the specification, this hasn't actually been released yet, so I think we have a small amount of latitude in fixing this "properly", right?  I imagine actual users in the wild are limited to the two cases already identified, and probably only really `ENOSYS` or `EPERM`? :sweat_smile:

(I doubt this has been an extremely popular feature for users to take advantage of in their custom profiles -- the number of times I've seen someone write a non-trivial custom seccomp profile that wasn't just "download the default from my runtime and change one minor thing" I can count on a single hand. :see_no_evil:)

giuseppe:
> From the perspective of the specification, this hasn't actually been released yet, so I think we have a small amount of latitude in fixing this "properly", right? I imagine actual users in the wild are limited to the two cases already identified, and probably only really `ENOSYS` or `EPERM`?

I think we can probably drop `defaultErrnoRet` and `errnoRet` once the new version that uses a string settles down.

Would you be fine if I add two fields `defaultErrno` and `errno` that accept a string and have higher precedence over `defaultErrnoRet` and `errnoRet` that will be marked as obsolete?

thaJeztah:
> From the perspective of the specification, this hasn't actually been released yet, so I think we have a small amount of latitude in fixing this "properly", right?

Good call; looks indeed like it was not in a release yet; https://github.com/opencontainers/runtime-spec/compare/v1.0.2...main

From that perspective, I'd be 👍 to fix it before it's released. _With the added note_ that we need to check what projects already shipped with this feature in use (mostly concerned if, e.g., `runc` uses it already and/or what happens if containers were created with this option set, and it's upgraded 🤔)

PR #1090, opened 2021-02-25
Support mount_setattr(2) attributes (mostly for recursively-readonly mounts)

AkihiroSuda:
`mount_setattr(2)` introduced in [kernel 5.12](https://github.com/torvalds/linux/commit/2a1867219c7b27f928e2545782b86daaf9ad50bd) is especially useful for creating recursively-readonly bind mounts:
```c
struct mount_attr {
	__u64 attr_set;
	__u64 attr_clr;
	__u64 propagation;
	__u64 userns_fd;
};

int mount_setattr(int dfd, const char *path, unsigned flags,
                  struct mount_attr *uattr, size_t usize);

```

e.g.,
```c
struct mount_attr attr = {
  .attr_set   = MOUNT_ATTR_RDONLY,
};
rc = mount_setattr(-1, "/mnt/ro", AT_RECURSIVE, &attr, sizeof(attr));
```

This commit updates `config.md` to add OCI support for `mount_setattr(2)`.

e.g.,
```json
"mounts": [
    {
        "destination": "/mnt/ro",
        "type": "none",
        "source": "/src",
        "options": ["rbind"],
        "attr": {
          "flags": ["AT_RECURSIVE"],
          "attr_set": ["MOUNT_ATTR_RDONLY"],
        }
    }
]
```

### DIFF
--- a/config.md
+++ b/config.md
@@ -139,6 +139,41 @@ For POSIX platforms the `mounts` structure has the following fields:
 ]
 ```
 
+### <a name="configLinuxMounts" />Linux Mounts
+
+For Linux, the `mounts` structure mounts MAY have the following additional fields:
+
+* **`attr`** (object, OPTIONAL) specifies `mount_setattr(2)` attributes. Requires kernel 5.12 or later.
+  The runtime MUST [generate an error](runtime.md#errors) for any values which cannot be mapped to a relevant kernel interface.
+    * **`flags`** (array of strings, OPTIONAL) specifies the `flags` argument of `mount_setattr(2)` syscall.
+       e.g.,  "AT_RECURSIVE"
+    * **`attr_set`** (array of strings, OPTIONAL) specifies the `attr_set` field of `mount_attr` struct.
+       e.g., "MOUNT_ATTR_RDONLY"
+    * **`attr_clr`** (array of strings, OPTIONAL) specifies the `attr_clr` field of `mount_attr` struct.
+    * **`propagation`** (string, OPTIONAL) specifies the `propagation` field of `mount_attr` struct.
+       See [Rootfs Mount Propagation](./config-linux.md#configLinuxRootfsMountPropagation) for the string format.
+
+<!-- TODO: consider supporting mount_attr.userns_fd -->
+
+### Example
+
+```json
+"mounts": [
+    {
+        "destination": "/ro",
+        "type": "none",
+        "source": "/volumes/ro",
+        "options": ["rbind"],
+        "attr": {
+          "flags": ["AT_RECURSIVE"],
+          "attr_set": ["MOUNT_ATTR_RDONLY"],
+          "attr_clr": ["MOUNT_ATTR_NOEXEC"],
+          "propagation": "private"
+        }
+    }
+]
+```
+
 ## <a name="configProcess" />Process
 
 **`process`** (object, OPTIONAL) specifies the container process.
--- a/schema/defs.json
+++ b/schema/defs.json
@@ -141,11 +141,31 @@
                 },
                 "type": {
                     "type": "string"
+                },
+                "attr": {
+                    "$ref": "#/definitions/LinuxMountAttr"
                 }
             },
             "required": [
                 "destination"
             ]
+        },
+        "LinuxMountAttr": {
+          "type": "object",
+          "properties": {
+                "flags": {
+                    "$ref": "#/definitions/ArrayOfStrings"
+                },
+                "attr_set": {
+                    "$ref": "#/definitions/ArrayOfStrings"
+                },
+                "attr_clr": {
+                    "$ref": "#/definitions/ArrayOfStrings"
+                },
+                "propagation": {
+                    "type": "string"
+                }
+            }
         },
         "ociVersion": {
             "description": "The version of Open Container Initiative Runtime Specification that the document complies with",
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -117,6 +117,22 @@ type Mount struct {
 	Source string `json:"source,omitempty"`
 	// Options are fstab style mount options.
 	Options []string `json:"options,omitempty"`
+	// Attr specifies mount_setattr(2) attributes.
+	Attr *LinuxMountAttr `json:"attr,omitempty"`
+}
+
+// LinuxMountAttr specifies mount_setattr(2) attributes.
+// Requires Linux kernel 5.12 or later.
+type LinuxMountAttr struct {
+	// Flags specifies the flags argument of mount_setattr(2) syscall. e.g., "AT_RECURSIVE"
+	Flags []string `json:"flags,omitempty"`
+	// AttrSet specifies the attr_set field of mount_attr struct. e.g., "MOUNT_ATTR_RDONLY"
+	AttrSet []string `json:"attr_set,omitempty"`
+	// AttrClr specifies the attr_clr field of mount_attr struct.
+	AttrClr []string `json:"attr_clr,omitempty"`
+	// Propagation specifies the propagation field of mount_attr struct.
+	// Format corresponds to Linux.RootfsPropagation .
+	Propagation string `json:"propagation,omitempty"`
 }
 
 // Hook specifies a command that is run at a particular event in the lifecycle of a container


AkihiroSuda:
cc @brauner @cyphar @kolyshkin

AkihiroSuda:
Travis is hitting Docker Hub limit :cry:
```
$ docker pull vbatts/pandoc
Using default tag: latest
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
The command "docker pull vbatts/pandoc" failed and exited with 1 during .
```
https://travis-ci.org/github/opencontainers/runtime-spec/jobs/760421044

giuseppe:
the change LGTM, could you also update the go bindings and the schema?

config.md
```diff
@@ -139,6 +139,38 @@ For POSIX platforms the `mounts` structure has the following fields:
 ]
 ```
 
+### <a name="configLinuxMounts" />Linux Mounts
+
+For Linux the `mounts` structure has the following fields:
```
kolyshkin, line 144:
```suggestion
For Linux, the `mounts` structure MAY have the following additional fields:
```

config.md
```diff
@@ -139,6 +139,38 @@ For POSIX platforms the `mounts` structure has the following fields:
...
+
+* **`attr`** (object, OPTIONAL) specifies `mount_setattr(2)` attributes
+    * **`flags`** (array of strings, OPTIONAL) specifies the `flags` argument of `mount_setattr(2)` syscall.
+       e.g.,  "AT_RECURSIVE"
+    * **`set`** (array of strings, OPTIONAL) specifies the `attr_set` field of `mount_attr` struct.
+       e.g., "MOUNT_ATTR_RDONLY"
+    * **`clr`** (array of strings, OPTIONAL) specifies the `attr_clr` field of `mount_attr` struct.
```
kolyshkin, line 151: As all other fields are named exactly as in `mount_setattr(2)`, perhaps we should do the same here?

IOW
* `set` -> `attr_set`
* `clr` -> `attr_clr`

config.md
AkihiroSuda: Updated, sorry for delay

config.md
AkihiroSuda: Updated, sorry for delay

AkihiroSuda:
Added go binding and schema, sorry for delay

config.md
```diff
@@ -139,6 +139,40 @@ For POSIX platforms the `mounts` structure has the following fields:
 ]
 ```
 
+### <a name="configLinuxMounts" />Linux Mounts
+
+For Linux, the `mounts` structure mounts MAY have the following additional fields:
```
thaJeztah, line 144: We should probably describe what the expected behaviour is of the runtime if a spec includes attr's that are not supported by the kernel; should it ignore those attributes, or should it fail?

config.md
```diff
@@ -139,6 +139,40 @@ For POSIX platforms the `mounts` structure has the following fields:
...
+        "destination": "/ro",
+        "type": "none",
+        "source": "/volumes/ro",
+        "options": ["rbind"],
+        "attr": {
+          "flags": ["AT_RECURSIVE"],
+          "attr_set": ["MOUNT_ATTR_RDONLY"],
+          "attr_clr": ["MOUNT_ATTR_NOEXEC"],
+          "propagation": "private"
+        }
```
thaJeztah, line 171: I'm a bit concerned about how this is just a 1:1 mapping of `mount_setattr` (although I realise we already crossed that path on many other options). The advantage is that it's flexible, but it we only need it for the `recursively-readonly`, I'm wondering if it would be good to design around that case, making it somewhat more less verbose?

It makes it a bit confusing in some areas, e.g., `options` would take  `(r)private`, `ro` and `noexec`, but `attr` takes `MOUNT_ATTR_RDONLY` and `MOUNT_ATTR_NOEXEC` (in addition to using a "diff" of attributes to _add_ and to _remove_).

config.md
AkihiroSuda: Added `The runtime MUST [generate an error](runtime.md#errors) for any values which cannot be mapped to a relevant kernel interface.`

config.md
AkihiroSuda: > The advantage is that it's flexible, but it we only need it for the recursively-readonly, I'm wondering if it would be good to design around that case, making it somewhat more less verbose?

I don't think this is going to be recursively-readonly, though we can prioritize recursively-readonly over other potential usecases

> It makes it a bit confusing in some areas, e.g., options would take (r)private, ro and noexec, but attr takes MOUNT_ATTR_RDONLY and MOUNT_ATTR_NOEXEC (in addition to using a "diff" of attributes to add and to remove).

`MOUNT_ATTR_XXX` form corresponds to `RLIMIT_XXX` form that has already existed.

thaJeztah:
> Added `The runtime MUST [generate an error](runtime.md#errors) for any values which cannot be mapped to a relevant kernel interface.`

Implicitly, this means that "higher level" runtimes then become responsible for checking kernel features (something I'd like to avoid if not needed).

For the feature mentioned ("recursively-readonly"), I can somewhat see it being (possibly?) desirable to error if not supported (if the runtime wouldn't apply the option, that would break a security expectation); question is if the same applies for any `attr` that can be _possibly_ set. A generic option like this makes it harder to describe expectations. (some options may be ok to ignore, other ones may not).

AkihiroSuda:
>> Added The runtime MUST [generate an error](runtime.md#errors) for any values which cannot be mapped to a relevant kernel interface.
>
> Implicitly, this means that "higher level" runtimes then become responsible for checking kernel features (something I'd like to avoid if not needed).

Runtime doesn't really need to check kernel stuff, it will just error out when `mount_setattr(2)` invocation fails.

AkihiroSuda:
Btw, w.r.t., recursively read-only mounts, if `mount(8)` is going to have a `mount -o OPTION` option string for `MOUNT_ATTR_RDONLY`, we can support it without modifying the OCI spec.

@brauner @cyphar Is there any discussion to support `MOUNT_ATTR_RDONLY` in `mount(8)`?

brauner:
> Btw, w.r.t., recursively read-only mounts, if `mount(8)` is going to have a `mount -o OPTION` option string for `MOUNT_ATTR_RDONLY`, we can support it without modifying the OCI spec.
> 
> @brauner @cyphar Is there any discussion to support `MOUNT_ATTR_RDONLY` in `mount(8)`?

This week I started working on initial support for idmapped mounts in `mount(8)`. I would expect that as part of that work `mount(8)` will gain read-only too.

thaJeztah:
> Runtime doesn't really need to check kernel stuff, it will just error out when mount_setattr(2) invocation fails.

Right, but that would lead to some obscure `OCI runtime failed to start` error. My train of thought here is that (from a user's perspective), someone would;

```bash
docker run -v /some/path:/foo:ro foobar

# or possibly
docker run --read-only foobar
```

And would receive such an error, so the "higher up" runtime (docker / containerd) would either:

- unconditionally set these options and get the obscure error (depending on what kernel is in use)
- first check kernel support and special-case the option (either "not set" the option, and warn, or produce a specific error)

> > Is there any discussion to support MOUNT_ATTR_RDONLY in mount(8)?
>
> This week I started working on initial support for idmapped mounts in mount(8). I would expect that as part of that work mount(8) will gain read-only too.

That sounds great, and would make things a lot let complicated.

AkihiroSuda:
> > Btw, w.r.t., recursively read-only mounts, if `mount(8)` is going to have a `mount -o OPTION` option string for `MOUNT_ATTR_RDONLY`, we can support it without modifying the OCI spec.
> > @brauner @cyphar Is there any discussion to support `MOUNT_ATTR_RDONLY` in `mount(8)`?
> 
> This week I started working on initial support for idmapped mounts in `mount(8)`. I would expect that as part of that work `mount(8)` will gain read-only too.

Thanks @brauner , what will the `OPTION` string look like?

AkihiroSuda:
- Opened https://github.com/util-linux/util-linux/issues/1501 for discussing the mount option string ("rro")

giuseppe:
I like the idea of `rro`.

Do you think we could also have something like `id-map` to set up the id mapped mount using the same mappings used for the container user namespace?

brauner:
> I like the idea of `rro`.
> 
> Do you think we could also have something like `id-map` to set up the id mapped mount using the same mappings used for the container user namespace?

I think having this ultimately available in the `mount` tool would be good. But as a more general option where the user can also specify idmappings for a mount on the command line when they e.g. want to idmap a directory on the host.

giuseppe:
yeah, probably for the `mount` tool it would not make sense, but for the OCI configuration, I doubt it will be necessary to offer more flexibility than just using the current container user namespace.  If needed it can always be added later :-)

brauner:
> yeah, probably for the `mount` tool it would not make sense, but for the OCI configuration, I doubt it will be necessary to offer more flexibility than just using the current container user namespace. If needed it can always be added later :-)

Oh, sorry, I misread your comment then. :) I didn't realize that you were talking about the OCI spec. Yes, that makes sense to me. :)

rata:
I wonder how to discover from upper layers if the fs for the bind mounts support id mapped mounts or not. Has anyone give any thought to this?

For example, when creating a container from kubernetes (we are working on a KEP for k8s to allow this with @giuseppe), the kubelet will have to decide if id mapped mounts should be used or not, and if they do that will be sent over the CRI interface to the contianer runtime and the container runtime will do the proper thing for the config.json that we are discussing here.

From this proposal, if the kubelet wanted to start with id mapped mounts and the volume is not supported, this will cause a hard fail currently. Catching that from kubernetes is not nice (seems wasteful of resources, etc.). So, we really want to know from the kubelet if we should ask for this or not, at least while if not available it is a hard fail. Although, I think if it is not a hard fail, we still want to know from higher levels if id mapped mounts will work for all fs, as in that way we can use non-overlapping mappings for pods/containers.

As fs support for id mapped mounts will be expanding over time (may be backported, etc.), the kubelet might need to probe for support for all the fs present in the volumes used by the container, or (guessing, not familiar with the CSI interface) the CSI drivers can maybe report if each of one supports id mapped mounts or something.

But I wonder, has anyone give any thought on how will the higher levels use this (or what adjustments might be useful) without much pain?

AkihiroSuda:
crun introduced support for "idmap" mount option, without requiring explicit `mount_setattr` objects in OCI Runtime `config.json` .


https://github.com/containers/crun/commit/827b8731899d4febea3f27907bcde5e6065bb65b

rata:
Regarding my previous question: I think the answer should be that user-space will probably have to probe for support with some dummy params (maybe the kubelet or the CSI driver in case of k8s), and having this be fatal as it is proposed here it is fine. After all, that is the case for most of the other features too. But if anyone has given more thought to this, I'm curious to know :)

rata:
@AkihiroSuda just curious: why was this closed?

AkihiroSuda:
Because runc and crun now support `rro` mounts without adopting this PR, and crun also supports `idmap`

rata:
Wouldn't it be better to have it as part of the spec? It is an important feature that userns implementation might rely on, it will be nice for it to be part of the spec so other runtimes implement it in compatible ways.

@AkihiroSuda what do you think?

AkihiroSuda:
The `rro` mount option  is now defined in:
- https://github.com/opencontainers/runtime-spec/pull/1181

The `idmap` mount option is replaced by:
- https://github.com/opencontainers/runtime-spec/pull/1143

PR #1095, opened 2021-03-12
Introduce zos as platform

najohnsn:
As discussed in the 1/27/2021 OCI developer meeting, these changes introduce IBM z/OS as a Native Platform for OCI. Changes to code, JSON schema, and markdown documentation is included.

Other than `zos` being added a _platform_ and having its own platform-specific configuration, nothing new is really added with this. The expectation is that additional pull requests (preceded by development discussions, if needed) will be used to request any z/OS-unique configuration properties as the IBM initiative to support OCI containers on z/OS progresses.

*Note on z/OS devices:* it is the only property in the z/OS-specific configuration object. This was copied from how devices can be specified for Linux. Specifying devices like this would be a perfectly valid way to define/create devices in a container on z/OS. However, its usefulness on z/OS is questionable. It is being included in this PR so that _something_ is defined in the z/OS-specific configuration object. It very well may turn out that the ability to specify devices to be created in a z/OS container may be unnecessary and that the definition should be replaced with more useful properties.

Signed-off-by: Neil Johnson <najohnsn@us.ibm.com>
Contributors: @SteeleDesmond and @najohnsn

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -15,7 +15,7 @@ type Spec struct {
 	// Mounts configures additional mounts (on top of Root).
 	Mounts []Mount `json:"mounts,omitempty"`
 	// Hooks configures callbacks for container lifecycle events.
-	Hooks *Hooks `json:"hooks,omitempty" platform:"linux,solaris"`
+	Hooks *Hooks `json:"hooks,omitempty" platform:"linux,solaris,zos"`
 	// Annotations contains arbitrary metadata for the container.
 	Annotations map[string]string `json:"annotations,omitempty"`
 
@@ -27,6 +27,8 @@ type Spec struct {
 	Windows *Windows `json:"windows,omitempty" platform:"windows"`
 	// VM specifies configuration for virtual-machine-based containers.
 	VM *VM `json:"vm,omitempty" platform:"vm"`
+	// ZOS is platform-specific configuration for z/OS based containers.
+	ZOS *ZOS `json:"zos,omitempty" platform:"zos"`
 }
 
 // Process contains information to start a specific application inside the container.
@@ -49,7 +51,7 @@ type Process struct {
 	// Capabilities are Linux capabilities that are kept for the process.
 	Capabilities *LinuxCapabilities `json:"capabilities,omitempty" platform:"linux"`
 	// Rlimits specifies rlimit options to apply to the process.
-	Rlimits []POSIXRlimit `json:"rlimits,omitempty" platform:"linux,solaris"`
+	Rlimits []POSIXRlimit `json:"rlimits,omitempty" platform:"linux,solaris,zos"`
 	// NoNewPrivileges controls whether additional privileges could be gained by processes in the container.
 	NoNewPrivileges bool `json:"noNewPrivileges,omitempty" platform:"linux"`
 	// ApparmorProfile specifies the apparmor profile for the container.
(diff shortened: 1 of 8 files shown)


tianon:
> It very well may turn out that the ability to specify devices to be created in a z/OS container may be unnecessary and that the definition should be replaced with more useful properties.

Given that this is speculative, and anything we merge might end up in a release, perhaps this should wait until there's something concrete to include?  Alternatively/also, it might be prudent to mark all the bits related to this platform as "experimental" or "in-progress" in some way until it's deemed at least minimally feature complete (if the goal is to merge adjustments incrementally before the full functionality is known/working).

najohnsn:
Sorry for the delay. Yes, just as you said, the goal is the merge adjustments incrementally.

Let me explain a bit more about /dev on z/OS. Although the major/minor number of devices are different on z/OS when compared to Linux, they can be created in the same way - `mknod`. The more interesting thing on z/OS is that the number of different kinds of devices is fairly limited and that they are created by the system either during initialization or on demand. For example, if a request to open `/dev/console` is done and the _file_ is not there, the system will create a `/dev/console` with an appropriate major/minor number. It is this on-demand creation that led me to the statement that it is not very useful on z/OS.

Do you think that just an _in progress_ note on the `zos` line in the list of Platforms in `spec.md` would satisfy the raised concern?

najohnsn:
I added a commit to _Indicate work in progress_.
It's not the changes I was leaning toward in my previous comment, but I think it covers more:
Added "_This document is a work in progress._" to `config-zos.md`.

tianon (approved):
That's good enough for me! :sweat_smile:  Thanks :+1:

LGTM

(Does this mean we're going to get some sweet Z systems to play/CI with? :trollface:)

najohnsn:
IBM is looking at providing z/OS systems for CI. 🙂

It's been a while. Is anything else needed for this PR to get approved/merged?

tianon:
I'd suggest a rebase (to try and fix the CI failure) and a squash (since I don't think we want three commits that together amount to "this is coming"). :sweat_smile:

(Maybe that will help encourage other members of @opencontainers/runtime-spec-maintainers to opine/review.)

specs-go/config.go
```diff
@@ -27,6 +27,8 @@ type Spec struct {
 	Windows *Windows `json:"windows,omitempty" platform:"windows"`
 	// VM specifies configuration for virtual-machine-based containers.
 	VM *VM `json:"vm,omitempty" platform:"vm"`
+	// Zos is platform-specific configuration for z/OS based containers.
+	Zos *Zos `json:"zos,omitempty" platform:"zos"`
```
samuelkarp, line 31: In Go, [initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) generally remain capitalized.  Seems like the field (and type) should be `ZOS` instead of `Zos`.

```suggestion
	// ZOS is platform-specific configuration for z/OS based containers.
	ZOS *ZOS `json:"zos,omitempty" platform:"zos"`
```

najohnsn:
@tianon @samuelkarp Thanks for the suggestions! The 3 commits have been replaced by a single commit that incorporates your suggestions.

samuelkarp (approved):
LGTM, though I'm not a maintainer of the spec.

tianon (approved):

vbatts (approved):

tianon:
Would love to check in on this and see what's the latest status -- this seemed like an effort with a lot of steam behind it three years ago, but we (in the OCI / runtime-spec) haven't really heard anything since.  Should we consider removing this temporary/experimental `zos` from the spec?  Should it be updated?

najohnsn:
@tianon, thanks for checking in!

Certainly don't want to remove. My team wants to push forward to remove the _temporary/experimental status_. [IBM z/OS Container Platform](https://www.ibm.com/products/zos-container-platform) is available to z/OS customers! We've been waiting for a while for all the pieces to come together, but everything is available now.

There are some changes from this initial PR. I believe it mostly amounts to the addition of namespaces on z/OS. I expect that someone on my team will open a PR in the coming weeks. We _will_ want to remove the devices. If it is felt we need to stage that kind of change, we can discuss in the upcoming PR.

PR #1096, opened 2021-03-18
Fix seccomp notify inconsistencies

rata:
In PR https://github.com/opencontainers/runtime-spec/pull/1074 I missed some consts, sorry!

Funny thing is we didn't catch them even when we used this in runc and containerd patches because containerd just passes a string (see commit for a detailed explanation) and runc doesn't use these constants (again, full explanation in the commit msg).

cc @vbatts @giuseppe as you LGTM to #1074.

### DIFF
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -70,8 +70,7 @@
             "enum": [
                 "SECCOMP_FILTER_FLAG_TSYNC",
                 "SECCOMP_FILTER_FLAG_LOG",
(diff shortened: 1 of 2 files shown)


giuseppe (approved):
LGTM

rata:
@giuseppe pushed another fix to another inconsistency, removing SECCOMP_FILTER_FLAG_NEW_LISTENER from the valid flags. That should have never been added there in the first place :see_no_evil:

giuseppe (approved):
LGTM

vbatts:
LGTM

PR #1099, opened 2021-03-30
Proposal: Flatten Rdma to array instead of using additionalProperties. [Breaking-change]

flouthoc:
Following PR aims to reduce unnecessary complications of type `mapStringObject` around Rdma by moving it to simpler array based structure.

Reasons:
1) Number of devices is limited, cant really understand use case of a `additonalProperties` here, I feel current `additonalProperties` here is anomaly and  forces parsers to implement `mapStringObject` which is not strongly typed and is not helpful.

### DIFF
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -184,8 +184,8 @@
                         }
                     },
                     "rdma": {
-                        "type": "object",
-                        "additionalProperties": {
+                        "type": "array",
+                        "items": {
                             "$ref": "defs-linux.json#/definitions/Rdma"
                         }
                     }
--- a/schema/defs-linux.json
+++ b/schema/defs-linux.json
@@ -277,6 +277,9 @@
         "Rdma": {
             "type": "object",
             "properties": {
+                "device": {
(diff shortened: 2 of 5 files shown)


flouthoc:
@giuseppe could you please take a look.

flouthoc:
I think we can also change it to mapStringRdma but dont know if it would be standard convention

AkihiroSuda (changes requested):
I don't think we can have any breaking change in this repo (for v1.x.x, at least)

flouthoc:
@AkihiroSuda Sure I understand. Probably this pr can never be merged. Just wanted to understand the rationale for choosing map for Rdma devices. I might be missing something but i feel array of struct would be a better choice. As of now strongly-typed parsers will need exception case to handle this as compared to any other type which is added in runtime-spec.

giuseppe:
@flouthoc are you fine with closing the PR since it would be a breaking change?

flouthoc:
@giuseppe Sure 😄

Issue #1067, opened 2020-09-25
better provide a interface for linuxBlockIODevice


ranchothu:
hi,  when i try to modify linuxBlockIODevice outside package, it seems not convenient for its anonymous.
code slice is below.
What about provide a interface like GetlinuxBlockIODevice here or make it public?
If ok, i will provide a merge later.

> // linuxBlockIODevice holds major:minor format supported in blkio cgroup
type linuxBlockIODevice struct {
	// Major is the device's major number.
	Major int64 `json:"major"`
	// Minor is the device's minor number.
	Minor int64 `json:"minor"`
}
// LinuxThrottleDevice struct holds a `major:minor rate_per_second` pair
type LinuxThrottleDevice struct {
	linuxBlockIODevice
	// Rate is the IO rate limit per cgroup per device
	Rate uint64 `json:"rate"`
}

THX.

PR #1103, opened 2021-04-16
specs-go: export LinuxBlockIODevice

askervin:
Previously unexported linuxBlockIODevice prevented fully defined
LinuxWeightDevice and LinuxThrottleDevice literals, making them
less user-friendly than the rest of the interface.

This solves issue #1067 (closed without a fix).

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -250,8 +250,8 @@ type LinuxInterfacePriority struct {
 	Priority uint32 `json:"priority"`
 }
 
-// linuxBlockIODevice holds major:minor format supported in blkio cgroup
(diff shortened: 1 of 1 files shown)


tianon (approved):
LGTM

Tracing the history of this, I think it was originally made private by @vbatts, so maybe he can provide additional opinion (and review :nerd_face:), but it appears the logic was because it's never used directly, and is rather only a shared ancestor for other types to avoid duplication.

vbatts:
haha tianon, i have no idea. Perhaps just erring on the side of make it
private unless required otherwise.

Issue #1077, opened 2020-11-13
Q: What is the meaning of SECCOMP_FILTER_FLAG_TSYNC for containers?


mauriciovasquezbernal:
https://github.com/opencontainers/runtime-spec/pull/1018 introduced support for specifying the seccomp flags  `SECCOMP_FILTER_FLAG_TSYNC`, `SECCOMP_FILTER_FLAG_LOG` & `SECCOMP_FILTER_FLAG_SPEC_ALLOW`. 

I agree that `SECCOMP_FILTER_FLAG_LOG` & `SECCOMP_FILTER_FLAG_SPEC_ALLOW` make sense as stated in https://github.com/opencontainers/runtime-spec/pull/1018#issuecomment-529626215. However I'm not able to understand what is the meaning of `SECCOMP_FILTER_FLAG_TSYNC` in the OCI context. According to my understanding the runtime should guarantee that the seccomp filters are applied to the initial container process, whether `TSYNC` has to be used or not depends on if the thread calling `seccomp()` is the same executing the container init process or not and it's an internal implementation detail of the runtimes.  

ref https://github.com/containers/crun/pull/85#issuecomment-529627607

@giuseppe 

cc @alban @rata

cyphar:
Yeah I'm not sure that it makes sense since it only is related to the loading mechanism of a container runtime and not the final container process. Though you could argue a correct implementation could also just ignore the flag because the user has no way of determining if the flag was actually used.

We can't remove it unfortunately (this was included in the `1.0.2` release) but we could deprecate it.

PR #1108, opened 2021-07-16
specs-go: add consts for seccomp flags

thaJeztah:
Relates to https://github.com/moby/moby/issues/42619

Commit d1ef109cd0b39239ff82c267df314f7ed2da576b (https://github.com/opencontainers/runtime-spec/pull/1018) added the ability to specify
flags that must be passed to seccomp(2) when installing the filter, and
defined an enum in the specification.

This patch adds corresponding consts for the Go implementation of the Spec.

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -613,6 +613,13 @@ type Arch string
 // LinuxSeccompFlag is a flag to pass to seccomp(2).
 type LinuxSeccompFlag string
 
+// LinuxSeccompFlag options
+const (
+	FlagTSync LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_TSYNC"
+	FlagLog   LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
+	FlagAllow LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_SPEC_ALLOW"
+)
+
 // Additional architectures permitted to be used for system calls
 // By default only the native architecture of the kernel is permitted
 const (


thaJeztah:
@giuseppe  @tianon @crosbymichael PTAL

/cc @sporksmith

giuseppe (approved):
LGTM

tianon:
I'm a little worried about how generic the naming convention here is -- I realize we did the same thing below with `ArchFoo`, but it feels like maybe we should use something a tad more specific since "flags" seem like something the specs might have others of outside Linux / seccomp.  Thoughts?

thaJeztah:
> I'm a little worried about how generic the naming convention here is -- I realize we did the same thing below with `ArchFoo`, but it feels like maybe we should use something a tad more specific since "flags" seem like something the specs might have others of outside Linux / seccomp. Thoughts?

Yes, I was actually looking at that as well; I did not add a prefix, because all existing ones did not have so, but all of these are in a single "namespace", and don't have prefixes to identify what they're for.

We could decide to

- add `Seccomp` or `Scmp` prefixes to all of them
- move them to a `seccomp` or `seccompoptions` dir / package

Either of the above would be a "breaking change", but we can of course add aliases and deprecate the existing ones.


Let me know what you think

specs-go/config.go
```diff
@@ -613,6 +613,13 @@ type Arch string
 // LinuxSeccompFlag is a flag to pass to seccomp(2).
 type LinuxSeccompFlag string
 
+// LinuxSeccompFlag options
+const (
+	FlagTSync LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_TSYNC"
```
rata, line 618: I'm not sure we want to add a const for this. This flag should probably not be used, as we asked here: https://github.com/opencontainers/runtime-spec/issues/1077.

Maybe it is better to _not_ facilitate using it, or even better mark it as deprecated in the spec? (the latter out of scope for this PR, but a follow-up PR to discuss it would be great IMHO).

thaJeztah: Good point; I'm fine with removing this one; I wasn't aware of the other discussion 👍

tianon:
Closing in favor of https://github.com/opencontainers/runtime-spec/pull/1138 :bow: :heart:

Issue #1021, opened 2019-11-06
Add systemd cgroup path spec


AkihiroSuda:
https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path

The systemd cgroup path convention implemented in runc/crun should be added to the spec.

This convention is becoming important for cgroup v2, because rootless containers depends on systemd for cgroup delegation in most environments.

kailun-qin:
Do we have any plan to make this forward? Thanks!

AkihiroSuda:
Thanks for opening #1115, but I meant adding proper spec for the existing implementation of systemd cgroupsPath, e.g. `user.slice:nerdctl:deadbeef`
https://github.com/containerd/nerdctl/blob/07c3ca23743308987aa3b26b1fb7b560804ff9a4/cmd/nerdctl/run_cgroup_linux.go#L53

I didn't mean adding new fields for systemd.

PR #1115, opened 2021-08-05
specs-go/config: add systemd cgroup support

kailun-qin:
The `--systemd-cgroup` flag and the systemd cgroup path convention
currently implemented in `runc/crun` should be added to the spec.

This patch adds in the spec:
* the option to enable systemd cgroup;
* the configs of systemd units and the containing slice to map to the
  cgroup tree.

Fixes https://github.com/opencontainers/runtime-spec/issues/1021


AkihiroSuda (changes requested):
We should just have a documentation for the existing implementation of systemd `cgroupsPath`, e.g. `user.slice:nerdctl:deadbeef`
https://github.com/containerd/nerdctl/blob/07c3ca23743308987aa3b26b1fb7b560804ff9a4/cmd/nerdctl/run_cgroup_linux.go#L53

I don't think we need to add new fields to the spec.

kailun-qin:
> We should just have a documentation for the existing implementation of systemd `cgroupsPath`, e.g. `user.slice:nerdctl:deadbeef`
> https://github.com/containerd/nerdctl/blob/07c3ca23743308987aa3b26b1fb7b560804ff9a4/cmd/nerdctl/run_cgroup_linux.go#L53
> 
> I don't think we need to add new fields to the spec.

@AkihiroSuda Thanks for the comments!

Actually I thought over this a bit. My original intention was trying to make it consistent for all implementations, since I'm not certain if they all follow the exactly same cgroups path convention and use it the same way. I went through runc, crun and youki to double check:

1. they all follow "slice:prefix:name" convention;
2. the behavior of parsing and using scope/slice maybe different, e.g., crun seems not support slice unit name.

I agree it's fine to only add docs, just need more clarification. Let me do an update.

Besides, do you think it's reasonable to add a field for enabling/disabling "systemd-cgroup"? Thanks!

kailun-qin:
> We should just have a documentation for the existing implementation of systemd `cgroupsPath`, e.g. `user.slice:nerdctl:deadbeef`
> https://github.com/containerd/nerdctl/blob/07c3ca23743308987aa3b26b1fb7b560804ff9a4/cmd/nerdctl/run_cgroup_linux.go#L53
> 
> I don't think we need to add new fields to the spec.

@AkihiroSuda Updated. PTAL, thanks!

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@kolyshkin PTAL

kolyshkin (changes requested):
This is somewhat similar to work I did documenting the runc systemd cgroup driver; see https://github.com/opencontainers/runc/blame/main/docs/systemd.md

In fact, I see that some wording is similar so this might be a derived work. If that is the case, perhaps it makes sense to at least refer to the source document(s) in the commit message.

What's missing from this description is
 - all three components are optional;
 - what the defaults for omitted components are (or can be)

config-linux.md
```diff
@@ -176,11 +176,31 @@ For more information, see the [kernel cgroups documentation][cgroup-v1].
 **`cgroupsPath`** (string, OPTIONAL) path to the cgroups.
 It can be used to either control the cgroups hierarchy for containers or to run a new process in an existing container.
 
-The value of `cgroupsPath` MUST be either an absolute path or a relative path.
+If the runtime creates cgroups and sets cgroup limits on its own (aka. fs cgroup driver mode), the value of `cgroupsPath` MUST be either an absolute path or a relative path.
```
kolyshkin, line 179: I would rephrase it as
```
If the runtime manages cgroups on its own (i.e. works with cgroupfs directly), the value of ...
```

config-linux.md
```diff
@@ -176,11 +176,31 @@ For more information, see the [kernel cgroups documentation][cgroup-v1].
...
 
 * In the case of an absolute path (starting with `/`), the runtime MUST take the path to be relative to the cgroups mount point.
 * In the case of a relative path (not starting with `/`), the runtime MAY interpret the path relative to a runtime-determined location in the cgroups hierarchy.
 
+If the runtime use systemd cgroup driver to create cgroups and set cgroup limits, the value of `cgroupsPath` MUST be in the "slice:prefix:name" form (e.g. "system.slice:runtime:434234").
```
kolyshkin, line 184:
```
If the runtime manages cgroups indirectly, via systemd, the value of ...
```

kailun-qin:
> perhaps it makes sense to at least refer to the source document(s) in the commit message.

Added in the commit message.

> What's missing from this description is
> 
> * all three components are optional;
> * what the defaults for omitted components are (or can be)

Updated and reflected in the spec descriptions.

kolyshkin:
@kailun-qin thanks!

Have you checked this against crun? IOW, is crun fully conformant to this spec?

config-linux.md
```diff
@@ -186,11 +186,30 @@ containers.
...
+By specifying with the transient systemd unit to create for the container and the containing slice which hosts the unit, the systemd units directly map to objects in the cgroup tree.
+When these units are activated, they map directly to cgroup paths built from the unit names.
+
+This form specifies the following systemd cgroup properties which are all optional:
+
+* `slice` - name of the parent slice systemd unit, under which the container is placed.
+    Note that `slice` can contain dashes to denote a sub-slice (e.g. `user-1000.slice` is a correct
+    notation, meaning a subslice of `user.slice`), but it must not contain slashes (e.g.
+    `user.slice/user-1000.slice` is invalid). A `slice` of `-` represents a root slice.
+    If not specified, it can default to:
```
AkihiroSuda, line 204:
```suggestion
    If not specified, it SHOULD default to:
```

Issue #1050, opened 2020-06-07
Add initial support for rsvd accounting hugetlb cgroup 


odinuge:
The previous non-rsvd max/limit_in_bytes does not account for reserved
huge page memory, making it possible for a processes to reserve all the
huge page memory, without being able to allocate it (due to cgroup
restrictions).

In practice this makes it possible to successfully mmap more huge page
memory than allowed via the cgroup settings, but when using the memory
the process will get a SIGBUS and crash. This is bad for applications
trying to mmap at startup (and it succeeds), but the program crashes
when starting to use the memory. eg. postgres is doing this by default.

More info can be found here: https://lkml.org/lkml/2020/2/3/1153

In order to solve this, I think we have to main ways to do it:
- Add writes (when supported) to rsvd for the current `hugepageLimits` found [here](https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#huge-page-limits). Silently ignore when rsvd is not supported.
- Add another element called something like `hugepageLimitsRsvd` to enforce the rsvd. value, silently fail or return error when rsvd is not supported.

I lean toward the first approach, since adding a new item makes it harder to understand, and may lead into "bad" implementations. The pro for the last one, for having a separate entity is that it is then up to the user of the runtime to decide, giving the "user" a full choice, even tho. i see no real reason to enforce the "old" value and not the reserved one. The current behavior makes a cgroup limited process able to reserve all the huge page memory available on a node, making it inaccessible to others. 

No matter the decition, we should then update the config-linux.md docs to clarify how it should work.

Any thoughts?

Simple WIP in runc to add support for enforcing it using the `hugepageLimits` is here: https://github.com/opencontainers/runc/pull/2360/files

kailun-qin:
@odinuge 
I think it's reasonable to go w/ way#1 since
1) in general reservation limits are superior to page fault limits as the limits are enforced at reservation time and never causes the application to get SIGBUS signal if the memory was reserved before hand which is more friendly to applications, so it seems fine to default it to reserved limits and fallback when unsupported;
2) it reduces implementation efforts and risks by not introducing a new field/struct.

There might be concerns for compatibility. Picking this up by firstly drafting an update on the spec. Let's see if any feedbacks from the community. Any comments are much appreciated, thanks!

PR #1116, opened 2021-08-06
config-linux: add support for rsvd hugetlb cgroup

kailun-qin:
The previous non-rsvd max/limit_in_bytes does not account for reserved
huge page memory, making it possible for a process to reserve all the
huge page memory, without being able to allocate it (due to hugetlb
cgroup page fault accounting restrictions).

In practice this makes it possible to successfully mmap more huge page
memory than allowed via the cgroup settings, but when using the memory
the process will get a SIGBUS and crash. This is bad for applications
trying to mmap at startup (and it succeeds), but the program crashes
when starting to use the memory. eg. postgres is doing this by default.

This patch updates and clarifies `LinuxResources.HugepageLimits` and
`LinuxHugepageLimit` by defaulting the configurations go to rsvd hugetlb
cgroup (when supported) and fallback to page fault accounting if not
supported.

Fixes https://github.com/opencontainers/runtime-spec/issues/1050

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -233,12 +233,13 @@ type POSIXRlimit struct {
 	Soft uint64 `json:"soft"`
 }
 
-// LinuxHugepageLimit structure corresponds to limiting kernel hugepages
+// LinuxHugepageLimit structure corresponds to limiting kernel hugepages.
+// Default to reservation limits if supported. Otherwise fallback to page fault limits.
 type LinuxHugepageLimit struct {
-	// Pagesize is the hugepage size
-	// Format: "<size><unit-prefix>B' (e.g. 64KB, 2MB, 1GB, etc.)
+	// Pagesize is the hugepage size.
+	// Format: "<size><unit-prefix>B' (e.g. 64KB, 2MB, 1GB, etc.).
 	Pagesize string `json:"pageSize"`
-	// Limit is the limit of "hugepagesize" hugetlb usage
+	// Limit is the limit of "hugepagesize" hugetlb reservations (if supported) or usage.
 	Limit uint64 `json:"limit"`
 }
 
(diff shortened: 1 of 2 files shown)


snecklifter:
@kailun-qin I'm confused, this patch only seems to include code comments and doc changes?

specs-go/config.go
```diff
@@ -364,7 +365,7 @@ type LinuxResources struct {
 	Pids *LinuxPids `json:"pids,omitempty"`
 	// BlockIO restriction configuration
 	BlockIO *LinuxBlockIO `json:"blockIO,omitempty"`
-	// Hugetlb limit (in bytes)
+	// Hugetlb limits (in bytes). Default to reservation limits if supported.
```
cbandy, line 368: Singular reads correctly to me. Similar to the files, `hugetlb.<hugepagesize>.limit_in_bytes` and `hugetlb.<hugepagesize>.rsvd.limit_in_bytes` of https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/hugetlb.html.

```suggestion
	// Hugetlb limit (in bytes). Default to reservation limit if supported.
```

kolyshkin: The field name is `HugepageLimits`, and it is a slice of limits, so I think the correct way to describe it is "limits" not "limit".

kolyshkin (approved):
LGTM

kolyshkin:
This (together with runtime implementation) should fix the real issue with some software, described in https://github.com/opencontainers/runtime-spec/issues/1050.

specs-go/config.go
cbandy: Ah, yes. Agreed!

cbandy (approved):

giuseppe (approved):
LGTM

AkihiroSuda:
@kailun-qin @odinuge Do you have a PR for runc?
(https://github.com/opencontainers/runc/pull/2360 seems closed)

PR #1117, opened 2021-08-07
defs-zos: [Fix] prevent schema parsers from hitting recursion-loop while resolving types.

flouthoc:
Hi Team, 

Following PR makes sure that `des-zos.json` is consistent with schema definition w.r.t to entire spec.  

I think following PR ( https://github.com/opencontainers/runtime-spec/pull/1095 )  has introduced type-resolution issues on downstream parsers. 
Specifying definition-files for types present in existing `.json` would result in infinite type resolution loop. However existing convention is just using `#` instead of specifying def-file for example  https://github.com/opencontainers/runtime-spec/blob/master/schema/defs-linux.json

Thanks

### DIFF
--- a/schema/defs-zos.json
+++ b/schema/defs-zos.json
@@ -32,16 +32,16 @@
                   "$ref": "defs.json#/definitions/FilePath"
                 },
                 "type": {
-                  "$ref": "defs-zos.json#/definitions/FileType"
+                  "$ref": "#/definitions/FileType"
                 },
                 "major": {
-                  "$ref": "defs-zos.json#/definitions/Major"
+                  "$ref": "#/definitions/Major"
                 },
                 "minor": {
-                  "$ref": "defs-zos.json#/definitions/Minor"
+                  "$ref": "#/definitions/Minor"
                 },
                 "fileMode": {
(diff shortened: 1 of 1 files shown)


giuseppe:
@najohnsn could you please take a look?

flouthoc:
Hi @najohnsn Were you able to take a look at this.

najohnsn:
Sorry, I was on vacation. This looks good to me. Not sure why a _relative_ reference wasn't used before.
@SteeleDesmond, you should take a look too.

tianon:
This seems like it's doing the _opposite_ of what https://github.com/opencontainers/runtime-spec/blob/82ab996a5bf91804a440391bff43140b17bbadd4/schema/config-linux.json does? (lots of `"$ref": "defs-linux.json#/...` in there)

najohnsn:
@tianon, yes, `config-linux.json` has references to `defs-linux.json`. But `defs-linux.json` has the _relative reference_ that is being proposed here for `defs-zos.json`.

tianon:
:facepalm: doh, thanks for clarifying :+1:

tianon (approved):

giuseppe (approved):
LGTM

PR #1120, opened 2021-08-11
config-linux: add CFS bandwidth burst

kailun-qin:
Burstable CFS controller is introduced in Linux 5.14. This helps with
parallel workloads that might be bursty. They can get throttled even
when their average utilization is under quota. And they may be latency
sensitive at the same time so that throttling them is undesired.

This feature borrows time now against the future underrun, at the cost
of increased interference against the other system users, by introducing
`cfs_burst_us` into CFS bandwidth control to enact the cap on unused
bandwidth accumulation, which will then used additionally for burst.

The patch adds the support/control for CFS bandwidth burst.

Fixes https://github.com/opencontainers/runtime-spec/issues/1119

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -360,6 +360,9 @@ The following parameters can be specified to set up the controller:
 
 * **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup
 * **`quota`** *(int64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+    If specified with any (valid) positive value, it MUST be no smaller than `burst` (runtimes MAY generate an error).
+* **`burst`** *(uint64, OPTIONAL)* - specifies the maximum amount of accumulated time in microseconds for which all tasks in a cgroup can run additionally for burst during one period (as defined by **`period`** below)
+    If specified, this value MUST be no larger than any positive `quota` (runtimes MAY generate an error).
 * **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only)
 * **`realtimeRuntime`** *(int64, OPTIONAL)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources
 * **`realtimePeriod`** *(uint64, OPTIONAL)* - same as **`period`** but applies to realtime scheduler only
@@ -373,6 +376,7 @@ The following parameters can be specified to set up the controller:
 "cpu": {
     "shares": 1024,
     "quota": 1000000,
+    "burst": 1000000,
     "period": 500000,
     "realtimeRuntime": 950000,
     "realtimePeriod": 1000000,
--- a/schema/test/config/good/spec-example.json
+++ b/schema/test/config/good/spec-example.json
@@ -275,6 +275,7 @@
             "cpu": {
                 "shares": 1024,
                 "quota": 1000000,
+                "burst": 1000000,
                 "period": 500000,
                 "realtimeRuntime": 950000,
                 "realtimePeriod": 1000000,
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -327,6 +327,9 @@ type LinuxCPU struct {
 	Shares *uint64 `json:"shares,omitempty"`
 	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
 	Quota *int64 `json:"quota,omitempty"`
+	// CPU hardcap burst limit (in usecs). Allowed accumulated cpu time additionally for burst in a
+	// given period.
+	Burst *uint64 `json:"burst,omitempty"`
 	// CPU period to be used for hardcapping (in usecs).
 	Period *uint64 `json:"period,omitempty"`
 	// How much time realtime scheduling may use (in usecs).
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -110,6 +110,9 @@
                             "quota": {
                                 "$ref": "defs.json#/definitions/int64"
                             },
+                            "burst": {
+                                "$ref": "defs.json#/definitions/uint64"
+                            },
                             "realtimePeriod": {
                                 "$ref": "defs.json#/definitions/uint64"
                             },


config-linux.md
```diff
@@ -300,7 +300,12 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
...
-* **`quota`** *(int64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+* **`quota`** *(int64, OPTIONAL)* - specifies the replenished amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+    A value of `-1` indicates that the cgroup does not have any bandwidth restriction in place (i.e., an unconstrained bandwidth group).
+    Any (valid) positive value(s) no smaller than `burst` will enact the specified bandwidth limit.
+* **`burst`** *(uint64, OPTIONAL)* - specifies the maximum amount of accumulated time in microseconds for which all tasks in a cgroup can run additionally for burst during one period (as defined by **`period`** below)
+    A value of `0` for indicates that the cgroup can not accumulate any unused bandwidth. It makes the traditional bandwidth control behavior unchanged.
+    Any (valid) positive value(s) no larger than `quota` will enact the cap on unused bandwidth accumulation.
```
m-yosefpor, line 308: Just a question on this line (actually the original linux commit (https://github.com/torvalds/linux/commit/f418371#diff-cc1a82129952a910fdc4292448c2a097a2ba538bebefcf3c06381e45639ae73eR9814), why values larger than quota should return EINVAL? Do you know the reason?

kailun-qin: @m-yosefpor Thanks for the question!

In short, a burst value much larger than the quota is breaking the design purpose of bwc (as isolation). The limit put on the burst aims to control the interference to other groups.

FYI, this was discussed in the v4 patch set of this feature: https://lwn.net/ml/linux-kernel/YFG4hEOe65cbCo26@hirez.programming.kicks-ass.net/.

m-yosefpor: @kailun-qin Thanks for your explanation. Completely makes sense.

kailun-qin:
Dear spec maintainers @crosbymichael @cyphar @dqminh @giuseppe @hqhq @mrunalp @tianon @vbatts,

I received this feature request several times from runtime users of different orgnizations, would you please kindly take a look at this one so that we can move forward (also https://github.com/opencontainers/runc/pull/3205)?

specs-go/config.go
```diff
@@ -320,6 +320,9 @@ type LinuxCPU struct {
 	Shares *uint64 `json:"shares,omitempty"`
 	// CPU hardcap limit (in usecs). Allowed cpu time in a given period.
 	Quota *int64 `json:"quota,omitempty"`
+	// CPU hardcap burst limit (in usecs). Allowed accumulated cpu time additionally for burst in a
+	// given period.
+	Burst *uint64 `json:"burst,omitempty"`
```
giuseppe, line 325: You also need to add it to `schema/config-linux.json`

kailun-qin: Done, thanks!

config-linux.md
```diff
@@ -359,7 +359,12 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
 The following parameters can be specified to set up the controller:
 
 * **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup
-* **`quota`** *(int64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+* **`quota`** *(int64, OPTIONAL)* - specifies the replenished amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+    A value of `-1` indicates that the cgroup does not have any bandwidth restriction in place (i.e., an unconstrained bandwidth group).
```
giuseppe, line 363: I am not sure about having `-1` here.  On cgroup v2 `quota` and `period` are written together.  Does `-1` mean it has the same value as `period`?

kailun-qin: For cgroup v2, I suppose `-1` means `max`, which can be written together w/ `period`.

According to the the `cpu.max` section of [cgroup v2 doc](https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/cgroup-v2.rst#cpu):
```
cpu.max
A read-write two value file which exists on non-root cgroups. The default is "max 100000".

The maximum bandwidth limit. It's in the following format:
$MAX $PERIOD
which indicates that the group may consume upto $MAX in each $PERIOD duration. "max" for $MAX indicates no limit. If only one number is written, $MAX is updated.
````

**Comment by Zheaoli** [comment]
In my opinion, I think cpu_burst mode is a cgroupv2 only feature. Is there any compatibility issue if we change the all schema which is used in both cgroupv1 and v2?

**Comment by Zheaoli** [comment]
Sorry for the bad information, my memory is wrong

I have confirmed the burst mode support both v1 and v2 since Linux 5.14, FYI https://www.kernel.org/doc/html/latest/scheduler/sched-bwc.html

File: config-linux.md [hunk 4]
```diff
@@ -359,7 +359,12 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
...
+    Any (valid) positive value(s) no smaller than `burst` will enact the specified bandwidth limit.
+* **`burst`** *(uint64, OPTIONAL)* - specifies the maximum amount of accumulated time in microseconds for which all tasks in a cgroup can run additionally for burst during one period (as defined by **`period`** below)
```
Zheaoli, line 365:
```suggestion
* **`burst`** *(uint64, OPTIONAL)* - specifies the maximum amount of accumulated time in microseconds for which all tasks in a cgroup can run additionally for burst during one period (as defined by **`period`** below). This feature just be supported in the Linux 5.14 or newer
```

gregkh:
> But for this, maybe we should think about the kernel issue? The last LTS kernel is 5.10 which is not supported for burst mode(People may use 510 with the newest version of runc . Maybe we should make burst mode as an experimental feature before the Linux 5.15 has been released

The 5.15 kernel was released last year, and has already been marked as "LTS" for quite some time now.  I do not understand why you are not using it now already :)

Zheaoli:
> The 5.15 kernel was released last year, and has already been marked as "LTS" for quite some time now. I do not understand why you are not using it now already :)

Many of our machines are using 510 not 515, it's a tough way to upgrade the kernel version(And my bad again, 515 has been released in  2021-10-31).

config-linux.md
kailun-qin: Normally we do not highlight the kernel version requirement in runtime spec.

config-linux.md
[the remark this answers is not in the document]
kailun-qin: `quota` is defined to be in int64 here, so it's OCI runtime implementation's responsiblity to handle this differently for v1 and v2.

config-linux.md
Zheaoli: OK, Got it

Zheaoli:
I think the spec is LGTM to me now.

For my personal idea, I'm very glad to see this spec will be accepted ASAP, because it's very useful for some performance-needed circumstances.

giuseppe (approved):
LGTM

Zheaoli:
Hello all, would you mind telling me if this PR is mergeable?

config-linux.md
kailun-qin: I think this applies to all negative integers actually (i.e., not only `-1`). Any suggestions? Thanks!

config-linux.md
tianon: Personally, I think it would be valuable for us to list the relevant kernel version that added each feature (above some global minimum we assume, which I don't think is currently codified anywhere in the spec even though we definitely do assume _some_ minimum), but I agree that it doesn't make sense to do it just for this one instance. :smile:

config-linux.md
tianon: I would guess the question here is really why it's important to `burst` support that we include support for `-1` here too -- I'd say this is a breaking change, since the spec doesn't currently say anything about negative numbers, and instead implies that this optional field be left unspecified for no restriction.  Does `cgroupv1` support negative values?

To be clear, I agree that we should clarify this, especially that unspecified needs to be `max` in `cgroupv2`, but I'm not sure how it relates to `burst` except that it's adjacent.

config-linux.md
```diff
@@ -359,7 +359,12 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
 The following parameters can be specified to set up the controller:
 
 * **`shares`** *(uint64, OPTIONAL)* - specifies a relative share of CPU time available to the tasks in a cgroup
-* **`quota`** *(int64, OPTIONAL)* - specifies the total amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
+* **`quota`** *(int64, OPTIONAL)* - specifies the replenished amount of time in microseconds for which all tasks in a cgroup can run during one period (as defined by **`period`** below)
```
tianon, line 362: IMO, "the total amount ... during one period" already implies that this is "replenished" and changing this word makes it less clear :see_no_evil:

config-linux.md
```diff
@@ -359,7 +359,12 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
...
+    A value of `-1` indicates that the cgroup does not have any bandwidth restriction in place (i.e., an unconstrained bandwidth group).
+    Any (valid) positive value(s) no smaller than `burst` will enact the specified bandwidth limit.
```
tianon, line 364: My understanding so far is that `burst` is either excess time "earned" from past under-usage or borrowed from future periods where we expect/hope we'll be under-utilizing (I'm not actually clear from this PR which it is).  Either way, in this specific case, I think the restriction is really on the value of _`burst`_, not a restriction on the value of `quota` (so doesn't seem helpful to mention here).

config-linux.md
tianon, line 365: > This feature borrows time now against the future underrun ...

Given this, I don't think "accumulated" is the right choice of word here -- this feature is about borrowing from the future, not accumulating from the past:

```suggestion
* **`burst`** *(uint64, OPTIONAL)* - specifies the maximum amount of burst time in microseconds for which all tasks in a cgroup can borrow from future periods during one period (as defined by **`period`** below)
```

config-linux.md
```diff
@@ -359,7 +359,12 @@ For more information, see the kernel cgroups documentation about [cpusets][cgrou
...
+* **`burst`** *(uint64, OPTIONAL)* - specifies the maximum amount of accumulated time in microseconds for which all tasks in a cgroup can run additionally for burst during one period (as defined by **`period`** below)
+    A value of `0` for indicates that the cgroup can not accumulate any unused bandwidth. It makes the traditional bandwidth control behavior unchanged.
+    Any (valid) positive value(s) no larger than `quota` will enact the cap on unused bandwidth accumulation.
```
tianon, line 367: I think this would be more consistent if we assume that not specifying this optional field is the way to get "no limit", and I think it would also make sense to be more clear about values bigger than `quota` are not just ignored but rather an error, perhaps something like this?

```suggestion
    If specified, this value MUST be no larger than `quota` (runtimes MAY generate an error).
```

**Edit:** coming back to this after https://github.com/opencontainers/runc/pull/3205#discussion_r705484770, I'm guessing that's related? (And likely why the "negative values" bits were added to `quota`?)

tianon (commented):
Overall, I agree we should add this, and I think the suggested name/format are fine (they seem to match what the kernel uses, from what I can tell), but I think we need to make some adjustments to the description -- I'm getting mixed signals about how this feature works from this PR, and I think we can improve that. :smile:

tianon:
Perhaps it would help me (and other spec reviewers) if we level-set on how the feature works.  Here's my current understanding -- is this accurate?

> there's at most `quota`+`burst` in any one given `period`, and any use of `burst` in a `period` counts against a future `period`'s available `quota`

Zheaoli:
Ping? Is there any update about this issue?

Zheaoli:
ping

config-linux.md
kailun-qin: `This feature borrows time now against the future underrun ...` is directly from the [kernel doc](https://www.kernel.org/doc/html/latest/scheduler/sched-bwc.html) and it's actually `accumulating from the past` rather than borrowing from future. It intends to say that `borrowing time now to prepare for the future underestimation`.

kailun-qin:
> Perhaps it would help me (and other spec reviewers) if we level-set on how the feature works. Here's my current understanding -- is this accurate?
> 
> > there's at most `quota`+`burst` in any one given `period`, and any use of `burst` in a `period` counts against a future `period`'s available `quota`

@tianon Please kindly take another look and see if it addresses the question. Thanks!

Zheaoli:
@tianon ping.

Hello guys, is there any blocking issue with this pull request?

AkihiroSuda (approved):
LGTM, but please consider squashing commits

kailun-qin:
> LGTM, but please consider squashing commits

Squashed and rebased to the latest, PTAL, thanks!

kailun-qin:
> LGTM, but please consider squashing commits

Squashed and rebased, PTAL, thanks!

AkihiroSuda (approved):

Zheaoli:
@tianon Hi is there any potential blocking issue for this PR?

Zheaoli:
@tianon ping...

kaffarell:
Can we merge this? Looks ready...

Zheaoli:
ping... I think it's ready for merge

Zheaoli:
ping

Zheaoli:
Is there any chance to merge this PR in 2022?(A lot of people really need this(

Zheaoli:
ping, ping, ping...

schema/test/config/good/spec-example.json
```diff
@@ -275,6 +275,7 @@
             "cpu": {
                 "shares": 1024,
                 "quota": 1000000,
+                "burst": 1000000,
```
marquiz, line 278: Period **MUST be no smaller than**(?)

kailun-qin: > Period MUST be no smaller than burst

Why is this contraint needed? `burst` can be greater than `period`.

marquiz: You're right. Sorry, my bad, I misread the new content on field description above 🙈

marquiz (approved):
lgtm

giuseppe (approved):
LGTM

Zheaoli:
There is two LGTM from the people who have the write access to this repo. Do we need more LGTM?

PR #1123, opened 2021-09-10
specify cgroup ownership semantics

frasertweedale:
cgroups v2 supports secure delegation of cgroups.  Accordingly,
control over a cgroup (that is, creation of new child cgroups and
movement of processes and threads among the cgroup subtree exposed
to a container) can be safely delegated to a container.  Adjusting
the ownership enables real-world use cases like systemd-based
containers fully isolated in user namespaces.

To encourage adoption of this feature, and secure implementation,
define the semantics of cgroup ownership.  Changing/setting the
cgroup ownership is only allowed on cgroups v2, and the specific
files whose ownership can be change are mentioned.

In terms of current practice, this is already the behaviour of crun
(which also chown's the memory.oom.group file), and there is a pull
request for runc: https://github.com/opencontainers/runc/pull/3057
(the behaviour is enabled by an annotation).

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -196,6 +196,51 @@ For example, to run a new process in an existing container without updating limi
 
 Runtimes MAY attach the container process to additional cgroup controllers beyond those necessary to fulfill the `resources` settings.
 
+### Cgroup ownership
+
+Runtimes MAY, according to the following rules, change (or cause to
+be changed) the owner of the container's cgroup to the host uid that
+maps to the value of `process.user.uid` in the [container
+namespace](glossary.md#container-namespace); that is, the user that
+will execute the container process.
+
+Runtimes SHOULD NOT change the ownership of container cgroups when
+cgroups v1 is in use.  Cgroup delegation is not secure in cgroups
+v1.
+
+A runtime SHOULD NOT change the ownership of a container cgroup
+unless it will also create a new cgroup namespace for the container.
+Typically this occurs when the `linux.namespaces` array contains an
+object with `type` equal to `"cgroup"` and `path` unset.
+
+Runtimes SHOULD change the cgroup ownership if and only if the
+cgroup filesystem is to be mounted read/write; that is, when the
+configuration's `mounts` array contains an object where:
+
+- The `source` field is equal to `"cgroup"`
+- The `destination` field is equal to `"/sys/fs/cgroup"`
+- The `options` field does not contain the value `"ro"`
+
+If the configuration does not specify such a mount, the runtime
+SHOULD NOT change the cgroup ownership.
+
+A runtime that changes the cgroup ownership SHOULD only change the
+ownership of the container's cgroup directory and files within that
+directory that are listed in `/sys/kernel/cgroup/delegate` (see
+`cgroups(7)` for details about this file).  If the
+`/sys/kernel/cgroup/delegate` file does not exist, the runtime MUST
+fall back to using the following list of files:
+
+```
+cgroup.procs
+cgroup.subtree_control
+cgroup.threads
+```
+
+The runtime SHOULD NOT change the ownership of any other files.
+Changing other files may allow the container to elevate its own
+resource limits or perform other unwanted behaviour.
+
 ### Example
 
 ```json


config-linux.md
```diff
@@ -196,6 +196,27 @@ For example, to run a new process in an existing container without updating limi
...
+
+Runtimes MAY change (or cause to be changed) the owner of the
+container's cgroup to the host uid that maps to uid 0 in the
+container's user namespace, if cgroups v2 is in use.
+
+Runtimes MUST NOT change the ownership of container cgroups when
+cgroups v1 is in use.  Cgroup delegation is not secure in cgroups
+v1.
+
+Runtimes MUST only change the ownership of the container's cgroup
```
AkihiroSuda, line 209: By default runtime MUST NOT change anything.
We need a new config field to opt-in to this.

frasertweedale: That would put some existing implementations in immediately violation of the spec - `crun`, at least, already chowns the cgroup.  I'd appreciate more perspectives on this proposal.

AkihiroSuda: But if we adopt the crun behavior, that would put runc in violation of the spec.

As runc is the reference implementation of the spec,  I suggest modifying this PR and crun implementation.

frasertweedale: The proposal does not say runtimes MUST chown the cgroup - only that if they do, they MUST chown only the particular files mentioned.  I will update the language to clarify.

giuseppe: by default `/sys/fs/cgroup` is mounted as read-only so in practice the cgroup ownership doesn't change much in what the container is allowed to do.  In order to use the `/sys/fs/cgroup` the mount must be writeable.

Maybe we can do the chowning when `/sys/fs/cgroup` is mounted as `rw`?  Otherwise it is unhelpful to have `/sys/fs/cgroup` writeable from an user not present in the current user namespace.

AkihiroSuda: > Maybe we can do the chowning when /sys/fs/cgroup is mounted as rw?

SGTM :+1:
I think when the mount is RW && cgroup is v2, we can safely say that the runtime `SHOULD` chown the cgroup.

AkihiroSuda: @kolyshkin @cyphar PTAL

frasertweedale: @giuseppe @AkihiroSuda thanks for your suggestion.  I updated the text to specify that the runtime SHOULD chown the cgroup when the cgroupfs is mounted RW (and is cgroups v2), otherwise MUST NOT.

AkihiroSuda (approved):

config-linux.md
```diff
@@ -196,6 +196,39 @@ For example, to run a new process in an existing container without updating limi
...
 
+### Cgroup ownership
+
+Runtimes MAY change (or cause to be changed) the owner of the
+container's cgroup to the host uid that maps to uid 0 in the
+container's user namespace, according to the following rules.
+
+Runtimes MUST NOT change the ownership of container cgroups when
+cgroups v1 is in use.  Cgroup delegation is not secure in cgroups
+v1.
```
cyphar, line 207: This should be a SHOULD NOT imho. While it might be a bad idea for security reasons, making this be something the spec explicitly disallows is a bit too extreme -- SHOULD NOT is strong enough, and including a justification is sufficient. There are reasons why you might want to do something different, and explicitly not allowing it a bit too extreme (not to mention adding MUST NOTs like this could be argued to be a SemVer-breaking change).

config-linux.md
```diff
@@ -196,6 +196,39 @@ For example, to run a new process in an existing container without updating limi
...
+ownership of the container's cgroup directory and the following
+files within that directory:
+
+- `cgroup.procs`
+- `cgroup.subtree_control`
+- `cgroup.threads`
+
+The runtime MUST NOT change the ownership of any other files.
+Changing other files may allow the container to elevate its own
+resource limits or perform other unwanted behaviour.
```
cyphar, line 230: Same comment about MUST and MUST NOT here. It should be SHOULD and SHOULD NOT. It's also entirely possible that it would be safe to allow the container to control the cgroup it's in (for instance if the runtime put the container in a nested cgroup and enforced the restrictions in the parent -- AFAIK this kind of thing is necessary to run systemd in containers so making it something you are not allowed to do is way too extreme IMHO).

config-linux.md
```diff
@@ -196,6 +196,39 @@ For example, to run a new process in an existing container without updating limi
...
+Runtimes SHOULD change the cgroup ownership if and only if the
+cgroup filesystem is to mounted read/write; that is, when the
+configuration's `mounts` array contains an object where:
+
+- The `source` field is equal to `"cgroup"`
+- The `destination` field is equal to `"/sys/fs/cgroup"`
+- The `options` field does not contain the value `"ro"`
+
+If the configuration does not specify such a mount, the runtime MUST
+NOT change the cgroup ownership.
```
cyphar, line 218: Same comment about MUST NOT here -- but in addition, the use of the phrase "if and only if" above makes this MUST NOT requirement a contradiction -- if I SHOULD do something if and only if X, then if not X I SHOULD NOT do it.

cyphar (changes requested):
Seems reasonable but using MUSTs is far too strong IMHO.

frasertweedale:
@cyphar thanks for your comments. I'll push an update later today, or tomorrow.

config-linux.md
```diff
@@ -196,6 +196,39 @@ For example, to run a new process in an existing container without updating limi
...
+Runtimes MAY change (or cause to be changed) the owner of the
+container's cgroup to the host uid that maps to uid 0 in the
+container's user namespace, according to the following rules.
+
+Runtimes MUST NOT change the ownership of container cgroups when
+cgroups v1 is in use.  Cgroup delegation is not secure in cgroups
+v1.
+
+Runtimes SHOULD change the cgroup ownership if and only if the
+cgroup filesystem is to mounted read/write; that is, when the
```
kolyshkin, line 210: nit: `s/to mounted/to be mounted/`

config-linux.md
frasertweedale: relaxed to SHOULD.

config-linux.md
frasertweedale: relaxed to SHOULD.

config-linux.md
frasertweedale: relaxed to SHOULD.

config-linux.md
frasertweedale: Fixed, thanks!

frasertweedale:
PR updated to fix a grammar issue and relax MUST to SHOULD per @cyphar's comments.

kolyshkin:
One question: should we also require cgroupns? It is assumed with cgroup v2, but what it it is not set in config?

I have just checked that by removing cgroup ns from config.json, and the whole host /sys/fs/cgroup is mounted (for both crun and runc). Meaning we definitely need to add cgroupns to the list of conditions.

frasertweedale:
> One question: should we also require cgroupns? It is assumed with cgroup v2, but what it it is not set in config?
> 
> I have just checked that by removing cgroup ns from config.json, and the whole host /sys/fs/cgroup is mounted (for both crun and runc). Meaning we definitely need to add cgroupns to the list of conditions.

@kolyshkin good point, I agree.  Should we further restrict it to a cgroupns *without* `path` specified?  That is, where the runtime must create a new cgroupns?  Otherwise `path` could be a reference to an arbitrary, pre-existing cgroupns (possibly the "root" host cgroupns).

frasertweedale:
Are there any further comments about this PR.  Any further concerns?  What else needs to be done for this PR to be ready to merge?

tianon (approved):
Seems sane to me

@cyphar @kolyshkin willing to take another look? 😅🙏

frasertweedale:
Dear maintainers / reviewers, it would be great if we could get some more eyes on this and hopefully, approval and merge.
Ping @giuseppe @mrunalp ?

giuseppe (approved):
LGTM

dqminh:
LGTM

frasertweedale:
Thank you to reviewers.  We now have the requisite number of approvals.  @cyphar has an older review with changes requested; I believe I have resolved all of the feedback.

Can this be merged now?

frasertweedale:
Weekly bump.  Could someone please merge it, or explain what prevents it?

frasertweedale:
Weekly bump (Nov 22).   Could someone please merge it, or explain what else needs to be done or is preventing this PR from being merged?

vbatts:
@frasertweedale thank you for the weekly bumps 😬 
Looks like the "MUST" that @cyphar commented on was never marked resolved. Perhaps that's why the other maintainers didn't merge? 🤷 

LGTM

frasertweedale:
@vbatts thank you so much!  Thanks to all for their feedback and insights.

PR #1124, opened 2021-09-22
relocate GOVERNANCE to oci/org

vbatts:
This doc is a left over from a time that existed where the original
github.com/opencontainers/specs was the only place to particularly have
"TDC" maintainers.

This needs to live org-wide, so it can offer broader guidance.

Relates to https://github.com/opencontainers/.github/pull/33


tianon (approved):
LGTM

cyphar (changes requested):
Putting in an explicit NACK -- see my above comment.

vbatts:
:+1: 

This came up in the calls as well. This current templated project goveranance is too wordy, and difficult to track the changes per project. 

Some things are going to be generic, like 2/3 vote quorum for specs and simple majority (or even 2 LGTM) for source-code projects. 

I'll close this, and expect that we can prune this doc, hopefully putting in place something more modular

PR #1131, opened 2021-12-29
Introduce mdbook.

utam0k:
I have always used this specification to create youki. Thank you very much.
However, the ToC is only in one place, so it is difficult to grasp the whole structure. Also, I have to use github to search. How about introducing mdbook to solve these problems? You can use github pages for operation, so it won't be that hard.
Here's what it looks like in practice.
https://www.utam0k.jp/runtime-spec/

After merging, you need to do this in the repository settings.
image

### DIFF
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -1,0 +1,27 @@
+name: mdbook
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup mdBook
+        uses: peaceiris/actions-mdbook@v1
+        with:
+          mdbook-version: "latest"
+      - name: Build mdbook
+        working-directory: .
+        run: mdbook build
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        if: ${{ github.ref == 'refs/heads/main' }}
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./book
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 output
 schema/validate
 version.md
+book/
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 The [Open Container Initiative][oci] develops specifications for standards on Operating System process and application containers.
 
-The specification can be found [here](spec.md).
+The specification can be found [here](https://opencontainers.github.io/runtime-spec).
 
 ## Table of Contents
 
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -1,0 +1,19 @@
+# Summary
+
+[Open Container Initiative Runtime Specification](./spec.md)
+
+- [The 5 principles of Standard Containers](./principles.md)
+- [Filesystem Bundle](./bundle.md)
+- [Runtime and Lifecycle](./runtime.md)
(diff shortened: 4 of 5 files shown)


tianon:
We've got something like this already in https://specs.opencontainers.org/runtime-spec/ - can you comment on the differences?

cc @vsoch @jdolitsky

vsoch:
It's a nice book! The main difference (as I see it) is that this still is going to have the repos as separate. E.g., you can't easily click from the runtime docs to the distribution docs. Another layer would be needed to bring them together, and that still would be ultimately clicking between different sites. We also would want more customization to be able to open an issue, edit on GitHub etc.

@utam0k is there a way to combine a bunch of repos into one? I also suspect if one repo is linked from the other, it would just maintain the link and jump to the other GitHub.

utam0k:
> We've got something like this already in https://specs.opencontainers.org/runtime-spec/ - can you comment on the differences?
> 
> cc @vsoch @jdolitsky

@tianon 
Hi! Thanks for your comment. Although I don't know this page on which you comment, The links on this site may be causing broken links. The mdbook doesn't need to take care of between links on the original markdown.

utam0k:
> It's a nice book! The main difference (as I see it) is that this still is going to have the repos as separate. E.g., you can't easily click from the runtime docs to the distribution docs. Another layer would be needed to bring them together, and that still would be ultimately clicking between different sites. We also would want more customization to be able to open an issue, edit on GitHub etc.
> 
> @utam0k is there a way to combine a bunch of repos into one? I also suspect if one repo is linked from the other, it would just maintain the link and jump to the other GitHub.

Thanks for your review! I'm very sorry but I'm not sure to understand what you want to say however how about preparing a new markdown which references other specifications. I'm sorry if I'm missing the point which you pointed out 😭 
Also, How are you currently dealing with that problem?

vsoch:
> Thanks for your review! I'm very sorry but I'm not sure to understand what you want to say however how about preparing a new markdown which references other specifications. I'm sorry if I'm missing the point which you pointed out sob
Also, How are you currently dealing with that problem?

Okay - let me try again! So the example here is one specification repository, the runtime-spec, in very pretty docs. However there are many specs - the image, distribution, etc. This solution would create a _separate_ website for each specification. What would be required for a full transition of the current solution to the one here is to create a full set of repos, one per spec, _plus_ some unifying interface that brings them together.

My solution isn't magic - it's a simple template that basically allows you to register a spec repository, and then it shows up rendered into an RFC template. You can look at the source code if that helps! https://github.com/opencontainers/specs.opencontainers.org.

utam0k:
@vsoch @tianon 
Thanks for explaining again. The way it is now, the spec itself and the home page are separate repositories, right? That seems quite difficult to maintain. In this PR, a `gh-pages` branch is automatically created by github actions and static html is pushed to it. So there is no need for a new repository. In addition, unlike the current homepage, one repository can automatically keep up with the latest specifications. What do you think?

vsoch:
@utam0k the approach was adopted because people found it hard to navigate the specs because they weren't in one place. I also want to note that no manual work is needed - the spec files are loaded on demand when the page is loaded. You basically just register a spec repo there and forget about it. 

I think I'd be open to an approach that automates deployment of a spec docs from its respective repository, but to fulfill the original need of having a holistic set of docs I'd also want a central place / means to link them together.

utam0k:
@vscoch The current page is useful enough now, so I'll close this PR. I'd love for you to add it to the README or something!

vsoch:
Sure, do you mean just a link from the respective repos? Or maybe better to have it in the description (as a URL) so that it isn't then rendered and links to itself? :laughing: 

And I just want to say I really like the design of what you did here! Thank you for that, and if you ever wanted to come back and rethink what we currently have I think everyone is very open to that granted that we make tiny improvements toward something better.

Issue #1135, opened 2022-02-15
Proposal: add idle option for container cgroup


wineway:
[Cgroup SCHED_IDLE support](https://lore.kernel.org/lkml/162971078674.25758.15464079371945307825.tip-bot2@tip-bot2/#r)  was [merged](https://github.com/torvalds/linux/commit/5d3c0db4598c5de511824649df2aa976259cf10a) in Linux mainline since 5.15.

with this feature, we can preempt low priority containers when there is some container with higher priority or expect less latency

this is necessary for mixture deployment of offline service and online service on the container environment

PR #1136, opened 2022-02-16
config-linux: add idle option for container cgroup

wineway:
[sched: Cgroup SCHED_IDLE support](https://lore.kernel.org/lkml/162971078674.25758.15464079371945307825.tip-bot2@tip-bot2/#r) is introduced in Linux 5.15

we can mark a container SCHED_IDLE with this feature which means its sched weight are configured to the minimum

see also #1135

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -361,6 +361,7 @@ The following parameters can be specified to set up the controller:
 * **`realtimePeriod`** *(uint64, OPTIONAL)* - same as **`period`** but applies to realtime scheduler only
 * **`cpus`** *(string, OPTIONAL)* - list of CPUs the container will run in
 * **`mems`** *(string, OPTIONAL)* - list of Memory Nodes the container will run in
+* **`idle`** *(int64, OPTIONAL)* - cgroups are configured with minimum weight, 0: default behavior, 1: SCHED_IDLE.
 
 #### Example
 
@@ -372,7 +373,8 @@ The following parameters can be specified to set up the controller:
     "realtimeRuntime": 950000,
     "realtimePeriod": 1000000,
     "cpus": "2-3",
-    "mems": "0-7"
+    "mems": "0-7",
+    "idle": 0
 }
 ```
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -330,6 +330,8 @@ type LinuxCPU struct {
 	Cpus string `json:"cpus,omitempty"`
 	// List of memory nodes in the cpuset. Default is to use any available memory node.
 	Mems string `json:"mems,omitempty"`
+	// cgroups are configured with minimum weight, 0: default behavior, 1: SCHED_IDLE. Default 0
+	Idle *int64 `json:"idle,omitempty"`
 }
 
 // LinuxPids for Linux cgroup 'pids' resource management (Linux 4.3)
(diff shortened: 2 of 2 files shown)


giuseppe (approved):
LGTM

tianon:
Looking at https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee, I think this was even in _5.15_. :sweat_smile:

config-linux.md
```diff
@@ -361,6 +361,7 @@ The following parameters can be specified to set up the controller:
 * **`realtimePeriod`** *(uint64, OPTIONAL)* - same as **`period`** but applies to realtime scheduler only
 * **`cpus`** *(string, OPTIONAL)* - list of CPUs the container will run in
 * **`mems`** *(string, OPTIONAL)* - list of Memory Nodes the container will run in
+* **`idle`** *(int64, OPTIONAL)* - cgroups are configured with minimum weight, 0: default behavior, 1: SCHED_IDLE.
```
tianon, line 364: I'm getting mixed signals on the actual "type" of this value from https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee#diff-1b49efb1a709264f46f4ac518b965e9620acc2f5b928209547313b8dfc9f1286R400 (`int`) vs https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee#diff-1b49efb1a709264f46f4ac518b965e9620acc2f5b928209547313b8dfc9f1286R511 (`long`) -- it doesn't make a _huge_ difference since the only values are 0 and 1, but I think any future changes/improvements will benefit if we match the kernel's type size from the get go. :sweat_smile:

Any ideas which one is "canonical" ? (unfortunately I'm not very familiar with the kernel source to be able to parse this diff better than the two conflicting lines I found :see_no_evil:)

thaJeztah: Doest it support magic values like `-1` ? (wondering if it would make it easier to use to use an uintXX if it only supports 0 and 1

specs-go/config.go
```diff
@@ -330,6 +330,8 @@ type LinuxCPU struct {
 	Cpus string `json:"cpus,omitempty"`
 	// List of memory nodes in the cpuset. Default is to use any available memory node.
 	Mems string `json:"mems,omitempty"`
+	// cgroups are configured with minimum weight, 0: default behavior, 1: SCHED_IDLE. Default 0
+	Idle *int64 `json:"idle,omitempty"`
```
thaJeztah, line 334: Do we want `const` / enums for these values (for easier consumption)?

thaJeztah: Also, if `0` is the default behavior, it may not be needed to define it as a pointer

wineway:
> Looking at [torvalds/linux@3040003](https://github.com/torvalds/linux/commit/304000390f88d049c85e9a0958ac5567f38816ee), I think this was even in _5.15_. 😅

yes, updated issue & pr comment.

config-linux.md
tianon: The text around it implies that it might someday?  It does _not_ currently.  Currently it supports 0 and 1.  What I'd like to do is make sure the type we choose matches the type the kernel uses, but I guess `int64` fits both of C's `int` and `long` on all platforms I'm aware of, so I guess it should be plenty. :+1:

tianon (approved):

vbatts (approved):

PR #1138, opened 2022-02-23
Add available `LinuxSeccompFlag`s

saschagrunert:
We now list the available `LinuxSeccompFlag` values as part of the runtime spec.

cc @alban

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -615,6 +615,19 @@ type Arch string
 // LinuxSeccompFlag is a flag to pass to seccomp(2).
 type LinuxSeccompFlag string
 
+const (
+	// LinuxSeccompFlagLog is a seccomp flag to request all returned
+	// actions except SECCOMP_RET_ALLOW to be logged. An administrator may
+	// override this filter flag by preventing specific actions from being
+	// logged via the /proc/sys/kernel/seccomp/actions_logged file. (since
+	// Linux 4.14)
+	LinuxSeccompFlagLog LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
+
+	// LinuxSeccompFlagSpecAllow can be used to disable Speculative Store
+	// Bypass mitigation. (since Linux 4.17)
+	LinuxSeccompFlagSpecAllow LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_SPEC_ALLOW"
+)
+
 // Additional architectures permitted to be used for system calls
 // By default only the native architecture of the kernel is permitted
 const (


specs-go/config.go
```diff
@@ -615,6 +615,25 @@ type Arch string
...
+	// filter flag by preventing specific actions from being logged via the
+	// /proc/sys/kernel/seccomp/actions_logged file. (since Linux 4.14)
+	LinuxSeccompFlagLog LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
+
+	// LinuxSeccompFlagTsync is the filter to synchronize all other threads of
+	// the calling process to the same seccomp filter tree. A "filter tree" is
+	// the ordered list of filters attached to a thread. (Attaching identical
+	// filters in separate seccomp() calls results in different filters from
+	// this perspective.)
+	LinuxSeccompFlagTsync LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_TSYNC"
```
alban, line 630: With @rata we were discussing whether to remove tsync from the spec. In the context of container runtimes,  the process will call seccomp() before exec() to the entrypoint, so all other threads, if any, will be terminated anyway, so tsync does not make a difference.

Examples:

* AFAIU crun is not multithreaded, so tsync is meaningless
* runc is written in Go so multithreaded, but uses libseccomp-golang that always set tsync: it's not configurable because threads and Go routine are not necessarily tightly coupled.

specs-go/config.go
```diff
@@ -615,6 +615,25 @@ type Arch string
...
+
+	// LinuxSeccompFlagSpecALlow can be used to disable Speculative Store
+	// Bypass mitigation. (since Linux 4.17)
+	LinuxSeccompFlagSpecALlow LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_SPEC_ALLOW"
```
alban, line 634:
```suggestion
	// LinuxSeccompFlagSpecAllow can be used to disable Speculative Store
	// Bypass mitigation. (since Linux 4.17)
	LinuxSeccompFlagSpecAllow LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_SPEC_ALLOW"
```

specs-go/config.go
```diff
@@ -615,6 +615,25 @@ type Arch string
 // LinuxSeccompFlag is a flag to pass to seccomp(2).
 type LinuxSeccompFlag string
 
+const (
+	// LinuxSeccompFlagLog is the filter to return actions except
+	// SECCOMP_RET_ALLOW should be logged. An administrator may override this
+	// filter flag by preventing specific actions from being logged via the
+	// /proc/sys/kernel/seccomp/actions_logged file. (since Linux 4.14)
+	LinuxSeccompFlagLog LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
```
alban, line 623:
```suggestion
	// LinuxSeccompFlagLog is a seccomp flag to request all returned
	// actions except SECCOMP_RET_ALLOW to be logged. An administrator may
	// override this filter flag by preventing specific actions from being
	// logged via the /proc/sys/kernel/seccomp/actions_logged file. (since
	// Linux 4.14)
	LinuxSeccompFlagLog LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
```

specs-go/config.go
giuseppe: I've added it without thinking if it made any sense as I've copied all the existing flags.  I agree that it is pointless in the OCI specs, and we could drop it.

rata: I don't think it should be removed, I think it should be marked as deprecated.

See this: https://github.com/opencontainers/runtime-spec/issues/1077

specs-go/config.go
```diff
@@ -615,6 +615,26 @@ type Arch string
...
+	// override this filter flag by preventing specific actions from being
+	// logged via the /proc/sys/kernel/seccomp/actions_logged file. (since
+	// Linux 4.14)
+	LinuxSeccompFlagLog LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
+
+	// LinuxSeccompFlagTsync is the filter to synchronize all other threads of
+	// the calling process to the same seccomp filter tree. A "filter tree" is
+	// the ordered list of filters attached to a thread. (Attaching identical
+	// filters in separate seccomp() calls results in different filters from
+	// this perspective.)
```
rata, line 630: As I pointed out in another PR doing the same, I don't think we should add a const for this: https://github.com/opencontainers/runtime-spec/pull/1108#discussion_r680080483

rata (commented):
As a side note, there is already an open PR doing this: https://github.com/opencontainers/runtime-spec/pull/1108

specs-go/config.go
saschagrunert: Good, removed it from the code now.

rata:
@saschagrunert thanks, this SGTM. It is also handling the comments that were raised by @tianon  [in the other PR](https://github.com/opencontainers/runtime-spec/pull/1108#issuecomment-881778314). 

I'd cc @thaJeztah and @tianon just in case. I don't mind which PR ends up being merged, but IMO there should be some coordination with authors of the other PR :)

specs-go/config.go
```diff
@@ -615,6 +615,19 @@ type Arch string
 // LinuxSeccompFlag is a flag to pass to seccomp(2).
 type LinuxSeccompFlag string
 
+const (
+	// LinuxSeccompFlagLog is a seccomp flag to request all returned
+	// actions except SECCOMP_RET_ALLOW to be logged. An administrator may
+	// override this filter flag by preventing specific actions from being
+	// logged via the /proc/sys/kernel/seccomp/actions_logged file. (since
+	// Linux 4.14)
+	LinuxSeccompFlagLog LinuxSeccompFlag = "SECCOMP_FILTER_FLAG_LOG"
```
vbatts, line 624: @thaJeztah since the name of these variables is different than in [your PR](https://github.com/opencontainers/runtime-spec/pull/1108/files#diff-9790b3b8b39fc583eab8f6d97ac27ae3f35dbe339000c746faf64615125a8ec7R619) (`FlagLog` and `FlagAllow`) I'm wondering if Docker/moby has already _used_ your variables?
I'm inclined to merge this PR given the robust comments and that it drops the TS_SYNC flag that @rata called out

vbatts (approved):

AkihiroSuda:

tianon (approved):

PR #1139, opened 2022-03-09
config-linux: Fix example of cpu quota

hqhq:
Quota should be smaller than period length.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -371,8 +371,8 @@ The following parameters can be specified to set up the controller:
 ```json
 "cpu": {
     "shares": 1024,
-    "quota": 1000000,
-    "period": 500000,
+    "quota": 500000,
+    "period": 1000000,
     "realtimeRuntime": 950000,
     "realtimePeriod": 1000000,
     "cpus": "2-3",


giuseppe (approved):
LGTM

config-linux.md
```diff
@@ -371,8 +371,8 @@ The following parameters can be specified to set up the controller:
 ```json
 "cpu": {
     "shares": 1024,
-    "quota": 1000000,
-    "period": 500000,
+    "quota": 500000,
```
kailun-qin, line 374: Why quota needs to be smaller than period? The config seems valid on a multi-CPU machine.

config-linux.md
```diff
@@ -371,8 +371,8 @@ The following parameters can be specified to set up the controller:
 ```json
```
kailun-qin, line 371: Do we need to change values in `schema/test/config/good/spec-example.json` accordingly in this case?

config-linux.md
tianon: Indeed -- this change would effectively mean a single container could never use more than one core's worth of CPU time.

When I use Docker's `--cpus` shorthand for setting these options, here's what `/sys/fs/cgroup/cpu.max` looks like:

```console
$ docker run --rm --cpus 2 bash cat /sys/fs/cgroup/cpu.max
200000 100000
$ docker run --rm --cpus 0.5 bash cat /sys/fs/cgroup/cpu.max
50000 100000
```

(Both very legal and sane configurations.)

tianon (changes requested):
(for clarity, adding a "review" as an explicit NACK -- see discussion above)

config-linux.md
hqhq: Never mind, not saying it was a mistake, just figured quota less than period would be more understandable, extra description might needed to explain the multi cpu case (like examples in: https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt) which I thought can be avoided.

Thanks for your comments.

PR #1140, opened 2022-03-10
Hook: add type attribute to spec

liubin:
For vm-based runtimes, like Kata Containers, the hooks become
confused, some hooks may need run on the host side, but others
may need to run on the guest side.

Add type attribute identity where the hooks should run, the
host side or guest side, and let runtime treat them separately.

Signed-off-by: bin liu <liubin0329@gmail.com>

### DIFF
--- a/config.md
+++ b/config.md
@@ -390,6 +390,8 @@ For POSIX platforms, the configuration structure supports `hooks` for configurin
             * **`env`** (array of strings, OPTIONAL) with the same semantics as [IEEE Std 1003.1-2008's `environ`][ieee-1003.1-2008-xbd-c8.1].
             * **`timeout`** (int, OPTIONAL) is the number of seconds before aborting the hook.
                 If set, `timeout` MUST be greater than zero.
+            * **`type`** *(string, OPTIONAL)* - hook type, which defines where should the hooks run.
+                Valid values are `host` or `guest`, defaults to `host`.
         * The value of `path` MUST resolve in the [runtime namespace](glossary.md#runtime-namespace).
         * The `prestart` hooks MUST be executed in the [runtime namespace](glossary.md#runtime-namespace).
     * **`createRuntime`** (array of objects, OPTIONAL) is an array of [`createRuntime` hooks](#createRuntime-hooks).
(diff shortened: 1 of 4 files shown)


giuseppe:
how are these used by Kata?  Is there any open PR that uses the new attribute?

flouthoc (commented):
Not at all sure but I think instead of this specific field `type` with hardcoded enums like `guest` or `host` in `hooks` it should be more generic field like ~~maybe `labels` or `annotations`~~ or atleast `type` should accept arbitrary `string` not hardcoded `guest` or `host` enum.

Reason:
Arbitrary implementations could use values to customize their own implementation on the basis of this field.

One issue I see with this approach is that following field is very localized to `kata` while in future other implementations might show up with their own needs and `runtime-spec` might not be the best place to store values for these fields.

But I might be missing things and I could be wrong :smiley:

liubin:
@giuseppe Indeed Kata only supports running hooks on the host side now. And Kata has a feature named guest hook that running in the VM. Users can add hooks to the guest image. When Kata containers start, the agent inside the guest will scan for hooks and add them to the OCI spec of the containers. Here is one issue https://github.com/kata-containers/kata-containers/issues/3729.

Adding the `type` attribute can make it easy to decide whether run hooks on the host or guest.

@flouthoc I understand what you are meaning. I have thought of some `workaround`, for example passing the host/guest flag by `ENV` field, or setting flags in annotations.

How about adding a new field `annotations` to `Hook`? It's will be more generic and extendable than fixed-named field names like `type`. Through `annotations`, users can pass more metadata to runtime for handling hooks.

AkihiroSuda:
> Through `annotations`, users can pass more metadata to runtime for handling hooks.

The state object already has annotations.

https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#state

Is this enough for your usecase?

AkihiroSuda:
ping @liubin

liubin:
@AkihiroSuda Sorry for the late response, I'm considering closing this one as comments above.

And I'll try to think about other plans including [your](https://github.com/opencontainers/runtime-spec/pull/1140#issuecomment-1199487076)  with real-world use cases later.

PR #1141, opened 2022-03-10
schema: update readme for how to validate manually

liubin:
The latest go is using gomod, but the repo has no go.mod,
adding GO111MODULE=off to disable gomod and build it locally.

Signed-off-by: bin liu <liubin0329@gmail.com>


schema/README.md
```diff
@@ -24,8 +24,8 @@ To build it:
 
 ```bash
 export GOPATH=`mktemp -d`
```
tianon, line 26: Maybe we should add it here instead?

```suggestion
export GOPATH=`mktemp -d` GO111MODULE=off
```

tianon (approved):
Oof, good catch -- this repository's Go code probably needs a heavier revamp so it works properly on newer Go versions, since this variable is supposed to only last another release or so. :grimacing: :see_no_evil:

schema/README.md
liubin: @tianon I have updated as your suggestion.

BTW Makefile has the same issue but GitHub workflow calls `go mod init` before calling `make validate`.

tianon (approved):
Long term we probably need more here, but this seems like a positive change for now 👍

Thanks!

vbatts:
oh thanks, @liubin. I think this PR is obsoleted by #1084 
Closing. Let me know if it is still an issue for you.

PR #1143, opened 2022-03-23
IDMapping field for mount point

AlexeyPerevalov:
IDMap for mount point allows to apply file system feature for changing FS entity owner, once and with minimal performance impact.
In runc it could be used for shared volumes (but kernel still doesn't support overlayfs, ext[2,3,4] fs is supported in kernel > 5.17)

This interface implies the ID mapping per mount point.
Technically mount_setattr syscall requires fd of the user namespace to write mount ID map into uid_map in the procfs. Practically it could be arbitrary user ns, e.g. user namespace which was create by runc for container's process, or temporary.
Approach with temporary user namespace allows to create ID mapping per mount point, which is more flexible.


AkihiroSuda:
@giuseppe @cyphar PTAL 🙏

specs-go/config.go
```diff
@@ -117,6 +117,10 @@ type Mount struct {
 	Source string `json:"source,omitempty"`
 	// Options are fstab style mount options.
 	Options []string `json:"options,omitempty"`
+
+	// UID/GID mappings used for changing file owners w/o calling chown, fs should support it
+	// every mount point could have its own mapping
```
kolyshkin, line 122: It seems the comment is missing some punctuation.

specs-go/config.go
```diff
@@ -117,6 +117,10 @@ type Mount struct {
...
+	IDMappings []LinuxIDMapping `json:"id_mappings,omitempty"`
```
kolyshkin, line 123: One thing I don't understand is why we do not have separate UID and GID mappings, can you please explain?

giuseppe:
as @kolyshkin pointed out, I think we need separate mappings for UIDs and GIDs.

specs-go/config.go
AlexeyPerevalov: I think I kept in mind one limited case, I agree in for general purpose, better to have both uid and gid.

giuseppe:
> Should you describe the new fields `config.md`, too?
> 
> This is Linux-specific addition to the `Mount` struct. Should we introduce a platform specific sub-struct `LinuxMountOpts` for clearly separating platform-specific stuff, WDYT?

could we just use something like `platform:"linux"` as we already do for `Type`?

giuseppe:
> Should you describe the new fields `config.md`, too?

as well as the json schema under `schema/`

marquiz:
> could we just use something like `platform:"linux"` as we already do for `Type`?

Mm, I didn't realize that 🙄  Sounds better

AlexeyPerevalov:
> Should you describe the new fields `config.md`, too?
> 
> This is Linux-specific addition to the `Mount` struct. Should we introduce a platform specific sub-struct `LinuxMountOpts` for clearly separating platform-specific stuff, WDYT?

Initially I though to just keep uid/gid as strings into Options.
Regarding LinuxMountOpts, if we'll have mount options for another platform, it will look like WindowsMountOpts etc., like now in Spec
So I incline to LinuxMountOpts...

specs-go/config.go
```diff
@@ -117,6 +117,11 @@ type Mount struct {
 	Source string `json:"source,omitempty"`
 	// Options are fstab style mount options.
 	Options []string `json:"options,omitempty"`
+
+	// UID/GID mappings used for changing file owners w/o calling chown, fs should support it.
+	// Every mount point could have its own mapping
```
kolyshkin, line 122: nit: missing `.` at EOL.

specs-go/config.go
```diff
@@ -117,6 +117,11 @@ type Mount struct {
...
+	UIDMappings []LinuxIDMapping `json:"uid_mappings,omitempty"`
+	GIDMappings []LinuxIDMapping `json:"gid_mappings,omitempty"`
```
kolyshkin, line 124: Please add `platform:"linux"` as suggested by @giuseppe (similar to how it is done for `Type` above).

kolyshkin:
This also needs an addition to `config.md`, describing the new fields.

specs-go/config.go
```diff
@@ -117,6 +117,11 @@ type Mount struct {
 	Source string `json:"source,omitempty"`
 	// Options are fstab style mount options.
 	Options []string `json:"options,omitempty"`
```
rata, line 119: The idmap option is one of the fstab options and that is why we don't need to mention it in any other part of the spec?

rata (commented):
@AlexeyPerevalov 
> Approach with temporary user namespace allows to create ID mapping per mount point, which is more flexible.

Can I ask what is the use case for that?

 I am not sure we want to do that. If we add one mapping per mount to the spec, then we either:
 * Runtimes will implement the logic to use a tmp userns for all mounts. Or maybe something more complex might be needed (I honestly don't know if having one userns per mount on each container can hit some limits), like see if we can reuse the userns the OCI container might create, if not create a new one. If other existing mounts need the same mapping, share the userns used...
 * Or have runtimes not implement it as it is more complex and no use case really needed it so far, in which case it is just silly to have it on the spec as some runtime might implement it and some others don't, which defeats the purpose of the spec in a way.

I honestly can't see any use case where using different mappings per mount is useful. I think in most cases we will want to use the same userns as the container. That even supports to migrate to userns without chowning the volumes, I guess almost everyone using userns will want that.

Also, the upside is that we can in the future add mappings per mount if needed.

What is the use case for having different mappings per mount? Also, aren't there limits to create potentially so many userns? I guess not, but unsure.

Maybe I'm looking at this from a very kubernetes centric POV and missing other important use cases, sorry if that is the case, just let me know :)

I'm all in for adding idmap to the runtime-spec, though :)

specs-go/config.go
giuseppe: I guess it is explicit when a mapping is specified?

rata:
Thanks! That helps, but it doesn't say really the use case, it just says they want to not map root for the volume. I'm unsure, for example, if they wouldn't benefit of not having root at all mapped (nor for the container nor for the volume), because it doesn't really say why.

But well, if crun is doing it maybe it is worth supporting it in OCI. What do others think?

AlexeyPerevalov:
> @AlexeyPerevalov
> 
> > Approach with temporary user namespace allows to create ID mapping per mount point, which is more flexible.
> 
> Can I ask what is the use case for that?

We might have e.g. log grabber which in container1 under user1 and it works with log producers in container2 and container3 working under user2 and user3 appropriately.
In most cases application could be redesigned to solve this issue.

> 
> I am not sure we want to do that. If we add one mapping per mount to the spec, then we either:
> 
>     * Runtimes will implement the logic to use a tmp userns for all mounts, which is more complex. Or maybe something more complex might be needed (I honestly don't know if having one userns per mount on each container can hit some limits), like see if we can reuse the userns the OCI container might create, if not create a new one. If other existing mounts need the same mapping, share the userns used...



> 
>     * Or have runtimes not implement it as it is more complex and no use case really needed it so far, in which case it is just silly to have it on the spec as some runtime might implement it and some others don't, which defeats the purpose of the spec in a way.
> 
> 
> I honestly can't see any use case where using different mappings per mount is useful for OCI. I think in most cases we will want to use the same userns as the container. That even supports to migrate to userns without chowning the volumes, I guess almost everyone using userns will want that.


Usage of the same userns for idmap as in container requires to have|request that user ns by OCI, but  idmap is attribute of mount, but not user namespace, yes, notwithstanding in linux kernel it was implemented by [gid|uid]_map of specific user namespace. But the initial purpose of [gid|uid]_map in /proc/$pid/[gid|uid]_map was to map uid of the processes.

> 
> Also, the upside of not doing it now, is that we can in the future add mappings per mount if needed. But we can't remove it later if we add it now.
> 
> What is the use case for having different mappings per mount? Also, aren't there limits to create potentially so many userns? I guess not, but unsure.
> 
> Maybe I'm looking at this from a very kubernetes centric POV and missing other important use cases, sorry if that is the case, just let me know :)
> 
> I'm all in for adding idmap to the runtime-spec, though :)

As a summary I chose per mount option because of:
1. Do not mix process (of child and parent namespace) and mount uid mapping. 
2. Avoid dependency on user ns request in OCI, since shared volume with arbitrary uid/gid in fs could be used w/o user namespace.
3. When I checked it, with Brauner's patches, I didn't manage to create 2 different mapping with one persistent user namespace. It was possible only by temporary user namespace, as guided in Brauner's sample. So if in future somebody will want this feature - the persistent container's username space is not suitable for multiple mappings.

giuseppe:
> As I said, I will not oppose to adding idmap to OCI, I do want to move idmap to OCI. What I'm not sure I see is the use case for this extra complexity. But I'm fine with this if other's think this is needed, of course :)

IMO the OCI PoV should be as generic as possible to expose the kernel features, even if a use case is not clear yet.  Upper layers will then decide if/how to use these knobs.

rata:
@giuseppe @AlexeyPerevalov Makes sense, I buy this now. Sorry for the noise :)

specs-go/config.go
rata: Oh, so we use idmap if a mapping is specified? Makes sense. Thanks!

AlexeyPerevalov:
> > > @AlexeyPerevalov
> > > Can I ask what is the use case for that?
> > 
> > 
> > We might have e.g. log grabber which in container1 under user1 and it works with log producers in container2 and container3 working under user2 and user3 appropriately. In most cases application could be redesigned to solve this issue.
> 
> Ok, I guess you are talking about a k8s pod, right? That use case wouldn't fit more naturally into using securityContext.supplementalGroups (it takes an array), to add the GIDs that are needed to the log grabber? Or just using fsGroup? Or just having logs o+r to containers in the pod, that is another option too.

I'm not sure the entire range of problems with DAC could be solved by groups. The initial intent of id mapping feature was to avoid any preparation steps with image, and take image as is, use it with any user with minimum operation to start container.

giuseppe:
@AlexeyPerevalov are you fine with the requested changes?

AlexeyPerevalov:
> @AlexeyPerevalov are you fine with the requested changes?

Yes, I'll update PR soon.

marquiz (commented):
Looks good. One note though: shouldn't you describe the new fields under [Mounts in config.md](https://github.com/opencontainers/runtime-spec/blob/main/config.md#mounts)?

schema/defs.json
```diff
@@ -141,6 +147,12 @@
                 },
                 "type": {
                     "type": "string"
+                },
+                "uidmappings": {
+                    "$ref": "#/definitions/ArrayOfIDMappings"
+                },
+                "gidmappings": {
+                    "$ref": "#/definitions/ArrayOfIDMappings"
```
giuseppe, line 155: could we just use here:

```json
                "uidMappings": {
                    "type": "array",
                    "items": {
                        "$ref": "#/definitions/IDMapping"
                    }
                },
                "gidMappings": {
                    "type": "array",
                    "items": {
                        "$ref": "#/definitions/IDMapping"
                    }
                }
```

so we do not need to define the new ArrayOfIDMappings type?

Alternatively let's change the names to `uidMappings` and `gidMappings`, and also update `config-linux.json` to use the new type as well

config.md
```diff
@@ -101,6 +101,11 @@ For POSIX platforms the `mounts` structure has the following fields:
 * **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
     * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). For bind mounts (when `options` include either `bind` or `rbind`), the type is a dummy, often "none" (not listed in */proc/filesystems*).
     * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+* **`uidmappings` (array of type LinuxIDMapping, OPTIONAL). The file system mapping to convert uid and gid from value written in the container's image into desired value. This mapping contains of following fields: ContainerID, HostID, Size. These values would be written into temporary user namespace's /proc/$pid/uid_map and /proc/$pid/gid_map as followes "ContainerID HostID Size". The uid and gid of the image file system would be mapped inside kernel from ContainerdID to HostID. The Size field is a range for mapping, as an example.
```
giuseppe, line 104:
```suggestion
* **`uidMappings`**, **`gidMappings`**  (array of type LinuxIDMapping, OPTIONAL). The file system mapping used to convert UIDs and GIDs from the source file system. This mapping contains the following fields: ContainerID, HostID, Size. These values would be written into temporary user namespace's /proc/$pid/uid_map and /proc/$pid/gid_map as follows "ContainerID HostID Size". The UID and GID of the source file system are mapped inside the kernel from ContainerID to HostID. The Size field is a range for mapping, as an example.
```

schema/defs.json
```diff
@@ -141,6 +141,18 @@
                 },
                 "type": {
                     "type": "string"
+                },
+                "uidmappings": {
```
giuseppe, line 145:
```suggestion
                "uidMappings": {
```

schema/defs.json
```diff
@@ -141,6 +141,18 @@
...
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/IDMapping"
+                    }
+                },
+                "gidmappings": {
```
giuseppe, line 151:
```suggestion
                "gidMappings": {
```

config.md
rata: Would it be written to /proc/$pid/uid_map? I don't know what pid refers to here, but if it is the container pid and the container userns uses a different mapping, I guess that is not true?

Can't we just avoid mentioning those implementation details?

rata, line 104: Missing gidMappings?

rata, line 104: Please follow the convention of one sentence per line

giuseppe: probably that is better, let's avoid the implementation details

AlexeyPerevalov: yes, it was questionable. For gidMapping I'll just add short reference to uidMapping.

config.md
```diff
@@ -101,6 +101,17 @@ For POSIX platforms the `mounts` structure has the following fields:
...
+This mapping contains of following fields: ContainerID, HostID, Size.
+These values would be written into temporary user namespace's /proc/$pid/uid_map as followes "ContainerID HostID Size".
+The uid of the image file system would be mapped inside kernel from ContainerdID to HostID.
+The Size field is a range for mapping, as an example.
+        1000 1001 1       -> map uid 1000 to uid 1001 in <target> and no other ids
+        20000 100000 1000 -> map uid 20000 to uid 100000, uid 20001 to uid 100001 [...] until uid 20999 in <target>
+Currently up to 340 separate idmappings may be specified by the kernel.
+* **`gidmappings` (array of type LinuxIDMapping, OPTIONAL) The file system mapping to convert gid from value written in the container's image into desired value.
+For more details see `uidmappings`.
+
```
giuseppe, line 114: please use **uidMappings** and **gidMappings** (with uppercase M and the missing closing **).

We shouldn't mention "container's image" since the mapping is performed from the source file system (and runtime specs don't know about images).

I'd use something simpler as the following:

```suggestion
* **`uidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert UIDs from the source file system to the destination mount point.
This mapping contains the following fields: ContainerID, HostID, Size.
The Size field is a range for mapping, as an example.
        1000 1001 1       -> map UID 1000 to uid 1001 in <target> and no other UIDs
        20000 100000 1000 -> map UID 20000 to uid 100000, uid 20001 to UID 100001 [...] until UID 20999 in <target>
* **`gidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert GIDs from the source file system to the destination mount point.
For more details see `uidMappings`.

```

schema/defs.json
giuseppe: still pending

schema/defs.json
giuseppe: still pending

specs-go/config.go
```diff
@@ -117,6 +117,11 @@ type Mount struct {
...
+
+	// UID/GID mappings used for changing file owners w/o calling chown, fs should support it.
+	// Every mount point could have its own mapping.
+	UIDMappings []LinuxIDMapping `json:"uid_mappings,omitempty" platform:"linux"`
+	GIDMappings []LinuxIDMapping `json:"gid_mappings,omitempty" platform:"linux"`
```
giuseppe, line 124: please use the name `uidMappings` and `gidMappings` here:

```suggestion
	UIDMappings []LinuxIDMapping `json:"uidMappings,omitempty" platform:"linux"`
	GIDMappings []LinuxIDMapping `json:"gidMappings,omitempty" platform:"linux"`
```

giuseppe (changes requested):

giuseppe (approved):
LGTM

marquiz (approved):
lgtm (non-binding 😊)

kolyshkin:
Looks like the commits need to be squashed.

giuseppe:
@vbatts @crosbymichael @cyphar @dqminh @hqhq @mrunalp @tianon PTAL

kolyshkin (approved):
LGTM, thanks!

brauner (approved):

rata (approved):

specs-go/config.go
```diff
@@ -117,6 +117,11 @@ type Mount struct {
 	Source string `json:"source,omitempty"`
 	// Options are fstab style mount options.
 	Options []string `json:"options,omitempty"`
+
+	// UID/GID mappings used for changing file owners w/o calling chown, fs should support it.
+	// Every mount point could have its own mapping.
+	UIDMappings []LinuxIDMapping `json:"uidMappings,omitempty" platform:"linux"`
+	GIDMappings []LinuxIDMapping `json:"gidMappings,omitempty" platform:"linux"`
```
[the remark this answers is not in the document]
giuseppe: `MOUNT_ATTR_IDMAP` is used to set the user namespace on the mount, and the way this feature is going to be implemented.

The point of having these attributes here is that you can specify a different mapping per mount.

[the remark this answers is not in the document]
jiangliu: Thanks @giuseppe , I understood the design now after reading the PR https://github.com/containerd/containerd/pull/5890/files

jiangliu (approved):

kolyshkin:
@tianon PTAL 🙏🏻

config.md
```diff
@@ -101,6 +101,14 @@ For POSIX platforms the `mounts` structure has the following fields:
 * **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
     * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). For bind mounts (when `options` include either `bind` or `rbind`), the type is a dummy, often "none" (not listed in */proc/filesystems*).
     * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+* **`uidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert UIDs from the source file system to the destination mount point.
+This mapping contains the following fields: ContainerID, HostID, Size.
+The Size field is a range for mapping, as an example.
+        1000 1001 1       -> map UID 1000 to uid 1001 in <target> and no other UIDs
+        20000 100000 1000 -> map UID 20000 to uid 100000, uid 20001 to UID 100001 [...] until UID 20999 in <target>
```
tianon, line 108: The formatting here renders a little bit odd (all squashed together into a single paragraph). :grimacing:

tianon: Since this `LinuxIDMapping` isn't a new type, is there somewhere else this description should perhaps go?

AlexeyPerevalov: I fixed formatting, just adding new line.
Do you suggest to add reference to some man page or commit message of kernel's patch?

kolyshkin: > Since this LinuxIDMapping isn't a new type, is there somewhere else this description should perhaps go?

@AlexeyPerevalov I guess that @tianon means that the format description should also be added to the existing documentation that mentions/uses LinuxIDMapping. Specifically, this is https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#user-namespace-mappings

Now I see that the way you describe it is the way it is expected by the kernel (three space-separated values). In here, we use three named fields, so space-separated format makes less sense.

I guess you can just ditch the example, instead referring to https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#user-namespace-mappings

tianon (approved):

rata:
This has 2 LGTM now, is there anything else we want before merging? :)

vbatts (approved):

vbatts:
bah, why is pullaprove still here. I thought we removed it.

brauner (approved):

rata:
Pullapprove seems to be gone now (replaced by build-pr/run maybe?), but all checks are green. Anything else missing to merge? :)

tianon:
Last I checked pullapprove was still blocking the merge, but can confirm it's indeed now gone! :+1:

config.md
```diff
@@ -101,6 +101,11 @@ For POSIX platforms the `mounts` structure has the following fields:
 * **`type`** (string, OPTIONAL) The type of the filesystem to be mounted.
     * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). For bind mounts (when `options` include either `bind` or `rbind`), the type is a dummy, often "none" (not listed in */proc/filesystems*).
     * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
+* **`uidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert UIDs from the source file system to the destination mount point.\
+The format is the same as [user namespace mappings](config-linux.md#user-namespace-mappings).
+* **`gidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert GIDs from the source file system to the destination mount point.
+For more details see `uidMappings`.
+
```
AkihiroSuda, line 108: Is there any specific reason that we don't have this for rootfs?

rata: Also, the spec currently says the runtime should not modify the rootfs permissions: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#user-namespace-mappings

> The runtime SHOULD NOT modify the ownership of referenced filesystems to realize the mapping.

PR #1144, opened 2022-04-13
Update Windows CPU comments

jterry75:
Updated the documentation about Windows CPU for containers and the expectations that each of the properties have. This is taken directly from the Microsoft wording at: https://docs.microsoft.com/en-us/virtualization/api/hcs/schemareference#Container_Processor and is also referenced in the markdown for clarity.

Signed-off-by: Justin Terry <jlterry@amazon.com>

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -77,11 +77,13 @@ The following parameters can be specified:
 
 `cpu` is an OPTIONAL configuration for the container's CPU usage.
 
-The following parameters can be specified:
+The following parameters can be specified (mutually exclusive):
 
-* **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container.
-* **`shares`** *(uint16, OPTIONAL)* - specifies the relative weight to other containers with CPU shares.
-* **`maximum`** *(uint16, OPTIONAL)* - specifies the portion of processor cycles that this container can use as a percentage times 100.
+* **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor `count` in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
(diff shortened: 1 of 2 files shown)


jterry75:
@kevpar - FYI

giuseppe (approved):
LGTM

config-windows.md
```diff
@@ -77,11 +77,13 @@ The following parameters can be specified:
 
 `cpu` is an OPTIONAL configuration for the container's CPU usage.
 
-The following parameters can be specified:
+The following parameters can be specified (mutually exclusive):
+
+* **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor count in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
+* **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor weight is a value between 0 and 10,000.
```
dcantah, line 83: nit: The OCI naming is shares but the last sentence is worded as if Weight, like on the HCS schema, is the field name here also. Maybe note a "also referred to as processor weight" or change the end to "Shares is a value between 0 and 10,000" or something similar.

dcantah (approved):
LGTM, one minor comment. Thanks Justin!

config-windows.md
dcantah: Not sure why this selected 3 extra lines.. This was directed at line 83 (shares) only

jterry75: Lol I get what you're saying. Will do an update. I've always found it confusing for OCI we used Shares even though the platform does not have it. Will update

jterry75: Ok done.

dcantah: Thanks! Resolving

dcantah: Oh never mind I can't, feel free to resolve 😆

TomSweeneyRedHat:
LGTM

vbatts (approved):
:window:

PR #1146, opened 2022-04-24
config-linux: update type of LinuxCPU.Idle  to *int64

wineway:
the config values should contain a value that means "do not modify it" within its range of values. for `int64` in other cases, we usually use `0` to express this meaning, but in `cpu.idle`, `0` can not be regarded as this meaning.

by changing the type from `int64`to `*int64`, we can use `nil` solve this problem

### DIFF
--- a/config.md
+++ b/config.md
@@ -840,6 +840,7 @@ Here is a full example `config.json` for reference.
                 "realtimeRuntime": 950000,
                 "realtimePeriod": 1000000,
                 "cpus": "2-3",
+                "idle": 1,
                 "mems": "0-7"
             },
             "devices": [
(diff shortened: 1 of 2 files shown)


wineway:
😂I found this when I try to implement it, if we keep this type, we can only choose a value like `-1` to mark it does not need to modify, but `-1` might be used by the kernel in the future, and also ugly

giuseppe (approved):
LGTM

tianon (approved):
LGTM (and matches the description in https://github.com/opencontainers/runtime-spec/blob/2cf6663ca299f8454318ea7c797a8039d013fe8c/config-linux.md#cpu, which IMO doesn't need any updates beyond the `OPTIONAL` it already has)

Issue #1147, opened 2022-05-03
Ensure that devices are under /dev


c3d:
In the [Linux device specification](https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#devices), all the examples indicate devices created under `/dev`. However, it is not specified if this is necessary or mandated.

We recently ran across a test that specifically puts devices under `/mnt`, and that makes me wonder if this is legitimate. This notably causes possible problems with VM-based runtimes such as Kata containers. Without specific precautions, they would be forced to scan volumes in search for device nodes, in order to make necessary adjustments.

I believe that the wording should be improved to explicitly state that devices can only be placed under `/dev` on Linux, and that having device nodes anywhere else has undefined behaviour.

kolyshkin:
In my humble opinion, `/dev` is just a convention, and UNIX/Linux do not impose any specific restriction as to where device nodes can or should be created, and therefore OCI spec should not enforce any restrictions either.

OTOH your scenario might be valid, too, so if you can either elaborate on it (does Kata containers need to explicitly find all device nodes in container filesystem? what for? is it currently only looking under /dev?), or just open a PR to modify/amend the existing wording in the spec, and we can discuss it further right there in the PR.

PR #1148, opened 2022-05-03
config-linux: Clarify where device nodes can be created

c3d:
Clarify that device nodes need not be under `/dev`, but that the runtime need to
be informed of all the device nodes that are used by the
container.

Virtual-machine based runtimes such as Kata Containers need to be able to
perform adjustment on device nodes, and cannot be required to deep-scan
file-systems to do so.

The proposed wording was chosen to avoid any regression for any workload
mounding nodes elsewhere, while at the same time clarifying that correct
behaviour cannot be guaranteed if a device node is created on the host and used
by the container without being passed in the devices list.

This fixes issue #1147.

Signed-off-by: Christophe de Dinechin <christophe@dinechin.org>

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -118,6 +118,7 @@ Each entry has the following structure:
     More info in [mknod(1)][mknod.1].
 * **`path`** *(string, REQUIRED)* - full path to device inside container.
     If a [file][] already exists at `path` that does not match the requested device, the runtime MUST generate an error.
+    The path MAY be anywhere in the container filesystem, notably outside of `/dev`.
 * **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - [major, minor numbers][devices] for the device.
 * **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
     You can also control access to devices [with cgroups](#configLinuxDeviceAllowedlist).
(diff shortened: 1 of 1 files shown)


c3d:
A relevant issue in OpenShift sandboxed containers can be found [here](https://issues.redhat.com/browse/KATA-671)

config-linux.md
```diff
@@ -126,6 +127,12 @@ Each entry has the following structure:
 
 The same `type`, `major` and `minor` SHOULD NOT be used for multiple devices.
 
+Containers MUST NOT access any device node that is not explicitly referenced in
```
giuseppe, line 130: the wording seems too strong here.  Maybe we can say that using a device node not referenced in the **devices** array is runtime-specific and not portable?

c3d: Sure. Any suggestion for the wording? Just changing MUST NOT to MAY NOT sufficient?

c3d: @giuseppe Updated, please check.

giuseppe (approved):
LGTM

config-linux.md
giuseppe: thanks, LGTM

TomSweeneyRedHat (commented):
LGTM

c3d:
@TomSweeneyRedHat If this looks good to you, could you please change the status of your review to match?

c3d:
Ping?

config-linux.md
```diff
@@ -126,6 +127,12 @@ Each entry has the following structure:
 
 The same `type`, `major` and `minor` SHOULD NOT be used for multiple devices.
 
+Containers MAY NOT access any device node that is not explicitly referenced in
+the **`devices`** array. Rationale: runtimes based on virtual machines need to
```
AkihiroSuda, line 131: Conflicts with the configLinuxDefaultDevices section

AkihiroSuda (changes requested):
Seems inconsistent with the configLinuxDefaultDevices section

config-linux.md
c3d: @AkihiroSuda Good point. Reworded to indicate this.

AkihiroSuda (approved):

Issue #1101, opened 2021-04-05
Audit of active maintainers + call for new maintainers


caniszczyk:
In regards to opencontainers/tob#95

If I do a quick audit of all of authors and last commit dates, i.e., git log --pretty=%an%x09%ad | awk -F$'\t' '!seen[$1]++'

Tianon Gravi	Fri Mar 26 12:09:08 2021 -0700
Vincent Batts	Fri Mar 19 07:46:52 2021 -0400
Rodrigo Campos	Thu Mar 18 17:29:02 2021 +0100
Sebastiaan van Stijn	Tue Mar 9 17:25:23 2021 +0100
Mrunal Patel	Mon Mar 8 14:51:45 2021 -0800
Kir Kolyshkin	Mon Mar 8 10:41:59 2021 -0800
Christian Felder	Wed Mar 3 21:26:04 2021 +0100
Giuseppe Scrivano	Thu Jan 21 13:20:57 2021 +0100
Iceber Gu	Sat Feb 13 14:45:23 2021 +0800
Alban Crequy	Fri Oct 23 12:44:54 2020 +0200
Aleksa Sarai	Tue Sep 29 16:35:07 2020 +1000
Sascha Grunert	Tue Aug 25 15:37:48 2020 +0200
Daniel J Walsh	Tue Jul 7 15:08:12 2020 -0400
John Bartholomew	Thu Jul 30 23:58:45 2020 +0100
Michael Crosby	Tue Jul 28 13:02:52 2020 -0400
Jia Zhang	Mon Jul 13 09:26:59 2020 +0800
Kenta Tada	Thu Jul 23 17:18:50 2020 +0900
Renaud Gaubert	Tue Jul 21 07:08:52 2020 +0000
Qiang Huang	Tue Jun 9 21:07:15 2020 +0800
Albin Vass	Sun May 31 12:50:33 2020 +0200
Paulo Gomes	Sun May 17 08:44:37 2020 +0100
H. Vetinari	Fri Mar 6 08:55:22 2020 +0100
Alexander Kanevskiy	Mon Mar 2 16:01:57 2020 +0200
Dongsu Park	Tue Jun 19 15:06:25 2018 +0200
zhouhao	Wed Oct 18 15:22:53 2017 +0800
Joowon Cheong	Tue Feb 25 15:42:09 2020 +0900
Ma Shimiao	Fri Dec 1 11:53:02 2017 +0800
[...]

PR #1150, opened 2022-05-31
maintainer updates as per #1101

h-vetinari:
Add "recently" added runc-maintainers also to runtime spec (emails taken from [runc](https://github.com/opencontainers/runc/blob/main/MAINTAINERS)). These 3 have a proven track record of excellent work in runc, and so I feel comfortable to propose them as one batch.

They have also all agreed in #1101 to join, see [this comment](https://github.com/opencontainers/runtime-spec/issues/1101#issuecomment-916972654) and below.

If someone wants to split the PR or propose their own one, I really really don't mind - I'm just trying to get the ball rolling. 🙃 

CC @AkihiroSuda 
CC @kolyshkin 
CC @thaJeztah 

Needs at least 6 approvals from the current 8 maintainers:
* [ ] crosbymichael 
* [x] cyphar 
* [ ] dqminh 
* [x] giuseppe 
* [x] hqhq 
* [x] mrunalp 
* [x] tianon 
* [ ] vbatts


kolyshkin (approved):
LGTM

AkihiroSuda (approved):
Thanks, LGTM

giuseppe (approved):
LGTM

samuelkarp:
Hey folks!

I wanted to chime in from the TOB here.  It looks like the [governance document](https://github.com/opencontainers/runtime-spec/blob/main/GOVERNANCE.md) for the runtime spec does not currently establish procedures for adding (or removing) maintainers.  Under the current set of rules I can see two paths forward:

1. Amend the governance document per the process outlined there (motion and voting on the dev@opencontainers.org mailing list) to add rules around how maintainers can be added (or removed).
2. Or, directly move to add the new maintainers via a single vote using the same process.  Note that this path doesn't address the general case of establishing rules around how maintainers are added (or removed) and eventually we'll likely need to revisit that again.

The current governance rules require a 2/3 approval for a motion to pass; with 8 current maintainers that means at least 6 must approve.

(Note: the governance document recommends voting on the mailing list with a "SHOULD" but not a "MUST" so I think it'd be reasonable to go forward with 6 approvals here _if_ the current maintainers prefer to do that.)

Thanks!
Sam

hqhq (approved):

h-vetinari:
PTAL @crosbymichael @cyphar @dqminh @mrunalp @tianon @vbatts

mrunalp (approved):

tianon (approved):

cyphar (approved):
LGTM

cyphar:
@samuelkarp I don't know why it's not mentioned in the governance document (the point of copy-pasting the same document was so we had rules for everything already), but adding maintainers with a 2/3rd vote is accepted practice and is covered under the general governance change rules. It would be a good idea to do a review of the governance documents for all the projects to make sure they aren't missing important information like this.

I suspect this information is missing because the rules for adding maintainers in runc were outlined in the `MAINTAINERS_GUIDE.md` rather than `GOVERNANCE.md` but runtime-spec didn't copy the former into this repo.

h-vetinari:
> Gentle ping @crosbymichael @dqminh @vbatts

☝️

vbatts (approved):
LGTM

Sorry I missed this PR. And then came here as I'm trying to determine and refresh myself on the process for adding and removing maintainers :-\

I generally agree with maintaining the 2/3 quorum for topics like adding/removing maintainers. And hope to see (or create) a PR for this language in the specs projects soon.

For the record, with my vote it makes the 2/3 quorum for _this_ vote. And effective upon merge we'll have **11** maintainers for an 7.3 majority? So, A) maybe we'll round that to 7?; B) perhaps some folks step down before being voted off?

vbatts:
congrats and welcome @AkihiroSuda, @kolyshkin, and @thaJeztah! :cupcake:

PR #1151, opened 2022-06-20
Add support for time namespace

KentaTada:
The time namespace is a new kernel feature available in 5.6+ to
isolate the system monotonic and boot-time clocks.

I'm sorry I mistook and closed the existing request https://github.com/opencontainers/runtime-spec/pull/1062

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -34,6 +34,7 @@ The following parameters can be specified to set up namespaces:
     * **`uts`** the container will be able to have its own hostname and domain name.
     * **`user`** the container will be able to remap user and group IDs from the host to local users and groups within the container.
     * **`cgroup`** the container will have an isolated view of the cgroup hierarchy.
+    * **`time`** the container will be able to have its own clocks.
 * **`path`** *(string, OPTIONAL)* - namespace file.
     This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
     The runtime MUST place the container process in the namespace associated with that `path`.
@@ -70,6 +71,9 @@ If a `namespaces` field contains duplicated namespaces with same `type`, the run
     },
     {
         "type": "cgroup"
+    },
+    {
+        "type": "time"
     }
 ]
 ```
@@ -106,6 +110,17 @@ Note that the number of mapping entries MAY be limited by the [kernel][user-name
     }
 ]
 ```
+
+## <a name="configLinuxTimeOffset" />Offset for Time Namespace
+
+**`timeOffsets`** (object, OPTIONAL) sets the offset for Time Namespace. For more information
+see the [time_namespaces](time_namespaces.7).
+
+The name of the clock is the entry key.
+Entry values are objects with the following properties:
+
+* **`secs`** *(int64, OPTIONAL)* - is the offset of clock (in seconds) in the container.
+* **`nanosecs`** *(uint32, OPTIONAL)* - is the offset of clock (in nanoseconds) in the container.
 
 ## <a name="configLinuxDevices" />Devices
 
@@ -770,3 +785,4 @@ subset of the available options.
 [zero.4]: http://man7.org/linux/man-pages/man4/zero.4.html
 [user-namespaces]: http://man7.org/linux/man-pages/man7/user_namespaces.7.html
 [intel-rdt-cat-kernel-interface]: https://www.kernel.org/doc/Documentation/x86/intel_rdt_ui.txt
+[time_namespaces.7]: https://man7.org/linux/man-pages/man7/time_namespaces.7.html
--- a/config.md
+++ b/config.md
@@ -905,6 +905,16 @@ Here is a full example `config.json` for reference.
                 }
             ]
         },
+        "timeOffsets": {
+            "monotonic": {
+                "secs": 172800,
+                "nanosecs": 0
+            },
+            "boottime": {
+                "secs": 604800,
+                "nanosecs": 0
+            }
+        },
         "namespaces": [
             {
                 "type": "pid"
@@ -926,6 +936,9 @@ Here is a full example `config.json` for reference.
             },
             {
                 "type": "cgroup"
+            },
+            {
+                "type": "time"
             }
         ],
         "maskedPaths": [
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -253,6 +253,12 @@
             "personality": {
                 "type": "object",
                 "$ref": "defs-linux.json#/definitions/Personality"
+            },
+            "timeOffsets": {
+                "type": "object",
+                "additionalProperties": {
+                    "$ref": "defs-linux.json#/definitions/TimeOffsets"
+                }
             }
         }
     }
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -182,6 +182,8 @@ type Linux struct {
 	IntelRdt *LinuxIntelRdt `json:"intelRdt,omitempty"`
 	// Personality contains configuration for the Linux personality syscall
 	Personality *LinuxPersonality `json:"personality,omitempty"`
+	// TimeOffsets specifies the offset for supporting time namespaces.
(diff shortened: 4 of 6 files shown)


KentaTada:
`monotonicNanosecs` and `boottimeNanosecs` do not use the negative number.
And I noticed that  the value of offset-nanosecs is not greater than 999,999,999.
So I changed those types from int64 to uint32 in this new PR.

specs-go/config.go
```diff
@@ -182,6 +182,8 @@ type Linux struct {
 	IntelRdt *LinuxIntelRdt `json:"intelRdt,omitempty"`
 	// Personality contains configuration for the Linux personality syscall
 	Personality *LinuxPersonality `json:"personality,omitempty"`
+	// LinuxTimeOffset specifies the offset for supporting time namespaces.
```
AkihiroSuda, line 185:
```suggestion
	// TimeOffset specifies the offset for supporting time namespaces.
```

AkihiroSuda:
@KentaTada could you update the PR?

config.md
```diff
@@ -905,6 +905,12 @@ Here is a full example `config.json` for reference.
                 }
             ]
         },
+        "timeOffset": {
+            "monotonicSecs": 172800,
+            "monotonicNanosecs": 0,
+            "boottimeSecs": 604800,
+            "boottimeNanosecs": 0
+        },
```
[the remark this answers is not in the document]
KentaTada: >(In Go, I'd personally represent this as a map[string]LinuxTimeOffset with a then-generic LinuxTimeOffset type to represent the kernel's offset-secs + offset-nanosecs.)

I agrree with that. But if some maintainers have opinons, please update it.

specs-go/config.go
KentaTada: Fixed although I used TimeOffsets

KentaTada:
@AkihiroSuda @tianon 
I updated this pr from your comments.  
If I remember correctly, `Secs` was required and `NanoSecs` was optional because of the kernel implementation.  
But we changed the definition of LinuxTimeOffset from specific to generic. So, I changed `Secs` to optional.  
I may be slow to respond, but I'll check reviews. Sorry.

config-linux.md
```diff
@@ -107,6 +111,17 @@ Note that the number of mapping entries MAY be limited by the [kernel][user-name
...
+## <a name="configLinuxTimeOffset" />Offset for Time Namespace
+
+**`timeOffsets`** (object, OPTIONAL) sets the offset for Time Namespace. For more information
+see the [time_namespaces](time_namespaces.7).
+
+The name of the clock is the entry key.
+Entry values are objects with the following properties:
+
+* **`secs`** *(int64, OPTIONAL)* - is the offset of clock (in secs) in the container.
+* **`nanosecs`** *(uint32, OPTIONAL)* - is the additional offset for secs (in nanosecs). The actual offset is secs plus nanosecs.
```
AkihiroSuda, line 123: > The actual offset is secs plus nanosecs.

Just to confirm, is this true for negative secs too?

Also curious why nanosecs is unsigned

KentaTada: updated

AkihiroSuda (approved):

config-linux.md
```diff
@@ -107,6 +111,17 @@ Note that the number of mapping entries MAY be limited by the [kernel][user-name
...
+## <a name="configLinuxTimeOffset" />Offset for Time Namespace
+
+**`timeOffsets`** (object, OPTIONAL) sets the offset for Time Namespace. For more information
+see the [time_namespaces](time_namespaces.7).
+
+The name of the clock is the entry key.
+Entry values are objects with the following properties:
+
+* **`secs`** *(int64, OPTIONAL)* - is the offset of clock (in seconds) in the container.
+* **`nanosecs`** *(int64, OPTIONAL)* - is the offset of clock (in nanoseconds) in the container.
```
marquiz, line 123: I guess this must be smaller than 10^9(?) Do we want to mention the limits and that this is for fratcions of the seconds?

config-linux.md
AkihiroSuda: Sorry for going back and forth, but nanosecs can remain unsigned, as it is stated in the kernel man page
https://man7.org/linux/man-pages/man7/time_namespaces.7.html

> These offsets are expressed
       relative to the clock values in the initial time namespace.  The
       offset-secs value can be negative, subject to restrictions noted
       below; offset-nanosecs is an unsigned value.

config-linux.md
AkihiroSuda: It's already explained in the kernel doc: https://github.com/opencontainers/runtime-spec/pull/1151

>   EINVAL An offset-nanosecs value is greater than 999,999,999.

config-linux.md
KentaTada: No problem. I re-updated to remain unsigned.

AkihiroSuda (approved):

thaJeztah:
(Silly) questions;

- Do Windows containers allow for time to be namespaced? (possibly not for this proposal, but in case it's added to the Windows counterpart in future)
- As the discussion above appears to be focussing on signedness and size of integers (_because_ we're trying to do a 1:1 mapping of the Linux kernel to the Spec), I'm wondering if we _MUST_ follow the linux Kernel here, or if it would be beneficial to pick our own definition (does [RFC3399](https://www.rfc-editor.org/rfc/rfc3339) officially support nanoseconds and a "duration", or is there an RFC for that?).

Perhaps it's not an option, and I know many of the options were a 1:1 mapping of kernel features (out of convenience?) I also know that _some_ of those choices made it less agnostic/portable than desired, which is why I was considering if another format would be beneficial / future proof (as long as it's not ambiguous, and documented how runtimes must interpret the config and convert it to (currently) what Linux provides).

AkihiroSuda:
> windows

I think windows is out of the scope of this PR, and windows support can be added in the `config-windows.md` spec separately in the future when somebody needs it.

thaJeztah:
> I think windows is out of the scope of this PR, and windows support can be added in the config-windows.md spec separately in the future when somebody needs it.

For sure! Mostly exploring _if_ there's an platform-agnostic way to specify this configuration (and if there is, and it wouldn't be "crippling" either side, the same format could be used for Windows)

tianon (approved):

hqhq (approved):

schema/config-linux.json
```diff
@@ -253,6 +253,12 @@
             "personality": {
                 "type": "object",
                 "$ref": "defs-linux.json#/definitions/Personality"
+            },
+            "timeOffsets": {
+                "type": "object",
+                "additionalProperties": {
+                    "$ref": "defs-linux.json#/definitions/TimeOffsets"
+                }
```
giuseppe, line 261: this doesn't seem correct.

I've opened a PR: https://github.com/opencontainers/runtime-spec/pull/1193

AkihiroSuda:
@KentaTada Do you plan to open a PR for runc (or other runtimes)?

KentaTada:
Hello @AkihiroSuda san

Recently, I don't have enough resources to focus on the container runtime.
But @chethanah and team members are interested in implementing time namespace in runc.
If it is ok to send the PR to runc instead of me, I'll ask him to take over my activity.

KentaTada:
@chethanah san

Especially, I have no idea about the special-handling case for time namespace.
For example, when it comes to IPC namespace, it affects the changeable kernel parameters(fs.mqueue.*)
https://github.com/opencontainers/runc/blob/v1.1.7/libcontainer/configs/validate/validator.go#L184
So we need to investigate the properties related with time namespace at first.
Could you create the new issue in runc? If it is difficult or takes a lot of time, please feel free to ask me in the issue's page.

AkihiroSuda:
> issue in runc

- https://github.com/opencontainers/runc/issues/2345

utam0k:
👋 Hi @KentaTada @chanezon FYI: crun's implementation https://github.com/containers/crun/pull/1184

PR #1155, opened 2022-08-04
config-linux: add sysDevBlock

grooverdan:
This is to expose the /sys/dev/block hierarchy to the container for devices of mounted filesystems and block devices of the containers.

In [rhbz#1772993](https://bugzilla.redhat.com/show_bug.cgi?id=1772993) the OCP decided to block the system storage information from unprivileged containers. As a move to re-expose this information back to the unprivileged containers, this add the option back, leaving the default behaviour to what the container runtime previously did.

The population and use of this option has a implementation under review https://github.com/containers/podman/pull/13708.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -894,6 +894,20 @@ subset of the available options.
 * **`flags`** *(array of strings, OPTIONAL)* - the additional flags to apply.
     Currently no flag values are supported.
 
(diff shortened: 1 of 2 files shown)


giuseppe:
The implementation seems to be in Podman.  Why do we need this change in the runtime specs?

grooverdan:
quite right. I got lost.

PR #1156, opened 2022-08-07
add domainname spec entity

cdoern:
add the domainname entity so that container runtimes can add special handling similar to hostname. The current workaround of adding a sysctl for kernel.domainname only works with rootful execution in most cases. This will allow for rootless execution.

container runtimes will be able to add special handling as they do for hostname, using setdomainname to add the entry to /proc/sys/kernel/domainname.

Signed-off-by: Charlie Doern <cdoern@redhat.com>

### DIFF
--- a/config.md
+++ b/config.md
@@ -355,6 +355,18 @@ For Windows based systems the user structure has the following fields:
 "hostname": "mrsdalloway"
 ```
 
+## <a name="configDomainname" />Domainname
+
+* **`domainname`** (string, OPTIONAL) specifies the container's domainname as seen by processes running inside the container.
+    On Linux, for example, this will change the domainname in the [container](glossary.md#container-namespace) [UTS namespace][uts-namespace.7].
+    Depending on your [namespace configuration](config-linux.md#namespaces), the container UTS namespace may be the [runtime](glossary.md#runtime-namespace) [UTS namespace][uts-namespace.7].
+
+### Example
+
+```json
+"domainname": "foobarbaz.test"
+```
+
 ## <a name="configPlatformSpecificConfiguration" />Platform-specific configuration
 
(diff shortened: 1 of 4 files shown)


cdoern:
@giuseppe PTAL

flouthoc (changes requested):
I think this should also update the config schema
```diff
diff --git a/schema/config-schema.json b/schema/config-schema.json
index a4d1274..cf66c65 100644
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -35,6 +35,9 @@
         "hostname": {
             "type": "string"
         },
+        "domainname": {
+            "type": "string"
+        },
         "mounts": {
             "type": "array",
             "items": {
```

flouthoc:
Nit:
>This will allow for rootless execution

I think in commit message following must be extended to specify how a `container-runtime` should utilize this field, two approach i could  think are runtime invokes `setdomainname(` or the second one is write `/proc/sys/kernel/domainname` before setting up UTS namespace but is it possible in rootless session ? (Requesting maintainers to confirm)

giuseppe:
as @flouthoc pointed out, please update the config schema as well

giuseppe (approved):
LGTM

AkihiroSuda (approved):

AkihiroSuda:
What should happen if the sysctl is also specified but with a different value? Should it fail?

flouthoc:
@AkihiroSuda Afaik In crun i think sysctl is not allowed at all if spec already has a knob to configure it for instance i can see for `kernel/hostname` : https://github.com/containers/crun/blob/main/src/libcrun/linux.c#L3198

AkihiroSuda:
@opencontainers/runtime-spec-maintainers Can we merge this?

vbatts (approved):

PR #1158, opened 2022-08-29
config-linux: add memory.checkBeforeUpdate

kolyshkin:
This setting can be used to mimic cgroup v1 behavior on cgroup v2,
when setting the new memory limit during update operation.

In cgroup v1, a limit which is lower than the current usage is rejected.

In cgroup v2, such a low limit is causing an OOM kill.

Ref: https://github.com/opencontainers/runc/issues/3509

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -336,6 +336,11 @@ The following properties do not specify memory limits, but are covered by the `m
     To disable it, specify a value of `true`.
 * **`useHierarchy`** *(bool, OPTIONAL)* - enables or disables hierarchical memory accounting.
     If enabled (`true`), child cgroups will share the memory limits of this cgroup.
+* **`checkBeforeUpdate`** *(bool, OPTIONAL)* - enables container memory usage check before setting a new limit.
+    If enabled (`true`), runtime MAY check if a new memory limit is lower than the current usage, and MUST
+    reject the new limit. Practically, when cgroup v1 is used, the kernel rejects the limit lower than the
+    current usage, and when cgroup v2 is used, an OOM killer is invoked. This setting can be used on
+    cgroup v2 to mimic the cgroup v1 behavior.
 
 #### Example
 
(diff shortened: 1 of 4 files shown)


config-linux.md
```diff
@@ -336,6 +336,11 @@ The following properties do not specify memory limits, but are covered by the `m
     To disable it, specify a value of `true`.
 * **`useHierarchy`** *(bool, OPTIONAL)* - enables or disables hierarchical memory accounting.
     If enabled (`true`), child cgroups will share the memory limits of this cgroup.
+* **`checkBeforeUpdate`** *(bool, OPTIONAL)* - enables container memory usage check before setting a new limit.
```
giuseppe, line 339: could we make it clearer that it is about the memory limit?  Perhaps `checkMemoryUsageBeforeUpdate`?

kolyshkin: Well, this is already under `.linux.memory.checkBeforeUpdate` so adding a second "memory" would be a tautology.

giuseppe (approved):
LGTM

config-linux.md
```diff
@@ -336,6 +336,11 @@ The following properties do not specify memory limits, but are covered by the `m
...
+    If enabled (`true`), runtime MAY check if a new memory limit is lower than the current usage, and MUST
+    reject the new limit. Practically, when cgroup v1 is used, the kernel rejects the limit lower than the
+    current usage, and when cgroup v2 is used, an OOM killer is invoked. This setting can be used on
+    cgroup v2 to mimic the cgroup v1 behavior.
```
mrunalp, line 343: Should we call out that there is TOCTOU race possible here and it won't be able to exactly match cgroups v1 behavior?

utam0k (approved):
Thanks @kolyshkin

AkihiroSuda:
Do you have a POC implementation of this?

AkihiroSuda (approved):

config-linux.md
kolyshkin: To me this is an implementation detail, and this is merely a spec. In fact it should not even explain why we need it, but I guess some context would not hurt.

PR #1159, opened 2022-09-02
config-linux.md: formalize the order of `seccomp.syscalls`

AkihiroSuda:
When the syscall matches multiple entries, only the first entry is effective.

Corresponds to the behavior of existing implementations such as runc.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -759,6 +759,11 @@ The following parameters can be specified to set up seccomp:
             * `SCMP_CMP_GT`
             * `SCMP_CMP_MASKED_EQ`
 
+    Syscalls in this list are not guaranteed to be unique, and MAY appear multiple
+    times. If a syscall appears multiple times, runtimes MUST use the first match,
+    and MUST ignore further occurrences. Runtimes MAY log a warning if duplicate entries
+    are present.
+
 ### Example
 
 ```json


config-linux.md
```diff
@@ -718,6 +718,7 @@ The following parameters can be specified to set up seccomp:
     This field MUST NOT be set if `listenerPath` is not set.
 
 * **`syscalls`** *(array of objects, OPTIONAL)* - match a syscall in seccomp.
+    When the syscall matches multiple entries, only the first entry is effective.
```
thaJeztah, line 721: > When the syscall matches multiple entries

Is any explanation needed on "matches"? As the same `syscall` with different `args` effectively is a different rule? (Suggestions welcome)


I'm somewhat inclined to move this its own paragraph below the definition of the properties. It feels a bit "off" to describe this _before_ describing the property to be optional, and describing what's in it.

I guess some of this may become more relevant with https://github.com/opencontainers/runtime-spec/pull/1102; perhaps that option may require stricter constraints though on duplicates (?). TBH, I think it's a bit unfortunate that this is how runtimes behave currently; Wondering if there's cases where duplicates were added unintentionally, and an error would've been more appropriate. 

Should we (can we?) add a recommendation for runtimes to log such cases?


```markdown
Syscalls in this list are not guaranteed to be unique, and MAY appear multiple
times. If a syscall appears multiple times, runtimes MUST use the first match,
MUST ignore further occurrences. Runtimes MAY log a warning if duplicate entries
are present.
```

AkihiroSuda: Thanks, updated, and added  `Co-authored-by:` line with your name in the commit log

config-linux.md
```diff
@@ -759,6 +759,11 @@ The following parameters can be specified to set up seccomp:
             * `SCMP_CMP_GT`
             * `SCMP_CMP_MASKED_EQ`
 
+    Syscalls in this list are not guaranteed to be unique, and MAY appear multiple
+    times. If a syscall appears multiple times, runtimes MUST use the first match,
+    MUST ignore further occurrences. Runtimes MAY log a warning if duplicate entries
```
thaJeztah, line 764: I think I may have missed an `and` here ☺️ 

```suggestion
    and MUST ignore further occurrences. Runtimes MAY log a warning if duplicate entries
```

AkihiroSuda: done

config-linux.md
thaJeztah: Thanks! 

> and added Co-authored-by: line with your name in the commit log

Oh, wasn't strictly needed for me, but thanks ☺️ 


I think the only part I could use input on is "first match"; I wonder if that needs further clarification (matching _only_ on sys call, or matching syscall  + args, which would make it "best match").

I should've probably dug into this myself, but perhaps others may have good wording for that to fill me in ❤️

Issue #1163, opened 2022-10-18
IMA Namespace support


IlyaHanov:
The Linux kernel community is now working on supporting [IMA namespaces](https://lore.kernel.org/linux-integrity/20220302134703.1273041-1-stefanb@linux.ibm.com/T/)
and it is almost done. It is a new kernel feature that allows isolation of  Platform Configuration Register (PCR) values, Measurement Logs (ML), etc. The related issue opened for [runC](https://github.com/opencontainers/runc/issues/3638).

Signed-off-by: Ilya Hanov <ilya.hanov@huawei-partners.com>

Advanced Software Technology Lab
Huawei

PR #1164, opened 2022-10-18
[RFC]: IMA Namespace support

IlyaHanov:
The Linux kernel community is now working on supporting [IMA namespaces](https://lore.kernel.org/linux-integrity/20220302134703.1273041-1-stefanb@linux.ibm.com/T/)
and it is almost done. It is a new kernel feature that allows isolation of  Platform Configuration Register (PCR) values, Measurement Logs (ML), etc. The related issue is #1163.

Signed-off-by: Ilya Hanov <ilya.hanov@huawei-partners.com>

Advanced Software Technology Lab
Huawei

### DIFF
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -218,6 +218,8 @@ const (
 	UTSNamespace LinuxNamespaceType = "uts"
 	// UserNamespace for isolating user and group IDs
 	UserNamespace LinuxNamespaceType = "user"
+	// ImaNamespace for isolating PCR values
+	ImaNamespace LinuxNamespaceType = "ima"
 	// CgroupNamespace for isolating cgroup hierarchies
 	CgroupNamespace LinuxNamespaceType = "cgroup"
 )


specs-go/config.go
```diff
@@ -218,6 +218,8 @@ const (
 	UTSNamespace LinuxNamespaceType = "uts"
 	// UserNamespace for isolating user and group IDs
 	UserNamespace LinuxNamespaceType = "user"
+	// ImaNamespace for isolating PCR values
+	ImaNamespace LinuxNamespaceType = "ima"
```
AkihiroSuda, line 222: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md needs to be updated too

AkihiroSuda:
Let me mark this as a draft, until the kernel patch gets merged into the upstream

tianon:
IMO this is a little bit premature -- ideally this functionality would be at least merged into the kernel (if not in an actual released version) before we implement or discuss it in this context.  What I would suggest in the future would be a posting to the OCI dev mailing list, OCI #general channel on Slack, etc if your goal is to get folks from the OCI aware of and potentially involved in the upstream kernel discussions.

DenisSemakin:
> IMO this is a little bit premature...
I think it's Yes and No simultaneously... Of course, the IMA namespace feature is not merged in kernel mainline and may be will not for some time.
But the way how to launch (activate) IMA namespace for given user namespace is stable ... more or less. Basically, this is the attempt to use IMA-ns in runC if the feature would be merged in kernel upstream and find out a number of problems and issues with this way.

IlyaHanov:
> IMO this is a little bit premature -- ideally this functionality would be at least merged into the kernel (if not in an actual released version) before we implement or discuss it in this context.  What I would suggest in the future would be a posting to the OCI dev mailing list, OCI #general channel on Slack, etc if your goal is to get folks from the OCI aware of and potentially involved in the upstream kernel discussions.

Yes, you're right in a way, IMA namespaces is not in upstream, but there's another interesting problem with _runC_. This is related to why IMA namespace uses _securityFS_ to create namespaces, but runC doesn't know about namespaces which are created not by using clone/unshare at all. This _patchset_ (one for runC, one for runtime-spec) presented not only IMA namespace creation scheme, but  tries to solve this problem as well, because if kernel community decides to invent another namespace, sooner or later this problem will show up.

PR #1167, opened 2022-10-31
runtime: move prestart hook in lifecycle section

corhere:
config.md specifies that "`prestart` hooks MUST be called after the `start` operation is called but before the user-specified program command is executed." However, the Lifecycle section in runtime.md contradicts this by showing that `prestart` hooks are called after the `create` operation is called and before the `start` operation. As the note in the config.md section for `createRuntime` hooks indicates, the order of operations in runtime.md#Lifecycle reflects the old runC behaviour which was contrary to the spec. Update the Lifecycle section to be congruent with the hook semantics as specified in config.md.

Signed-off-by: Cory Snider <csnider@mirantis.com>


h-vetinari:
It was my understanding from that despite being wrong, the prestart hooks could not be fixed (both in spec and implementation), because doing so would have broken a lot of projects relying on the old behaviour, hence why they were deprecated and replaced.

I think their placement in the lifecycle reflects that reality, and changing it is not really desirable.

I may be misremembering things from the sheer infinite #1008, but just my 2¢.

AkihiroSuda:
The actual behavior of runc can't be changed practically, so let's merge #1169

Issue #1020, opened 2019-10-29
[RFC]: allow to skip `setgroups(2)`


giuseppe:
There are cases where it would be necessary to skip the `setgroups(2)` syscall so that the original additional groups can be maintained.

It can be used, for example, by rootless containers to keep access to a storage directory that is accessible only by a secondary group.

runc already skips the `setgroups` in some cases: either if the user had euid != 0 or if `/proc/self/setgroups` is set to `deny`.  I'd like to add a third condition where the `setgroups` is skipped also if explicitly requested.

Do we need a new field under [`process/user`](https://github.com/opencontainers/runtime-spec/blob/master/config.md#user), e.g. `keepOriginalGroups`?  Would be enough to reuse `additionalGids` to have some special value (e.g. `-1` to keep current groups)?

cyphar:
If we do add an option, it needs to have a really scary name (`disableSetgroupSecurity` or something). Not dropping supplementary groups weakens the userns security boundary, and really is something that very few people should actually want to do (not least of all because it will confuse all sorts of programs to be touching unmapped files).

In my view, the best solution to the problem of such volumes is to do exactly what LXD does -- "punch out" the GID that the storage volume is owned by (by adding a single 1:1 mapping for that GID). The most ideal solution would be the next-gen "shiftfs" work that was discussed recently, but obviously we'll have to wait for that to actually land.

rhatdan:
I am skeptical, and think it could be a long wait, especially to get it upstream.

cyphar:
(Also I would seriously suggest that this is functionality that should be exposed through a runtime-specific annotation and not a first-class field in `config.json` -- the runtime-spec already has lots of really odd features we probably shouldn't have added, and this one just rubs me the wrong way.)

giuseppe:
> In my view, the best solution to the problem of such volumes is to do exactly what LXD does -- "punch out" the GID that the storage volume is owned by (by adding a single 1:1 mapping for that GID). The most ideal solution would be the next-gen "shiftfs" work that was discussed recently, but obviously we'll have to wait for that to actually land.

how would it work with rootless containers or in general with IDs that are not mapped in the current namespace?  I guess rootless containers won't still be able to map arbitrary IDs from the host.

tentator:
Hello @cyphar ,
 I also agree with Giuseppe since I have a customer that has exactly the same problem: he is wanting to use rootless podman but is currently limited by the missing subgroups to access mounted directories..
[...]

PR #1168, opened 2022-11-02
config: base GID must be present in the supplementary GIDs array

neersighted:
Currently, the spec is unclear whether the list of [supplementary GIDs][POSIX-sgids-def]
used to create a container process should include the 'base' GID
implicitly, or whether the config needs to specify this explicitly if
desired.

While [per POSIX][POSIX-sgids-rat] it is permissible for a system to
include or exclude the base GID from the list of supplementary GIDs, in
all Runtime Spec platforms the base GID is always added, and omitting it
can have [real security consequences][benthams-gaze] as fully dropping a
group is not typically allowed in Unix.

This recently led to a number of CVEs in OCI Runtime Spec
implementations, as it was concluded that it is necessary for a Unix
container to always include the base GID in the list of supplementary
GIDs, as originally established by 4.4BSD.

Some of the CVEs include:
* [Podman (CVE-2022-2989)][CVE-2022-2989]
* [Moby (CVE-2022-36109)][CVE-2022-36109]
* [Buildah (CVE-2022-2990)][CVE-2022-2990]
* [CRI-O (CVE-2022-2995)][CVE-2022-2995]

Some examples of how existing implementations handle this:
* util-linux [calls][util-linux] [initgroups(3)][initgroups.3-linux]
  with the user's primary GID.
* shadowutils (Linux) [calls][shadowutils]
  [initgroups(3)][initgroups.3-linux] with the user's primary GID.
* FreeBSD [calls][freebsd-setusercontext]
  [initgroups(3)][initgroups.3-freebsd] with the user's GID from the
  password file (aka the primary GID).
* Solaris [calls][solaris-setup_credentials]
  [initgroups(3)][initgroups.3-solaris] with the user's primary GID.
* Z/OS's session creation code is not available; however
  [initgroups(3)][initgroups.3-zos] specifies a convention of including
  including the real group ID from the user database (aka the primary
  GID).
* OpenSSH [calls][openssh] initgroups(3) with the user's primary GID; on
  all of the above platforms this will have the same result as a
  login(1), including the primary GID in the list of supplementary GIDs.

While login(1) has generally been used as the example above, the same
holds true for su(1) and other methods of starting a new session
(including OpenSSH, as explained above).

Given this seems clearly desirable and  the OCI runtime is effectively
equivalent of login(1)/su(1)/any other program that sets up a new
session, the OCI runtime is the best place to ensure that the list of
supplementary group IDs contains the base GID.

[POSIX-sgids-def]: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_378
[POSIX-sgids-rat]: https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xbd_chap03.html#tag_21_03_00_73

[CVE-2022-2989]: https://access.redhat.com/security/cve/cve-2022-2989
[CVE-2022-36109]: https://github.com/moby/moby/security/advisories/GHSA-rc4r-wh2q-q6c4
[CVE-2022-2990]: https://access.redhat.com/security/cve/cve-2022-2990
[CVE-2022-2995]: https://access.redhat.com/security/cve/cve-2022-2995
[benthams-gaze]: https://www.benthamsgaze.org/2022/08/22/vulnerability-in-linux-containers-investigation-and-mitigation/

[util-linux]: https://github.com/util-linux/util-linux/blob/96ccdc00e1fcf1684f9734a189baf90e00ff0c9a/login-utils/login.c#L1443
[shadowutils]: https://github.com/shadow-maint/shadow/blob/eaebea55a495a56317ed85e959b3599f73c6bdf2/libmisc/setugid.c#L55
[freebsd-setusercontext]: https://github.com/freebsd/freebsd-src/blob/eeaf9d562fe137e0c52b8c346742dccfc8bde015/lib/libutil/login_class.c#L486
[solaris-setup_credentials]: https://github.com/illumos/illumos-gate/blob/d9c3e05c2d8261e3f133b5e96a300b4fa6c0f1b7/usr/src/cmd/login/login.c#L1926
[openssh]: https://github.com/openssh/openssh-portable/blob/25bd659cc72268f2858c5415740c442ee950049f/session.c#L1379
[initgroups.3-linux]: https://man7.org/linux/man-pages/man3/initgroups.3.html
[initgroups.3-freebsd]: https://www.freebsd.org/cgi/man.cgi?initgroups(3)
[initgroups.3-solaris]: https://illumos.org/man/3C/initgroups
[initgroups.3-zos]: https://www.ibm.com/docs/en/zos/2.2.0?topic=functions-initgroups-initialize-supplementary-group-id-list-process

Signed-off-by: Bjorn Neergaard <bneergaard@mirantis.com>


neersighted:
cc @sjmurdoch @haircommander @p-rog

config.md
```diff
@@ -226,10 +226,17 @@ The user for the process is a platform-specific structure that allows specific c
...
 * **`umask`** (int, OPTIONAL) specifies the [umask][umask_2] of the user. If unspecified, the umask should not be changed from the calling process' umask.
-* **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs in the [container namespace](glossary.md#container-namespace) to be added to the process.
+* **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs in the [container namespace](glossary.md#container-namespace) to be added to the list of supplementary group IDs.
+
+On a POSIX platform, processes have both a 'base' GID (as specified in the `gid` field), and an array of supplementary group IDs as described in [IEEE Std 1003.1-2008][ieee-1003.1.2008-xbd-c3.378].
+Runtimes MUST ensure that all group IDs listed defined by `gid` and `additionalGids` are present in the array of supplementary group IDs.
+If the base GID (as specified in the `gid` field) is not present in `additionalGids`, Runtimes SHOULD place the base GID at the start of the supplementary group array. If the base GID is present in `additionalGids`, no additional action is required.
```
AkihiroSuda, line 236: > If the base GID is present in `additionalGids`, no additional action is required.

This sentence can be potentially misinterpreted as if an action to set "additional" GIDs was not required.
Probably this sentence can be just removed.

AkihiroSuda, line 236: > Runtimes SHOULD place the base GID at the start of the supplementary group array.

s/SHOULD/MUST/ for conssitency with "Runtimes MUST ensure that..." ?

neersighted: This was intended to allow some flexibility as to where the base GID is placed when it is not explicitly duplicated; I have rephrased this to be one sentence (and addressed your other review commit) in a way that hopefully removes this ambiguity.

AkihiroSuda (commented):
~LGTM but please squash commits~ (**EDIT**: Let me withdraw my LGTM, for compatibility sake)

neersighted:
The intention is to make review easier by not destroying review history (as Github does not have a concept of patch sets/series like Gerrit or `git format-patch`); approving reviews will not be dismissed when the commit series is squashed down (and the intended rebase can be seen by the fixup messages).

I'll likely wait until more maintainers have had time to review to squash, unless you think this is ready for merge now, in which case I can perform the rebase.

p-rog:
thank you @neersighted for opening this PR

giuseppe:
I am not fully convinced about this change.   It is a breaking change and there won't be a way to keep the existing behavior even if desired.

Personally, I still believe it is the caller's responsibility to ensure the gid is part of the additional gids.

neersighted:
For what its worth, I don't think that this precludes something like #1020. There is also the potential of adding a new field so we can have a backwards compatible behavior; however I do strongly feel that patching the existing spec is best as there does not seem to be any case where it is desirable to have the primary GID set but not in the supplementary GIDs.

giuseppe:
@cyphar what do you think about this change?

config.md
```diff
@@ -226,10 +226,16 @@ The user for the process is a platform-specific structure that allows specific c
...
+* **`gid`** (int, REQUIRED) specifies the group ID (GID) in the [container namespace](glossary.md#container-namespace).
 * **`umask`** (int, OPTIONAL) specifies the [umask][umask_2] of the user. If unspecified, the umask should not be changed from the calling process' umask.
-* **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs in the [container namespace](glossary.md#container-namespace) to be added to the process.
+* **`additionalGids`** (array of ints, OPTIONAL) specifies additional group IDs in the [container namespace](glossary.md#container-namespace) to be added to the list of supplementary group IDs.
+
+On a POSIX platform, processes have both a 'base' GID (as specified in the `gid` field), and an array of supplementary group IDs as described in [IEEE Std 1003.1-2008][ieee-1003.1.2008-xbd-c3.378].
+Runtimes MUST ensure that all group IDs specified by `gid` and `additionalGids` are present in the array of supplementary group IDs.
+Runtimes SHOULD preserve the order of `additionalGids`; when the base GID (as specified in the `gid` field) is absent from `additionalGids`, it SHOULD be positioned at the start of the supplementary group ID array.
```
thaJeztah, line 236: Minor nit; should it be s/when/if/ ?

```suggestion
Runtimes SHOULD preserve the order of `additionalGids`; if the base GID (as specified in the `gid` field) is absent from `additionalGids`, it SHOULD be positioned at the start of the supplementary group ID array.
```

neersighted: Both are valid syntactically and convoy the intent accurately; however, your version is a bit more obvious, so I think we should prefer it.

kolyshkin (approved):
@neersighted can you squash the commits? Otherwise LGTM

neersighted:
Pushed a squashed version with no other changes.

giuseppe (changes requested):
FWIW, I am still against this change as it is since there is no way to keep the current behavior.  Could we at least add a new knob to maintain the existing behavior?

AkihiroSuda:
Let met withdraw [my LGTM on Nov 3](https://github.com/opencontainers/runtime-spec/pull/1168#pullrequestreview-1166257746), for compatibility sake

neersighted:
Seems like there's a consensus that we should introduce a new field to preserve existing behavior. I'll try and work up a draft next week.

sjmurdoch:
There was a proposal by [Cory Snider](https://github.com/corhere) in the private discussion of [GHSA-rc4r-wh2q-q6c4](https://github.com/moby/moby/security/advisories/GHSA-rc4r-wh2q-q6c4#advisory-comment-73617) which I think intended to preserve backwards compatibility. I don't have any particular views on it, but it might be a useful starting point for discussion. Since it was private, I don't think I should copy and paste it here, but those who can see the issue can read it and Cory could be asked for permission to publish it.

neersighted:
Cory and I have discussed it outside of that security bulletin and iterated on the idea a bit more (we both work on the same team on Mirantis); I will be work-shopping something similar to his design with him before presenting it here 😄

corhere:
Here is what I had proposed in the GHSA discussion linked above:

> The trouble I see with pushing the responsibility to duplicate `gid` into the `additionalGids` array solely onto the container engines and other software which generates an OCI config is that there are a lot more of them than of OCI runtimes. Each and every one of them would need to get it right, on a very obscure piece of UNIX trivia. Excluding the user's gid from the supplemental groups is almost always a mistake, and with the subtle security consequences of such a mistake I think that would qualify it as a footgun. That's why I think that *both* the OCI runtimes and container engines should augment the supplementary groups list.
>
> I believe that due to the vague wording of the runtime spec, OCI runtimes augmenting the supplemental groups list would be compliant with runtime-spec v1.0.2. The spec [has a configuration field to specify "additional" group IDs to be added to the process on POSIX platforms,](https://specs.opencontainers.org/runtime-spec/config/?v=v1.0.2#posix-platform-user) but neither the OCI runtime spec nor POSIX define what "additional group IDs" are. POSIX does, on the other hand, [define "supplementary group IDs."](https://pubs.opengroup.org/onlinepubs/007904875/basedefs/xbd_chap03.html#:~:text=3.370%20Supplementary%20Group%20ID) As the OCI runtime spec does not define the mapping from a `user` structure to the process's supplementary group ID list, an implementation which sets it to the union of `user.gid` and `user.additionalGids` would be in compliance. I propose that the OCI runtime spec be updated roughly as follows:
> * Add a new field to the POSIX user structure: `sgids` (array of int, OPTIONAL). When set, the OCI runtime MUST create the process with supplementary group IDs set to exactly the set of groups listed in the `user.sgids` array. This field takes precedence over the `additionalGids` field; the OCI runtime MUST ignore the `additionalGids` field if `sgids` is provided.
> * When no `user.sgids` array is provided in the container config, the OCI runtime MUST create the process with supplementary group IDs set to all the GIDs listed in the `user.additionalGroups` config fields. The runtime MUST also add the group ID specified in `user.gid` to the process supplementary groups if the platform conventionally includes the user's primary group in the supplementary groups list for login users and the ID is not already listed in `user.additionalGroups`.
>   * Including the user's primary group in the supplementary group list is conventional on Linux, *BSD, Solaris and Darwin.
>   * Generators of OCI runtime configuration metadata SHOULD append the primary group ID (`user.gid`) to the `user.additionalGroups` array on host platforms where the convention is to include the user's primary group in the supplementary group list to maximize compatibility and consistency with OCI runtimes targeting older versions of the OCI runtime spec.
>
> A spec change like this would allow for defence in depth while still permitting full control over the supplementary group list to be opted into for advanced use cases. Container engines could still add the primary group to the `additionalGids` array to fix the vulnerability even when used with older runtimes, and runtimes could be updated to fix the vulnerability even when used with older container engines. And if someone writes a new container engine and forgets to append the primary GID to `additionalGids`, they'd still be safe so long as they're using it with an up-to-date runtime.

config.md
neersighted, line 1: Needs your `Co-authored-by:` as well.

corhere:
@neersighted not squashing was intentional. (Omitting the `Co-authored-by` trailer was not...) Since it's such a big change, I want to make it easy for reviewers to compare and contrast our drafts side by side.

AkihiroSuda:
I feel this is too complicated.
Can we just add "the `gid` value SHOULD be prepended to the `additionalGids` array"?

Runtime implementations may raise a warning when `additionalGids` does not contain the base `gid`.
Image linters, such as https://github.com/goodwithtech/dockle , may error out in that situation, depending on their policy.

corhere:
> I feel this is too complicated. Can we just add "the `gid` value SHOULD be prepended to the `additionalGids` array"?

That would still leave the generator of the container config as the single point of failure. The purpose of this proposal is to provide [defense-in-depth:](https://en.wikipedia.org/wiki/Swiss_cheese_model) the bundle generator (container engine) and the runtime would both have to be out-of-date or otherwise get things wrong for a container to run with broken negative group permissions.

My proposal, described simply, is:
- runtimes MUST prepend `gid` to the container process' supplemental group IDs if not already present in `additionalGids`
- generators of a container config are strongly encouraged to prepend `gid` to `additionalGids`
- config generators which require absolute control over supplemental groups for some reason have an escape hatch

The complexity of how that is specified comes from having to work within the confines of what is, and is not, defined by POSIX. Privileged operations including `setgroups` and `initgroups` are out of scope of POSIX, so I specified the runtime's behaviour in terms of `getgroups` in order to have the expected behaviour rigorously defined for all POSIX compliant platforms.

> Runtime implementations may raise a warning when `additionalGids` does not contain the base `gid`.

Logging a warning wouldn't be of much help as the container would still be started with a broader scope of privilege than intended. And runtime-spec doesn't mandate that warnings logged by the runtime reach the user.

> Image linters, such as https://github.com/goodwithtech/dockle , may error out in that situation, depending on their policy.

Image? This is runtime-spec. Images are out of scope for this discussion as there's no guarantee that a container bundle was even derived from an image.

giuseppe:
> I feel this is too complicated. Can we just add "the `gid` value SHOULD be prepended to the `additionalGids` array"?
> 
> Runtime implementations may raise a warning when `additionalGids` does not contain the base `gid`.

I agree, the current design is too complicated and IMO not worth it.

I still feel this is not the OCI runtime responsibility and the security issue is anyway difficult to find in reality as it affects only images that use negative DACs and setuid binaries.

corhere:
@giuseppe all those programs call `initgroups` when setting up a login session, not `setgroups`.

Issue #1069, opened 2020-10-01
Proposal: separate go implementation tagging from specification itself


thaJeztah:
Related to https://github.com/opencontainers/runtime-spec/issues/1052 and https://github.com/opencontainers/runtime-spec/pull/1068

Currently, this repository tags releases for the runtime spec itself, however, the go implementation of the specification also resides in this repository.

While it's good to have both living in the same repository, sometimes changes in the go implementation may be desirable that don't relate to a change in the specification itself (bugfixes, updated dependencies, enhancements).

My proposal is to consider to tag the go implementation separate from the specification.

Given that the go implementation lives in a subdirectory, this could be done by making it a submodule, and using prefixes for the tags; tags for the go implementation would be prefixed with `specs-go/`

- new prefixed tags can be created for each existing tag (`v1.0.0` -> `specs-go/v1.0.0`, `v1.0.1` -> `specs-go/v1.0.1`)
- a go.mod can be placed inside the `spec-go` directory, to make it its own module

Note that the Go implementation contains the version of the runtime-spec that it implements; https://github.com/opencontainers/runtime-spec/blob/v1.0.2/specs-go/version.go, so the version if the specification can be identified once the tagged releases of the go module diverge from the runtime spec.

Go modules require SemVer to be applied (at the Go API level), so breaking changes require the "major" version component to be updated, and feature additions in the Go code should increment the "minor" version component. Given that go modules look from the Go API perspective, it's likely that the versions will diverge (e.g., v2.0.0 of the specification may be fully backward compatible with v1.x.x, so strictly wouldn't require a major version update of the go implementation). 

I realise this could confuse users, so proper documentation / inclusion in the README around this would be good to have.

mikebrow:
SGTM

kolyshkin:
LGTM

vbatts:
I'm for this, but it adds to the release process, which I had to make a [checklist](https://github.com/opencontainers/runtime-spec/blob/main/RELEASES.md#checklist) for.

It seems some of this can be done in phases, right? Like, we don't need to split out a git-submodule initially. But could make tags for past releases?

AkihiroSuda:
image-spec and distribution-spec already have `go.mod`:
- https://github.com/opencontainers/image-spec/blob/main/go.mod
- https://github.com/opencontainers/distribution-spec/blob/main/specs-go/go.mod

PR #1179, opened 2023-02-02
Add Go 1.20 support

austinvazquez:
* Add Go 1.20 to GitHub actions matrix
* Updates actions/setup-go from v1 to v3. (Resolves deprecation warnings in GitHub actions around Node 12. Reference [here](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)) 

Signed-off-by: Austin Vazquez <macedonv@amazon.com>

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -11,22 +11,32 @@ jobs:
     steps:
       - name: checkout source code
         uses: actions/checkout@master
+
       - name: setup go environment
         uses: actions/setup-go@v1
         with:
           go-version: '1.17.2'
+      
+      - name: create go.mod
+        run: |
+          # Fix for "cannot find main module" issue
+          go mod init github.com/opencontainers/runtime-spec
+
+          go get -d ./schema/...
+
+      - name: run golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.51.2
+          args: --verbose
+
       - name: run tests
         run: |
           export PATH="$(go env GOPATH)/bin:${PATH}"
           set -x
           make install.tools
 
-          # Fix for "cannot find main module" issue
-          go mod init github.com/opencontainers/runtime-spec
-
-          go get -d ./schema/...
           make .govet
-          make .golint
 
           make .gitvalidation
           make docs
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -8,14 +8,19 @@ on:
 jobs:
   run:
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        go: [1.19.x, 1.20.x]
+
     steps:
       - name: checkout source code
-        uses: actions/checkout@master
+        uses: actions/checkout@v3
 
       - name: setup go environment
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v4
         with:
-          go-version: '1.17.2'
+          go-version: ${{ matrix.go }}
       
       - name: create go.mod
         run: |
@@ -32,7 +37,6 @@ jobs:
 
       - name: run tests
         run: |
-          export PATH="$(go env GOPATH)/bin:${PATH}"
           set -x
           make install.tools
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,22 +11,32 @@ jobs:
     steps:
       - name: checkout source code
         uses: actions/checkout@master
+
       - name: setup go environment
         uses: actions/setup-go@v1
         with:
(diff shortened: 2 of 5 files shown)


.github/workflows/build-pr.yml
```diff
@@ -8,13 +8,18 @@ on:
 jobs:
   run:
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        go: ['1.18', '1.19', '1.20']
```
thaJeztah, line 14: I'm always confused by the formats this action supports; is `1.18` equivalent to `1.18.0` or should it be `1.18.x`?

thaJeztah (approved):
LGTM (assuming the version format is correct)

thaJeztah (changes requested):
arf; looks like some changes are needed (see failures)

.github/workflows/build-pr.yml
austinvazquez: `1.18` should pull the latest; however, don't mind updating to `1.18.x` just to be more clear.

austinvazquez:
> arf; looks like some changes are needed (see failures)

Added a little extra scope to resolve the errors. Took the opportunity to update the lint tooling from golint to golangci-lint and resolved new errors. Included here since we have the hood open but could split this into a seperate commit/PR if that helps.

tianon (approved):
Adding `go.mod` explicitly to `.gitignore` is a little odd when I don't think anything here generates it (right?), but I don't care about that enough to block this.

The other combined CI fixes/updates seem fine to include here IMO 👍

austinvazquez:
> Adding `go.mod` explicitly to `.gitignore` is a little odd when I don't think anything here generates it (right?), but I don't care about that enough to block this.
> 
> The other combined CI fixes/updates seem fine to include here IMO 👍

Good call out I could have explained that better. CI generates them [here](https://github.com/opencontainers/runtime-spec/blob/main/.github/workflows/build-pr.yml#L25) and just wanted to be sure I didn't accidentally push those.

AkihiroSuda:
Why not use go.mod ?

austinvazquez:
> Why not use go.mod ?

I can make this happen. IIRC other specs had them.

AkihiroSuda (approved):

go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/opencontainers/runtime-spec
```
thaJeztah, line 1: This has a potential wide impact than updating go versions in CI; do we still need to discuss;

https://github.com/opencontainers/runtime-spec/issues/1069

(or was there a concensus on that yet w.r.t. "SemVer from a Go perspective vs Spec perspective")?

AkihiroSuda: image-spec and distribution-spec already have `go.mod`:
- https://github.com/opencontainers/image-spec/blob/main/go.mod
- https://github.com/opencontainers/distribution-spec/blob/main/specs-go/go.mod

And these struct definitions are unlikely to have breaking changes (until the spec itself has breaking changes)

AkihiroSuda: @thaJeztah WDYT?

AkihiroSuda: @austinvazquez @thaJeztah 
Do we need to split `go.mod` to another PR?

thaJeztah: I would consider moving it separate yes, as it's not directly related to updating / updating CI to test against a newer go version

austinvazquez: Okay, thanks for the feedback. Let me split and push.

tianon (approved):

kolyshkin (changes requested):
Changes look good overall, but they need to be structured.

1.  Fix using deprecated ioutil.
2. Switch to using golangci-lint.
3. Modify ci to use matrix of go versions

One other thing, for golangci-lint it is better to use their action directly from .github/workflows rather than through make.

AkihiroSuda:
@austinvazquez ↑ PTAL

austinvazquez:
> Changes look good overall, but they need to be structured.
> 
> 1. Fix using deprecated ioutil.
> 2. Switch to using golangci-lint.
> 3. Modify ci to use matrix of go versions
> 
> One other thing, for golangci-lint it is better to use their action directly from .github/workflows rather than through make.

@kolyshkin, let me draft up what golangci-lint through github action would look like and you can lmk your thoughts.

AkihiroSuda:
CI is failing

austinvazquez:
> CI is failing

Oof I've fallen victim to the go.mod issue. Moved the hack up to resolve.

tianon (approved):

thaJeztah (approved):
LGTM

AkihiroSuda (approved):

PR #1188, opened 2023-03-16
spec: add scheduler entity

giuseppe:
extend the process struct to represent scheduling attributes for a process based on the sched_setattr(2) syscall.

I've been playing with crun and it is currently implemented as an annotation:

- https://github.com/containers/crun/pull/1164
- https://github.com/containers/crun/pull/1166

but I think this makes sense as part of the runtime-specs without the custom hack.

### DIFF
--- a/config.md
+++ b/config.md
@@ -291,6 +291,33 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     This is a per-process setting, where as [`disableOOMKiller`](config-linux.md#memory) is scoped for a memory cgroup.
     For more information on how these two settings work together, see [the memory cgroup documentation section 10. OOM Contol][cgroup-v1-memory_2].
+* **`scheduler`** (object, OPTIONAL) is an object describing the scheduler properties for the process.  The `scheduler` contains the following properties:
+
+    * **`policy`** (string, REQUIRED) represents the scheduling policy.  A valid list of values is:
+
+        * `SCHED_OTHER`
+        * `SCHED_FIFO`
+        * `SCHED_RR`
+        * `SCHED_BATCH`
+        * `SCHED_ISO`
+        * `SCHED_IDLE`
+        * `SCHED_DEADLINE`
+
+    * **`nice`** (int32, OPTIONAL) is the nice value for the process, affecting its priority. A lower nice value corresponds to a higher priority. If not set, the runtime must use the value 0.
+    * **`priority`** (int32, OPTIONAL) represents the static priority of the process, used by real-time policies like SCHED_FIFO and SCHED_RR. If not set, the runtime must use the value 0.
+    * **`flags`** (array of strings, OPTIONAL) is an array of strings representing scheduling flags.  A valid list of values is:
+
+        * `SCHED_FLAG_RESET_ON_FORK`
+        * `SCHED_FLAG_RECLAIM`
+        * `SCHED_FLAG_DL_OVERRUN`
+        * `SCHED_FLAG_KEEP_POLICY`
+        * `SCHED_FLAG_KEEP_PARAMS`
+        * `SCHED_FLAG_UTIL_CLAMP_MIN`
+        * `SCHED_FLAG_UTIL_CLAMP_MAX`
+
+    * **`runtime`** (uint64, OPTIONAL) represents the amount of time in nanoseconds during which the process is allowed to run in a given period, used by the deadline scheduler. If not set, the runtime must use the value 0.
+    * **`deadline`** (uint64, OPTIONAL) represents the absolute deadline for the process to complete its execution, used by the deadline scheduler. If not set, the runtime must use the value 0.
+    * **`period`** (uint64, OPTIONAL) represents the length of the period in nanoseconds used for determining the process runtime, used by the deadline scheduler. If not set, the runtime must use the value 0.
 * **`selinuxLabel`** (string, OPTIONAL) specifies the SELinux label for the process.
     For more information about SELinux, see  [SELinux documentation][selinux].
 
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -147,6 +147,38 @@
                 "noNewPrivileges": {
                     "type": "boolean"
                 },
+                "scheduler": {
+                    "type": "object",
+                    "required": [
+                        "policy"
+                    ],
+                    "properties": {
+                        "policy": {
+                            "$ref": "defs-linux.json#/definitions/SchedulerPolicy"
+                        },
+                        "nice": {
+                            "$ref": "defs.json#/definitions/int32"
+                        },
+                        "priority": {
+                            "$ref": "defs.json#/definitions/int32"
+                        },
+                       "flags": {
+                           "type": "array",
+                           "items": {
+                               "$ref": "defs-linux.json#/definitions/SchedulerFlag"
+                           }
+                        },
+                        "runtime": {
+                            "$ref": "defs.json#/definitions/uint64"
+                        },
+                        "deadline": {
+                            "$ref": "defs.json#/definitions/uint64"
+                        },
+                        "period": {
+                            "$ref": "defs.json#/definitions/uint64"
+                        }
+                    }
+                },
                 "rlimits": {
                     "type": "array",
                     "items": {
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -33,6 +33,34 @@ type Spec struct {
 	ZOS *ZOS `json:"zos,omitempty" platform:"zos"`
 }
 
+// Scheduler represents the scheduling attributes for a process. It is based on
+// the Linux sched_setattr(2) syscall.
+type Scheduler struct {
+	// Policy represents the scheduling policy (e.g., SCHED_FIFO, SCHED_RR, SCHED_OTHER).
+	Policy LinuxSchedulerPolicy `json:"policy"`
+
+	// Nice is the nice value for the process, which affects its priority.
+	Nice int32 `json:"nice,omitempty"`
+
+	// Priority represents the static priority of the process.
+	Priority int32 `json:"priority,omitempty"`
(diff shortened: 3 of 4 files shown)


config.md
```diff
@@ -291,6 +291,16 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     This is a per-process setting, where as [`disableOOMKiller`](config-linux.md#memory) is scoped for a memory cgroup.
     For more information on how these two settings work together, see [the memory cgroup documentation section 10. OOM Contol][cgroup-v1-memory_2].
+* **`scheduler`** (object, OPTIONAL) is an object describing the scheduler properties for the process.  The `scheduler` contains the following properties:
+
+    * **`policy`** (string, OPTIONAL) represents the scheduling policy (e.g., SCHED_FIFO, SCHED_RR, SCHED_OTHER).
+    * **`nice`** (int32, OPTIONAL) is the nice value for the process, affecting its priority. A lower nice value corresponds to a higher priority.
+    * **`priority`** (int32, OPTIONAL) represents the static priority of the process, used by real-time policies like SCHED_FIFO and SCHED_RR.
+    * **`flags`** (array of strings, OPTIONAL) is an array of strings representing scheduling flags (e.g., SCHED_FLAG_RESET_ON_FORK, SCHED_FLAG_RECLAIM).
```
AkihiroSuda, line 299: Needs a complete list

specs-go/config.go
```diff
@@ -33,6 +33,34 @@ type Spec struct {
...
 }
 
+// Scheduler represents the scheduling attributes for a process. It is based on
+// the Linux sched_setattr(2) syscall.
+type Scheduler struct {
+	// Policy represents the scheduling policy (e.g., SCHED_FIFO, SCHED_RR, SCHED_OTHER).
+	Policy string `json:"policy"`
+
+	// Nice is the nice value for the process, which affects its priority.
+	Nice int32 `json:"nice"`
```
AkihiroSuda, line 43: Probably this should be a pointer to differentiate zero from nil

giuseppe: do we need that?  There is no unset for `sched_setattr`:

```
           struct sched_attr {
               u32 size;              /* Size of this structure */
               u32 sched_policy;      /* Policy (SCHED_*) */
               u64 sched_flags;       /* Flags */
               s32 sched_nice;        /* Nice value (SCHED_OTHER,
                                         SCHED_BATCH) */
               u32 sched_priority;    /* Static priority (SCHED_FIFO,
                                         SCHED_RR) */
               /* Remaining fields are for SCHED_DEADLINE */
               u64 sched_runtime;
               u64 sched_deadline;
               u64 sched_period;
           };
```

in this end this will be the value we set, unset will be more confusing

config.md
```diff
@@ -291,6 +291,16 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     This is a per-process setting, where as [`disableOOMKiller`](config-linux.md#memory) is scoped for a memory cgroup.
     For more information on how these two settings work together, see [the memory cgroup documentation section 10. OOM Contol][cgroup-v1-memory_2].
+* **`scheduler`** (object, OPTIONAL) is an object describing the scheduler properties for the process.  The `scheduler` contains the following properties:
+
+    * **`policy`** (string, OPTIONAL) represents the scheduling policy (e.g., SCHED_FIFO, SCHED_RR, SCHED_OTHER).
```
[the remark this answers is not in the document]
giuseppe: fixed

config.md
```diff
@@ -291,6 +291,33 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+
+    * **`policy`** (string, REQUIRED) represents the scheduling policy.  A valid list of values is:
+
+        * `SCHED_OTHER`
+        * `SCHED_FIFO`
+        * `SCHED_RR`
+        * `SCHED_BATCH`
+        * `SCHED_ISO`
+        * `SCHED_IDLE`
+        * `SCHED_DEADLINE`
```
AkihiroSuda, line 304: Needs to be defined in specs-go too

specs-go/config.go
```diff
@@ -788,3 +818,43 @@ type ZOSDevice struct {
 	// Gid of the device.
 	GID *uint32 `json:"gid,omitempty"`
 }
+
+// LinuxSchedulerPolicy represents different scheduling policies used with the Linux Scheduler
+type LinuxSchedulerPolicy string
+
+const (
+	// SCHED_OTHER is the default scheduling policy
+	SCHED_OTHER LinuxSchedulerPolicy = "SCHED_OTHER"
```
AkihiroSuda, line 827:
```suggestion
	SchedOther LinuxSchedulerPolicy = "SCHED_OTHER"
```
To follow the current coding style:
https://github.com/opencontainers/runtime-spec/blob/main/specs-go/config.go#L677

giuseppe: thanks, changed to camel case

AkihiroSuda (approved):

rhatdan:
LGTM
@kolyshkin @mrunalp PTAL

mrunalp (approved):
LGTM

utam0k:
Thanks, @giuseppe

specs-go/config.go
```diff
@@ -33,6 +33,34 @@ type Spec struct {
...
 }
 
+// Scheduler represents the scheduling attributes for a process. It is based on
+// the Linux sched_setattr(2) syscall.
+type Scheduler struct {
+	// Policy represents the scheduling policy (e.g., SCHED_FIFO, SCHED_RR, SCHED_OTHER).
+	Policy LinuxSchedulerPolicy `json:"policy"`
+
+	// Nice is the nice value for the process, which affects its priority.
+	Nice int32 `json:"nice"`
```
utam0k, line 43: If you know, please let me know, does this apply to process groups?:pray: Or only one process?

config.md
utam0k, line 304: SCHED_DEADLINE seems to sometimes return an EBUSY error. Do you want the error numbers to follow these when the set fails?
https://man7.org/linux/man-pages/man2/sched_setattr.2.html

There seems to be some difficulty in using SCHED_DEADLINE with cpuset. can we consider removing SCHED_DEADLINE from support?
https://www.kernel.org/doc/Documentation/scheduler/sched-deadline.txt

giuseppe: I don't think it is required to support it, I am just listing the possible values for now (e.g. `SCHED_ISO` is not even implemented yet)

specs-go/config.go
giuseppe: from my understanding that is only for one process or thread:

utam0k: I agree with you. I thought the user would expect the nice value to be applied to all processes in the container? What do you think?

config.md
utam0k: I will leave it up to you to specify about it in the oci-spec, but there are some things that probably cannot be handled. I thought users might expect to be able to use them.

utam0k (commented):
@giuseppe 
Sorry if I am causing trouble by leaving comments from the outside. If it doesn't seem necessary and the maintainer is satisfied, please ignore it and merge.

config.md
utam0k: How about specifying whether it applies to the container init process alone or to all processes in the container?

specs-go/config.go
utam0k: How about specifying whether it applies to the container init process alone or to all processes in the container?

giuseppe:
anything blocking or could we move forward with this PR?

schema/config-schema.json
```diff
@@ -147,6 +147,32 @@
                 "noNewPrivileges": {
                     "type": "boolean"
                 },
+                "scheduler": {
+                    "type": "object",
+                    "properties": {
+                        "policy": {
+                            "type": "string"
+                        },
```
thaJeztah, line 155: This should probably have an enum as well (`SCHED_OTHER`, `SCHED_FIFO` ...), and perhaps a definition (to match the Go type)

schema/config-schema.json
```diff
@@ -147,6 +147,32 @@
                 "noNewPrivileges": {
                     "type": "boolean"
                 },
+                "scheduler": {
+                    "type": "object",
+                    "properties": {
```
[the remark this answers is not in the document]
giuseppe: yes thanks for spotting it!

schema/config-schema.json
```diff
@@ -147,6 +147,32 @@
...
+                            "type": "string"
+                        },
+                        "nice": {
+                            "$ref": "defs.json#/definitions/int32"
+                        },
+                        "priority": {
+                            "$ref": "defs.json#/definitions/int32"
+                        },
+                        "flags": {
+                            "$ref": "defs.json#/definitions/ArrayOfStrings"
```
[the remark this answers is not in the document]
giuseppe: sure, fixed now

giuseppe:
@thaJeztah thanks for the review.  Addressed your comments and pushed a new version

giuseppe:
@thaJeztah does the new version look good to you?

specs-go/config.go
```diff
@@ -33,6 +33,34 @@ type Spec struct {
...
+
+	// Priority represents the static priority of the process.
+	Priority int32 `json:"priority"`
```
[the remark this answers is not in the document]
giuseppe: sure, that seems like a good idea.  Fixed and pushed a new version

giuseppe:
@thaJeztah could you PTAL?

thaJeztah (approved):
LGTM (thanks for the ping; sorry,  thought I already did 🙈 ❤️)

thaJeztah:
Any more reviewers needed, or are we good to go? (forgot how many we want for such changes in this repository 🙈)

thaJeztah:
Ah! Thanks @AkihiroSuda 😂

PR #1191, opened 2023-03-23
Add I/O Priority Configuration for Process Group in Linux Containers

utam0k:
This PR introduces a new configuration option to set the I/O priority for the entire process group within a Linux container. By allowing users to configure I/O priority, we enable better resource management and performance tuning for various workloads, particularly in Kubernetes environments where multiple containers may be running concurrently.

One specific use case is when running batch processing jobs alongside application Pods in a Kubernetes cluster. Batch jobs are often I/O-intensive and can cause resource contention, negatively affecting the performance of other containers running on the same node. By allowing I/O priority configuration for each container, cluster administrators can minimize the impact of I/O-intensive batch jobs on application Pods.

By adding I/O priority configuration to the OCI Runtime spec, we enable Kubernetes (and other container orchestration platforms) to take advantage of this feature and better manage container I/O resources. This can lead to improved overall performance and resource utilization on nodes running multiple containers with different I/O requirements.


AkihiroSuda:
- https://github.com/opencontainers/runtime-spec/blob/main/schema/config-linux.json has to be updated too.
- I'd like to see a POC of runc, crun, or youki before merging this

utam0k:
@AkihiroSuda I have implemented it to runc.
https://github.com/opencontainers/runc/pull/3783

AkihiroSuda (changes requested):
Please update https://github.com/opencontainers/runtime-spec/blob/main/schema/config-linux.json

utam0k:
> Please update https://github.com/opencontainers/runtime-spec/blob/main/schema/config-linux.json

@AkihiroSuda 
I completely forgot about it 🤦I have updated it.

AkihiroSuda (approved):

AkihiroSuda:

schema/config-schema.json
```diff
@@ -144,6 +144,10 @@
                 "selinuxLabel": {
                     "type": "string"
                 },
+                "ioPriority": {
+                    "class": "string",
+                    "priority": "integer"
+                },
```
giuseppe, line 150: should it be an enumeration of the supported values?

utam0k: @giuseppe Thanks for your review. It would be better! I have fixed it.

AkihiroSuda (approved):

giuseppe (approved):
LGTM

schema/config-schema.json
```diff
@@ -144,6 +144,15 @@
...
                 },
+                "ioPriority": {
+                    "class": "string",
+                    "enum": [
+                        "IOPRIO_CLASS_RT",
+                        "IOPRIO_CLASS_BE",
+                        "IOPRIO_CLASS_IDLE"
+                    ],
+                    "priority": "integer"
+                },
```
thaJeztah, line 155: I think the enum is defined in the wrong location, because this `enum` would now apply to `ioPriority` as a whole, but the enum defines values for `class` only.

So `class` should probably be defined as `"#/definitions/IOPriorityClass"` (e.g.) instead of a `"string"`. Similar to 

https://github.com/opencontainers/runtime-spec/blob/720792f25ae6e9ee6b1332db698f37659e69ce8d/schema/defs-linux.json#L102

https://github.com/opencontainers/runtime-spec/blob/504f70ef81a4f5bfb337e5a694c9552e22bf9935/schema/defs-linux.json#L77-L88

thaJeztah: Oh! I think https://github.com/opencontainers/runtime-spec/pull/1206 was meant to address that, correct @giuseppe ?

PR #1194, opened 2023-04-03
Clarify I/O throttling differences between cgroup v1 and v2

utam0k:
This PR adds a note to the OCI Runtime Specification documentation, clarifying the differences in I/O throttling behavior between cgroup v1 and cgroup v2. Specifically, it mentions that I/O throttling settings in cgroup v1 apply only to Direct I/O due to kernel implementation constraints, while this limitation does not exist in cgroup v2.

By providing this information, users will be better informed about these differences and can make more appropriate decisions when configuring.


giuseppe (approved):
LGTM

PR #1197, opened 2023-04-19
config: add support for org.opencontainers.image annotations

cyphar:
These annotations are specified as part of the official conversion process from an OCI image configuration to an OCI runtime configuration since v1.0.0-rc7 of the image-spec but they were never officially specified in

The fact these are not allowed by the current runtime-spec causes some issues with inter-spec compatibility. In order for tools like umoci to be able to do this generation properly, we need to allow this namespace to be used.

Ref: https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/conversion.md

### DIFF
--- a/config.md
+++ b/config.md
@@ -648,7 +648,21 @@ If there are no annotations then this property MAY either be absent or an empty 
 Keys MUST be strings.
 Keys MUST NOT be an empty string.
 Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
-Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
+
+The `org.opencontainers` namespace for keys is reserved for use by this specification, annotations using keys in this namespace MUST be as described in this section.
+The following keys in the `org.opencontainers` namespaces MAY be used:
+|                   Key                   | Definition                                                         |
+| --------------------------------------- | -----------------------------------------------------------------------------------------------------------------------------------|
+| `org.opencontainers.image.os`           | Indicates the operating system the container image was built to run on. The annotation value MUST have a valid value for the `os` property as defined in [the OCI image specification][oci-image-config-properties]. This annotation SHOULD only be used in accordance with the [OCI image specification's runtime conversion specification][oci-image-conversion]. |
+| `org.opencontainers.image.os.version`   | Indicates the operating system version targeted by the container image. The annotation value MUST have a valid value for the `os.version` property as defined in [the OCI image specification][oci-image-config-properties]. This annotation SHOULD only be used in accordance with the [OCI image specification's runtime conversion specification][oci-image-conversion]. |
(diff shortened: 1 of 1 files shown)


cyphar:
This is a minimal implementation of support for the `org.opencontainers.image` namespace of annotations ([as defined in and used by the image-spec](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc2/conversion.md)).

However, given that the list of annotations might grow in the future, it might make far more sense to copy the current set of annotations into the runtime-spec and require any additions to the list of annotations to be co-ordinated between specifications? That way we don't need to hard-code a reference to a specific image-spec version in the runtime-spec...

cyphar:
(One reason we need this PR is that umoci has been generating these configs for ages, but the version of `oci-runtime-tool` used to validate the generated `config.json` didn't support validating annotations. I've only just gotten around to updating the ancient version of `oci-runtime-tool` and it turns out this conversion causes verification failures.)

giuseppe (approved):
LGTM

cyphar:
I will switch to having a copy of the properties here, since otherwise it'll be difficult to guarantee compatibility with a specific runtime-spec version (what version of image-spec should be used?). We don't add annotations to the list very regularly so having to do updates in lock-step probably isn't too much of a hassle.

giuseppe:
Would these annotations matter anyway for an OCI runtime?

cyphar:
I think some runtimes might make use of them (`org.opencontainers.image.stopSignal` is the most obvious one that a runtime might care about).

The issue is that at the moment technically you're forbidden to use them for anything according to the runtime-spec, but the image-spec explicitly requires you to use them when generating a `config.json`. If we want to have proper validation in `runtime-tools` (rather than just allowing all of `org.opencontainers.image.*` without restrictions) I think we should at least have a list of the allowed annotations (and point to the image-spec for their semantics).

giuseppe (approved):
LGTM

config.md
```diff
@@ -648,7 +648,21 @@ If there are no annotations then this property MAY either be absent or an empty
...
+| `org.opencontainers.image.created`      | Indicates the date and time when the container image was created. The annotation value MUST have a valid value for the `created` property as defined in [the OCIimage specification][oci-image-config-properties]. This annotation SHOULD only be used in accordance with the [OCI image specification's runtime conversion specification][oci-image-conversion]. |
+| `org.opencontainers.image.stopSignal`   | Indicates signal that SHOULD be sent by the container runtimes to [kill the container](runtime.md#kill). The annotation value MUST have a valid value for the `config.StopSignal` property as defined in [the OCI image specification][oci-image-config-properties]. This annotation SHOULD only be used in accordance with the [OCI image specification's runtime conversion specification][oci-image-conversion]. |
+
+All other keys in the `org.opencontainers` namespace not specified in this above table are reserved and MUST NOT be used by subsequent specifications.
```
tianon, line 665: "subsequent specifications" here seems a little weird - that almost sounds like it's saying we won't use any others in the future, which I don't think is a promise we should make :shrug:

cyphar: This is the wording in the current spec so I tried to mirror it, but I agree that it sounds very wrong. I suspect that "subsequent specifications" is meant to mean "other specifications" or "other organisations" (I haven't looked at the git history but it's possible this comes from when we thought AppC might end up being an extension of runtime-spec.)

I'll figure out some nicer text.

AkihiroSuda (approved):

utam0k (approved):

Issue #1216, opened 2023-08-05
idmapped mounts: should they be applied recursively?


cyphar:
The current text for idmapped mounts doesn't specify whether the attributes should be applied recursively or not (`AT_RECURSIVE`). At the moment, runc never passes `AT_RECURSIVE`. The two options I can see are:

 * Make `rbind` imply that the attribute should be recursive, since that's the only case where `AT_RECURSIVE` is necessary. (This would imply that it's always recursive.) The main downside is that a user wouldn't be able to opt-out of it, and runc's current behaviour would be spec in-compliant. But I suspect this is what most users would expect.
 * Make it configurable with a new mount option (`ridmap`, maybe) which can only be set if you have the `uidMappings` and `gidMappings` options enabled. The main downside is that the current `rbind` (which is counter-intuitive imho) would remain, but that would avoid any possible backwards-compatibility issues with runtime-spec 1.1.0.

PR #1219, opened 2023-08-21
features-linux: Expose idmap information

rata:
High level container runtimes sometimes need to know if the OCI runtime supports idmap mounts or not, as the OCI runtime silently ignores unknown fields.

This means that if it doesn't support idmap mounts, a container with userns will be started, without idmap mounts, and the files created on the volumes will have a "garbage" owner/group. Furthermore, as the userns mapping is not guaranteed to be stable over time, it will be completely unusable.

Let's expose idmap support in the features subcommand, so high level container runtimes use the feature safely.

---
cc @giuseppe @AkihiroSuda

### DIFF
--- a/features-linux.md
+++ b/features-linux.md
@@ -209,3 +209,22 @@ Irrelevant to the availability of Intel RDT on the host operating system.
   "enabled": true
 }
 ```
+
+## <a name="linuxFeaturesMountExtensions" />MountExtensions
+
+**`mountExtensions`** (object, OPTIONAL) represents whether the runtime supports certain mount features, irrespective of the availability of the features on the host operating system.
+
+* **`idmap`** (object, OPTIONAL) represents whether the runtime supports idmap mounts using the `uidMappings` and `gidMappings` properties of the mount.
+  * **`enabled`** (bool, OPTIONAL) represents whether the runtime parses and attempts to use the `uidMappings` and `gidMappings` properties of mounts if provided.
+    Note that it is possible for runtimes to have partial implementations of id-mapped mounts support (such as only allowing mounts which have mappings matching the container's user namespace, or only allowing the id-mapped bind-mounts).
+    In such cases, runtimes MUST still set this value to `true`, to indicate that the runtime recognises the `uidMappings` and `gidMappings` properties.
+
+### Example
+
+```json
+"mountExtensions": {
+  "idmap":{
+    "enabled": true
+  }
+}
+```
--- a/specs-go/features/features.go
+++ b/specs-go/features/features.go
@@ -36,11 +36,12 @@ type Linux struct {
 	// Nil value means "unknown", not "no support for any capability".
 	Capabilities []string `json:"capabilities,omitempty"`
 
-	Cgroup   *Cgroup   `json:"cgroup,omitempty"`
-	Seccomp  *Seccomp  `json:"seccomp,omitempty"`
-	Apparmor *Apparmor `json:"apparmor,omitempty"`
-	Selinux  *Selinux  `json:"selinux,omitempty"`
-	IntelRdt *IntelRdt `json:"intelRdt,omitempty"`
+	Cgroup          *Cgroup          `json:"cgroup,omitempty"`
+	Seccomp         *Seccomp         `json:"seccomp,omitempty"`
+	Apparmor        *Apparmor        `json:"apparmor,omitempty"`
+	Selinux         *Selinux         `json:"selinux,omitempty"`
+	IntelRdt        *IntelRdt        `json:"intelRdt,omitempty"`
+	MountExtensions *MountExtensions `json:"mountExtensions,omitempty"`
 }
 
 // Cgroup represents the "cgroup" field.
@@ -120,6 +121,19 @@ type Selinux struct {
 type IntelRdt struct {
 	// Enabled is true if Intel RDT support is compiled in.
 	// Unrelated to whether the host supports Intel RDT or not.
+	// Nil value means "unknown", not "false".
+	Enabled *bool `json:"enabled,omitempty"`
+}
+
+// MountExtensions represents the "mountExtensions" field.
+type MountExtensions struct {
+	// IDMap represents the status of idmap mounts support.
+	IDMap *IDMap `json:"idmap,omitempty"`
+}
+
+type IDMap struct {
+	// Enabled represents whether idmap mounts supports is compiled in.
+	// Unrelated to whether the host supports it or not.
 	// Nil value means "unknown", not "false".
 	Enabled *bool `json:"enabled,omitempty"`
 }
--- a/schema/features-linux.json
+++ b/schema/features-linux.json
@@ -97,6 +97,19 @@
                         "type": "boolean"
                     }
                 }
+            },
+            "mountExtensions": {
+                "type": "object",
+                "properties": {
+                    "idmap": {
+                        "type": "object",
+                        "properties": {
+			    "enabled": {
+                                "type": "boolean"
+                            }
+                       }
+                    }
+                }
             }
         }
     }


giuseppe (approved):
LGTM

specs-go/features/features.go
```diff
@@ -36,11 +36,12 @@ type Linux struct {
...
-	Seccomp  *Seccomp  `json:"seccomp,omitempty"`
-	Apparmor *Apparmor `json:"apparmor,omitempty"`
-	Selinux  *Selinux  `json:"selinux,omitempty"`
-	IntelRdt *IntelRdt `json:"intelRdt,omitempty"`
+	Cgroup          *Cgroup          `json:"cgroup,omitempty"`
+	Seccomp         *Seccomp         `json:"seccomp,omitempty"`
+	Apparmor        *Apparmor        `json:"apparmor,omitempty"`
+	Selinux         *Selinux         `json:"selinux,omitempty"`
+	IntelRdt        *IntelRdt        `json:"intelRdt,omitempty"`
+	MountExtensions *MountExtensions `json:"mountExtensions,omitempty"`
```
AkihiroSuda, line 44: `MountOptions` is in the `Features` struct, so `MountExtensions` should follow there too?

AkihiroSuda: (But IDMap is Linux-specific, so the current PR might be good as-is)

AkihiroSuda (approved):
Thanks

features-linux.md
```diff
@@ -209,3 +209,17 @@ Irrelevant to the availability of Intel RDT on the host operating system.
   "enabled": true
 }
 ```
+
+## <a name="linuxFeaturesMountExtensions" />MountExtensions
+
+**`mountExtensions`** (object, OPTIONAL) represents the runtime's implementation status of different mount features.
+
+* **`idmap`** (bool, OPTIONAL) represents whether the runtime supports idmap mounts using the uidMappings and gidMappings properties of the mount.
```
cyphar, line 217: I don't think a boolean is sufficient to describe the state of idmapped mount support.

At the moment, runc only supports this for one specific case (which will be fixed by opencontainers/runc#3985). Once we fix this, runc will only support these flags for bind mounts. However, the specification allows you to set these fields for *any* mount, which runc doesn't currently support (because it would require reworking a lot of our mount infrastructure). I don't know if crun supports arbitrary mounts to have this flag added either (@giuseppe?). Same for youki (@utam0k?).

The problem with a boolean is that it's not clear to me whether this should be `true` or `false` in any of these cases. Does `true` mean that setting the idmapping flags is supported at all (even with restrictions), or that it is fully supported?

I don't know if there is a nice way to describe all of these cases. You could have a string saying which filesystem types are supported, but bind-mounts are not treated as a filesystem type by the spec...

cyphar (changes requested):

features-linux.md
AkihiroSuda: What about having two flags: `"mountExtensions": ["idmapFixed", "idmapFlexible"]`

cyphar: That would work for the level of idmapped supports, but for mount types I think we will need something as well.

For instance, in runc we would need to explicitly work to add support for `tmpfs` because we have special handling for `tmpcopyup` (as well as `proc`, `sys`, `cgroup` -- but none of those filesystems have `FS_ALLOW_IDMAP` so it's not super important).

Now that I think about it, we could just have a list of supported filesystem types (the runtime could use `/proc/filesystems` and filter out filesystems they need special handling for, or just have a fixed list they support). For bind-mounts we might be able have `idmapBind` as an extension? Or we could add `bind` as a filesystem type...

cyphar: Maybe this could be an object like:

```
"mountExtensions": {
  "idmap": {
    "modes": ["sameAsUserns", "arbitrary"],
    "filesystemTypes": ["ext4", "tmpfs", "foobarfs"],
    "supportsBind": true,
  }
}
```

wdyt?

Also we need to figure out how to deal with recursive mounts -- see #1216.

AkihiroSuda: SGTM, but probably `s/Userns/UserNS/`, as `Userns` may look like `Users`

giuseppe: > I don't know if crun supports arbitrary mounts to have this flag added either (@giuseppe?)

crun supports arbitrary mappings specified.  If its mappings are different than the target user namespace mappings then a temporary user namespace is created

specs-go/features/features.go
rata: Yes, this is Linux-only so IMHO where it is is better

features-linux.md
rata: @cyphar I can make this an object, so we can extend it in the future with whatever we need. But I don't think we need to expose any of that now.

No user needs that information now. All we need to know is if the OCI runtime supports idmap mounts, if it does, then we can trust it will throw an error if the idmap config requested can't be setup (due to runtime limitations or host limitations, etc.).

Also, I don't think exposing the fs is a good idea: the features section is set at compile time, not runtime, so the user can upgrade the kernel, have a new fs that supports `FS_ALLOW_IDMAP` and use it in a bind mount just fine. But the `filesystems` field reported will be inaccurate.

If we _really_ need to expose some of that info, we can expose maybe the negation (unsupported fs by the runtime, because we know we need work to do to support it, even if the kernel supports it). But I would not complicate things with stuff no one yet wants, we can add it later when we have an use case, and we will have a better understanding of how exposing that information makes sense for the user.

utam0k: > At the moment, runc only supports this for one specific case (which will be fixed by https://github.com/opencontainers/runc/pull/3985). Once we fix this, runc will only support these flags for bind mounts. However, the specification allows you to set these fields for any mount, which runc doesn't currently support (because it would require reworking a lot of our mount infrastructure). I don't know if crun supports arbitrary mounts to have this flag added either (@giuseppe?). Same for youki (@utam0k?).

As for youki, we haven't supported it yet, so you don't have to worry about youki's implementation. Thanks for considering youki.

cyphar: @giuseppe Sorry, by "arbitrary mounts", I mean do you support doing idmapped mounts for non-bindmounts? (In other words, do you use `fsopen`/`fsconfig` to configure mounts and apply idmapped mounts to them?

@rata 

> No user needs that information now. All we need to know is if the OCI runtime supports idmap mounts, if it does, then we can trust it will throw an error if the idmap config requested can't be setup (due to runtime limitations or host limitations, etc.).

If a user wants to configure a mount with custom mappings, `"idmap": true` doesn't give them enough information to know what exactly is supported. They will have to just try the config and see if it fails. To be fair, the fact this is silently ignored is a pretty serious issue (in fact, I think this would warrant us requiring users to specify an "idmap" mount option -- this would also allow us to know if the mapping request is recursive with "ridmap" which would fix #1216 as well... I'll open a PR for that.).

> Also, I don't think exposing the fs is a good idea: the features section is set at compile time, not runtime, so the user can upgrade the kernel, have a new fs that supports FS_ALLOW_IDMAP and use it in a bind mount just fine. But the filesystems field reported will be inaccurate.

I wouldn't expect runtimes to know what filesystems support FS_ALLOW_IDMAP. That wasn't what I was suggesting. The idea would be that runtimes which support setting up all mounts with `fsconfig` and `mount_setattr` would take the output of `/proc/filesystems` and remove filesystems that the runtime knows they cannot support (because they need special handling). If the underlying filesystem doesn't support the feature, the container setup will fail -- but that is how every other runtime-spec feature works as well.

We could invert the list so the runtime just specifies what filesystems they don't support. However, at the moment runc doesn't support doing this for _any_ filesystems so we would need to output the contents of `/proc/filesystems` as well. I think having a "here are the ones we support" list is more in the spirit of the `features` list (we don't have any negative feature lists ATM IIRC) -- but since the two approaches are equivalent I don't really care either way.

> But I would not complicate things with stuff no one yet wants, we can add it later when we have an use case, and we will have a better understanding of how exposing that information makes sense for the user.

We can't change the type of this entry once we've picked it. Since we expect to expand it in the future, it should be a struct. I disagree there isn't a clear usecase -- the point of `<runtime> features` is to allow users to know what is supported without doing things through trial-and-error -- given the various hurdles to implementing complete idmapped mounts support, providing this kind of information to users makes pretty clear sense to me.

rata: @cyphar 

> If a user wants to configure a mount with custom mappings, `"idmap": true` doesn't give them enough information to know what exactly is supported. They will have to just try the config and see if it fails

Yes, that is what I said.

> To be fair, the fact this is silently ignored is a pretty serious issue

What exactly is silently ignored and with which runc version(s)?

> We can't change the type of this entry once we've picked it. Since we expect to expand it in the future, it should be a struct.

That is exactly what I did, for this to be extensible. Can you have another look? :)

> the point of `<runtime> features` is to allow users to know what is supported without doing things through trial-and-error -- given the various hurdles to implementing complete idmapped mounts support, providing this kind of information to users makes pretty clear sense to me.

Sure, adding fields is backwards-compatible and we can add more fields to provide more information. Why would we not merge this, if this is extensible, and add more fields in other PRs?

cyphar: > > To be fair, the fact this is silently ignored is a pretty serious issue

> What exactly is silently ignored and with which runc version(s)?

I was referring to the issue you mention in the PR description -- the new fields are ignored on older runc versions (by design) and thus you end up with misconfigured mounts.

> That is exactly what I did, for this to be extensible. Can you have another look? :)

Ah, you pushed a new version since my last comment. My bad. :sweat_smile:

rata: @cyphar Also, crun is using open_tree(), mount_setattr() and move_mount(); not fsconfig, for idmap mounts.

giuseppe: > @giuseppe Sorry, by "arbitrary mounts", I mean do you support doing idmapped mounts for non-bindmounts? (In other words, do you use `fsopen`/`fsconfig` to configure mounts and apply idmapped mounts to them? Can I create a `type: tmpfs` mount with an id mapping?)

no, that is not supported at the moment.  idmapped mounts work only with bind mounts

features-linux.md
```diff
@@ -209,3 +209,21 @@ Irrelevant to the availability of Intel RDT on the host operating system.
   "enabled": true
 }
 ```
+
+## <a name="linuxFeaturesMountExtensions" />MountExtensions
+
+**`mountExtensions`** (object, OPTIONAL) represents the runtime's implementation status of different mount features.
+Irrelevant to the availability of the features on the host operating system.
```
cyphar, line 216:
```suggestion
**`mountExtensions`** (object, OPTIONAL) represents whether the runtime supports certain mount features, irrespective of the availability of the features on the host operating system.
```

features-linux.md
```diff
@@ -209,3 +209,21 @@ Irrelevant to the availability of Intel RDT on the host operating system.
...
+
+* **`idmap`** (object, OPTIONAL) represents whether the runtime supports idmap mounts using the uidMappings and gidMappings properties of the mount.
```
cyphar, line 218:
```suggestion
* **`idmap`** (object, OPTIONAL) represents whether the runtime supports idmap mounts using the `uidMappings` and `gidMappings` properties of the mount.
```

features-linux.md
```diff
@@ -209,3 +209,21 @@ Irrelevant to the availability of Intel RDT on the host operating system.
...
+  * **`enabled`** (bool, OPTIONAL) represents whether the feature is enabled.
```
cyphar, line 219: This needs to have a note saying that "enabled" may mean that the runtime has restrictions on its use. Something like:

```suggestion
  * **`enabled`** (bool, OPTIONAL) represents whether the runtime parses and attempts to use the `uidMappings` and `gidMappings` properties of mounts if provided.
    Note that it is possible for runtimes to have partial implementations of id-mapped mounts support (such as only allowing mounts which have mappings matching the container's user namespace, or only allowing the id-mapped bind-mounts).
    In such cases, runtimes MUST still set this value to `true`, to indicate that the runtime recognises the `uidMappings` and `gidMappings` properties.
```

cyphar (changes requested):
Aside from these changes, looks good.

features-linux.md
rata: Fixed, thanks

features-linux.md
rata: Fixed, thanks

features-linux.md
rata: Fixed, thanks

features-linux.md
rata: IIUC what is relevant for this PR is solved. Otherwise, please feel free to re-open.

cyphar: I'll keep the conversation as "unresolved" so that it's not hidden when you open the PR after it's been merged. I can open a PR to discuss additional fields we might need for users to fully understand what features are supported.

cyphar (approved):
LGTM, thanks!

features-linux.md
```diff
@@ -209,3 +209,22 @@ Irrelevant to the availability of Intel RDT on the host operating system.
...
 ```
+
+## <a name="linuxFeaturesMountExtensions" />MountExtensions
+
+**`mountExtensions`** (object, OPTIONAL) represents whether the runtime supports certain mount features, irrespective of the availability of the features on the host operating system.
+
+* **`idmap`** (object, OPTIONAL) represents whether the runtime supports idmap mounts using the `uidMappings` and `gidMappings` properties of the mount.
+  * **`enabled`** (bool, OPTIONAL) represents whether the runtime parses and attempts to use the `uidMappings` and `gidMappings` properties of mounts if provided.
+    Note that it is possible for runtimes to have partial implementations of id-mapped mounts support (such as only allowing mounts which have mappings matching the container's user namespace, or only allowing the id-mapped bind-mounts).
+    In such cases, runtimes MUST still set this value to `true`, to indicate that the runtime recognises the `uidMappings` and `gidMappings` properties.
```
AkihiroSuda, line 220: How to detect whether the runtime supports arbitrary mapping?

rata: @AkihiroSuda We can add other fields for that.

I'm not convinced it is useful, though. From the high-level container runtime, it is not that you will choose different mappings if that is supported or not. You need the mappings you need, and all you care about is that runc will not just ignore the setting and create a big mess (as files in volumes will be owned by the hostUID/GID, etc.).

If that is not supported, what you want is runc to throw an error, not discover it via features IMHO. So, I don't see why exposing that would be useful.

cyphar: Knowing the support level is useful for the same reason the rest of the features subcommand is useful -- it means you don't have to trial-and-error test which features are available.

@AkihiroSuda I will open an issue where we can flesh out which extra fields we want (based on the above discussion). I felt that merging this as-is is okay, as we can discuss the details of extending it separately.

rata: @cyphar can you cc me?

rata: Oh, it is already open

AkihiroSuda:
@cyphar Next time can we use the merge button on the GitHub web UI?
On the web UI this PR is marked as "closed" and caused a confusion to me.

cyphar:
Putting "closed" in the merge commit used to not cause issues, it seems GitHub has changed something. I will avoid putting it in the future.

I prefer merging from the cli because my merge commits are signed that way.

Issue #1220, opened 2023-08-21
dev versions don't respect semver


rata:
When we switch to dev after a release, we are not respecting semver (as defined in semver.org version 2.0.0). The issue is that we just change the [patchDev](https://github.com/opencontainers/runtime-spec/commit/4ee185a47ad5caf0ef82d45e78ddfe6587687ad5) to "-dev", without changing the minor or anything else.

After release x.y.0, we almost always switch that field to "-dev". The issue is that according the spec, x.y.0-dev is a lower version than x.y.0. Therefore, we should increase _some_ other component while switching to "-dev". Like x.y.1-dev.

This is not a big issue generally, as this dev versions are not included in releases, but it become an issue since we expose this in "runc features". This is incovenient, as using modules like https://pkg.go.dev/golang.org/x/mod/semver will not give the desired result when we compare versions (with our versioning, "1.0.2-dev" is greater than "1.0.2", but not with that module that correctly implements the spec).

Looking at the git history, we've almost always done that, the only exception is recent: https://github.com/opencontainers/runtime-spec/commit/3f552ce17e77ef28a21764f7c499b555517afa43 (what is currently on main). The `+` sign should be used to indicate a build, which is incorrect here.

In fact, if we look at the go.mod in runc release-1.1, it shows this that is in sync with what I propose in the next paragraph:
```
        github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
```

I propose the following:
 1. Switch main from 1.1.0+dev to 1.1.1-dev (respects semver and is a greater version than 1.1.0)
 1. The feature in runc 1.1 is experimental, so just in the next 1.1 patch release, change the version reported from "1.0.2-dev" to "1.0.3-dev" (hardcode the string in the release-1.1 branch, that is simple and should do it).
 1. Unsure how to add a test case that will enforce this, but at least add a big comment in the specs-go/version.go file about it.

Note there is nothing to do for crun (it implemented it in 1.8.6, that uses an rc runtime spec release all good in 1.8.7 and main)

---
cc @giuseppe @AkihiroSuda

cyphar:
The reasoning behind this change is in #1198. Changing back to `-dev` is not acceptable. Please read through the discussion there and if you feel there is a better solution, please propose that instead.

[...]

PR #1221, opened 2023-08-21
specs-go: Respect semver 2.0.0

rata:
According to semver.org, the 2.0.0 spec, the pre-release info should be delimited by a "-" ("+" is for build info) and we already released 1.1.0, so increment the minor so this version compares to something bigger than 1.1.0.

Updates: #1220

### DIFF
--- a/specs-go/version.go
+++ b/specs-go/version.go
@@ -8,10 +8,10 @@ const (
 	// VersionMinor is for functionality in a backwards-compatible manner
 	VersionMinor = 1
 	// VersionPatch is for backwards-compatible bug fixes
-	VersionPatch = 0
+	VersionPatch = 1
 
 	// VersionDev indicates development branch. Releases will be empty string.
-	VersionDev = "+dev"
+	VersionDev = "-dev"
 )
 
 // Version is the specification version that the package types support.


thaJeztah:
- Ah, I recall this was changed in https://github.com/opencontainers/runtime-spec/pull/1198#issuecomment-1517831399
- And there was some discussion on this; on that PR: https://github.com/opencontainers/runtime-spec/pull/1198#issuecomment-1517831399

/cc @cyphar

thaJeztah:
Note that `v1.0.1-dev` may not be a good choice either, as it would rank _higher_ than `v1.0.1-alpha.0` (or beta, etc).

One option could be to use `v1.0.1-alpha.0`, and create a tag after the (non-alpha) release is done. 


In all cases, it's a bit of a "chicken-and-egg" situation, because SemVer versions are based on the content of the changes, and the content of changes is not yet known until an actual release is prepared (update Major, Minor or Patch, depending on the changes that are included).

cyphar:
1. SemVer uses MAY in the text describing pre-release versions: "A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version." As per RFC 2119, not following this is not a spec violation. It is not a "should".
2. SemVer is concerned with versions that are actually shipped. We use `+dev` for the HEAD of our branch, which is not shipped. As far as I'm concerned, SemVer has nothing to say about this -- we could just as easily say that the version is "DO NOT USE" for non-released versions because it is not released.
3. Switching away from `-dev` was intentional. As mentioned in the original PR which changed this, the old behaviour was suboptimal and needed to be changed. For one, `-dev` ranks higher than `-alpha` and `-beta`. It also has the frustrating behaviour of making emergency updates cause issues with spec validation tools because the version (according to SemVer) is different -- there is no ideal solution for the spec validation problem, but `-dev` is the worst solution. The only alternative solution that seems reasonable is to switch to the Go model (`1.1.1-0-unreleased`) but this has other downsides.

cyphar (changes requested):
nack

giuseppe (approved):
LGTM

rata:
@cyphar 
1. And why you mention that? It is completely unrelated to what I say. Am I missing something? I just say we treat "1.0.2-dev" as older than "1.0.2", but on semver it isn't. I'm not saying something break without a pre-release info.
1. But as I said, this is exposed to users: `runc features` exposed version `1.0.2-dev`.

@thaJeztah thanks for the links, I've missed that. And I had a typo in my example when comparing "-dev" versions with "alpha.1", so I thought the comparisson worked fine, when it didn't :-D

rata:
@cyphar 
3. Why not use `1.1.1-0.unreleased` then? This compares to something bigger than 1.1.0, so users can see that is newer and it is smaller than all the pre-releases we might make (1.1.1-alpha, -alpha.1, beta, or even 1.2.0, all works fine)

I don't think using an unreleased version creates a _new_ problem for runtime-tools. It **only** supports to validate [the current spec-version vendored](https://github.com/opencontainers/runtime-tools/blob/e931285f4b699ed577fca8327cbb3c51954aac2a/validate/validate.go#L260-L262), so if the exact same version is vendored, it will work. If it isn't, it won't (same as with the curren "-dev" we have today).

rata:
Now that we merged the way to expose the features, this is not so important to me. If indeed exposing in runc different specs but whose versions compare to equal is problematic (I wouldn't be surprised if it is), we can revisit this decision.

PR #1222, opened 2023-08-22
config: add idmap and ridmap mount options

### LINKED ISSUE #1216: idmapped mounts: should they be applied recursively?

cyphar:
Adding new fields for MOUNT_ATTR_IDMAP had the flaw that users
specifying these fields with older runtimes would result in the fields
being ignored and incorrect mounts being configured. In addition, there
is no text in the specification indicating whether MOUNT_ATTR_IDMAP
should be applied with AT_RECURSIVE (which matters for rbind idmapped
mounts).

In retrospect, the addition of the fields should've included new (dummy)
mount options that would cause errors on older runtimes. Unfortunately,
we have had a runtime-spec release since then so we cannot MUST these
new mount options, but we can SHOULD them.

Fixes #1216
Resolves the main issue motivating the urgency of #1219 (though that feature is still something we should have).

Fixes: https://github.com/opencontainers/runtime-spec/commit/9d1130dc3bec2d2aabeb166f12e0ca027c3f404d ("IDMapping field for mount point")

### DIFF
--- a/config.md
+++ b/config.md
@@ -180,10 +180,12 @@ For POSIX platforms the `mounts` structure has the following fields:
     * Linux: filesystem types supported by the kernel as listed in */proc/filesystems* (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). For bind mounts (when `options` include either `bind` or `rbind`), the type is a dummy, often "none" (not listed in */proc/filesystems*).
     * Solaris: corresponds to "type" of the fs resource in [zonecfg(1M)][zonecfg.1m].
 * **`uidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert UIDs from the source file system to the destination mount point.
+  This SHOULD be implemented using [`mount_setattr(MOUNT_ATTR_IDMAP)`][mount_setattr.2], available since Linux 5.12.
   If specified, the `options` field of the `mounts` structure SHOULD contain either `idmap` or `ridmap` to specify whether the mapping should be applied recursively for `rbind` mounts, as well as to ensure that older runtimes will not silently ignore this field.
   The format is the same as [user namespace mappings](config-linux.md#user-namespace-mappings).
   If specified, it MUST be specified along with `gidMappings`.
 * **`gidMappings`** (array of type LinuxIDMapping, OPTIONAL) The mapping to convert GIDs from the source file system to the destination mount point.
+  This SHOULD be implemented using [`mount_setattr(MOUNT_ATTR_IDMAP)`][mount_setattr.2], available since Linux 5.12.
   If specified, the `options` field of the `mounts` structure SHOULD contain either `idmap` or `ridmap` to specify whether the mapping should be applied recursively for `rbind` mounts, as well as to ensure that older runtimes will not silently ignore this field.
   For more details see `uidMappings`.
   If specified, it MUST be specified along with `uidMappings`.
--- a/config.md
+++ b/config.md
@@ -146,6 +146,8 @@ Runtimes MUST/SHOULD/MAY implement the following option strings for Linux:
  `sync`           | MUST        | [^1]
  `tmpcopyup`      | MAY         | copy up the contents to a tmpfs
  `unbindable`     | MUST        | [^2] (bind mounts)
+ `idmap`          | SHOULD      | Indicates that the mount has `uidMappings` and `gidMappings` specified. This option SHOULD NOT be passed to the underlying [`mount(2)`][mount.2] call. If supported, the runtime MUST return an error if this option is provided and either of `uidMappings` or `gidMappings` are empty or not present.
+ `ridmap`         | SHOULD      | Indicates that the mount has `uidMappings` and `gidMappings` specified, and the mapping is applied recursively [^3]. This option SHOULD NOT be passed to the underlying [`mount(2)`][mount.2] call. If supported, the runtime MUST return an error if this option is provided and either of `uidMappings` or `gidMappings` are empty or not present.
 
 [^1]: Corresponds to [`mount(8)` (filesystem-independent)][mount.8-filesystem-independent].
 [^2]: Corresponds to [`mount(8)` (filesystem-specific)][mount.8-filesystem-specific].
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -146,8 +146,8 @@ Runtimes MUST/SHOULD/MAY implement the following option strings for Linux:
...
- `ridmap`         | SHOULD      | Indicates that the mount has `uidMappings` and `gidMappings` specified, and the mapping is applied recursively [^3]. This option SHOULD NOT be passed to the underlying [`mount(2)`] call. If supported, the runtime MUST return an error if this option is provided and either of `uidMappings` or `gidMappings` are empty or not present.
+ `idmap`          | SHOULD      | Indicates that the mount has `uidMappings` and `gidMappings` specified. This option SHOULD NOT be passed to the underlying [`mount(2)`][mount.2] call. If supported, the runtime MUST return an error if this option is provided and either of `uidMappings` or `gidMappings` are empty or not present.
+ `ridmap`         | SHOULD      | Indicates that the mount has `uidMappings` and `gidMappings` specified, and the mapping is applied recursively [^3]. This option SHOULD NOT be passed to the underlying [`mount(2)`][mount.2] call. If supported, the runtime MUST return an error if this option is provided and either of `uidMappings` or `gidMappings` are empty or not present.
```
rata, line 150: The mount link is being added in the wrong commit

rata (approved):
LGTM

config.md
cyphar: Already fixed :wink:

cyphar:
@rata The error message is fairly ugly, but the mount will fail because no filesystem supports the flag "idmap" nor "ridmap".

... Oh actually, now that I think about it, I think this flag will be ignored for bind-mounts due to how they are implemented in runc (extra mount options for bind-mounts are ignored because that's what the kernel does). Grrrrrrr..........

rata (changes requested):
Please check this does old runtimes fail (runc and crun at least) and post the error it shows here, so we can easily know the impact of this change.

cyphar:
This is the error you get if you add it to a regular mount (for runc):

```
ERRO[0000] runc run failed: unable to start container process: error during container init: error mounting "cgroup" to rootfs at "/sys/fs/cgroup": mount src=cgroup, dst=/sys/fs/cgroup, dstFd=/proc/self/fd/8, flags=0x20000f, data=idmap: invalid argument
```

But for bind-mounts there's no error.

rata:
Then this is not useful as-is, at least :(

cyphar:
I think we need `ridmap` anyway (because of #1216). I've added a patch to opencontainers/runc#3967 (`configs: validate: add validation for bind-mount fsflags`) which will cause runc to detect this properly:

```
ERRO[0000] runc run failed: invalid mount &{Source:/tmp Destination:foo Device:bind Flags:2101263 ClearedFlags:0 PropagationFlags:[] Data:idmap Relabel: RecAttr:<nil> Extensions:0 UIDMappings:[] GIDMappings:[]}: bind mounts cannot have any filesystem-specific options applied
```

Unfortunately this doesn't help with old versions of runc (though the same is true for #1219 because we need to update runtimes to say whether they support idmapped mounts), and I suspect other runtimes have the same behaviour because it's easy to miss this.

@giuseppe @utam0k How do crun and youki handle data flags being passed to bind-mounts? Do you return errors or are the flags silently ignored? (Try adding `idmap` to a bind-mount options list.)

AkihiroSuda:
Adding mount options which do NOP by themselves looks clumsy.

Why not just use the `mountExtensions` struct?

>  Maybe this could be an object like:
> 
> ```
> "mountExtensions": {
>   "idmap": {
>     "modes": ["sameAsUserns", "arbitrary"],
>     "filesystemTypes": ["ext4", "tmpfs", "foobarfs"],
>     "supportsBind": true
>   }
> }
> ```
> 
> wdyt?
> 
> Also we need to figure out how to deal with recursive mounts -- see #1216. If we add a separate flag for this, we will need to add an entry to this struct.
> 
> _Originally posted by @cyphar in https://github.com/opencontainers/runtime-spec/pull/1219#discussion_r1301251621_
>

AkihiroSuda (changes requested):
I'd prefer `mountExtensions` struct

https://github.com/opencontainers/runtime-spec/pull/1222#issuecomment-1688634840

cyphar:
@AkihiroSuda The features proposal doesn't help with:

* Indicating whether the mapping should be applied with `AT_RECURSIVE`. This requires some kind of flag on the mount.
* As far as I can tell, we don't have any other runtime-spec features that require you to check the features list in order to avoid a setting being silently ignored. The annotations stuff you're proposing is in the same ballpark but I think there's no other way to deal with annotations -- for mounts we have a specific way of specifying what features we need.

Also, the flags aren't NOP flags -- they indicate you need to do a certain mount operation (just like "rro" and the other recursive mount settings).

That being said, I think we need the features addition as well.

rata:
@cyphar 

> * As far as I can tell, we don't have any other runtime-spec features that require you to check the features list in order to avoid a setting being silently ignored

I guess there is some other. By design, [runtimes MUST ignore unknown fields](https://github.com/opencontainers/runtime-spec/blob/e8c413418efcb816b963be9762ddc202e181ba88/config.md#extensibility). Probably in practice most of the stuff is supported for years, so not a real issue, but some example probably exists (probably also not with such awful consequences as this one).

cyphar:
Hmm, now that I think about it some more, that's a good point @rata...

FWIW, I'm starting to come to the view that the way we planned for extensibility in the spec was a bit misguided -- runtimes do not reject configurations from specification versions they do not support, which is behaviour that just begging to cause a security issue in the future. The purpose of the "runtimes MUST ignore unknown fields" text was to allow for vendor extensions (I think this was also motivated by the fact that the Go stdlib JSON parser didn't have a way to block unknown fields when we first started working on the spec -- `DisallowUnknownFields` was added in Go 1.10 in 2018 -- which was after the runtime-spec 1.0 release in 2017).

We should've caught this issue when reviewing the idmapped mounts additions to the spec...

All of that being said, we need `ridmap` to resolve #1216 and if we have `ridmap` I don't see any downside to having `idmap` as well.

rata:
@cyphar I fully agree with that. I did a patch to runc, to fail on unknown fields, but never opened a PR as I found that note about extensibility. I completely share that that note creates more issues than it solves.

There might be some "escapes" (besides going for v2.0.0, that I don't think is a good idea), like adding a `failOnUnknownFields: true` somewhere in the config.json, that high-level runtimes could set (maybe also even query support for that using the features thing) and some years in the future, when all used OCI runtimes support that flag, then that would be universal. At that point, maybe high-level runtimes can just fail if that feature is not implemented, and not even worry about providing a fall-back to be safe in the face of OCI runtime unknown features.

That is a longer conversation and I'm not 100% sure what my thoughts are about it now.

AkihiroSuda (approved):

giuseppe:
> @giuseppe @utam0k How do crun and youki handle data flags being passed to bind-mounts? Do you return errors or are the flags silently ignored? (Try adding `idmap` to a bind-mount options list.)

crun for bind mounts would just ignore a flag it doesn't know about.

crun already supports `idmap` with a slightly different meaning though.  If it is specified then it uses the same mappings used for the container user namespace.  I was playing with it before `uidMappings/gidMappings` were added to the OCI specs: https://github.com/containers/crun/pull/780

rata (approved):
LGTM. But the PR description is out of date (it doesn't mitigate the urgency of the other feature, this doesn't fail on old runtimes as we first thought).

utam0k:
> @giuseppe @utam0k How do crun and youki handle data flags being passed to bind-mounts? Do you return errors or are the flags silently ignored? (Try adding idmap to a bind-mount options list.)

`Youki` ignores a flag that we don't know about.
https://github.com/containers/youki/blob/f2d95ca41c1d2c83af2f87e28b31c77c835dfad6/crates/libcontainer/src/rootfs/utils.rs#L80-L121

utam0k (approved):

AkihiroSuda:
> crun already supports idmap with a slightly different meaning though.

This PR had to be updated to avoid conflict with crun?

utam0k:
Sorry for missing it. I thought it was fine, especially since there was no objection, but I should have checked. 

@giuseppe The names of the mount options in crun seem to overlap, is that OK?
> If the idmap option is specified then the mount is ID mapped using the container target user namespace. This is an experimental feature and can change at any time without notice.
https://github.com/giuseppe/crun/blob/827b8731899d4febea3f27907bcde5e6065bb65b/crun.1.md#idmap-mount-options

cyphar:
We will need to change the MUST language to allow crun's behaviour. In fact, maybe we should make crun's behaviour the standard behaviour (it seems reasonable to me). I'm on vacation for the next two weeks, if someone else wants to pick this up in the meantime I'd appreciate it 😅

AkihiroSuda:
↑ @giuseppe WDYT? 🙏

giuseppe:
what would be the best wording to allow the crun behaviour?  Just dropping the following sentence?

```
If supported, the runtime MUST return an error if this option is provided and either of `uidMappings` or `gidMappings` are empty or not present.
```

giuseppe:
opened a PR, we can discuss there how it should be: https://github.com/opencontainers/runtime-spec/pull/1224

PR #1230, opened 2023-09-28
config-linux: add schemata field to IntelRdt

marquiz:
Add a new "schemata" field to the Linux IntelRdt configuration. This addresses the complexity of separate schema fields and resolves the issue of supporting currently uncovered RDT features like L2 cache allocation and CDP (Code and Data Prioritization).

The new field is for specifying the complete schemata (all schemas) to be written to the schemata file in Linux resctrl fs. The aim is for simple usage and runtime implementation (by not requiring any parsing/filtering of data or otherwise re-implement parsing or validation of the Linux resctrl interface) and also to support all RDT features now and in the future (i.e. schemas like L2, L2CODE, L2DATA, L3CODE and L3DATA and who knows L4 or something else in the future).

Behavior of existing fields is not changed but it is required that the new schemata field is applied last.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -645,6 +645,7 @@ The following parameters can be specified for the container:
     The value SHOULD start with `L3:` and SHOULD NOT contain newlines.
 * **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth per L3 cache id.
     The value MUST start with `MB:` and MUST NOT contain newlines.
+* **`schemata`** *(array of strings, OPTIONAL)* - specifies the schemata to be written to the `schemata` file in resctrlfs. Each element represents one line in the `schemata` file. The value MUST NOT contain newlines.
 
 The following rules on parameters MUST be applied:
 
@@ -654,15 +655,17 @@ The following rules on parameters MUST be applied:
 
 * If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the that sub-directory discussed in `closID`.
 
-* If neither `l3CacheSchema` nor `memBwSchema` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
+* If `schemata` field is set, runtimes MUST write the value to the `schemata` file in the that sub-directory discussed in `closID`. If also `l3CacheSchema` or `memBwSchema` is set the value of `schemata` field must be written last, after the values from `l3CacheSchema` and `memBwSchema` has been written.
+
+* If none of `l3CacheSchema`, `memBwSchema` or `schemata` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
 
 * If `closID` is not set, runtimes MUST use the container ID from [`start`](runtime.md#start) and create the `<container-id>` directory.
 
-* If `closID` is set, `l3CacheSchema` and/or `memBwSchema` is set
+* If `closID` is set, `l3CacheSchema` and/or `memBwSchema` and/or `schemata` is set
   * if `closID` directory in a mounted `resctrl` pseudo-filesystem doesn't exist, the runtimes MUST create it.
   * if `closID` directory in a mounted `resctrl` pseudo-filesystem exists, runtimes MUST compare `l3CacheSchema` and/or `memBwSchema` value with `schemata` file, and [generate an error](runtime.md#errors) if doesn't match.
 
-* If `closID` is set, and neither of `l3CacheSchema` and `memBwSchema` are set, runtime MUST check if corresponding pre-configured directory `closID` is present in mounted `resctrl`. If such pre-configured directory `closID` exists, runtime MUST assign container to this `closID` and [generate an error](runtime.md#errors) if directory does not exist.
+* If `closID` is set, and none of `l3CacheSchema`, `memBwSchema` or `schemata` are set, runtime MUST check if corresponding pre-configured directory `closID` is present in mounted `resctrl`. If such pre-configured directory `closID` exists, runtime MUST assign container to this `closID` and [generate an error](runtime.md#errors) if directory does not exist.
 
 * **`enableCMT`** *(boolean, OPTIONAL)* - specifies if Intel RDT CMT should be enabled:
     * CMT (Cache Monitoring Technology) supports monitoring of the last-level cache (LLC) occupancy
@@ -674,18 +677,27 @@ The following rules on parameters MUST be applied:
 
 ### Example
 
-Consider a two-socket machine with two L3 caches where the default CBM is 0x7ff and the max CBM length is 11 bits,
-and minimum memory bandwidth of 10% with a memory bandwidth granularity of 10%.
+Consider a two-socket machine with:
 
-Tasks inside the container only have access to the "upper" 7/11 of L3 cache on socket 0 and the "lower" 5/11 L3 cache on socket 1,
-and may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.
+- two L3 caches where the default CBM is 0x7ff (11 bits)
+- eight L2 caches where the default CBM is 0xFF (8 bits)
+- minimum memory bandwidth of 10% with a memory bandwidth granularity of 10%
+
+Tasks inside the container:
+
+- have access to the "upper" 7/11 of L3 cache on socket 0 and the "lower" 5/11 L3 cache on socket 1
+- have access to the "lower" 4/8 of L2 cache on socket 0 (socket 1 is left out from this example)
+- may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.
 
 ```json
 "linux": {
     "intelRdt": {
         "closID": "guaranteed_group",
-        "l3CacheSchema": "L3:0=7f0;1=1f",
-        "memBwSchema": "MB:0=20;1=70"
+        "schemata": [
+            "L3:0=7f0;1=1f",
+            "L2:0=f;1=f;2=f;3=f",
+            "MB:0=20;1=70"
+        ]
     }
 }
 ```
--- a/features-linux.md
+++ b/features-linux.md
@@ -201,12 +201,15 @@ Irrelevant to the availability of SELinux on the host operating system.
 Irrelevant to the availability of Intel RDT on the host operating system.
 
 * **`enabled`** (bool, OPTIONAL) represents whether the runtime supports Intel RDT.
+* **`schemata`** (bool, OPTIONAL) represents whether the
+  (`schemata` field of `linux.intelRdt` in `config.json`)[config-linux.md#intelrdt] is supported.
 
 ### Example
 
 ```json
 "intelRdt": {
-  "enabled": true
+  "enabled": true,
+  "schemata": true
 }
 ```
 
--- a/features.md
+++ b/features.md
@@ -337,7 +337,8 @@ Here is a full example for reference.
       "enabled": true
     },
     "intelRdt": {
-      "enabled": true
+      "enabled": true,
+      "schemata": true
     }
   },
   "annotations": {
--- a/schema/config-linux.json
+++ b/schema/config-linux.json
@@ -262,6 +262,9 @@
                     "closID": {
                         "type": "string"
                     },
+                    "schemata": {
+                        "$ref": "defs.json#/definitions/ArrayOfStrings"
+                    },
                     "l3CacheSchema": {
                         "type": "string"
                     },
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -795,14 +795,23 @@ type LinuxSyscall struct {
 type LinuxIntelRdt struct {
 	// The identity for RDT Class of Service
 	ClosID string `json:"closID,omitempty"`
+
+	// Schemata specifies the complete schemata to be written as is to the
+	// schemata file in resctrl fs. Each element represents a single line in the schemata file.
+	// NOTE: This will overwrite schemas specified in the L3CacheSchema and/or
+	// MemBwSchema fields.
+	Schemata []string `json:"schemata,omitempty"`
+
 	// The schema for L3 cache id and capacity bitmask (CBM)
 	// Format: "L3:<cache_id0>=<cbm0>;<cache_id1>=<cbm1>;..."
+	// NOTE: Should not be specified if Schemata is non-empty.
 	L3CacheSchema string `json:"l3CacheSchema,omitempty"`
 
 	// The schema of memory bandwidth per L3 cache id
 	// Format: "MB:<cache_id0>=bandwidth0;<cache_id1>=bandwidth1;..."
 	// The unit of memory bandwidth is specified in "percentages" by
 	// default, and in "MBps" if MBA Software Controller is enabled.
+	// NOTE: Should not be specified if Schemata is non-empty.
 	MemBwSchema string `json:"memBwSchema,omitempty"`
 
 	// EnableCMT is the flag to indicate if the Intel RDT CMT is enabled. CMT (Cache Monitoring Technology) supports monitoring of


thaJeztah:
(disclaimer: I only very briefly glanced over, so apologies for any mistake in interpreting); IIUC, the new field effectively is to replace the other field(s), but those already shipped in a release.

Should we consider deprecating the other fields, or add wording to discourage use of those going forward (only for backward compatibility)?

marquiz:
Thanks @thaJeztah for the quick feedback

> (disclaimer: I only very briefly glanced over, so apologies for any mistake in interpreting); IIUC, the new field effectively is to replace the other field(s), but those already shipped in a release.

Yes, I would say so. This is effectively to make those useless/unneeded, without breaking backwards compatibility. The current model of one field per "schema" is infeasible as we can already see that there are many features (L2 and CDP) not covered by the OSI spec.

> Should we consider deprecating the other fields, or add wording to discourage use of those going forward (only for backward compatibility)?

I think we could consider that, at least suggest people use the new field. I dunno how deprecation is reacted to in OCI spec.

marquiz:
@thaJeztah et al, one more consideration is that I don't believe any runtime does any parsing/filtering of the existing `l3Schema` or `mbSchema` fields. So in practice the existing fields work just like the new `schema` field I suggested. An alternative approach would be to make a backwards-incompatible change in the spec and just state the "you can put anything you like in the {l3,mb}Schema fields and the runtime will write those (sequentially, two separate writes) into resctrl schemata file". However, I didn't do that because the backwards-incompatibility and the somewhat misleading naming.

config-linux.md
```diff
@@ -654,15 +655,17 @@ The following rules on parameters MUST be applied:
 
 * If either `l3CacheSchema` or `memBwSchema` is set, runtimes MUST write the value to the `schemata` file in the that sub-directory discussed in `closID`.
 
-* If neither `l3CacheSchema` nor `memBwSchema` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
+* If `schemata` field is set, runtimes MUST write the value to the `schemata` file in the that sub-directory discussed in `closID`. If also l3CacheSchema` or `memBwSchema` is set the value of `schemata` field must be written last, after the values from `l3CacheSchema` and `memBwSchema` has been written.
```
thaJeztah, line 658: I'll write  a longer comment, but I'm considering that we should document that if `schemata` is set, runtimes `MUST` ignore both the `l3CacheSchema` and `memBwSchema` fields, and instead to use the `schemata` field instead (and write its content "as-is" (that should probably also be documented on the `l3CacheSchema` and `memBwSchema` fields)

config-linux.md
```diff
@@ -674,18 +677,23 @@ The following rules on parameters MUST be applied:
...
+- have access to the "lower" 4/8 of L2 cache on socket 0 (socket 1 is left out from this example)
+- may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.
 
 ```json
 "linux": {
     "intelRdt": {
         "closID": "guaranteed_group",
-        "l3CacheSchema": "L3:0=7f0;1=1f",
-        "memBwSchema": "MB:0=20;1=70"
+        "schemata": "L3:0=7f0;1=1f\nL2:0=f;1=f;2=f;3=f\nMB:0=20;1=70"
```
thaJeztah, line 696: Just for my understanding, none of the other fields (`ClosID`, `enableCMT`, `enableMBM`) are to be included in the `schemata` file? https://github.com/opencontainers/runtime-spec/blob/c0e90434df2acfc0f75e689098d865d1ece16ff8/specs-go/config.go#L796-L814

config-linux.md
thaJeztah: This field also needs to be added to the JSON-schema; https://github.com/opencontainers/runtime-spec/blob/c0e90434df2acfc0f75e689098d865d1ece16ff8/schema/config-linux.json#L259-L278

As well as the go implementation; https://github.com/opencontainers/runtime-spec/blob/c0e90434df2acfc0f75e689098d865d1ece16ff8/specs-go/config.go#L795

thaJeztah:
> An alternative approach would be to make a backwards-incompatible change in the spec and just state the "you can put anything you like in the {l3,mb}Schema fields and the runtime will write those

Yeah, not sure that's great, as that would effectively make both an "anything goes" field, but they would give the _impression_ they're only to be used for a specific purpose.

I think the existing fields are a bit of a leaky abstraction already; they give the impression that it's an abstraction, but in the end (but I'm not deeply familiar with this matter) it looks like it's just artificially splitting 

So I could see two possible directions;

- either go "all the way in", and every possible RDT feature must have a counterpart in the specification
    - :+1: runtimes can parse, and validate options
    - :+1: runtimes (and the spec) can advertise exactly what RDT options it supports (https://github.com/opencontainers/runtime-spec/pull/1130)
    - :point_up: this can improve portability
    - :-1: runtimes (and the runtime spec) must be updated for changes in the RDT feature-set (new features, feature deprecation)
- make it clear that there's no abstraction _at all_;
    - the `schemata` field is a 1:1 mapping to the `schemata` file
    - runtimes SHOULD NOT perform _any_ validation of the content (perhaps some restrictions w.r.t. accepted characters and/or length)
    - runtimes MUST write the content of the `schemata` field to the `schemata` file _as-is_

For the `schemata` field, and deprecation of the existing fields (I'd have to look up the correct wording for "deprecated");

- mention that `l3Schema` and `memBwSchema` `SHOULD NOT` be used, for reasons other than backward compatibility with older runtimes.
- mention that `l3Schema` and `memBwSchema` `MUST NOT` be added to  if `schemata` if set (i.e., if `schemata` is used, it replaces any use of `l3Schema` and `memBwSchema`)
- :question: TBD: what would be the best approach to provide a "hybrid" spec for backward compatibility (i.e., an OCI spec that both older (using the `l3Schema` and `memBwSchema`) and newer (through `schemata`) runtimes would accept?
    - older runtimes (taking the existing spec into account) would only be able to use L3 (not L2)
    - which means that _either_ the spec will result in a different outcome (older runtimes only L3, newer runtimes (from `schemata`) potentially using both `L3` and `L2` features)
    - so if we want a "hybrid" spec to produce the _SAME_ result, then runtimes MUST validate that concatenating `l3Schema` and `memBwSchema` MUST be identical to the content of the `schemata` field
    - if, on the other hand, we are ok with the result being different from older runtimes, we could either just "document" that they may differ, or document that `l3Schema` and `memBwSchema` MUST be empty if `schemata` is non-empty (i.e. can either use new features, or no RDT at all).

For the `schemata` field itself

- We could refer to Intel's documentation as canonical reference for the accepted format (i.e., what's supported being outside of the OCI runtime specification).
- However; if the format itself is well-documented and stable (i.e., new features may be added, but all follow the same format), we _COULD_ document the accepted format (accepted characters, overall format).
    - I don't think the runtime-spec's go-implementation generally provides utilities for this ("runtime-tools" might)
    - _Maybe_ we could provide / document a (suggested) regular expression that captures the format (which runtimes and runtime-tools can implement).
- :question: does RDT provide any form of feature detection? And if so, is that something which could be included in the `features.md` (or would that be beyond scope?)
    - This information could be used by "higher level" runtimes (runtimes that generate the OCI spec for a container)

config-linux.md
marquiz: Correct 👍

config-linux.md
marquiz: Woops 🙄 Added

marquiz:
@thaJeztah appreciate your thoughtful review. My comments in-lined below

> Yeah, not sure that's great, as that would effectively make both an "anything goes" field, but they would give the _impression_ they're only to be used for a specific purpose.

I agree

> I think the existing fields are a bit of a leaky abstraction already; they give the impression that it's an abstraction, but in the end (but I'm not deeply familiar with this matter) it looks like it's just artificially splitting

Exactly my thinking/feeling. We could possibly use something like

```golang
type LinuxIntelRdt struct {
...
	Schemata []LinuxIntelRdtSchema
...
}

type LinuxIntelRdtSchema struct {
	Type   String  // "L2", "L3", "MB" etc
	Schema String  // Depends on the Type, e.g. "0=f;1=f"
}
```

But I don't think that makes anyone's life any easier.


> So I could see two possible directions;
> 
> * either go "all the way in", and every possible RDT feature must have a counterpart in the specification
>   
>   * 👍 runtimes can parse, and validate options
>   * 👍 runtimes (and the spec) can advertise exactly what RDT options it supports ([Add `features.md` to formalize the `runc features` JSON #1130](https://github.com/opencontainers/runtime-spec/pull/1130))
>   * ☝️ this can improve portability
>   * 👎 runtimes (and the runtime spec) must be updated for changes in the RDT feature-set (new features, feature deprecation)

I think this would be a maintenance nightmare. Even with the current very limited validation/filtering specified I think this hasn't been implemented. Plus for a full validation the runtime will need virtually to replicate the validation code in linux kernel. Who knows what the actual format for future "FOOBAR" technology will look like. And e.g. in the case of RDT the min/max width of the bitmask varies between HW etc.

> * make it clear that there's no abstraction _at all_;
>   
>   * the `schemata` field is a 1:1 mapping to the `schemata` file
>   * runtimes SHOULD NOT perform _any_ validation of the content (perhaps some restrictions w.r.t. accepted characters and/or length)
>   * runtimes MUST write the content of the `schemata` field to the `schemata` file _as-is_

I'm a strong proponent of this approach. Let kernel do the validation because it can do it correctly 😇 

 
> For the `schemata` field, and deprecation of the existing fields (I'd have to look up the correct wording for "deprecated");
> 
> * mention that `l3Schema` and `memBwSchema` `SHOULD NOT` be used, for reasons other than backward compatibility with older runtimes.
> * mention that `l3Schema` and `memBwSchema` `MUST NOT` be added to  if `schemata` if set (i.e., if `schemata` is used, it replaces any use of `l3Schema` and `memBwSchema`)
> * ❓ TBD: what would be the best approach to provide a "hybrid" spec for backward compatibility (i.e., an OCI spec that both older (using the `l3Schema` and `memBwSchema`) and newer (through `schemata`) runtimes would accept?
>   
>   * older runtimes (taking the existing spec into account) would only be able to use L3 (not L2)
>   * which means that _either_ the spec will result in a different outcome (older runtimes only L3, newer runtimes (from `schemata`) potentially using both `L3` and `L2` features)
>   * so if we want a "hybrid" spec to produce the _SAME_ result, then runtimes MUST validate that concatenating `l3Schema` and `memBwSchema` MUST be identical to the content of the `schemata` field
>   * if, on the other hand, we are ok with the result being different from older runtimes, we could either just "document" that they may differ, or document that `l3Schema` and `memBwSchema` MUST be empty if `schemata` is non-empty (i.e. can either use new features, or no RDT at all).

I'd strongly suggest to keep it simple. Any schema validation will be a tricky business to do right with all the corner cases like  `L3:0=f;1=f0` and `L3:1=f0;0=f` will have the same end result. So either
-`l3Schema` and `memBwSchema` `MUST NOT` be set to  if `schemata` if set
- or then just accept that the end result may be different

In any case the "portability" of the schemata is very poor even between reboots/remount of the resctrlfs as the format and accepted fields in the schemata file depends e.g. on the mount options.


 
> For the `schemata` field itself
> 
> * We could refer to Intel's documentation as canonical reference for the accepted format (i.e., what's supported being outside of the OCI runtime specification).
> * However; if the format itself is well-documented and stable (i.e., new features may be added, but all follow the same format), we _COULD_ document the accepted format (accepted characters, overall format).
>   
>   * I don't think the runtime-spec's go-implementation generally provides utilities for this ("runtime-tools" might)
>   * _Maybe_ we could provide / document a (suggested) regular expression that captures the format (which runtimes and runtime-tools can implement).
> * ❓ does RDT provide any form of feature detection? And if so, is that something which could be included in the `features.md` (or would that be beyond scope?)
>   
>   * This information could be used by "higher level" runtimes (runtimes that generate the OCI spec for a container)

The documentation is not Intel's but in Linux kernel (https://docs.kernel.org/arch/x86/resctrl.html). But 👍 for adding a reference. I would refer to that directly instead of describing it here. With new kernel versions there may be (backwards-compatible) extensions and new technologies enabled that the OCI spec would then be missing. I think that also applies to any regular expression we tried to add here.

RDT does provide information about the currently active setup (what features are enabled and in which configuration) under `/sys/fs/resctrl/info/`. That might be something useful to add into `features.md` but I think that would be a subject of a separate PR – to think about thoroughly.

thaJeztah:
> I think this would be a maintenance nightmare. Even with the current very limited validation/filtering specified I think this hasn't been implemented

:100: I fully agree. My main intent here was to mention options we _considered_ (for future visitors, including "future self"), and _rejected_ (for reasons mentioned). It's good to try to provide an abstraction, but in some case, such as this one, I don't think that's possible, and we just need to acknowledge that some parts of the OCI just are very platform-specific and tightly coupled to kernel/platform features.

> I'd strongly suggest to keep it simple. Any schema validation will be a tricky business

Yes, that's what I somewhat expected the case to be; I guess the recommendation would be "use the new option", and only in exceptional cases consider using "both" (if you like schrödingers cat).

> The documentation is not Intel's but in Linux kernel (https://docs.kernel.org/arch/x86/resctrl.html). But 👍 for adding a reference.

Heh, yes, let myself go on the "canonical reference" side; definitely "+1" on keeping docs in this spec as minimal as possible, and refer to the kernel reference as source of truth.

>  I think that also applies to any regular expression we tried to add here.

If it's well-defined and a regex is possible (either in docs and/or the JSON-schema), I guess that would be good to have.

> RDT does provide information about the currently active setup (what features are enabled and in which configuration) under /sys/fs/resctrl/info/. That might be something useful to add into features.md but I think that would be a subject of a separate PR – to think about thoroughly.

Yes, that's for sure not a blocker (from my point of view) for this PR. 

My main motivation for some of these comments is to explore what amount of validation can reasonably be performed by (higher-level) runtimes before "the spec hits the kernel". 

My general rule of thumb is to keep that as minimal as possible for most cases (unless we can absolutely be sure), but a "well-formedness" check (where possible) can help. Having some amount of early error-handling can reduce burden on maintainers (as all OCI / kernel / runtime errors result in a "OCI runtime" error, and those tend to be piled on by users "I have this same issue!" (no you don't 😂))

marquiz:
> If it's well-defined and a regex is possible (either in docs and/or the JSON-schema), I guess that would be good to have.

My feeling is that it's not well-defined enough. The currently supported format for currently supported technologies is described (https://docs.kernel.org/arch/x86/resctrl.html#l3-schemata-file-details-code-and-data-prioritization-disabled) but I wouldn't count on that. Say next kernel version will start supporting id ranges like `L2:0-4=ff;5-8=ff00` or the new FOOBAR technology supporting some new format `FOOBAR:xyzy;1234` we're screwed

marquiz:
@thaJeztah I think I'll sleep over this and push an update after the weekend, trying to capture all the review comments

thaJeztah:
For sure! Mostly exploring here if it's possible (within reason). Having a quick look at examples on that link;

```
# echo "L3:0=3;1=3\nMB:0=1024;1=500" > /sys/fs/resctrl/p1/schemata
# echo -e "L3:0=f8000;1=fffff\nMB:0=20;1=100" > p0/schemata
```

Which would translate to;

```
L3:0=3;1=3
MB:0=1024;1=500
```

and;

```
L3:0=f8000;1=fffff
MB:0=20;1=100
```

Based on that the format "roughly" is a newline-delimited set of features;

```
<feature>:<[key=value;]...>
<feature>:<[key=value;]...>
```

- first column is the name of the feature, followed by a colon (`:`)
- followed by (one or more / zero or more?) semi-colon (`;`) separated `key=value` pairs.

So I guess the blanks to fill in is what are accepted characters for `feature`, `key`, and `value`
 
Whether `<key>=<value>` pairs are compulsory (i.e. is `<feature>:` without any options valid), might be _useful_ and to some extent if `=<value>` is compulsory (or "key-only" are allowed), but I guess any of that may be going too much into details for the spec itself (we can treat `[key=value;]` mostly as a opaque element, other than "valid characters")

marquiz:
> Based on that the format "roughly" is a newline-delimited set of features;
> 
> ```
> <feature>:<[key=value;]...>
> ```

Yeah, the currently availabe technologies/features follow this format. My fear is that we could get bitten by future extensions (like `L3:1,3,5-10=f`) or new stuff, like you speculated `<feature>:<key1>[;<key2>...]`.

One option would be to validate the currently known features and just not validate unknown stuff. Another is to follow the currently ubiquitous format `<feature>:<[key=value;]...>`. A third variant as you suggested would be to validate `<feature>:<blob>[;<blob>]` (where blob could be `<key>=<value>` or smth else) 

In any case, we could make the schemata field an array of schemas

```golang
Schemata []string
```

config-linux.md
```diff
@@ -654,15 +655,17 @@ The following rules on parameters MUST be applied:
...
+
+* If none of `l3CacheSchema`, `memBwSchema` or `schemata` is set, runtimes MUST NOT write to `schemata` files in any `resctrl` pseudo-filesystems.
 
 * If `closID` is not set, runtimes MUST use the container ID from [`start`](runtime.md#start) and create the `<container-id>` directory.
 
-* If `closID` is set, `l3CacheSchema` and/or `memBwSchema` is set
+* If `closID` is set, `l3CacheSchema` and/or `memBwSchema` and/or `schemata` is set
   * if `closID` directory in a mounted `resctrl` pseudo-filesystem doesn't exist, the runtimes MUST create it.
-  * if `closID` directory in a mounted `resctrl` pseudo-filesystem exists, runtimes MUST compare `l3CacheSchema` and/or `memBwSchema` value with `schemata` file, and [generate an error](runtime.md#errors) if doesn't match.
+  * if `closID` directory in a mounted `resctrl` pseudo-filesystem exists, runtimes MUST compare `l3CacheSchema` and/or `memBwSchema` and/or `schemata` value with `schemata` file, and [generate an error](runtime.md#errors) if doesn't match.
```
giuseppe, line 666: too late too relax it to "SHOULD"?  It is not so obvious how the comparison should be done from a runtime

marquiz: @giuseppe good point. For the old fields (`l3CacheSchema` and `memBwSchema`) let's just keep it as is. But for the new `schemata` field we could just drop this requirement. As you said it's not at all clear how this comparison should be implemented (I'd consider that as internal implementation details of the Linux kernel).

This might actually even be desirable. If you think that the higher-level runtime would be managing the CLOS configuration. With the old behavior you cannot actually change the configuration of existing groups, you'd need to delete it and re-create it (and if there were some running containers there you'd need to freeze them, list the pids, delete CLOS, re-create CLOS with new config, assign the pids to the new group, unfreeze...).

marquiz:
Resurrecting this PR. This is still an open issue and new features (like Slow Memory Bandwidth Allocation (SMBA) from AMD) that are not supported by the spec are piling up.

A small change since the last update is that I changed the `schemata` field to an array of strings (each element representing one line written into the `schemata` file in resctrlfs).

Any thoughts on the validation (https://github.com/opencontainers/runtime-spec/pull/1230#issuecomment-1742917186? All features (L3, L2, MB, SMBA, ...) in the kernel so far follow the `<feature>:<[key=value;]...>` format but I'm not sure if that can be treated as a given contract for all future extensions :thinking: 
 
 Ping @thaJeztah @giuseppe

config-linux.md
giuseppe: nit: missing "`" at the beginning of l3CacheSchema

giuseppe (approved):
LGTM

marquiz:
Updated example in `config-linux.md` and `schema/` and `specs-go/` to match the array of strings format

marquiz:
> LGTM

ping @AkihiroSuda @utam0k any thoughts on this?

config-linux.md
```diff
@@ -645,6 +645,7 @@ The following parameters can be specified for the container:
     The value SHOULD start with `L3:` and SHOULD NOT contain newlines.
 * **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth per L3 cache id.
     The value MUST start with `MB:` and MUST NOT contain newlines.
+* **`schemata`** *(array of strings, OPTIONAL)* - specifies the schemata to be written to the `schemata` file in resctrlfs. Each element represents one line in the `schemata` file. The value MUST NOT contain newlines.
```
AkihiroSuda, line 648: https://github.com/opencontainers/runtime-spec/blob/main/features-linux.md needs to cover the availability of this

marquiz: @AkihiroSuda PTAL

AkihiroSuda (approved):

utam0k (approved):

PR #1234, opened 2023-10-06
clarify kill and delete operation

lifubang:
In fact, `container-runtime kill [-a,--all] container_id 'signal]` and `container-runtime delete [-f,--force] container_id` has been used by upstream projects for many years, but it isn't defined in `runtime-spec`. It may cause some break changes when we do code refactor in this area.
So, maybe we should define it clearly in `runtime-spec`.

background: https://github.com/opencontainers/runc/issues/4049


runtime.md
```diff
@@ -129,18 +129,19 @@ This operation MUST run the user-specified program as specified by [`process`](c
...
 ### <a name="runtimeDelete" />Delete
-`delete <container-id>`
+`delete [-f,--force] <container-id>`
 
 This operation MUST [generate an error](#errors) if it is not provided the container ID.
-Attempting to `delete` a container that is not [`stopped`](#state) MUST have no effect on the container and MUST [generate an error](#errors).
-Deleting a container MUST delete the resources that were created during the `create` step.
+Without `-f`, attempting to `delete` a container that is not [`stopped`](#state) MUST have no effect on the container and MUST [generate an error](#errors).
+With `-f`, runtime should try its best to delete the container no matter what the state is.
+Deleting a container MUST delete the resources that were created during the `create` step, specially, if the container uses `shared pid namespace`, all processes inside the container's cgroup should also be killed first.
```
giuseppe, line 144: I'd rephrase it as "if the container doesn't use a PID namespace`

lifubang: Sorry, I haven't understood your meaning yet.
You mean change "if the container uses `shared pid namespace`" to "if the container doesn't use a PID namespace`"?

lifubang: > You mean change "if the container uses `shared pid namespace`" to "if the container doesn't use a PID namespace`"?

If this is your wanted change, I think it can't include all the situations, because in here `shared pid namespace` means:
1. Doesn't use a PID namespace (Using the host pid namespace);
2. Shares a PID namespaces with other container's processes(Using a specified process' pid namespace, the process may be not in this container).

So maybe we should keep "if the container uses `shared pid namespace`"?
@giuseppe PTAL, thanks very much.

lifubang:
@opencontainers/runtime-spec-maintainers PTAL, we need this PR to make a decision how to fix a bug in `runc`(https://github.com/opencontainers/runc/issues/4047).
Looking forward your feedback, thanks.

runtime.md
```diff
@@ -129,18 +129,19 @@ This operation MUST run the user-specified program as specified by [`process`](c
 This operation MUST generate an error if `process` was not set.
 
 ### <a name="runtimeKill" />Kill
-`kill <container-id> <signal>`
+`kill <container-id> [-a,--all] <signal>`
 
 This operation MUST [generate an error](#errors) if it is not provided the container ID.
-Attempting to send a signal to a container that is neither [`created` nor `running`](#state) MUST have no effect on the container and MUST [generate an error](#errors).
-This operation MUST send the specified signal to the container process.
+This operation MUST send the specified signal to the container process(without `-a`) or to all processes in the container(with `-a`).
```
kolyshkin, line 135: As per https://github.com/opencontainers/runc/issues/3864, I think that `-a` should be removed.

In short, the runtime should already know how to kill a container:
 - if the container has own private PID namespace, kill the init;
 - if the container shares the PID namespace, kill all processes in the container's cgroup;
 - the container which does have neither the cgroup nor private PID namespace is broken and should not be created in the first place.
 
Option `-a` was added as a hack to kill the container which does not have own private PID namespace. Since the runtime already knows about it, there is no need for `-a`.

runtime.md
kolyshkin: @lifubang I think the best wording would be
```
s/if the container uses `shared pid namespace`/if the container does not use its own private PID namespace/
```
or somesuch.

runtime.md
lifubang: > I think that `-a` should be removed.

I agree this when the signal is `SIGKILL`, but for signals, which is not `SIGKILL`, can't be send to all processes in the container like before.

After remove `-a`, we must treat `SIGKILL` as a special signal, it looks like unreasonable. Maybe all signals should be equal。

lifubang: > we must treat `SIGKILL` as a special signal

If we agree to remove `-a`, I think we should clarify this special signal `SIGKILL` for private pid ns container here.
So, do you have a strong opinion to remove `-a` in `kill` operation?

runtime.md
```diff
@@ -129,18 +129,19 @@ This operation MUST run the user-specified program as specified by [`process`](c
 This operation MUST generate an error if `process` was not set.
 
 ### <a name="runtimeKill" />Kill
-`kill <container-id> <signal>`
+`kill <container-id> [-a,--all] <signal>`
```
AkihiroSuda, line 132: Duplicate of 
- https://github.com/opencontainers/runtime-spec/pull/1190

AkihiroSuda: https://github.com/opencontainers/runtime-spec/pull/1190#pullrequestreview-1350798670

AkihiroSuda (changes requested):
https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/runtime.md?plain=1#L95

runtime.md
lifubang: OK, I changed this PR to `clarify kill and delete operation for shared pid namespace container`, let's discuss `-a` in #1190 .

lifubang:
> https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/runtime.md?plain=1#L95

I wonder we should change this rule or not?
Before we change this rule, I changed my PR to clarify kill and delete operation for shared pid namespace container, if we don't think this type container should be described in runtime-spec, feel free to close this PR.

fuweid:
@lifubang the operation can be function or command-line. `-a` is kind of command-line api. I think the maintainer can accept `<all>` here.

AkihiroSuda:
> However, the runc kill -a has been used for many years. <all> can be -a in the command-line implementation, is it correct? @AkihiroSuda

Yes, but the command line implementation is (currently) out of the scope of the OCI Runtime Spec.

lifubang:
> Yes, but the command line implementation is (currently) out of the scope of the OCI Runtime Spec.

How about add `runtime-cmd-api.md` to define all the basic command apis?

AkihiroSuda:
> > Yes, but the command line implementation is (currently) out of the scope of the OCI Runtime Spec.
> 
> How about add `runtime-cmd-api.md` to define all the basic command apis?

Adding the CLI spec to the Runtime Spec was once rejected in 2017 (https://github.com/opencontainers/runtime-spec/pull/513#issuecomment-278550496) and accepted in the runtime-tools repo: https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md

I guess we can revisit this, but it is beyond the scope of this PR. (Feel free to open an issue/PR)

fuweid:
> > > Yes, but the command line implementation is (currently) out of the scope of the OCI Runtime Spec.
> 
> > 
> 
> > How about add `runtime-cmd-api.md` to define all the basic command apis?
> 
> 
> 
> Adding the CLI spec to the Runtime Spec was once rejected in 2017 (https://github.com/opencontainers/runtime-spec/pull/513#issuecomment-278550496) and accepted in the runtime-tools repo: https://github.com/opencontainers/runtime-tools/blob/master/docs/command-line-interface.md
> 
> 
> 
> I guess we can revisit this, but it is beyond the scope of this PR. (Feel free to open an issue/PR)

Thanks. We should discuss it in other pr (I also received the runc maintainer's comment about introducing new flag https://github.com/opencontainers/runc/pull/4045)

runtime.md
```diff
@@ -132,16 +132,26 @@ This operation MUST generate an error if `process` was not set.
 `kill <container-id> <signal>`
 
 This operation MUST [generate an error](#errors) if it is not provided the container ID.
+
 Attempting to send a signal to a container that is neither [`created` nor `running`](#state) MUST have no effect on the container and MUST [generate an error](#errors).
-This operation MUST send the specified signal to the container process.
+
+This operation MUST send the specified signal to the container's init process.
+
+Specially, if the signal is `SIGKILL` and the container does not use its own private PID namespace, this operation MUST send the `SIGKILL` signal to all the processes in the container, even if the container's state is `stopped`. If there is no process left in this type container, the operation MUST [generate an error](#errors).
```
kolyshkin, line 140: 1. `s/does not use/does not have/`
2. Traditionally, `runc kill -a` does not result in an error even if there are no processes left. I'm not sure if we want to change it here.

runtime.md
```diff
@@ -132,16 +132,26 @@ This operation MUST generate an error if `process` was not set.
...
 ### <a name="runtimeDelete" />Delete
 `delete <container-id>`
 
 This operation MUST [generate an error](#errors) if it is not provided the container ID.
+
 Attempting to `delete` a container that is not [`stopped`](#state) MUST have no effect on the container and MUST [generate an error](#errors).
+
 Deleting a container MUST delete the resources that were created during the `create` step.
+
+Specially, when deleting a container, which does not use its own private PID namespace, the operation should ensure kill all processes in this type container, and ensure no process left in it.
```
kolyshkin, line 151: `s/does not use/does not have/`

`s/should ensure kill all processes/SHOULD kill all processes/`

`s/and ensure no process left in it./and return an error if those processes can not be killed./`

lifubang:

PR #1237, opened 2023-12-06
clarify user ns mappings and time ns offset configurations

lifubang:
Nowadays, when we let the container to join an existing `user` or `time` namespace, we should not provide `uidMappings`, `gidMappings`, and `timeOffsets` configurations.
Let's add some description to clarify user ns mappings and time ns offset configurations.

----
background:
1. https://github.com/opencontainers/runc/issues/4122
2. https://github.com/opencontainers/runc/pull/3985#issuecomment-1842477942

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -37,7 +37,7 @@ The following parameters can be specified to set up namespaces:
     * **`time`** the container will be able to have its own clocks.
 * **`path`** *(string, OPTIONAL)* - namespace file.
     This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
-    The runtime MUST place the container process in the namespace associated with that `path`.
+    The runtime MUST let the container process join in the namespace associated with that `path`.
     The runtime MUST [generate an error](runtime.md#errors) if `path` is not associated with a namespace of type `type`.
 
     If `path` is not specified, the runtime MUST create a new [container namespace](glossary.md#container-namespace) of type `type`.
@@ -80,6 +80,9 @@ If a `namespaces` field contains duplicated namespaces with same `type`, the run
 
 ## <a name="configLinuxUserNamespaceMappings" />User namespace mappings
 
+If the runtime should create an new user namespace for the container, `uidMappings` and `gidMappings` should be provided, otherwise, these two fields should not be specified, 
+and it will be ignored by the runtime.
+
 **`uidMappings`** (array of objects, OPTIONAL) describes the user namespace uid mappings from the host to the container.
 **`gidMappings`** (array of objects, OPTIONAL) describes the user namespace gid mappings from the host to the container.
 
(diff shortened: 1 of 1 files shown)


config-linux.md
```diff
@@ -37,7 +37,7 @@ The following parameters can be specified to set up namespaces:
     * **`time`** the container will be able to have its own clocks.
 * **`path`** *(string, OPTIONAL)* - namespace file.
     This value MUST be an absolute path in the [runtime mount namespace](glossary.md#runtime-namespace).
-    The runtime MUST place the container process in the namespace associated with that `path`.
+    The runtime MUST let the container process join in the namespace associated with that `path`.
```
rata, line 40: why this wording change?

config-linux.md
```diff
@@ -80,6 +80,9 @@ If a `namespaces` field contains duplicated namespaces with same `type`, the run
 
 ## <a name="configLinuxUserNamespaceMappings" />User namespace mappings
 
+If the runtime should create an new user namespace for the container, `uidMappings` and `gidMappings` should be provided, otherwise, these two fields should not be specified, 
+and it will be ignored by the runtime.
```
rata, line 84:
```suggestion
and they MUST be ignored by the runtime.
```

config-linux.md
```diff
@@ -113,6 +116,9 @@ Note that the number of mapping entries MAY be limited by the [kernel][user-name
 
 ## <a name="configLinuxTimeOffset" />Offset for Time Namespace
 
+If the runtime should create an new time namespace for the container, `timeOffsets` should be provided, otherwise, it should not be specified, 
+and it will be ignored by the runtime.
```
rata, line 120: idem

config-linux.md
cyphar: The old wording was better, "let the container process join" is not an accurate description.

config-linux.md
cyphar, line 84: In the runtime-spec, we don't hard-wrap lines. Each line is meant to be a complete sentence (to make diffs less messy).

I think the text here is also a little confusing. Maybe something like this would better explain things:

```suggestion
If the runtime is creating a new user namespace for this container (in other words, the [`user` namespace entry in `namespaces`](#configLinuxNamespaces) does not have a `path` specified), `uidMappings` and `gidMappings` MUST be provided.

However, if the container will join an existing user namespace, `uidMappings` and `gidMappings` SHOULD NOT be specified (user namespaces cannot have their mappings changed after being configured).
If specified, the specified `uidMappings` and `gidMappings` SHOULD be the same mappings as the existing user namespace, and MUST be ignored by the runtime.
Runtimes MAY [generate an error](runtime.md#errors) if the specified `uidMappings` and `gidMappings` do not match the existing user namespace.

> **NOTE**: runc 1.1 and earlier, as well as some other spec-compliant runtimes, incorrectly required users to specify `uidMappings` and `gidMappings` even if the container was joining an existing namespace. The provided mappings were completely ignored until runc 1.2.
```

config-linux.md
cyphar, line 120:
```suggestion
If the container will join an existing time namespace, `timeOffsets` SHOULD NOT be specified (time namespaces cannot have their mappings changed after being used), and MUST be ignored by the runtime.
Runtimes MAY [generate an error](runtime.md#errors) if the `timeOffsets` is specified and the container is joining an existing time namespace.
```

We can be a bit more strict here because there's no released version of runc with timens support yet.

cyphar (changes requested):

PR #1241, opened 2024-01-02
Carry #1111: specs-go/config: add Landlock LSM support

Zheaoli:
Carry #1111

### DIFF
--- a/config.md
+++ b/config.md
@@ -353,51 +353,51 @@ For Linux-based systems, the `process` object supports the following process-spe
       CPU affinity after the process is moved to container's cgroup, and the
       final affinity is determined by the Linux kernel.
 * **`landlock`** (object, OPTIONAL) specifies the Landlock unprivileged access control settings for the container process.
-    Note that `noNewPrivileges` must be set to true to use this feature.
-    For more information about Landlock, see [Landlock documentation][landlock].
-    `landlock` contains the following properties:
+   Note that `noNewPrivileges` must be set to true to use this feature.
+   For more information about Landlock, see [Landlock documentation][landlock].
+   `landlock` contains the following properties:
 
-    * **`ruleset`** (object, OPTIONAL) the `ruleset` field identifies a set of rules (i.e., actions on objects) that need to be handled (i.e., restricted).
-    The `ruleset` currently contains the following types:
-        * **`handledAccessFS`** (array of strings, OPTIONAL) is an array of FS typed actions that are handled by a ruleset.
-        If no rule explicitly allow them, they should then be forbidden.
-        * **`handledAssessNetwork`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are handled by a ruleset. (The NETWORK typed actions are avaliable when the ABI version >= 4. the behavior of the NETWORK typed actions is not used when the ABI version is less than 4 will depend on the **`disableBestEffort`**)
-    * **`rules`** (object, OPTIONAL) the `rules` field specifies the security policies (i.e., actions allowed on objects) to be added to an existing ruleset.
-    The `rules` currently contains the following types:
-        * **`pathBeneath`** (array of objects, OPTIONAL) is an array of the file-hierarchy typed rules.
-        Entries in the array contain the following properties:
-            * **`allowedAccess`** (array of strings, OPTIONAL) is an array of FS typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
-                1. ABI version >= 1:
-                    1. exectute
-                    2. write_file
-                    3. read_file
-                    4. read_dir
-                    5. remove_dir
-                    6. remove_file
-                    7. make_char
-                    8. make_dir
-                    9. make_reg
-                    10. make_sock
-                    11. make_fifo
-                    12. make_block
-                    13. make_sym
-                2. ABI version >= 2:
-                    1. refer
-                3. ABI version >= 3:
-                    1. truncate
-            * **`paths`** (array of strings, OPTIONAL) is an array of files or parent directories of the file hierarchies to restrict.
-        * **`portBeneath`** (array of objects, OPTIONAL) is an array of the network-hierarchy typed rules.
-        Entries in the array contain the following properties:
-            * **`allowedAccess`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
-                1. ABI version >= 4:
-                    1. bind
-                    2. connect
-            * **`ports`** (array of strings, OPTIONAL) is an array of network ports to restrict.
-    * **`disableBestEffort`** (bool, OPTIONAL) the `disableBestEffort` field disables the best-effort security approach for Landlock access rights.
-    This is for conditions when the Landlock access rights explicitly configured by the container are not supported or available in the running kernel.
-    If the best-effort security approach is enabled (`false`), the runtime SHOULD enforce the strongest rules configured up to the current kernel support, and only be [logged as a warning](runtime.md#warnings) for those not supported.
-    If disabled (`true`), the runtime MUST [generate an error](runtime.md#errors) if one or more rules specified by the container is not supported.
-	Default is `false`, i.e., following a best-effort security approach.
+   * **`handledAccess`** (object, OPTIONAL) specifies the access rights that will be restricted by the ruleset.
+   The `handledAccess` currently contains the following types:
+       * **`handledAccessFS`** (array of strings, OPTIONAL) is an array of FS typed actions that are handled by a ruleset.
+       If no rule explicitly allow them, they should then be forbidden.
+       * **`handledAccessNetwork`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are handled by a ruleset. (The NETWORK typed actions are available when the ABI version >= 4. The behavior when the ABI version is less than 4 will depend on the **`enableBestEffort`**)
+   * **`rules`** (object, OPTIONAL) specifies the security policies (i.e., actions allowed on objects) to be enforced.
+   The `rules` currently contains the following types:
+       * **`pathBeneath`** (array of objects, OPTIONAL) is an array of the file-hierarchy typed rules.
+       Entries in the array contain the following properties:
+           * **`allowedAccess`** (array of strings, OPTIONAL) is an array of FS typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
+               1. ABI version >= 1:
+                   1. execute
+                   2. write_file
+                   3. read_file
+                   4. read_dir
+                   5. remove_dir
+                   6. remove_file
+                   7. make_char
+                   8. make_dir
+                   9. make_reg
+                   10. make_sock
+                   11. make_fifo
+                   12. make_block
+                   13. make_sym
+               2. ABI version >= 2:
+                   1. refer
+               3. ABI version >= 3:
+                   1. truncate
+           * **`paths`** (array of strings, OPTIONAL) is an array of files or parent directories of the file hierarchies to restrict.
+       * **`networkPort`** (array of objects, OPTIONAL) is an array of the network socket rules.
+       Entries in the array contain the following properties:
+           * **`allowedAccess`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
+               1. ABI version >= 4:
+                   1. bind
+                   2. connect
+           * **`ports`** (array of strings, OPTIONAL) is an array of network ports to restrict.
+   * **`enableBestEffort`** (bool, OPTIONAL) the `enableBestEffort` field disables the best-effort security approach for Landlock access rights.
+   This is for conditions when the Landlock access rights explicitly configured by the container are not supported or available in the running kernel.
+   If the best-effort security approach is enabled (`false`), the runtime SHOULD enforce the strongest rules configured up to the current kernel support, and only be [logged as a warning](runtime.md#warnings) for those not supported.
+   If disabled (`true`), the runtime MUST [generate an error](runtime.md#errors) if one or more rules specified by the container is not supported.
+   Default is `true`, i.e., following a best-effort security approach.
 
 ### <a name="configUser" />User
 
@@ -444,7 +444,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
         "priority": 4
     },
     "landlock": {
-        "ruleset": {
+        "handledAccess": {
             "handledAccessFS": [
                 "execute",
                 "write_file",
@@ -462,7 +462,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
                 "refer",
                 "truncate"
             ],
-            "handledAssessNetwork": [
+            "handledAccessNetwork": [
                 "bind",
                 "connect"
             ]
@@ -501,7 +501,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
                     ]
                 }
             ],
-            "portBeneath": [
+            "networkPort": [
                 {
                     "allowedAccess": [
                         "bind",
@@ -514,7 +514,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
                 }
             ]
         },
-        "disableBestEffort": false
+        "enableBestEffort": true
     },
     "noNewPrivileges": true,
     "capabilities": {
--- a/config.md
+++ b/config.md
@@ -352,6 +352,52 @@ For Linux-based systems, the `process` object supports the following process-spe
       for `initial`. If omitted or empty, runtime SHOULD NOT change process'
       CPU affinity after the process is moved to container's cgroup, and the
       final affinity is determined by the Linux kernel.
+* **`landlock`** (object, OPTIONAL) specifies the Landlock unprivileged access control settings for the container process.
+    Note that `noNewPrivileges` must be set to true to use this feature.
+    For more information about Landlock, see [Landlock documentation][landlock].
+    `landlock` contains the following properties:
+
+    * **`ruleset`** (object, OPTIONAL) the `ruleset` field identifies a set of rules (i.e., actions on objects) that need to be handled (i.e., restricted).
+    The `ruleset` currently contains the following types:
+        * **`handledAccessFS`** (array of strings, OPTIONAL) is an array of FS typed actions that are handled by a ruleset.
+        If no rule explicitly allow them, they should then be forbidden.
+        * **`handledAssessNetwork`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are handled by a ruleset. (The NETWORK typed actions are avaliable when the ABI version >= 4. the behavior of the NETWORK typed actions is not used when the ABI version is less than 4 will depend on the **`disableBestEffort`**)
+    * **`rules`** (object, OPTIONAL) the `rules` field specifies the security policies (i.e., actions allowed on objects) to be added to an existing ruleset.
+    The `rules` currently contains the following types:
+        * **`pathBeneath`** (array of objects, OPTIONAL) is an array of the file-hierarchy typed rules.
+        Entries in the array contain the following properties:
+            * **`allowedAccess`** (array of strings, OPTIONAL) is an array of FS typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
+                1. ABI version >= 1:
+                    1. exectute
+                    2. write_file
+                    3. read_file
+                    4. read_dir
+                    5. remove_dir
+                    6. remove_file
+                    7. make_char
+                    8. make_dir
+                    9. make_reg
+                    10. make_sock
+                    11. make_fifo
+                    12. make_block
+                    13. make_sym
+                2. ABI version >= 2:
+                    1. refer
+                3. ABI version >= 3:
+                    1. truncate
+            * **`paths`** (array of strings, OPTIONAL) is an array of files or parent directories of the file hierarchies to restrict.
+        * **`portBeneath`** (array of objects, OPTIONAL) is an array of the network-hierarchy typed rules.
+        Entries in the array contain the following properties:
+            * **`allowedAccess`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
+                1. ABI version >= 4:
+                    1. bind
+                    2. connect
+            * **`ports`** (array of strings, OPTIONAL) is an array of network ports to restrict.
+    * **`disableBestEffort`** (bool, OPTIONAL) the `disableBestEffort` field disables the best-effort security approach for Landlock access rights.
+    This is for conditions when the Landlock access rights explicitly configured by the container are not supported or available in the running kernel.
+    If the best-effort security approach is enabled (`false`), the runtime SHOULD enforce the strongest rules configured up to the current kernel support, and only be [logged as a warning](runtime.md#warnings) for those not supported.
+    If disabled (`true`), the runtime MUST [generate an error](runtime.md#errors) if one or more rules specified by the container is not supported.
+	Default is `false`, i.e., following a best-effort security approach.
 
 ### <a name="configUser" />User
 
@@ -396,6 +442,79 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
     "ioPriority": {
         "class": "IOPRIO_CLASS_IDLE",
         "priority": 4
+    },
+    "landlock": {
+        "ruleset": {
+            "handledAccessFS": [
+                "execute",
+                "write_file",
+                "read_file",
+                "read_dir",
+                "remove_dir",
+                "remove_file",
+                "make_char",
+                "make_dir",
+                "make_reg",
+                "make_sock",
+                "make_fifo",
+                "make_block",
+                "make_sym",
+                "refer",
+                "truncate"
+            ],
+            "handledAssessNetwork": [
+                "bind",
+                "connect"
+            ]
+        },
+        "rules": {
+            "pathBeneath": [
+                {
+                    "allowedAccess": [
+                        "execute",
+                        "read_file",
+                        "read_dir"
+                    ],
+                    "paths": [
+                        "/usr",
+                        "/bin"
+                    ]
+                },
+                {
+                    "allowedAccess": [
+                        "execute",
+                        "write_file",
+                        "read_file",
+                        "read_dir",
+                        "remove_dir",
+                        "remove_file",
+                        "make_char",
+                        "make_dir",
+                        "make_reg",
+                        "make_sock",
+                        "make_fifo",
+                        "make_block",
+                        "make_sym"
+                    ],
+                    "paths": [
+                        "/tmp"
+                    ]
+                }
+            ],
+            "portBeneath": [
+                {
+                    "allowedAccess": [
+                        "bind",
+                        "connect"
+                    ],
+                    "ports": [
+                        80,
+                        443
+                    ]
+                }
+            ]
+        },
+        "disableBestEffort": false
     },
     "noNewPrivileges": true,
     "capabilities": {
@@ -1151,7 +1270,8 @@ Here is a full example `config.json` for reference.
 
 [apparmor]: https://wiki.ubuntu.com/AppArmor
 [cgroup-v1-memory_2]: https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt
-[selinux]:https://selinuxproject.org/page/Main_Page
+[selinux]:http://selinuxproject.org/page/Main_Page
+[landlock]: https://landlock.io
 [no-new-privs]: https://www.kernel.org/doc/Documentation/prctl/no_new_privs.txt
 [proc_2]: https://www.kernel.org/doc/Documentation/filesystems/proc.txt
 [umask.2]: https://pubs.opengroup.org/onlinepubs/009695399/functions/umask.html
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -166,13 +166,13 @@
                 "landlock": {
                     "type": "object",
                     "properties": {
-                        "ruleset": {
-                            "$ref": "defs.json#/definitions/LandlockRuleset"
+                        "handledAccess": {
+                            "$ref": "defs.json#/definitions/LandlockHandledAccess"
                         },
                         "rules": {
                             "$ref": "defs.json#/definitions/LandlockRules"
                         },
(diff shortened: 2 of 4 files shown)


Zheaoli:
cc @l0kod @AkihiroSuda @vbatts @gnoack @cyphar

AkihiroSuda:
cc @kailun-qin (author of #1111) PTAL

AkihiroSuda:
Before merging this PR, can we have a POC implementation of runc (or crun, youki) to confirm the design?

Zheaoli:
https://github.com/opencontainers/runc/pull/3194 POC is here. Still need polish some code and add tests for it.

I will carry it

config.md
```diff
@@ -340,6 +340,26 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
+* **`landlock`** (object, OPTIONAL) specifies the Landlock unprivileged access control settings for the container process.
```
kailun-qin, line 343: I think this PR should be carefully updated following the latest advances of landlock (rather than simply carried) as some of the spec proposals might be stale now. Pls see https://docs.kernel.org/userspace-api/landlock.html and https://github.com/landlock-lsm/go-landlock for details.

Zheaoli: Do you got time to continue this PR or I can continue carry(

kailun-qin: Pls go ahead updating this PR. I'll need some time to follow up and dive into the updates of landlock (even for review).

l0kod:
Thanks for working on this!

You might want to take a look at this talk about backward and forward compatibility challenges: https://archive.fosdem.org/2023/schedule/event/rust_backward_and_forward_compatibility_for_security_features/

FYI, I plan to review these changes next week.

gnoack:
> Thanks for working on this!
> 
> You might want to take a look at this talk about backward and forward compatibility challenges: https://archive.fosdem.org/2023/schedule/event/rust_backward_and_forward_compatibility_for_security_features/
> 
> FYI, I plan to review these changes next week.

FWIW the most common implementation problem so far has been the use of the "refer" right, whose logic unfortunately works slightly differently than for other access rights.  I tried to explain this in https://blog.gnoack.org/post/landlock-best-effort/ last year.

With Linux 6.7, Landlock also has gained support for restricting the use of bind(2) and connect(2) for TCP.  (Landlock ABI version 4)

Zheaoli:
> FYI, I plan to review these changes next week.

Thanks for the important information, I will update this spec in this week ASAP I can

utam0k:
I wonder what the behavior should be when a new process(e.g. docker exec) jumps into the existing container.

Zheaoli:
I have update the  PR,  PTAL @kailun-qin @gnoack @l0kod

specs-go/config.go
```diff
@@ -94,7 +94,86 @@ type Process struct {
...
+	// AllowedAccess contains a list of allowed network actions for the network sockets.
+	AllowedAccess []LandlockNetworkAction `json:"allowedAccess,omitempty" platform:"linux"`
+	// Ports are the network ports to restrict.
+	Ports []string `json:"ports,omitempty" platform:"linux"`
+}
+
+// LandlockFSAction used to specify the FS actions that are handled by a ruleset or allowed by a rule.
+type LandlockFSAction string
+
+// Define actions on files and directories that Landlock can restrict a sandboxed process to.
```
gnoack, line 152: Documentation nit: The word "define" is usually unnecessary in docstrings (all of the exported symbols are defining something after all)

specs-go/config.go
```diff
@@ -94,7 +94,86 @@ type Process struct {
...
+}
+
+type LandlockRulePortBeneath struct {
+	// AllowedAccess contains a list of allowed network actions for the network sockets.
+	AllowedAccess []LandlockNetworkAction `json:"allowedAccess,omitempty" platform:"linux"`
+	// Ports are the network ports to restrict.
+	Ports []string `json:"ports,omitempty" platform:"linux"`
+}
+
+// LandlockFSAction used to specify the FS actions that are handled by a ruleset or allowed by a rule.
```
gnoack, line 149: "LandlockFSAction specifies the FS actions that are handled by a ruleset or allowed by a rule."

("used to specify" is also past tense, which is a bit confusing to read)

specs-go/config.go
```diff
@@ -94,7 +94,86 @@ type Process struct {
...
+	// forbidden if no rule explicitly allow them.
+	HandledAccessNetwork []LandlockNetworkAction `json:"handledAccessNetwork,omitempty" platform:"linux"`
+}
+
+// LandlockRules represents the security policies (i.e., actions allowed on objects).
+type LandlockRules struct {
+	// PathBeneath specifies the file-hierarchy typed rules.
+	PathBeneath []LandlockRulePathBeneath `json:"pathBeneath,omitempty" platform:"linux"`
+	// PortBeneath specifies the network-socket typed rules.
+	PortBeneath []LandlockRulePortBeneath `json:"portBeneath,omitempty" platform:"linux"`
```
gnoack, line 130: The two types of rules that exist in Landlock are "path beneath" and "net port", in the Linux headers.  I would encourage you to stay consistent with that naming, as it will make it easier for users to map your API to the Landlock documentation.

The word "beneath" is referring to the hierarchical nature of file systems -- when an access right is granted on a directory, it also applies to the files *beneath* that directory.  The word "beneath" is inappropriate when used together with ports, as we have no such hierarchy there.

specs-go/config.go
```diff
@@ -94,7 +94,86 @@ type Process struct {
...
+	// Rules are the security policies (i.e., actions allowed on objects) to be added to an existing ruleset.
+	Rules *LandlockRules `json:"rules,omitempty" platform:"linux"`
+	// DisableBestEffort disables the best-effort security approach for Landlock access rights.
+	// This is for conditions when the Landlock access rights explicitly configured by the container are not
+	// supported or available in the running kernel.
+	// Default is false, i.e., following a best-effort security approach.
+	DisableBestEffort bool `json:"disableBestEffort,omitempty" platform:"linux"`
+}
+
+// LandlockRuleset identifies a set of rules (i.e., actions on objects) that need to be handled.
```
gnoack, line 115: Documentation nit: "The Ruleset identifies a set of rules" is a trivial statement which does not provide much additional value, IMHO.

gnoack: BTW, I think the naming of the "Landlock" and "LandlockRuleset" types is a bit backwards: In Landlock, the "ruleset" (which is a file descriptor under the hood), *is* actually the thing where users add their "path beneath" and "net port" rules.  (Whereas, in this proposal, the "Ruleset" type does not actually hold the rules...? %-))

The "handled access rights" are just the information needed when creating the initial (empty) ruleset.  If the kernel, the struct with the access rights is called the "ruleset attributes", but you might as well call it "handled access rights" or something like that, if you have the flexibility in your code to change that name later on.

schema/defs.json
```diff
@@ -165,6 +176,98 @@
...
+                }
+            }
+        },
+        "ArrayOfLandlockRulePathBeneaths": {
+            "type": "array",
+            "items": {
+                "$ref": "#/definitions/LandlockRulePathBeneath"
+            }
+        },
+        "ArrayOfLandlockRulePortBeneaths": {
```
gnoack, line 258: See comment on the other file about the word "beneath" when talking about network ports

config.md
```diff
@@ -340,6 +340,52 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+    Note that `noNewPrivileges` must be set to true to use this feature.
+    For more information about Landlock, see [Landlock documentation][landlock].
+    `landlock` contains the following properties:
+
+    * **`ruleset`** (object, OPTIONAL) the `ruleset` field identifies a set of rules (i.e., actions on objects) that need to be handled (i.e., restricted).
+    The `ruleset` currently contains the following types:
+        * **`handledAccessFS`** (array of strings, OPTIONAL) is an array of FS typed actions that are handled by a ruleset.
+        If no rule explicitly allow them, they should then be forbidden.
+        * **`handledAssessNetwork`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are handled by a ruleset. (The NETWORK typed actions are avaliable when the ABI version >= 4. the behavior of the NETWORK typed actions is not used when the ABI version is less than 4 will depend on the **`disableBestEffort`**)
```
gnoack, line 352: I'm sure you meant "Access" here?

config.md
```diff
@@ -340,6 +340,52 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+                    1. truncate
+            * **`paths`** (array of strings, OPTIONAL) is an array of files or parent directories of the file hierarchies to restrict.
+        * **`portBeneath`** (array of objects, OPTIONAL) is an array of the network-hierarchy typed rules.
+        Entries in the array contain the following properties:
+            * **`allowedAccess`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
+                1. ABI version >= 4:
+                    1. bind
+                    2. connect
+            * **`ports`** (array of strings, OPTIONAL) is an array of network ports to restrict.
+    * **`disableBestEffort`** (bool, OPTIONAL) the `disableBestEffort` field disables the best-effort security approach for Landlock access rights.
```
gnoack, line 384: (This is slightly confusing in my mind that the boolean condition has the "inverted" logic here.  I assume that in this config format, it is not possible to call this variable "bestEffortFallback" and to make `true` the default value instead?)

gnoack:
High level remark: Looks good so far.  I added a bunch of documentation nits.

As for Landlock-specific comments, I would recommend to re-think some of the type names, and I left comments in these places.

I think the bigger chance for mistakes is in the implementation itself.  - Are you planning to add that in the same PR, or will that come separately?

Zheaoli:
> I think the bigger chance for mistakes is in the implementation itself. - Are you planning to add that in the same PR, or will that come separately?

I think it would be great to implement it in a single PR(Actually, I don't know which is better to split the PR. Depend on the ABI version or else)

config.md
```diff
@@ -340,6 +340,52 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
+* **`landlock`** (object, OPTIONAL) specifies the Landlock unprivileged access control settings for the container process.
```
rata, line 343: What happens if the runtime doesn't know this field? I guess in runc it will be completely ignored, right? In that case, I think we at least need to expose landlock support via the features command too.

config.md
```diff
@@ -340,6 +340,52 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+    Note that `noNewPrivileges` must be set to true to use this feature.
+    For more information about Landlock, see [Landlock documentation][landlock].
+    `landlock` contains the following properties:
+
+    * **`ruleset`** (object, OPTIONAL) the `ruleset` field identifies a set of rules (i.e., actions on objects) that need to be handled (i.e., restricted).
+    The `ruleset` currently contains the following types:
+        * **`handledAccessFS`** (array of strings, OPTIONAL) is an array of FS typed actions that are handled by a ruleset.
+        If no rule explicitly allow them, they should then be forbidden.
```
rata, line 351: How does this work on overlayfs? https://docs.kernel.org/userspace-api/landlock.html#bind-mounts-and-overlayfs

It is very widely used, can you clarify what works and what doesn't?

rata (commented):
> [opencontainers/runc#3194](https://github.com/opencontainers/runc/pull/3194) POC is here. Still need polish some code and add tests for it.
> 
> I will carry it

As there are no tests there, can you share the example config you used to test it?

I'm curious about overlayfs and the behavior in old runtimes.

config.md
Zheaoli: Yes, the runtime will ignore the field. I think it's not necessary to add feature command like other optional features.

config.md
Zheaoli: Actually, I'm not testing it on overlayfs yet. But I think it still work on overlayfs. Because we will always target to restrict the rule on the merged hierarchy.

Zheaoli: Maybe you want to figure out is it working on the following circumstance?

1. We got directories A and B, A and B will merged into C
3. We give the process write permission for A and B
4. The process still have write permission for C

config.md
rata: @Zheaoli why not? This is a security feature, I can see use cases where upper layers want to run enforce this.

config.md
rata: My understandig of the link is, if C is the merged dir by overlayfs, then no, you don't have permissions to write even if you can write to A and B.

But I think if we use C for the rules, it might work. I don't know if we also need permissions in A and B (not 100% clear to me when reading the doc), in which case it will be tricky.

IMHO, it will be nice to understand that, as overlay is a widely used fs for containers. But not sure that will change something on how we support it in OCI.

Zheaoli: https://github.com/Zheaoli/py-landlock/blob/main/examples/demo.py

I have some overlayfs test case here. PTAL if you got time

rata: @Zheaoli I can try to download the python modules and run that locally. But it would be way simpler if you just share what works and what doesn't :)

At the OCI layer we don't know the lowerdir nor the upperdir, we just know the merged dir. It will be great to find if we apply a landlock policy restricting the merged dir, what effect have on reading AND writing files.

Knowing if it works, or its limitations, is key to evaluate a PR.

Zheaoli: Sorry for the confusion. Here's the full test result:

1. A and B are lower dirs, and C is a merged dirs.
2. Policy to C would no longer affect A and B. The reverse action would be the same(policy on A and B would not affect C). 
3. The write and read (or any permission) on C is not depend on the permission on A and B. The permission between C and AB is independent

rata: @Zheaoli thanks! This LGTM.

I was playing with this, to make sure I'm not missing any case and indeed it seems to work just fine.

To play further, I used the example from: https://github.com/landlock-lsm/go-landlock/blob/efb66220540a9ef86aa0160d15e55f429d5b94d9/examples/go-landlock-configurable/main.go

With this json config:
```
JSON config:
{
  "forbidden_access": [
    "read_dir",
    "write_file",
    "make_reg"
  ],
  "exceptions": [
    {
      "paths": [
        "./rootfs-merged/tmp"
      ],
      "permitted_access": [
        "read_dir",
        "make_reg",
        "write_file"
      ]
    }
  ],
  "best_effort": false
}
```

And rootfs-merged is an overlayfs that I mounted manually. I tried it works as expected (can't create files on any paths except on ./rootfs-merged/tmp/, can't ls/read files except in that path), also if I chroot into that (`chroot rootfs-merged /bin/sh`).

The files are created by overlayfs in the upper dir and all is as expected.

config.md
```diff
@@ -340,6 +340,52 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+                    10. make_sock
+                    11. make_fifo
+                    12. make_block
+                    13. make_sym
+                2. ABI version >= 2:
+                    1. refer
+                3. ABI version >= 3:
+                    1. truncate
+            * **`paths`** (array of strings, OPTIONAL) is an array of files or parent directories of the file hierarchies to restrict.
+        * **`portBeneath`** (array of objects, OPTIONAL) is an array of the network-hierarchy typed rules.
```
rata, line 377: portBeneath? Is this for all ports below the number? Or is this just a c&p mistake?

rata (commented):
I had a cursory look, I need to look later at what field are exactly proposed. But this seems in the right direction to me.

Added some comments/issues that I see.

config.md
Zheaoli: Sorry for replying late. I misunderstand something before. I think it's necessary to export the feature we support for landlock in features command

config.md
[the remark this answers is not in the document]
Zheaoli: Yep, we can remove the disableBestEffort field and export the features we support in features command instead

config.md
Zheaoli: Yes, it's for all ports below the number. Actually I follow style with `pathBeneath`

rata: Hmm, okay, thanks. I'm in an off-site this week, so I'll check this up next week.

config.md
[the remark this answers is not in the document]
rata: We can do boths, right?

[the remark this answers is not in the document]
Zheaoli: maybe we can output the feature check like this ?

```json
{
    "linux": {
        "landlock": {
            "enabled": true,
            "bestEffort": {
                "enabled": true // False
            },
            "fs":{
                "enabled": true,
                "actions": [
                    "exectute",
                    "write_file",
                    "read_file",
                    "read_dir",
                    "remove_dir",
                    "remove_file",
                    "make_char",
                    "make_dir",
                    "make_reg",
                    "make_sock",
                    "make_fifo",
                    "make_block",
                    "make_sym",
                    "refer",
                    "truncate"
                ]
            },
            "network": {
                "enabled": true,
                "actions": [
                    "bind",
                    "connect"
                ]
            }
        }
    }
}
```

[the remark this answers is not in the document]
rata: I don't think we need to expose all of that. I think just the landlock object with the enabled field will be enough.

Then, if the runtime doesn't recognize some of the constants or if the kernel ABI is not the one needed, then make the runtime fail?

config.md
rata: I can't make sense out of that.

If I want to have a rule for port 21, 22, 53 and 80, it seems super complicated to have a portBeneath rule. For filesystems that makes sense, because is all beneath this (i.e. subdirs and everything below that), this is applied.

But what does it mean for ports? It seems like a very complicated way to express it.

Oh, it seems @gnoack already commented on this. I fully agree: https://github.com/opencontainers/runtime-spec/pull/1241/files#r1508607132

rata (commented):
@Zheaoli thanks for the PR! Added some comments.

I think the config.json interface needs more work, but I'd love for this to be ready.

I read @gnoack comments, that weren't addressed, and I fully agree with them. Please take a look their review comments.

AkihiroSuda:
Needs rebase

config.md
gnoack: Thanks, exactly -- this is still confusing and would better be named "net port" [in line with the `landlock.h` header file](https://github.com/torvalds/linux/blob/master/include/uapi/linux/landlock.h#L98).

It does *not* permit access for all ports below a given number, but only for that specific port number itself, and the word "hierarchy" is also misleading, as there is nothing hierarchical about that (probably copied over from the file system access rights documentation).

AkihiroSuda:
Needs another rebase

AkihiroSuda:
@Zheaoli Could you rebase this? 🙏 
Let's merge this soon

Zheaoli:
> Could you rebase this? 🙏
> Let's merge this soon

Sure, I will carry this today and fix the review idea during this week.

(BTW I'm watching new job opportunities)(Free man today

Zheaoli:
cc @AkihiroSuda @gnoack @utam0k @rata I made a huge update based on the reivew idea. PTAL

utam0k:
I'd like to know the use cases with LandLock. 
I'm worried that it will become a feature that no one uses, and it will only increase the burden of maintenance. I'm not familiar with this, so please let me know 🙏

AkihiroSuda:
Can we get the POC https://github.com/opencontainers/runc/pull/3194 synced with this PR, so as to verify the feasibility and usability?

Zheaoli:
> Can we get the POC [opencontainers/runc#3194](https://github.com/opencontainers/runc/pull/3194) synced with this PR, so as to verify the feasibility and usability?

SGTM, Let me carry the runc#3194

config.md
rata: This wasn't done, right?

config.md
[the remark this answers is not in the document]
rata: ?

rata:
@Zheaoli thanks! It's very hard to review if you just "do huge updates" but don't really answer the comments. It's not easy to know what you tackled, what you didn't and why, and what you might just forgotten.

I had a quick look, it does look better from a 10k feet point of view, but the part of exposing this in features is completely missed. I fear without it it might not be possible to use it in some scenarios, as old runtimes without this will just ignore this. This doesn't play well with the `disable_best_effort` thing. It seems to do one thing (enforce it), but it doesn't (the old runtime will just ignore things it doesn't know about).

rata (changes requested):
This is a great improvement on the previous iterations, but I don't think we should merge this, I think we still need changes :(

Besides some errors on how landlock behaves (and some typos that I removed the comments to make the review shorter, as there are other more important blockers IMHO), I don't like the idea of having the possible words in the spec as the only way to configure landlock (e.g. read_file and all allowed_access, the same for the network section, etc.).

In a lot of kernel releases (I think most) there were landlock changes, I mentioned the example of ioctls as the last recent example (this PR was updated in Dec and is already old). I think we need a better way to have something that is more "future proof", because the linux kernel is released every 6-8 weeks, we can't update the spec and runtimes so often for any new thing that can be restricted in a new kernel.

I suggested using the plain numbers, as defined in the landlock.h uapi header. This way, we can expose nice names in the go bindings, but users like containerd and so can still use new features if they just use the numbers (they can define a friendly constant for the number, etc.). This removes the runtime-spec and runtimes to create new releases to support every new "string" supported. We will need a new release for things like "now we can restrict a whole new category", like the network stuff  that was added at some point in the past, but that isn't as often and seems like a reasonable trade-off for me. But I'm open to other ideas, of course :-)

config.md
```diff
@@ -352,6 +352,52 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+                   1. truncate
+           * **`paths`** (array of strings, OPTIONAL) is an array of files or parent directories of the file hierarchies to restrict.
+       * **`networkPort`** (array of objects, OPTIONAL) is an array of the network socket rules.
+       Entries in the array contain the following properties:
+           * **`allowedAccess`** (array of strings, OPTIONAL) is an array of NETWORK typed actions that are allowed by a rule. The actions are grouped by the ABI version in the following description:
+               1. ABI version >= 4:
+                   1. bind
+                   2. connect
+           * **`ports`** (array of strings, OPTIONAL) is an array of network ports to restrict.
+   * **`enableBestEffort`** (bool, OPTIONAL) the `enableBestEffort` field disables the best-effort security approach for Landlock access rights.
```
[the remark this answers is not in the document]
l0kod: I agree, we need to be careful to not give users a false sense of security.

rata:
@l0kod +1 on not creating two different json specs to describe a landlock policy. I'll monitor this for when you post the link to the experimental release :)

l0kod:
Here is an initial experimental version: https://github.com/landlock-lsm/landlockconfig
I'm working on improving the compatibility configuration to make it simpler to tweak (if required).

Zheaoli:
> Here is an initial experimental version: https://github.com/landlock-lsm/landlockconfig
> I'm working on improving the compatibility configuration to make it simpler to tweak (if required), see [landlock-lsm/landlockconfig#3](https://github.com/landlock-lsm/landlockconfig/issues/3).

Thanks for the work. I think it would be great if we have a common landlock spec

config.md
```diff
@@ -352,6 +352,52 @@ For Linux-based systems, the `process` object supports the following process-spe
       for `initial`. If omitted or empty, runtime SHOULD NOT change process'
       CPU affinity after the process is moved to container's cgroup, and the
       final affinity is determined by the Linux kernel.
+* **`landlock`** (object, OPTIONAL) specifies the Landlock unprivileged access control settings for the container process.
```
rata, line 355: I think this doesn't belong to this file. This should be in config-linux. The same with the secion of the struct this is, this should be inside the Linux section, as seccomp is. Not as a general feature, as this is a linux-only feature.

specs-go/config.go
```diff
@@ -96,8 +96,86 @@ type Process struct {
 	IOPriority *LinuxIOPriority `json:"ioPriority,omitempty" platform:"linux"`
 	// ExecCPUAffinity specifies CPU affinity for exec processes.
 	ExecCPUAffinity *CPUAffinity `json:"execCPUAffinity,omitempty" platform:"linux"`
+	// Landlock specifies the Landlock unprivileged access control settings for the container process.
+	// `noNewPrivileges` must be enabled to use Landlock.
+	Landlock *Landlock `json:"landlock,omitempty" platform:"linux"`
```
rata, line 101: Idem, this is the wrong section. This should be inside the linux-only part, not the generic part. This only applies to linux

rata:
Added some other comments so the next iteration/PR doesn't do those mistakes too.

config.md
rata: Ouch, there is a section for Linux here too. Sorry!

specs-go/config.go
rata: Ouch, there is a section for Linux here too. Sorry!

rata:
@l0kod I'm no maintainer here, but having a quick look the json schema seems mostly okay to me. Some comments:
1. IIUC it is not handling numbers for the constants in landlock.h from uapi. I guess you want to change that?
1. The `additionalproperties: false` is not something we can do in runtime-spec, sadly. I think the lib will need to have a way to accept additional properties to be able to use it in OCI runtimes like runc or crun. Maaaybe there is some edge case, I haven't look in detail
1. What would be the difference in practice between "soft_requirement" and "best_effort"? Maybe an example that behaves differently by changing this option?

Thanks!

l0kod:
> @l0kod I'm no maintainer here, but having a quick look the json schema seems mostly okay to me. Some comments:
> 
> 1. IIUC it is not handling numbers for the constants in landlock.h from uapi. I guess you want to change that?

Correct, I [reverted](https://github.com/landlock-lsm/landlockconfig/commit/be06b24f0dddcb7d4bf4c76f7329ac2ee45ce781) this part for now because it is not implemented yet, but I will.

> 2. The `additionalproperties: false` is not something we can do in runtime-spec, sadly. I think the lib will need to have a way to accept additional properties to be able to use it in OCI runtimes like runc or crun. Maaaybe there is some edge case where we can reject, I haven't look in detail. Let's look at that later, it seems something easy to change later too.

OK, I'd like to make it optional in the library then, it would be a shame to start a new schema with a legacy issue. :wink: 
It should be easy to only remove `additionalproperties: false` for the OCI runtime spec and still keep it in sync. Anyway, we'll need to integrate them somehow, so this could be part of the merge process.

> 3. What would be the difference in practice between "soft_requirement" and "best_effort"? Maybe an example that behaves differently by changing this option?

`soft_requirement` is to completely disable the sandbox, whereas `best_effort` is to enforce as much restrictions as possible. However, I plan to replace that with options that make more sense, see https://github.com/landlock-lsm/landlockconfig/issues/3#issuecomment-2694673887

rata:
@l0kod recently when discussing another rust lib to integrate in runc, there were some concerns with that from other maintainers.

@AkihiroSuda Do you have any thoughts on using a rust lib (with go bindings) for landlock support in runc?
other @opencontainers/runc-maintainers any thoughts on this?

haircommander:
@Zheaoli are you still interested in driving this? I believe there are still some open points to address

MikeZappa87:
@haircommander I would like to get this over the finish line. I can't dedicate a ton of time to this however I will try and keep this one alive

haircommander:
great thanks! cc @saschagrunert

schema/config-schema.json
```diff
@@ -163,6 +163,20 @@
                         }
                     }
                 },
+                "landlock": {
```
giuseppe, line 166: how to differentiate between applying landlock before pivot_root and after?  I believe both cases can be useful

l0kod:
FYI, I'm still working on stabilizing [LandlockConfig](https://github.com/landlock-lsm/landlockconfig). I would help to get some reviews of the current and ongoing spec.

schema/config-schema.json
rata: I agree both are useful and I'd like to have a way to do it before pivot root in the future. But today if you use landlock, you can't pivot_root or at least you can't mount IIRC

l0kod: Correct, pivot_root (and any privileged actions) should happen before Landlock enforcement, at least for now. However, the Landlock ruleset should be built before pivot_root to use absolute paths.

giuseppe: is that going to be relaxed at some point though?  I am asking because for the OCI specs, it will be easier to think about this now than have to worry later

l0kod: We can already create a Landlock sandbox that would allow pivot_root/mount operations but the cost is that this sandbox would not restrict filesystem operations (e.g. it can still restrict network, IPC, ptrace...). Fine-grained control of mount-related operations is an ongoing work though, but it will not land this year. What are you thinking about exactly?

giuseppe: I am only wondering/bikeshedding on how to make the specs future-proof.  As you've pointed out, there is on going work to have a better control of mount-related operations.  How will the OCI specs be affected by that once it lands?  Can we foresee that in the future it will be possible to use path restrictions after a pivot_root too or that is never going to change?

IMO, the OCI runtime specs should allow to define all the possible configurations, then it is up to the user to choose what and how to use them

l0kod: > Can we foresee that in the future it will be possible to use path restrictions after a pivot_root too or that is never going to change?

Landlock filesystem access control is independant from the FS namespace or the FS root, so path restrictions works after a pivot_root too. pivot_root is currently not allowed when there are FS access control in place because Landlock use ephemeral (e.g. not xattr) tags and identify file hierarchies, taking into account stacked sandboxes. If a sandboxed process would be allow to change itself its FS topology, the enforced restrictions would lose their initial semantic (i.e. the security policy might not be correct anymore).

That being said, the same way Landlock still allows (restricted) links/renames, it could also allow (limited) pivot_root and mount. Work an that direction started but we need to bring important changes: https://github.com/landlock-lsm/linux/issues/14

> IMO, the OCI runtime specs should allow to define all the possible configurations, then it is up to the user to choose what and how to use them

Yes, but we should be careful to not give the false impression of security neither, which means some combinations may be restricted.

giuseppe: Thanks, so that is exactly my point.  If pivot_root will be allowed, even with limitations, then the paths will look different since Landlock can be applied in two different moments: before the OCI runtime does the pivot_root, or after the pivot_root is done. In the former case they are absolute paths that resolve in the host namespace, in the latter the paths are relative to the container root, so we need to distinguish between the two cases.

PR #1252, opened 2024-05-17
config: add annotation for exec isolated CPU affinity

bartwensley:
The runc container runtime recently introduced functionality to direct exec operations to the first isolated CPU in a container cgroup cpuset. This is enabled through a new runc specific container annotation: org.opencontainers.runc.exec.isolated-cpu-affinity-transition

See https://github.com/opencontainers/runc/commit/afc23e3397 for the details.

This new functionality is also required in the crun container runtime and will be added soon. This purpose of this commit is to add a new annotation to the OCI runtime specification that would be used by any runtime that implements this new functionality:
org.opencontainers.runtime.exec.isolated-cpu-affinity-transition

The runc implementation should switch to using the new annotation once it is accepted.


AkihiroSuda (changes requested):
This should be a proper JSON field

tianon:
Agree -- a runtime-specific annotation seems like a really great way for one runtime to experiment with adding functionality like this before it's fully standardized or proven out, but as soon as it's something desirable for interoperability, it seems worth making a proper "feature" instead.

PR #1253, opened 2024-05-18
Add CPU affinity to executed processes

kolyshkin:
This allows to set initial and final CPU affinity for a process being run in a container, which is needed to solve the issue described in https://github.com/opencontainers/runc/issues/3922.

This is an alternative to #1252.

### DIFF
--- a/config.md
+++ b/config.md
@@ -340,6 +340,17 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
+* **`execCPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+    This setting is not applicable to the container's init process.
+    The following properties are available:
+    * **`initial`** (string, OPTIONAL) is a list of CPUs a runtime parent
+      process to be run on initially, before the transition to container's
+      cgroup. This is a a comma-separated list, with dashes to represent
+      ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7.
+    * **`final`** (string, OPTIONAL) is a list of CPUs the process will be run
+      on after the transition to container's cgroup. The format is the same as
+      for `initial`. If omitted or empty, the container's default CPU affinity,
+      as defined by [cpu.cpus property](./config.md#configLinuxCPUs)), is used.
 
 ### <a name="configUser" />User
 
@@ -416,7 +427,11 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are 
             "hard": 1024,
             "soft": 1024
         }
-    ]
+    ],
+    "execCPUAffinity": {
+        "initial": "7",
+        "final": "0-3,7"
+    }
 }
 ```
 ### Example (Solaris)
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -94,6 +94,8 @@ type Process struct {
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
 	// IOPriority contains the I/O priority settings for the cgroup.
 	IOPriority *LinuxIOPriority `json:"ioPriority,omitempty" platform:"linux"`
+	// ExecCPUAffinity specifies CPU affinity for exec processes.
+	ExecCPUAffinity *CPUAffinity `json:"execCPUAffinity,omitempty" platform:"linux"`
 }
 
 // LinuxCapabilities specifies the list of allowed capabilities that are kept for a process.
@@ -126,6 +128,12 @@ const (
 	IOPRIO_CLASS_BE   IOPriorityClass = "IOPRIO_CLASS_BE"
 	IOPRIO_CLASS_IDLE IOPriorityClass = "IOPRIO_CLASS_IDLE"
 )
+
+// CPUAffinity specifies process' CPU affinity.
+type CPUAffinity struct {
+	Initial string `json:"initial,omitempty"`
+	Final   string `json:"final,omitempty"`
+}
 
 // Box specifies dimensions of a rectangle. Used for specifying the size of a console.
 type Box struct {
--- a/config-linux.md
+++ b/config-linux.md
@@ -395,8 +395,8 @@ The following parameters can be specified to set up the controller:
 * **`period`** *(uint64, OPTIONAL)* - specifies a period of time in microseconds for how regularly a cgroup's access to CPU resources should be reallocated (CFS scheduler only)
 * **`realtimeRuntime`** *(int64, OPTIONAL)* - specifies a period of time in microseconds for the longest continuous period in which the tasks in a cgroup have access to CPU resources
 * **`realtimePeriod`** *(uint64, OPTIONAL)* - same as **`period`** but applies to realtime scheduler only
-* **`cpus`** *(string, OPTIONAL)* - list of CPUs the container will run in
-* **`mems`** *(string, OPTIONAL)* - list of Memory Nodes the container will run in
+* **`cpus`** *(string, OPTIONAL)* - list of CPUs the container will run on. This is a comma-separated list, with dashes to represent ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7.
+* **`mems`** *(string, OPTIONAL)* - list of memory nodes the container will run on. This is a comma-separated list, with dashes to represent ranges. For example, `0-3,7` represents memory nodes 0,1,2,3, and 7.
 * **`idle`** *(int64, OPTIONAL)* - cgroups are configured with minimum weight, 0: default behavior, 1: SCHED_IDLE.
 
 #### Example
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -220,7 +220,20 @@
                             }
                         }
                     }
-                }
+                },
+                "execCPUAffinity": {
+                    "type": "object",
+                    "properties": {
+                        "initial": {
+                            "type": "string",
+			    "pattern": "^[0-9, -]*$"
+                        },
+                        "final": {
+                            "type": "string",
+			    "pattern": "^[0-9, -]*$"
+                        }
+                    }
+		}
             }
         },
         "linux": {


config.md
```diff
@@ -340,6 +340,10 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
+* **`cpuAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+    The following properties are available:
+    * **`initial`** (string, REQUIRED) is a list of CPUs the process will be run on initially, before the transition to container's cgroup.
```
AkihiroSuda, line 345: The string format needs to be explained (with an example)

kolyshkin: Thanks; updated.

AkihiroSuda (approved):

config.md
```diff
@@ -340,6 +340,17 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
+* **`cpuAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+    The following properties are available:
+    * **`initial`** (string, REQUIRED) is a list of CPUs a runtime parent
+      process to be run on initially, before the transition to container's
+      cgroup. This is a a comma-separated list, with dashes to represent
+      ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7.
```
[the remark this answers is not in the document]
kolyshkin: Quite the opposite, this is for exec only. In runc terms, this is a CPU affinity used to start `runc init` (see e.g. https://github.com/opencontainers/runc/blob/main/libcontainer/process_linux.go#L1219).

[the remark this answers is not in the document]
giuseppe: thanks.  I guess we can do something similar to `startCommandWithCPUAffinity` when `CLONE_INTO_CGROUP` is used.

giuseppe (approved):
LGTM

kolyshkin:
@opencontainers/runtime-spec-maintainers PTAL (I'm going to merge this this week if there are no other comments)

tianon (approved):

bartwensley (approved):
LGTM

config.md
```diff
@@ -340,6 +340,17 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+    * **`final`** (string, OPTIONAL) is a list of CPUs the process will be run
+      on after the transition to container's cgroup. The format is either the
+      same as for `initial`, or a special value `cgroup` which means container's
+      CPU affinity, as defined by [cpu.cpus property](./config.md#configLinuxCPUs)).
+      If omitted or empty, the `initial` setting is kept.
```
kad, line 353: shouldn't it be the opposite? I mean "if ommitted, cpu.cpus will be in effect"?
The `initial` seems to allow to set affinity to cpu cores that are outside of the set which is specified by `cpu.cpus`.
This also trigger the question of validation of the values between `cpuAffinity` and `cpu.cpus`.

kolyshkin:
@cclerget PTAL

utam0k (approved):
LGTM. It's interesting.

config.md
kolyshkin: There are two questions here.

1. What is the default if `final` is not set. You are probably right that `cpu.cpus` makes more sense. UPDATED.

2. Whether to allow affinity that's outside of container's cpu.cpus. While it seems theoretically possible for the `initial` cpuset, I haven't checked that it works, and I don't want to add any specific constraints to the spec (or to runc/crun).

kolyshkin: PTAL @kad

cclerget (approved):
LGTM

kad:
@kolyshkin please, give us some time, we would like to do few experiments with @klihub, as we also observed corner cases scenarios with DPDK-style apps in the past, but we need some time to replicate that environment and see behaviour in case of invalid arguments between requests via cgroups or `setaffinity`.

Regarding PR in general:
1. naming of the fields: `initial` and `final`: would it make sense to have more consistent names with other parts of the spec that corresponds to same lifecycle stages of the container? e.g. normalize names with hook names: `createRuntime`/`createContainer` (or `startContainer`)? This also helps spec consumers to understand at which stage this CPU affinity is supposed to be applied in container lifecycle.
2. `initial` seems to be mandatory argument right now, even were it is applied before entering container namespace. We seen in the past that bigger problem for execution inside container namespace, not before. I think it would make more sense if both `initial`/`final` will be optional. WDYT?
3. Overall error handling with `setaffinity()`: it might be good to add clause in the description that incorrect setting of those parameters would lead to container creation failure? because if spec requests some value in those fields, but corresponding function call fails (invalid arguments/permissions/etc), we can't silently ignore such errors.

kolyshkin:
@kad thanks for reviewing this!

> 1. naming of the fields: `initial` and `final`: would it make sense to have more consistent names with other parts of the spec that corresponds to same lifecycle stages of the container? e.g. normalize names with hook names: `createRuntime`/`createContainer` (or `startContainer`)? This also helps spec consumers to understand at which stage this CPU affinity is supposed to be applied in container lifecycle.

I think you are misunderstanding. These two are for `exec`, not `create/start/run`. The problem we're trying to solve with these is laid out at https://github.com/opencontainers/runc/issues/3922.

> 2. `initial` seems to be mandatory argument right now, even were it is applied before entering container namespace. We seen in the past that bigger problem for execution inside container namespace, not before. I think it would make more sense if both `initial`/`final` will be optional. WDYT?

See, they (`initial` and `final`) are both part of `cpuAffinity` object, which itself is optional. By saying that `initial` is mandatory we're just saying that if `cpuAffinity` is specified, it should not be empty.

> 3. Overall error handling with `setaffinity()`: it might be good to add clause in the description that incorrect setting of those parameters would lead to container creation failure? because if spec requests some value in those fields, but corresponding function call fails (invalid arguments/permissions/etc), we can't silently ignore such errors.

Again, this is about `exec` not `create/start/run`. Unfortunately, this distinction (between starting container's PID 1 and executing an arbitrary process in a running container) is not laid out very well in the runtime-spec.

I think it should be up to implementation what to do it case of errors. IOW the spec should not be way too rigid. Maybe some implementation could have a flag saying "ignore this" or "ignore error from that" or "if this won't work, fall back to that" etc. If the spec is too rigid, such hacks render an implementation incompatible, which I guess is not good.

kolyshkin:
I guess I am merging this later today. @kad if you have something, please open a followup PR.

kolyshkin:
We also need to add a flag into runtime features.

kolyshkin:
> We also need to add a flag into runtime features.

I can't find a good place to add such a flag, so maybe the upper-level runtime (cri-o, containerd) can check the `features.ociVersionMax field instead`? WDYT @haircommander

haircommander:
if the feature piece isn't sanctified in the runtime spec that's fine, as long as runtimes that support it agree on the name of the feature

kolyshkin:
@kad thanks, changed like this:
```diff
diff --git a/config.md b/config.md
index eab6eec..b9b5573 100644
--- a/config.md
+++ b/config.md
@@ -340,9 +340,10 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
-* **`cpuAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+* **`execCPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+    This setting is not applicable to the container's init process.
     The following properties are available:

-    * **`initial`** (string, REQUIRED) is a list of CPUs a runtime parent
+    * **`initial`** (string, OPTIONAL) is a list of CPUs a runtime parent
       process to be run on initially, before the transition to container's
       cgroup. This is a a comma-separated list, with dashes to represent
       ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7.
@@ -427,7 +427,7 @@ _Note: symbolic name for uid and gid, such as uname and gname respectively, are
             "soft": 1024
         }
     ],
-    "cpuAffinity": {
+    "execCPUAffinity": {
         "initial": "7",
         "final": "0-3,7"
     }
diff --git a/schema/config-schema.json b/schema/config-schema.json
index 39cc71e..cb74342 100644
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -221,11 +221,8 @@
                         }
                     }
                 },
-                "cpuAffinity": {
+                "execCPUAffinity": {
                     "type": "object",
-                    "required": [
-                        "initial"
-                    ],
                     "properties": {
                         "initial": {
                             "type": "string",
diff --git a/specs-go/config.go b/specs-go/config.go
index 00bf946..290b6dc 100644
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -94,8 +94,8 @@ type Process struct {
        SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
        // IOPriority contains the I/O priority settings for the cgroup.
        IOPriority *LinuxIOPriority `json:"ioPriority,omitempty" platform:"linux"`
-       // CPUAffinity specifies CPU affinity for the process.
-       CPUAffinity *CPUAffinity `json:"cpuAffinity,omitempty" platform:"linux"`
+       // ExecCPUAffinity specifies CPU affinity for exec processes.
+       ExecCPUAffinity *CPUAffinity `json:"execCPUAffinity,omitempty" platform:"linux"`
 }
 
 // LinuxCapabilities specifies the list of allowed capabilities that are kept for a process.
```

specs-go/config.go
```diff
@@ -127,6 +129,12 @@ const (
 	IOPRIO_CLASS_IDLE IOPriorityClass = "IOPRIO_CLASS_IDLE"
 )
 
+// CPUAffinity specifies process' CPU affinity.
+type CPUAffinity struct {
+	Initial string `json:"initial"`
```
kad, line 134: ,omitempty?

config.md
```diff
@@ -340,6 +340,17 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+    This setting is not applicable to the container's init process.
+    The following properties are available:
+    * **`initial`** (string, OPTIONAL) is a list of CPUs a runtime parent
+      process to be run on initially, before the transition to container's
+      cgroup. This is a a comma-separated list, with dashes to represent
+      ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7.
+    * **`final`** (string, OPTIONAL) is a list of CPUs the process will be run
+      on after the transition to container's cgroup. The format is the same as
+      for `initial`. If omitted or empty, the container's default CPU affinity,
+      as defined by [cpu.cpus property](./config.md#configLinuxCPUs)), is used.
```
kad, line 353: Based on the tests, if you don't call `sched_setaffinity()`, the kernel itself will reset it to cgroups cpuset.cpu, so, maybe here statement can be simplified, that if not specified, after transition into container namespace, default affinity depends on kernel implementation?

kad (commented):
thanks for update. looks better now. I'm still not convinced about names `initial` and `final`, maybe something more descriptive can be done like `execRuntime` / `execContainer` for consistency with hooks?
But for that I'll let others to chime in.

config.md
kolyshkin: I think it is nicer to have a reference to another way of setting cpu affinity here. If removed, it still says the same thing:

```diff
       on after the transition to container's cgroup. The format is the same as
-      for `initial`. If omitted or empty, the container's default CPU affinity,
-      as defined by [cpu.cpus property](./config.md#configLinuxCPUs)), is used.
+      for `initial`. If omitted or empty, the container's default CPU affinity
+      applies.
 
```

so I think it's better to keep it.

kolyshkin: > Based on the tests, if you don't call sched_setaffinity(), the kernel itself will reset it to cgroups cpuset.cpu

@kad It seems your tests are wrong. Mine are showing this only happens when the initial (pre cgroup-move) affinity is not a subset of cgroup's affinity.

So maybe we need to change the spec to say something like 

> If omitted or empty, no attempt to change the CPU affinity of the process after moving it to container's cgroup is made, and its final affinity is determined by the Linux kernel.

I'd rather not go into more details here, because otherwise I'll be documenting the kernel.

kad: In my tests, it is not the case. (-p is "pause" where I'm attaching in another terminal process to cgroup with cpuset.cpus=0-10): 

```
r640-2:/home/akanevsk # ~akanevsk/getaffinity -g -s "0-5" -g -p -g
get_affinity: 0-87
set_affinity: affinity set to 0-5
get_affinity: 0-5
get_affinity: 0-10
r640-2:/home/akanevsk # ~akanevsk/getaffinity -g -s "50-60" -g -p -g
get_affinity: 0-87
set_affinity: affinity set to 50-60
get_affinity: 50-60
get_affinity: 0-10
```

kolyshkin: To anyone reading this thread: the post-cgroup-move affinity depends on the kernel version being used.

PR #1258, opened 2024-06-20
Add support for windows CPU affinity

kiashok:
This PR adds support for certain CRI fields in OCI runtime spec for supporting CPU affinity for windows containers.

CRI PR : https://github.com/kubernetes/kubernetes/pull/124285 
Github issue: https://github.com/kubernetes/kubernetes/issues/125262
KEP: https://github.com/kubernetes/kubernetes/issues/125262

### DIFF
--- a/config-windows.md
+++ b/config-windows.md
@@ -82,6 +82,14 @@ The following parameters can be specified (mutually exclusive):
 * **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor `count` in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinity`** *(array of objects, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
+
+  Each entry has the following structure:
+
+  Ref: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+
+  * **`mask`** *(uint64, REQUIRED)* - specifies the CPU mask relative to this CPU group.
+  * **`group`** *(uint32, REQUIRED)* - specifies the processor group this mask refers to, as returned by GetLogicalProcessorInformationEx.
 
 Ref: https://docs.microsoft.com/en-us/virtualization/api/hcs/schemareference#Container_Processor
 
--- a/specs-go/config.go
+++ b/specs-go/config.go
@@ -635,6 +635,17 @@ type WindowsCPUResources struct {
 	// cycles per 10,000 cycles. Set processor `maximum` to a percentage times
 	// 100.
 	Maximum *uint16 `json:"maximum,omitempty"`
+	// Set of CPUs to affinitize for this container.
+	Affinity []WindowsCPUGroupAffinity `json:"affinity,omitempty"`
+}
+
+// Similar to _GROUP_AFFINITY struct defined in
+// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+type WindowsCPUGroupAffinity struct {
+	// CPU mask relative to this CPU group.
+	Mask uint64 `json:"mask,omitempty"`
+	// Processor group the mask refers to, as returned by GetLogicalProcessorInformationEx.
+	Group uint32 `json:"group,omitempty"`
 }
 
 // WindowsStorageResources contains storage resource management settings.
--- a/schema/config-windows.json
+++ b/schema/config-windows.json
@@ -38,6 +38,17 @@
                             },
                             "maximum": {
                                 "$ref": "defs.json#/definitions/uint16"
+                            },
+                            "affinity": {
+                                "type": "object",
+                                "properties": {
+                                    "mask": {
+                                        "$ref": "defs.json#/definitions/uint64"
+                                    },
+                                    "group": {
+                                        "$ref": "defs.json#/definitions/uint32"
+                                    }
+                                }
                             }
                         }
                     },


kiashok:
cc @jsturtevant @kevpar @aravindhp @marosset

kiashok:
@AkihiroSuda @thaJeztah could you please take a look? :) thanks!

specs-go/config.go
AkihiroSuda, line 1: Markdown has to be updated too

specs-go/config.go
```diff
@@ -627,6 +627,19 @@ type WindowsCPUResources struct {
 	// cycles per 10,000 cycles. Set processor `maximum` to a percentage times
 	// 100.
 	Maximum *uint16 `json:"maximum,omitempty"`
+	// Set of CPUs to affinitize for this container.
+	AffinityCPUs []WindowsCPUGroupAffinity `json:"affinityCPUs,omitempty"`
+	// Specifies preferred set of numa node numbers to affinitize for this container.
+	AffinityPreferredNumaNodes []int64 `json:"affinityPreferredNumaNodes,omitempty"`
```
kevpar, line 633: What is actually done with this value?

specs-go/config.go
```diff
@@ -627,6 +627,19 @@ type WindowsCPUResources struct {
...
+}
+
+// Similar to _GROUP_AFFINITY struct defined in
+// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+type WindowsCPUGroupAffinity struct {
+	// CPU mask relative to this CPU group.
+	CPUMask uint64 `json:"cpuMask,omitempty"`
+	// CPU group that this CPU belongs to.
+	CPUGroup uint32 `json:"cpuGroup,omitempty"`
```
kevpar, line 642: I assume to be meaningful these must match exactly what the corresponding groups you can get info on from calling GetLogicalProcessorInformationEx? That would be good to explicitly state, both here and in the CRI PR.

specs-go/config.go
```diff
@@ -627,6 +627,19 @@ type WindowsCPUResources struct {
...
+	// Specifies preferred set of numa node numbers to affinitize for this container.
+	AffinityPreferredNumaNodes []int64 `json:"affinityPreferredNumaNodes,omitempty"`
+}
+
+// Similar to _GROUP_AFFINITY struct defined in
+// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+type WindowsCPUGroupAffinity struct {
+	// CPU mask relative to this CPU group.
+	CPUMask uint64 `json:"cpuMask,omitempty"`
+	// CPU group that this CPU belongs to.
```
kevpar, line 641: We aren't talking about a single CPU here, so this comment is confusing. I would say something like `Processor group the mask refers to, as returned by GetLogicalProcessorInformationEx`.

specs-go/config.go
utam0k: This PR would help you to update PR as an example.
https://github.com/opencontainers/runtime-spec/pull/1253

specs-go/config.go
jsturtevant: This would be used with https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-updateprocthreadattribute to set the Preffered numa node when using the memory manager.  

I have some of the details in https://github.com/kubernetes/enhancements/pull/4738/files#diff-69a2ffb8618949f1eddc42d0810c7afa17f3a82d8d8d7dbf07ba85495fd7137a

tianon:
Is there a compelling reason not to roughly match the naming convention adopted for Linux in #1253?

Also, please don't mark conversations with useful discussion/links as "resolved" especially if the PR itself didn't change as a result of the discussion but the discussion is still useful and potentially something someone else might reasonably ask about or want to discuss -- it makes them much more difficult to discover when reading the PR's comments.

specs-go/config.go
kevpar: I just commented here: https://github.com/kubernetes/enhancements/pull/4738/files#r1661653261

If we know preferred NUMA node via PROC_THREAD_ATTRIBUTE and affinity via job object are effectively the same thing (except one works at node granularity and the other at CPU granularity) do we need to offer both as options? Can't we just expose CPU affinity?

kiashok: @kevpar I don't see enough evidence to say PROC_THREAD_ATTRIBUTE  under covers is actually using cpu affinity mask in SetInformationJobObjectEx(). The KEP is going to be reworded a little. So I think it would be good to still have the numa node numbers passed down so hcsshim make the decision to set the masks appropriately. @jsturtevant does this sound ok?

kevpar: I guess I don't understand why we need both `AffinityCPUS` and `AffinityPreferredNumaNodes`. If these are mapping to the same setting under the hood, can we just expose the CPUs option, and let kubelet determine the node->cpu mapping? If these are mapping to different settings, what do they do differently such that we need both?

jsturtevant: This would be possible.  Due to the memory managers design today, to do this with no additional changes I believe it would require an additional query to remap the numa nodes that are calculated to the cpu's required.  

What do you see a major downside to including `AffinityPreferredNumaNodes`?

jsturtevant:
> Is there a compelling reason not to roughly match the naming convention adopted for Linux in #1253?

For cpu's, Linux is a simple list where as on windows it is a bitmap with a group number (https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity)  and the naming/structure was used to map to those Windows specific concepts

For memory, I think it was the same idea, making sure it was known this is "perferred" but not guarenteed do to the Windows API's.  Depending on outcome of https://github.com/opencontainers/runtime-spec/pull/1258#discussion_r1648430504 we might not have this field

specs-go/config.go
kevpar: I don't see a downside specifically to including `AffinityPreferredNumaNodes`, but I think if we are including it, we should understand what it really does and how it interacts with `AffinityCPUs`.

jsturtevant: I looked at it again, we could provide a way to map what the memory manager returns (which is a list of numanodes) to the CPU that numa node requires without any changes to memory manager. 

We can either make the system call or potentially even cache the CPU/numanode layout so we don't make the call for each container start (where numa nodes are required)

I don't really have a strong opinion so it comes down to where we want that conversion coming from. After some consideration, I am thinking that it may be more appropriate to do it in kubelet to make this interface simpler and easier to understand.

jsturtevant: We got the prototype working this week, After trying out a few difference scenarios, I am in favor of dropping this field.  I've updated the memory manager section in the KEP updates to identity how different scenarios would work: https://github.com/kubernetes/enhancements/pull/4738 PTAL

config-windows.md
```diff
@@ -82,6 +82,7 @@ The following parameters can be specified (mutually exclusive):
 * **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor `count` in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinityCPUs`** *(object, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
```
kevpar, line 85: I think we should describe the expected structure of `affinityCPUs` here, rather than relying on the go bindings to document it.

kiashok:
@AkihiroSuda @kevpar addressed review comments. Would you be able to take a look when you have some time please? Thanks!

config-windows.md
```diff
@@ -82,6 +82,12 @@ The following parameters can be specified (mutually exclusive):
 * **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor `count` in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinityCPUs`** *(array of objects, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
```
kolyshkin, line 85: nit: why not just `affinity` (since this is a part of cpu struct/group, and other (ajacent) fields do not have CPU in them)

kolyshkin: Answering to myself.

I see that earlier version had `affinityCPUs` and `AffinityPreferredNumaNodes`, but the latter was dropped.

Still, `affinity` would work, and, if we're to add NUMA affinity later, a name like `NUMAaffinity` (or `affinityNUMA`) would still work.

tianon: The sub fields could also be `mask` and `group` accordingly 👀

kolyshkin: Indeed. The less tautology the better.

config-windows.md
```diff
@@ -82,6 +82,12 @@ The following parameters can be specified (mutually exclusive):
...
+
+Each entry has the following structure:
+
+* **`CPUMask`** *(uint64, REQUIRED)* - specifies CPU mask relative to this CPU group.
+* **`CPUGroup`** *(uint32, REQUIRED)* - CPU group that this CPU belongs to.
```
kolyshkin, line 90: I'd place `group` first as the `mask` refers to the group (so it is slightly easier to read this way).

config-windows.md
kiashok: @tianon  @kolyshkin  keeping it similar to how we've defined this in k/k proto file. There was a similar comment in that PR : https://github.com/kubernetes/kubernetes/pull/124285#discussion_r1648429800 

group alone is a keyword in .proto so we couldn't just have mask or group. Thought uniformity would be good across the fields- prefix "cpu" in both fields or neither.

kiashok: similar rationale to why affinityCPUs was used : 
image


It is part of WindowsContainerResources struct which can mean cpus/memory or any other resource. So we named it affinityCPUs there. I understand that this struct in runtimespec is specifically for CPU alone. So if you would prefer me to drop "cpu" and just use "affinity" let me know, I can have that updated. Thanks.

config-windows.md
kevpar, line 90: Someone coming in and reading this spec would have no idea what "mask" and "group" mean here. Can we define specifically what we're talking about, or link to a doc which does?

Basically think of the audience as someone who is familiar with containers, but not familiar with Windows CPU concepts. How do they know what these mean, and how they map to OS concepts?

kiashok: Updated with link to msdocs that explains these concepts as it might be too much to add into this file.

kiashok:
@tianon @kolyshkin @kevpar @jsturtevant could you please take a look at this PR when you have some time please? Thanks!

jsturtevant:
/lgtm
for the functionality, I will defer to the reviewers here on the naming discussion

config-windows.md
tianon: I imagine there's got to be a way to convince Protobuf to use a keyword as an identifier, right?  Some way to escape it?

Regardless, _this_ spec is for a canonically-JSON-format, which doesn't have any constraints like that (or reserved keys at all), so I definitely still agree with @kolyshkin that we should ditch the duplication in the names and go for the simpler and more consistent all-lowercase versions (`affinity` or `affinities`, `group`, and `mask`).

specs-go/config.go
```diff
@@ -635,6 +635,17 @@ type WindowsCPUResources struct {
 	// cycles per 10,000 cycles. Set processor `maximum` to a percentage times
 	// 100.
 	Maximum *uint16 `json:"maximum,omitempty"`
+	// Set of CPUs to affinitize for this container.
+	AffinityCPUs []WindowsCPUGroupAffinity `json:"affinityCPUs,omitempty"`
```
rchincha, line 639: Why is the specific to "Windows"? CPU affinity is more general?

tianon: The specific structure/fields supported are Windows-specific -- see https://github.com/opencontainers/runtime-spec/blob/8f3fbc881602d85699e5c448634ec1288860d966/config.md#:~:text=to%207%20(lowest).-,execCPUAffinity,-(object%2C%20OPTIONAL)%20specifies (`execCPUAffinity`) for how this is specified on Linux (part of the `process` object).

config-windows.md
kiashok: @tianon  have addressed this. Please take a look and let me know if it looks good. Thanks.

config-windows.md
```diff
@@ -82,6 +82,8 @@ The following parameters can be specified (mutually exclusive):
 * **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor `count` in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinity`** *(object, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
```
kevpar, line 85: I think we need to define the fields of this object (`Mask` and `Group`), too.

specs-go/config.go
```diff
@@ -627,6 +627,19 @@ type WindowsCPUResources struct {
...
+	AffinityPreferredNumaNodes []uint32 `json:"affinityPreferredNumaNodes,omitempty"`
+}
+
+// Similar to _GROUP_AFFINITY struct defined in
+// https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+type WindowsCPUGroupAffinity struct {
+	// CPU mask relative to this CPU group.
+	Mask uint64 `json:"Mask,omitempty"`
+	// Processor group the mask refers to, as returned by GetLogicalProcessorInformationEx.
+	Group uint32 `json:"Group,omitempty"`
```
kevpar, line 642: The JSON names here should be camel case (`mask`, `group`).

config-windows.md
```diff
@@ -82,6 +82,8 @@ The following parameters can be specified (mutually exclusive):
...
+* **`affinityPreferredNumaNodes`** *(uint32, OPTIONAL)* - specifies referred set of numa node numbers to affinitize for this container.
```
kevpar, line 86: I think someone using this would naturally ask "how do I know which node is 0, 1, etc?". How would they determine this? Is there an API we can tell them to use that will match the indexing expected here?

config-windows.md
```diff
@@ -82,6 +82,14 @@ The following parameters can be specified (mutually exclusive):
...
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinity`** *(object, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
+
+Each entry has the following structure:
+
+Ref: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+
+* **`mask`** *(string, REQUIRED)* - specifies the CPU mask relative to this CPU group.
+* **`group`** *(string, REQUIRED)* - specifies the processor group this mask refers to, as returned by GetLogicalProcessorInformationEx.
```
kevpar, line 92: These lines need to be indented with two spaces, so they show up nested under `affinity`

specs-go/config.go
```diff
@@ -627,6 +627,19 @@ type WindowsCPUResources struct {
 	// cycles per 10,000 cycles. Set processor `maximum` to a percentage times
 	// 100.
 	Maximum *uint16 `json:"maximum,omitempty"`
+	// Set of CPUs to affinitize for this container.
+	Affinity []WindowsCPUGroupAffinity `json:"affinity,omitempty"`
+	// Specifies preferred set of numa node numbers to affinitize for this container.
+	AffinityPreferredNumaNodes []uint32 `json:"affinityPreferredNumaNodes,omitempty"`
```
kevpar, line 633: Is this supposed to be here still?

kiashok: Not sure why the PR is not updating :( was able to confirm from my local branch that these changes were infact removed. Checking what's happening

jsturtevant:
LGTM

@tianon @kevpar @kolyshkin could you give a final look?

config-windows.md
```diff
@@ -82,6 +82,14 @@ The following parameters can be specified (mutually exclusive):
...
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinity`** *(object, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
+
+  Each entry has the following structure:
+
+  Ref: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
+
+  * **`mask`** *(string, REQUIRED)* - specifies the CPU mask relative to this CPU group.
+  * **`group`** *(string, REQUIRED)* - specifies the processor group this mask refers to, as returned by GetLogicalProcessorInformationEx.
```
kevpar, line 92: Should these be listed as uint64/32 instead of string?

config-windows.md
```diff
@@ -82,6 +82,14 @@ The following parameters can be specified (mutually exclusive):
 * **`count`** *(uint64, OPTIONAL)* - specifies the number of CPUs available to the container. It represents the fraction of the configured processor `count` in a container in relation to the processors available in the host. The fraction ultimately determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles.
 * **`shares`** *(uint16, OPTIONAL)* - limits the share of processor time given to the container relative to other workloads on the processor. The processor `shares` (`weight` at the platform level) is a value between 0 and 10,000.
 * **`maximum`** *(uint16, OPTIONAL)* - determines the portion of processor cycles that the threads in a container can use during each scheduling interval, as the number of cycles per 10,000 cycles. Set processor `maximum` to a percentage times 100.
+* **`affinity`** *(object, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
```
tianon, line 85:
```suggestion
* **`affinity`** *(array of objects, OPTIONAL)* - specifies the set of CPU to affinitize for this container.
```

config-windows.md
```diff
@@ -82,6 +82,14 @@ The following parameters can be specified (mutually exclusive):
...
+
+  Each entry has the following structure:
+
+  Ref: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/miniport/ns-miniport-_group_affinity
```
tianon, line 89: I think this ref should probably be after the list (like the other one in this same section), but it is a little odd then to have the ref for count+shares+maximum follow immediately. :thinking:  Maybe it's fine (since the indentation should make it clear which is which)?  Maybe that's why you chose to put it here instead of matching the outer pattern?

tianon (commented):
A couple really minor nits, but I think this is probably fine.

Maybe we need to update the JSON schema also?

https://github.com/opencontainers/runtime-spec/blob/09fcb39bb7185b46dfb206bc8f3fea914c674779/schema/config-windows.json#L30-L43

config-windows.md
kiashok: yeah that's right - I thought keeping it closest to the declaration might help. The indentation is helping keep it readable.

kiashok:
> A couple really minor nits, but I think this is probably fine.
> 
> Maybe we need to update the JSON schema also?
> 
> https://github.com/opencontainers/runtime-spec/blob/09fcb39bb7185b46dfb206bc8f3fea914c674779/schema/config-windows.json#L30-L43

Addressed this!

kiashok:
@tianon would you be able to take a look at this PR when you have a chance please? Thanks!

jsturtevant:
LGTM

AkihiroSuda (approved):

Issue #1039, opened 2020-03-31
Follow-up to under-specification of createRuntime / createContainer hooks


h-vetinari:
Since #1008, `config.md` [states](https://github.com/opencontainers/runtime-spec/blob/master/config.md#createruntime-hooks) [my bold]:
> The definition of `createRuntime` hooks is currently **underspecified** and hooks authors, should only expect from the runtime that the mount namespace have been created and the mount operations performed. Other operations such as cgroups and SELinux/AppArmor labels might not have been performed by the runtime.

[as well as](https://github.com/opencontainers/runtime-spec/blob/master/config.md#createcontainer-hooks) [my bold]:
> The definition of `createContainer` hooks is currently **underspecified** and hooks authors, should only expect from the runtime that the mount namespace and different mounts will be setup. Other operations such as cgroups and SELinux/AppArmor labels might not have been performed by the runtime.

This was seen - AFAIU - as a necessary compromise to finally finish #1008 and unblock `runc-1.0.0`. Now that `runtime-spec-1.0.2` is out, we should at least track this state of under-specification somewhere though, hence this issue.

PR #1262, opened 2024-07-17
runtime: fail when a poststart hook fails

kolyshkin:
Poststart hooks exist in runc since 2015 [1], and since that time until today, if a hook returned an error, runc kills the container.

In 2020, commit c1662686cff159 (PR #1008) added the following text (which became part of runtime-spec release v1.0.2):

> 9. The `poststart` MUST be invoked by the runtime. If any
> `poststart` hook fails, the runtime MUST log a warning, but the
> remaining hooks and lifecycle continue as if the hook had succeeded.

Now, this text conflicted with the pre-existing runtime (runc) behavior, and it still conflicts with the current runc behavior.

At this point, we can either fix runtimes or the spec.

To my mind, fixing the spec is a better approach, because:
 - initial implementation predates the spec wording by a few years;
 - the wording in the spec was never implemented (in runc);
 - returning an error (and stopping the container) seems like a more versatile approach, since a hook can usually choose whether to return an error or not.

[1]: https://github.com/opencontainers/runc/pull/392

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -70,7 +70,7 @@ The lifecycle describes the timeline of events that happen from when a container
     If any `startContainer` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
 8. The runtime MUST run the user-specified program, as specified by [`process`](config.md#process).
 9. The [`poststart` hooks](config.md#poststart) MUST be invoked by the runtime.
-    If any `poststart` hook fails, the runtime MUST [log a warning](#warnings), but the remaining hooks and lifecycle continue as if the hook had succeeded.
+    If any `poststart` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
 10. The container process exits.
     This MAY happen due to erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
 11. Runtime's [`delete`](runtime.md#delete) command is invoked with the unique identifier of the container.


kolyshkin:
OTOH crun does not kill the container if a poststart hook has failed (and the code is from about 2017). Cc @giuseppe

lifubang (commented):
In `poststop` section, there is a text: `but the remaining hooks and lifecycle continue`.
Do you think we should change this description or let container runtimes to meet this rule?

giuseppe:
the spec behavior is validated by: https://github.com/opencontainers/runtime-tools/blob/master/validation/poststart_fail/poststart_fail.go and that is probably the reason I've implemented it this way for crun.

I've no preference on the two behaviors, with the spec behavior the user has more control to decide when to stop the container at the risk of just not seeing/ignoring the warning.  The runc behavior instead makes easier to detect when the poststart hook fails, but it terminates the container process once it is already started (is it always safe to do so?)

ningmingxiao:
shall we refer k8s poststart https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/

rata:
@ningmingxiao I think k8s poststart is unrelated. I think the post-start is just the kubelet doing an exec in the container.

kolyshkin:
> we call the hooks before exec'ing the container init process

This is another thing to fix in spec. It currently says

> The `poststart` hooks MUST be called [after the user-specified process is executed](runtime.md#lifecycle) but before the [`start`](runtime.md#start) operation returns.

but practically runc call hooks before exec'ing container init. If crun is doing the same, maybe we should revise the spec.

ningmingxiao:
user-specified process is "sleep 6"
post start is "sleep 10"
```
TIME(s) PCOMM            PID    PPID   RET ARGS
3.459   crun             2626   2534     0 /usr/bin/crun --debug run test119
3.528   sleep            2629   2626     0 /usr/bin/sleep 6
3.531   sleep            2628   2626     0 /bin/sleep 10
```
crun run poststart after user-specified process.

h-vetinari:
While you're taking a look at the whole hook situation, you might also want to take a glimpse at #1039.

abel-von:
I think the same situation also happened in `poststop` hook #1265

lifubang:
Can we come to a conclusion? Thanks.

lifubang:
> If crun is doing the same, maybe we should revise the spec.

Quite agree, the other reason is that maybe we can’t know the user-specified process is really executed successfully because of ‘execve’(of cause we can know it fails).

utam0k:
I don't have a strong opinion on this change, but failing is better because it makes it easier for users to detect errors.

utam0k:
📝 We need to update runtime-tools if we merge this PR.

utam0k (approved):
Sounds good from my perspective.

marquiz (approved):
lgtm

PR #1268, opened 2024-10-29
doc: fix expired link for "Naming a Volume" in config.md

taro0513:
This PR addresses an expired link for "Naming a Volume" in the config.md file. The original link was no longer accessible, and based on the context, I have identified a new link that likely aligns with the documentation's intended reference.


AkihiroSuda (approved):

thaJeztah (approved):
LGTM

Surprised that they let those short URLs expire.

PR #1269, opened 2024-11-04
update http links to https

thaJeztah:
noticed these while reviewing https://github.com/opencontainers/runtime-spec/pull/1268

Most of these either redirect (so changing saves an extra redirect), or have a TLS version available.


thaJeztah:
FWIW this one is broken, so we need to either use a wayback-machine URL or find a replacement; https://github.com/opencontainers/runtime-spec/blob/a55ebd9527ed472e510c1393e5f86689118e84a1/config-vm.md?plain=1#L67

AkihiroSuda (approved):

tianon (approved):

tianon:
> FWIW this one is broken, so we need to either use a wayback-machine URL or find a replacement;
> 
> https://github.com/opencontainers/runtime-spec/blob/a55ebd9527ed472e510c1393e5f86689118e84a1/config-vm.md?plain=1#L67

Maybe we just link to https://en.wikipedia.org/wiki/VMDK and call it good?

thaJeztah:
> Maybe we just link to https://en.wikipedia.org/wiki/VMDK and call it good?


Yeah, perhaps; I also found (through WikiPedia) this repository, which could also be an alternative; https://github.com/libyal/libvmdk/blob/20240510/documentation/VMWare%20Virtual%20Disk%20Format%20(VMDK).asciidoc

PR #1273, opened 2024-12-30
zos updates

kershawmehta:
This PR proposes updates to the OCI runtime spec with z/OS platform-specific details, including adding **namespaces**, adding **noNewPrivileges** flag, and removing **devices**.  These changes are currently in use by the IBM z/OS Container Platform (zOSCP) product - details can be found here: [https://www.ibm.com/products/zos-container-platform](https://www.ibm.com/products/zos-container-platform).

**Note:** The "removing of devices" is not considered to be an incompatible change since no user would have been using the z/OS-specific portion of the runtime spec, prior to the initial zOSCP release.

### DIFF
--- a/config-zos.md
+++ b/config-zos.md
@@ -1,20 +1,56 @@
-_This document is a work in progress._
-
 # <a name="ZOSContainerConfiguration" />z/OS Container Configuration
 
 This document describes the schema for the [z/OS-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+The z/OS container specification uses z/OS UNIX kernel features like namespaces and filesystem jails to fulfill the spec.
 
-## <a name="configZOSDevices" />Devices
+Applications expecting a z/OS environment will very likely expect these file paths to be set up correctly.
 
-**`devices`** (array of objects, OPTIONAL) lists devices that MUST be available in the container.
-The runtime MAY supply them however it likes.
+The following filesystems SHOULD be made available in each container's filesystem:
 
-Each entry has the following structure:
+| Path     | Type   |
+| -------- | ------ |
+| /proc    | [proc][] |
 
-* **`type`** *(string, REQUIRED)* - type of device: `c`, `b`, `u` or `p`.
-* **`path`** *(string, REQUIRED)* - full path to device inside container.
-    If a file already exists at `path` that does not match the requested device, the runtime MUST generate an error.
-* **`major, minor`** *(int64, REQUIRED unless `type` is `p`)* - major, minor numbers for the device.
-* **`fileMode`** *(uint32, OPTIONAL)* - file mode for the device.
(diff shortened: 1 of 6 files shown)


thaJeztah:
Only glancing over; would it be clearer to move the removal of devices to a separate commit? Currently it made me almost consider that "namespaces" was a _replacement_ for devices which was a bit confusing.

config-zos.md
```diff
@@ -1,20 +1,55 @@
...
+A namespace wraps a global system resource in an abstraction that makes it appear to the processes within the namespace that they have their own isolated instance of the global resource.
+Changes to the global resource are visible to other processes that are members of the namespace, but are invisible to other processes.
+
+Namespaces are specified as an array of entries inside the `namespaces` root field.
+The following parameters can be specified to set up namespaces:
 
-## <a name="configZOSDevices" />Devices
+* **`type`** *(string, REQUIRED)* - namespace type. The following namespace types SHOULD be supported:
+    * **`pid`** processes inside the container will only be able to see other processes inside the same container or inside the same pid namespace.
+    * **`mount`** the container will have an isolated mount table.
```
AkihiroSuda, line 24: Is there a public man page like https://man7.org/linux/man-pages/man7/mount_namespaces.7.html ?

schema/test/config/good/zos-example.json
```diff
@@ -0,0 +1,138 @@
+{
+    "ociVersion": "0.5.0-dev",
```
AkihiroSuda, line 2: nit: invalid version

AkihiroSuda (approved):

config-zos.md
kershawmehta: On the z/OS platform, we do not have published static man pages. Rather our man pages are created dynamically (JIT) when the man is issued for a specific command, using a specially formatted version of the documentation. However, we have publicly available documentation and I will push a commit that adds a URL which points to this documentation.

schema/test/config/good/zos-example.json
kershawmehta: The ociVersion is the same as used in `spec-example.json`.

kershawmehta:
> Or is it a replacement? 🙈

The deletion of devices and the addition of namespaces are unrelated. However, I thought that having these in a single PR would be more efficient, given this is the resultant of what is supported by z/OS Container Platform. 
If this really needs to be split into separate PRs, we can do that also.

kolyshkin (changes requested):
Please amend the first commit to provide a better commit message than "zos updates". You can start by taking the description of this PR and copy-paste it.

kershawmehta:
> Please amend the first commit to provide a better commit message than "zos updates". You can start by taking the description of this PR and copy-paste it.

Done.

kolyshkin (approved):
lgtm

PR #1275, opened 2025-01-13
Correct `prestart` hook description in summary

ssst0n3:
Correct `prestart` hook description in summary

It looks like the previous description was copied from the wrong line.   Update the `prestart (Deprecated)` hook's timing from being called "after the start operation is invoked but before the user-specified command executes"  to "during the create operation, after the runtime environment is created and before pivot root or any equivalent operation."


utam0k (approved):
I've confirmed the behavior of `prestart` and this change looks fine to me.
https://github.com/opencontainers/runc/blob/610aa88ab201f289c05c2e262912d0630f46eb35/libcontainer/process_linux.go#L754-L759

kolyshkin (approved):
LGTM

PR #1277, opened 2025-02-03
Fix description of errnoRet in Seccomp

z63d:
There is a subtle difference in syntax between Seccomp's defaultErrnoRet and errnoRet, so i will unify the syntax.

https://github.com/opencontainers/runtime-spec/blob/a5b01166ad3fdf111b1e427abe35ce2538f85f50/config-linux.md?plain=1#L722

https://github.com/opencontainers/runtime-spec/blob/a5b01166ad3fdf111b1e427abe35ce2538f85f50/config-linux.md?plain=1#L797

Or you could just reference one of them like this...

https://github.com/opencontainers/runtime-spec/blob/a5b01166ad3fdf111b1e427abe35ce2538f85f50/config-linux.md?plain=1#L718

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -719,7 +719,7 @@ The following parameters can be specified to set up seccomp:
 * **`defaultErrnoRet`** *(uint, OPTIONAL)* - the errno return code to use.
     Some actions like `SCMP_ACT_ERRNO` and `SCMP_ACT_TRACE` allow to specify the errno code to return.
     When the action doesn't support an errno, the runtime MUST print and error and fail.
-    If not specified then its default value is `EPERM`.
+    The default is `EPERM`.
 * **`architectures`** *(array of strings, OPTIONAL)* - the architecture used for system calls.
     A valid list of constants as of libseccomp v2.6.0 is shown below.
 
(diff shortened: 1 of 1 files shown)


kolyshkin (commented):
I would just say "The default is `EPERM`" in both cases.

Cc @cyphar

giuseppe:
agreed.  "The default is `EPERM`" should be enough

kolyshkin:
@z63d PTAL

z63d:
I just felt strange that there was only one difference in such a sentence.

```
the errno return code to use.
Some actions like SCMP_ACT_ERRNO and SCMP_ACT_TRACE allow to specify the errno code to return.
When the action doesn't support an errno, the runtime MUST print and error and fail.
If not specified then its default value is EPERM.
```

If there are no problems, close the PR.

kolyshkin:
@z63d what I (and @giuseppe) meant is, can you please change the commit do just say "The default is `EPERM`" in both cases?

z63d:
Ah, my understanding was lacking. Sorry.

kolyshkin (approved):
lgtm, thanks!

tianon (approved):

AkihiroSuda (approved):

PR #1282, opened 2025-03-27
Add support for Linux memory policy

askervin:
Enable setting a NUMA memory policy for the container. New linux.mempolicy object contains inputs to the set_mempolicy(2) syscall.

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -791,6 +791,45 @@ and may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.
 }
 ```
 
+## <a name="configLinuxMemoryPolicy" />Memory policy
+
+**`memoryPolicy`** (object, OPTIONAL) sets the NUMA memory policy for the container.
+For more information see the [set_mempolicy(2)][set_mempolicy.2] man page.
+
+* **`mode`** *(string, REQUIRED)* -
+
+    A valid list of constants is shown below.
+
+    * `MPOL_DEFAULT`
+    * `MPOL_BIND`
+    * `MPOL_INTERLEAVE`
+    * `MPOL_WEIGHTED_INTERLEAVE`
+    * `MPOL_PREFERRED`
+    * `MPOL_PREFERRED_MANY`
+    * `MPOL_LOCAL`
+
+* **`nodes`** *(string, REQUIRED)* - list of memory nodes from which nodemask is constructed to set_mempolicy(2). This is a comma-separated list, with dashes to represent ranges. For example, `0-3,7` represents memory nodes 0,1,2,3, and 7.
+
+* **`flags`** *(array of strings, OPTIONAL)* - list of flags to use with set_mempolicy(2).
+
+    A valid list of constants is shown below.
+
+    * `MPOL_F_NUMA_BALANCING`
+    * `MPOL_F_RELATIVE_NODES`
+    * `MPOL_F_STATIC_NODES`
+
+### Example
+
+```json
+"linux": {
+    "memoryPolicy": {
+        "mode": "MPOL_INTERLEAVE",
+        "nodes": "2-3"
+        "flags": ["MPOL_F_STATIC_NODES"],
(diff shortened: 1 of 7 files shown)


config-linux.md
AkihiroSuda, line 1: https://github.com/opencontainers/runtime-spec/blob/main/features-linux.md needs to be updated too

kad:
Small nit, I'd suggest to use `MemoryPolicy` and `memoryPolicy` instead of `mempolicy`. It would be more readably IMHO.

giuseppe:
LGTM after the changes suggested above

config-linux.md
askervin: Thanks @AkihiroSuda, added supported memoryPolicy modes and flags to be listed in features.

askervin:
> Small nit, I'd suggest to use `MemoryPolicy` and `memoryPolicy` instead of `mempolicy`. It would be more readable IMHO.

Thanks @kad, fixed. Definitely better.

utam0k (commented):
It seems set_mempolicy(2) is only effective agains for called threads. I'm not sure how to do it for processes created with exec. Any ideas? 

> set_mempolicy() sets the NUMA memory policy of the calling thread,
       which consists of a policy mode and zero or more nodes, to the
       values specified by the mode, nodemask, and maxnode arguments.

The behavior of several other system calls is the same, so it might be a good idea to define their behavior as well.

utam0k:
> It seems set_mempolicy(2) is only effective agains for called threads. 

I'm not sure but we may need to implement it in nsexec.c of runc because of the thread limitation. I recommend to implement PoC in runc.

giuseppe:
> It seems set_mempolicy(2) is only effective agains for called threads. I'm not sure how to do it for processes created with exec. Any ideas?

the man page says:

```
The process memory policy is preserved across an [execve](https://linux.die.net/man/2/execve)(2), and is inherited by child processes created using [fork](https://linux.die.net/man/2/fork)(2) or [clone](https://linux.die.net/man/2/clone)(2).
```

so I don't think it is a problem for the spec

utam0k:
@giuseppe Oh, I missed it. Looks good.

AkihiroSuda (approved):

utam0k (approved):

giuseppe (approved):
LGTM

config-linux.md
```diff
@@ -791,6 +791,45 @@ and may use a maximum memory bandwidth of 20% on socket 0 and 70% on socket 1.
...
+
+    * `MPOL_DEFAULT`
+    * `MPOL_BIND`
+    * `MPOL_INTERLEAVE`
+    * `MPOL_WEIGHTED_INTERLEAVE`
+    * `MPOL_PREFERRED`
+    * `MPOL_PREFERRED_MANY`
+    * `MPOL_LOCAL`
+
+* **`nodes`** *(string, REQUIRED)* - list of memory nodes from which nodemask is constructed to set_mempolicy(2). This is a comma-separated list, with dashes to represent ranges. For example, `0-3,7` represents memory nodes 0,1,2,3, and 7.
```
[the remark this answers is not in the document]
fabbione: no problem at all, thanks for being so fast in fixing the docs etc.

specs-go/config.go
```diff
@@ -855,6 +857,19 @@ type LinuxIntelRdt struct {
...
 }
 
+// LinuxMemoryPolicy represents input for the set_mempolicy syscall.
+type LinuxMemoryPolicy struct {
+	// Mode for the set_mempolicy syscall.
+	Mode MemoryPolicyModeType `json:"mode"`
+
+	// Nodes representing the nodemask for the set_mempolicy syscall in comma separated ranges format.
+	// Format: "<node0>-<node1>,<node2>,<node3>-<node4>,..."
+	Nodes string `json:"nodes"`
```
cyphar, line 867: Should this have been `omitempty` too?

PR #1271, opened 2024-11-07
Define Linux Network Devices

aojea:
The proposed "netdevices" field provides a declarative way to specify which host network devices should be moved into a container's network namespace.

This approach is similar than the existing "devices" field used for block devices but uses a dictionary keyed by the interface name instead.

The proposed scheme is based on the existing representation of network device by the `struct net_device`
https://docs.kernel.org/networking/netdevices.html.

This proposal focuses solely on moving existing network devices into the container namespace. It does not cover the complexities of network configuration or network interface creation, emphasizing the separation of device management and network configuration.

A list of real use cases that justify this proposal is:

1.  **Pre-Configuring Physical Devices:**

    * Scenario: A container requires a specific physical network interface with a complex IP configuration, RDMA or SR-IOV
    * Implementation:
        * Configure the physical interface on the host with the desired IP addresses, routing, and other settings. In kubernetes this can be done with DRA or Device Plugins.
        * Use `netDevices` to move the pre-configured interface into the container.

2.  **Creating and Moving Virtual Interfaces:**

    * Scenario: A container needs to have its own unique MAC address on an existing physical network, without bridging.
    * Implementation:
        * Create the `macvlan` interface on the host, based on an existing physical interface.
        * Use `netDevices` to move the MACVLAN interface into the container.

3.  **Network Function Containers:**

    * Scenario: A container acts as a network router or firewall.
    * Implementation:
        * Use `netDevices` to move multiple physical or virtual interfaces into the container.
        * The container's processes manage the network configuration, routing, and firewall rules.

### References

- [Linux network namespaces and interfaces](https://gist.github.com/aojea/a5371456177ae85765714fd52db55fdf)
- [runc implementation](https://github.com/opencontainers/runc/pull/4538)

Fixes: https://github.com/opencontainers/runtime-spec/issues/1239

PR #1286, opened 2025-05-05
Add FreeBSD as a platform

dfr:
This uses FreeBSD jails to implement container isolation.

### DIFF
--- a/config-freebsd.md
+++ b/config-freebsd.md
@@ -1,0 +1,137 @@
+# <a name="FreeBSDContainerConfiguration" />FreeBSD Container Configuration
+
+This document describes the schema for the [FreeBSD-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+
+## <a name="configFreeBSDDevices" />Devices
+
+Devices in FreeBSD are accessed via the `devfs` filesystem. Each container SHOULD have a `devfs` filesystem mounted into its `/dev` directory. Often, a minimal set of devices is exposed to the container using ruleset 4 from `/etc/defaults/devfs.rules` - the ruleset is specified as a mount option.
+
+Optionally, additional devices can be exposed to the container using an array of entries inside the `devices` root field:
+
+* **`path`** _(string, REQUIRED)_ - the device path relative to `/dev`
+* **`mode`** _(uint32, OPTIONAL)_ - file mode for the device.
+
+Note that JSON numbers must be represented in decimal. The value `448` below is the decimal representation of octal `0700` and this is used to request file mode `rwx------` for the device.
+
+### Example
+```json
+"devices": [
+	{
+        "path": "pf",
+        "mode": 448
+    }
+]
+```
+
+## <a name="configFreeBSDJail" />Jail
+
+On FreeBSD, containers are implemented using the platform's jail subsystem.
+Each jail is configured using a set of name/value pairs passed to the kernel using the `jail(2)` system calls.
+The `jail` root field contains values which are passed to the kernel when the container is created.
+
+* **`parent`** _(string, OPTIONAL)_ - parent jail.
+    If set, the value is the name of a jail which should be this container's parent, otherwise the container's parent is the host. This can be used to share namespaces such as `vnet` with another container.
+* **`host`** _(string, OPTIONAL)_ - allow overriding hostname, domainname, hostuuid and hostid.
+    The value can be "new" which allows these values to be overridden in the container or "inherit" to use the host values (or parent container values). If set to "new", the values for hostname and domainname are taken from the base config, if present.
+* **`ip4`** _(string, OPTIONAL)_ - control the availability of IPv4 addresses.
+    Set to "inherit" to allow access to host (or parent container) addresses or set to "disable" to stop use of IPv4 entirely. This is typically left unset when **`vnet`** is used (see below).
+* **`ip4Addr`** _(array of strings, OPTIONAL)_ - restrict the set of IPv4 addresses which the container can use. These addresses should be in numeric form (e.g. `"10.11.12.13"`). This can be used to allow restricted use of the host network. A common pattern with FreeBSD jails is to add alias addresses to a loopback interface and restrict each jail to a subset of addresses.
+* **`ip6`** _(string, OPTIONAL)_ - control the availability of IPv6 addresses.
+    Set to "inherit" to allow access to host (or parent container) addresses or set to "disable" to stop use of IPv6 entirely. This is typically left unset when **`vnet`** is used (see below).
+* **`ip6Addr`** _(array of strings, OPTIONAL)_ - restrict the set of IPv6 addresses which the container can use. These addresses should be in numeric form (e.g. `"fd10::11:12:13"`). This can be used to allow restricted use of the host network. A common pattern with FreeBSD jails is to add alias addresses to a loopback interface and restrict each jail to a subset of addresses.
+* **`vnet`** _(string, OPTIONAL)_ - control the vnet used for this container.
+    The value can be "new" which causes a new vnet to be created for the container or "inherit" which shares the vnet for the parent container (or host if there is no parent).
+* **`interface`** _(string, OPTIONAL)_ A network interface to add the container's IP addresses (**`ip4Addr`** and **`ip6Addr`**) to.  An alias for each address will be added to the interface when the container is created, and will be removed from the interface after the container is stopped. This is typically used when **`vnet`** is not set.
+* **`vnetInterfaces`** _(array of strings, OPTIONAL)_ - a set of network interfaces which are added to the container's vnet during its lifetime.
+* **`sysvmsg`** _(string, OPTIONAL)_ - allow access to SYSV IPC message primitives.
+    If set to "inherit", all IPC objects in the host (or parent container) are visible to this container, whether they were created by the container itself, the base system, or other containers.  If set to "new", the container will have its own key namespace, and can only see the objects that it has created; the system (or parent container) has access to the container's objects, but not to its keys.  If set to "disable", the container cannot perform any sysvmsg-related system calls. Defaults to "new".
+* **`sysvsem`** _(string, OPTIONAL)_ - allow access to SYSV IPC semaphore primitives, in the same manner as sysvmsg. Defaults to "new".
+* **`sysvshm`** _(string, OPTIONAL)_ - allow access to SYSV IPC shared memory primitives, in the same manner as sysvmsg. Defaults to "new".
+* **`enforceStatfs`** _(integer, OPTIONAL)_ - control visibility of mounts in the container.
+    A value of 0 allows visibility of all host mounts, 1 allows visibility of mounts nested under the container's root and 2 only allows the container root to be visible. If unset, the default value is 2.
+* **`allow`** _(object, OPTIONAL)_ - Some restrictions of the container environment may be set on a per-container basis.  With the exception of **`setHostname`** and **`reservedPorts`**, these boolean parameters are off by default.
+  - **`setHostname`** _(bool, OPTIONAL)_ - Allow the container's hostname to be changed. Defaults to `false`.
+  - **`rawSockets`** _(bool, OPTIONAL)_ - Allow the container to use raw sockets to support network utilities such as ping and traceroute. Defaults to `false`.
+  - **`chflags`** _(bool, OPTIONAL)_ - Allow the system file flags to be changed. Defaults to `false`.
+  - **`mount`** _(array of strings, OPTIONAL)_ - Allow the listed filesystem types to be mounted and unmounted in the container.
+  - **`quotas`** _(bool, OPTIONAL)_ - Allow the filesystem quotas to be changed in the container. Defaults to `false`.
+  - **`socketAf`** _(bool, OPTIONAL)_ - Allow socket types other than IPv4, IPv6 and unix. Defaults to `false`.
+  - **`mlock`** _(bool, OPTIONAL)_ - Allow the container to use `mlock(2)` and `munlock(2)` system calls. Defaults to `false`.
+  - **`reservedPorts`** _(bool, OPTIONAL)_ - Allow the jail to bind to ports lower than 1024. Defaults to `false`.
+  - **`suser`** _(bool, OPTIONAL)_ - The value of the jail's security.bsd.suser_enabled sysctl. The super-user will be disabled automatically if its parent system has it disabled.  The super-user is enabled by default.
+
+These fields SHOULD be mapped to a corresponding set of `jail(8)` parameters which can be used to create the container jail.
+A typical jail-based OCI implementation on FreeBSD MAY use the following mapping:
+
+| Jail parameter   | JSON equivalent      |
+| --------------   | -------------------- |
+| `jid`            | -                    |
+| `name`           | see below            |
+| `path`           | `root.path`          |
+| `ip4.addr`       | `freebsd.jail.ip4Addr` |
+| `ip4.saddrsel`   | -                    |
+| `ip4`            | `freebsd.jail.ip4`   |
+| `ip6.addr`       | `freebsd.jail.ip6Addr` |
+| `ip6.saddrsel`   | -                    |
+| `ip6`            | `freebsd.jail.ip6`   |
+| `vnet`           | `freebsd.jail.vnet`  |
+| `interface`      | `freebsd.jail.interface` |
+| `vnet.interface` | see below            |
+| `host.hostname`  | `hostname`           |
+| `host`           | `freebsd.jail.host`  |
+| `sysvmsg`        | `freebsd.jail.sysvmsg` |
+| `sysvsem`        | `freebsd.jail.sysvsem` |
+| `sysvshm`        | `freebsd.jail.sysvshm` |
+| `securelevel`    | -                    |
+| `devfs_ruleset`  | see below            |
+| `children.max`   | see below            |
+| `enforce_statfs` | `freebsd.jail.enforceStatfs` |
+| `persist`        | -                    |
+| `parent`         | `freebsd.jail.parent`  |
+| `osrelease`      | -                    |
+| `osreldate`      | -                    |
+| `allow.set_hostname` | `freebsd.jail.allow.setHostname` |
+| `allow.sysvipc`  | `freebsd.jail.allow.sysvipc` |
+| `allow.raw_sockets`  | `freebsd.jail.allow.rawSockets` |
+| `allow.chflags`  | `freebsd.jail.allow.chflags` |
+| `allow.mount`    | `freebsd.jail.allow.mount` |
+| `allow.quotas`    | `freebsd.jail.allow.quotas` |
+| `allow.read_msgbuf` | -                       |
+| `allow.socket_af` | `freebsd.jail.allow.socketAf` |
+| `allow.mlock`    | `freebsd.jail.allow.mlock` |
+| `allow.nfsd`     | - |
+| `allow.reserved_ports` | `freebsd.jail.allow.reservedPorts` |
+| `allow.unprivileged_proc_debug` | - |
+| `allow.suser`    | `freebsd.jail.allow.suser` |
+| `allow.mount.*`  | see below            |
+
+The jail name SHOULD be set to the create command's `container-id` argument.
+
+The `vnet.interface` jail pseudo parameter is not handled in the kernel but rather is implemented in user space (e.g. in `jail(8)`). In traditional jail configs, this parameter can be repeated several times and each instance specifies a network interface which is moved into the jail's vnet during the lifetime of the jail using the `ifconfig(8)` utility on the host. For OCI containers, this is managed using the `freebsd.jail.vnetInterfaces` field which is an array of interface names.
+
+A container which needs its own network namespace SHOULD set `"vnet"` to `"new"` and leave `"ip4"` and `"ip6"` unchanged.
+A container which shares the parent/host vnet SHOULD leave `"vnet"` unchanged and set `"ip4"` and `"ip6"` to `"inherit"`.
+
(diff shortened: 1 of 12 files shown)


AkihiroSuda:
Cc @samuelkarp

config.md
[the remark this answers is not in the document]
dfr: When I originally looked at features.md, it seemed that the only platform-specific features are for Linux. So far, I don't think we have a need for platform features on FreeBSD.

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+
+* **`path`** _(string, REQUIRED)_ - the device path relative to `/dev`
+* **`mode`** _(integer, OPTIONAL)_ - device permissions as an integer which is interpreted as in `chmod(1)`.
+
+### Example
+```json
+"devices": [
+	{
+        "path": "pf",
+        "mode": "0700"
```
tianon, line 19: Is it an integer or a string? 😅

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+    }
+]
+```
+
+## <a name="configFreeBSDJail" />Jail
```
tianon, line 24: This section feels to me more like documentation for an existing implementation than specification from which a new compatible implementation could be implemented. 🤔

Are there specific features of FreeBSD that could be explicitly named here to help connect these things more strongly?  Are there any places here where SHOULD or MUST language would be appropriate?

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+  - **`chflags`** _(bool, OPTIONAL)_ - Allow the system file flags to be changed.
+  - **`mount`** _(array of strings, OPTIONAL)_ - Allow the listed filesystem types to be mounted and unmounted in the container.
+  - **`quotas`** _(bool, OPTIONAL)_ - Allow the filesystem quotas to be changed in the container.
+  - **`socketAf`** _(bool, OPTIONAL)_ - Allow socket types other than IPv4, IPv6 and unix.
+  - **`reservedPorts`** _(bool, OPTIONAL)_ - Allow the jail to bind to ports lower than 1024.
+  - **`suser`** _(bool, OPTIONAL)_ - The value of the jail's security.bsd.suser_enabled sysctl. The super-user will be disabled automatically if its parent system has it disabled.  The super-user is enabled by default.
+
+### Mapping from jail(8) config file
+
+This table shows how to map settings from a typical `jail(8)` config file to the container configuration:
```
tianon, line 56: Is this advisory/suggestion, or prescriptive? (It currently reads as the former)

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+| `allow.read_msgbuf` | -                       |
+| `allow.socket_af` | `freebsd.jail.allow.socketAf` |
+| `allow.mlock`    | - |
+| `allow.nfsd`     | - |
+| `allow.reserved_ports` | `freebsd.jail.allow.reservedPorts` |
+| `allow.unprivileged_proc_debug` | - |
+| `allow.suser`    | `freebsd.jail.allow.suser` |
+| `allow.mount.*`  | see below            |
+
+The jail name is set to the create command's `container-id` argument.
```
tianon, line 98: "is", SHOULD, or MUST?  (Another place that feels like documentation, not spec)

tianon (commented):
A few initial (high level) thoughts 👍❤️

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
+# <a name="FreeBSDContainerConfiguration" />FreeBSD Container Configuration
+
+This document describes the schema for the [FreeBSD-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+
+## <a name="configFreeBSDDevices" />Devices
+
+Devices in FreeBSD are accessed via the `devfs` filesystem. Typically, each container will have a `devfs` filesystem mounted into it's `/dev` directory. Often, a minimal set of devices is exposed to the container using ruleset 4 from `/etc/defaults/devfs.rules` - the ruleset is specified as a mount option.
```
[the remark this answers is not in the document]
dch: This seems fine to me.

devfs includes `/dev/std{in,out,err}` `/dev/null` and more, optionally `/dev/pf` for firewall (FreeBSD has a container-enabled virtualised network & firewall stack), and `/dev/zfs` filesystem functionality is exposed this way, etc.

@dfr `s/it's/its/` though?

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+| `allow.mlock`    | - |
+| `allow.nfsd`     | - |
+| `allow.reserved_ports` | `freebsd.jail.allow.reservedPorts` |
+| `allow.unprivileged_proc_debug` | - |
+| `allow.suser`    | `freebsd.jail.allow.suser` |
+| `allow.mount.*`  | see below            |
+
+The jail name is set to the create command's `container-id` argument.
+
+Network addresses are typically managed by the host (e.g. using CNI or netavark) to we do not include a mapping for `ip4.addr` or `ip6.addr`.
```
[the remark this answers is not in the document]
dch: @dfr `s/to we do/so we do/`

config-freebsd.md
dch: Unless there is another mechanism to (externally from the runtime) link `jail name` with `container-id` this should be `MUST`.

config-freebsd.md
dch: Good point, I think we can say `This tables defines the mappings from a typical` here. Does that read better?

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+This document describes the schema for the [FreeBSD-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+
+## <a name="configFreeBSDDevices" />Devices
+
+Devices in FreeBSD are accessed via the `devfs` filesystem. Typically, each container will have a `devfs` filesystem mounted into it's `/dev` directory. Often, a minimal set of devices is exposed to the container using ruleset 4 from `/etc/defaults/devfs.rules` - the ruleset is specified as a mount option.
+
+Optionally, additional devices can be exposed to the container using an array of entries inside the `devices` root field:
+
+* **`path`** _(string, REQUIRED)_ - the device path relative to `/dev`
+* **`mode`** _(integer, OPTIONAL)_ - device permissions as an integer which is interpreted as in `chmod(1)`.
```
dch, line 12: Yes.. should `mode` be annotated as string here?

What about:

"device permissions as an exact string representation of permissions expressed in octal, with leading "0" as in `chmod(1)`, for example `"0700"`. 

Maybe more precise, but feels clumsy to me. Suggestions?

config-freebsd.md
dch: for the whole table, I'm not sure. If any of these parameters are used, the mappy is prescriptive. Is that `MUST` or `SHOULD`?

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+* **`vnet`** _(string, OPTIONAL)_ - control the vnet used for this container.
+    The value can be "new" which causes a new vnet to be created for the container or "inherit" which shares the vnet for the parent container (or host if there is no parent).
+* **`sysvmsg`** _(string, OPTIONAL)_ - allow access to SYSV IPC message primitives.
+    If set to "inherit", all IPC objects on the system are visible to this container, whether they were created by the container itself, the base system, or other containers.  If set to "new", the container will have its own key namespace, and can only see the objects that it has created; the system (or parent container) has access to the container's objects, but not to its keys.  If set to "disable", the container cannot perform any sysvmsg-related system calls.
+* **`sysvsem`** _(string, OPTIONAL)_ - allow access to SYSV IPC semaphore primitives, in the same manner as sysvmsg.
+* **`sysvshm`** _(string, OPTIONAL)_ - allow access to SYSV IPC shared memory primitives, in the same manner as sysvmsg.
```
davidchisnall, line 41: These options are listed as optional but do not define a default.  I would suggest that 'new' is a good balance between security and usability as a default.

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+* **`enforceStatfs`** _(integer, OPTIONAL)_ - control visibility of mounts in the container.
+    A value of 0 allows visibility of all host mounts, 1 allows visibility of mounts nested under the container's root and 2 only allows the container root to be visible. If unset, the default value is 2.
+* **`allow`** _(object, OPTIONAL)_ - Some restrictions of the container environment may be set on a per-container basis.  With the exception of **`setHostname`** and **`reservedPorts`**, these boolean parameters are off by default.
+  - **`setHostname`** _(bool, OPTIONAL)_ - Allow the container's hostname to be changed.
+  - **`rawSockets`** _(bool, OPTIONAL)_ - Allow the container to use raw sockets to support network utilities such as ping and traceroute.
+  - **`chflags`** _(bool, OPTIONAL)_ - Allow the system file flags to be changed.
```
davidchisnall, line 47: Defaults to false?

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+  - **`mount`** _(array of strings, OPTIONAL)_ - Allow the listed filesystem types to be mounted and unmounted in the container.
+  - **`quotas`** _(bool, OPTIONAL)_ - Allow the filesystem quotas to be changed in the container.
+  - **`socketAf`** _(bool, OPTIONAL)_ - Allow socket types other than IPv4, IPv6 and unix.
+  - **`reservedPorts`** _(bool, OPTIONAL)_ - Allow the jail to bind to ports lower than 1024.
```
davidchisnall, line 51: Defaults to false?

config-freebsd.md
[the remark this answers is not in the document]
dfr: > > Typically, each container will have a `devfs` filesystem mounted into it's `/dev` directory.
> 
> What are the cases where they wouldn't? Does any of what's specified here no longer work in those cases? (Do we need a stronger requirement here, or is the status quo fine?)

In theory, you could do without `/dev` for very simple workloads which don't need /dev/null, /dev/zero, good random numbers etc. but in practice, the suggested devfs ruleset for jails (look for ruleset 4 in /etc/defaults/devfs.conf on a FreeBSD host) is sufficient for most workloads. Container engines such as Podman, containerd etc. all add a `/dev` mount point using ruleset 4.

config-freebsd.md
dfr: This is a string which is interpreted as in chmod(1) which supports octal numeric modes as well as symbolic. I will fix the text to make this a little clearer.

config-freebsd.md
dfr: This section is informational and is intended as a guide for readers familiar with FreeBSD jails to help them understand the mapping from the FreeBSD OCI config settings to the corresponding jail.

Since this proposal only covers the use of FreeBSD jails to implement OCI containers, there is very little freedom for implementors. The text following the table would benefit from using SHOULD/MUST language.

dfr:
Fixed the type of FreeBSDDevice.Mode and fixed a typo in the json mapping for FreeBSDJail.SysVShm.

giuseppe (commented):
I've no specific comments on the FreeBSD part, but I can help to move this forward.

I've left an inline comment

dfr:
> I've no specific comments on the FreeBSD part, but I can help to move this forward.
> 
> I've left an inline comment

Thanks! I will check out your comment on the PR but my internet connection is flaky at the moment which is slowing me down a bit (hopefully that will be fixed over the weekend).

config-freebsd.md
dfr: This is a very good point. In this case, the parsing is handled by the devfs(8) system utility but that doesn't mean that this specification needs to allow anything more than a simple three digit octal number. I will rearrange things to share the definition of fileMode with config-linux.json and change this field to use that.

tianon (dismissed):
This looks pretty decent to me, although admittedly I'm not super familiar with FreeBSD / jails internals.

If we're not in a hurry to merge, I'd love to wait until @samuelkarp has a chance to review too, but if we've got something urgent that needs this, I think it's probably in a reasonable state (and hopefully he doesn't disagree).

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
...
+## <a name="configFreeBSDJail" />Jail
+
+On FreeBSD, containers are implemented using the platform's jail subsystem. Configuration for the container jail is via fields inside the `jail` root field:
+
+* **`parent`** _(string, OPTIONAL)_ - parent jail.
+    The value is the name of a jail which should be this container's parent (defaults to none). This can be used to share namespaces such as `vnet` with another container.
+* **`host`** _(string, OPTIONAL)_ - allow overriding hostname, domainname, hostuuid and hostid.
+    The value can be "new" which allows these values to be overridden in the container or "inherit" to use the host values (or parent container values). If set to "new", the values for hostname and domainname are taken from the base config, if present.
+* **`ip4`** _(string, OPTIONAL)_ - control the availability of IPv4 addresses.
+    This is typically left unset if the container has a vnet, set to "inherit" to allow access to host (or parent container) addresses or set to "disable" to stop use of IPv4 entirely.
```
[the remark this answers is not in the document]
dfr: Allowed values for "ip4" are "inherit", "new" or "disable". The interactions between vnet and ip4 are not clearly documented but by experimentation, I discovered that if "vnet" is set to "new", then "ip4" must be left unset, otherwise we get an EINVAL error.

config-freebsd.md
```diff
@@ -0,0 +1,124 @@
+# <a name="FreeBSDContainerConfiguration" />FreeBSD Container Configuration
+
+This document describes the schema for the [FreeBSD-specific section](config.md#platform-specific-configuration) of the [container configuration](config.md).
+
+## <a name="configFreeBSDDevices" />Devices
+
+Devices in FreeBSD are accessed via the `devfs` filesystem. Typically, each container will have a `devfs` filesystem mounted into its `/dev` directory. Often, a minimal set of devices is exposed to the container using ruleset 4 from `/etc/defaults/devfs.rules` - the ruleset is specified as a mount option.
```
samuelkarp, line 7: Do we want to have a SHOULD here ("containers SHOULD have a `devfs` filesystem mounted into its `/dev` directory")?

samuelkarp (approved):
~A couple minor comments, but generally this LGTM.  I may not be available to review again after my suggestions are addressed, so please consider this to be my approval either way.~

Edit: see further comments.

specs-go/config.go
```diff
@@ -923,3 +925,66 @@ const (
...
+	Mode *os.FileMode `json:"mode,omitempty"`
+}
+
+// FreeBSDJail describes how to configure the container's jail
+type FreeBSDJail struct {
+	// Parent jail name - this can be used to share a single vnet
+	// across several containers
+	Parent string `json:"parent,omitempty"`
+	// Whether to use parent UTS names or override in the container
+	Host FreeBSDSharing `json:"host,omitempty"`
```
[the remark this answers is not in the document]
dfr: These values were defined in the kernel in 2009 (https://github.com/freebsd/freebsd-src/commit/7cbf72137fadbb75662d02cabe4c7a0b1d51ab6d) and have remained unchanged since then. These are part of the kernel ABI so values cannot change without discussion and probably a long deprecation cycle.

In theory, values could be added but doing so must not break backward compatibility so this specification would be unaffected.

samuelkarp (commented):
I started trying to migrate runj to this and immediately ran into challenges with [vnet interfaces](https://samuel.karp.dev/blog/2022/12/docker-style-networking-for-freebsd-jails-with-runj/).  Are we expecting to require CNI in order to do networking?  That seems at odds with the work that was just merged for Linux in https://github.com/opencontainers/runtime-spec/pull/1271.

config-freebsd.md
```diff
@@ -0,0 +1,128 @@
...
+
+These fields SHOULD be mapped to a corresponding set of `jail(8)` parameters which can be used to create the container jail.
+A typical jail-based OCI implementation on FreeBSD MAY use the following mapping:
+
+| Jail parameter   | JSON equivalent      |
+| --------------   | -------------------- |
+| `jid`            | -                    |
+| `name`           | see below            |
+| `path`           | `root.path`          |
+| `ip4.addr`       | -                    |
```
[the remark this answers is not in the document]
samuelkarp: Since IP address management is supported by the kernel's jail interface directly, I don't think it's reasonable to require an OCI runtime to depend on CNI or netavark.

config-freebsd.md
```diff
@@ -0,0 +1,128 @@
...
+| `ip4.saddrsel`   | -                    |
+| `ip4`            | `freebsd.jail.ip4`   |
+| `ip6.addr`       | -                    |
+| `ip6.saddrsel`   | -                    |
+| `ip6`            | `freebsd.jail.ip6`   |
+| `vnet`           | `freebsd.jail.vnet`  |
```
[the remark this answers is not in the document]
samuelkarp: As above, I don't think it's reasonable to require CNI or netavark here.  This is especially true in light of https://github.com/opencontainers/runtime-spec/pull/1271.

samuelkarp:
> Currently it seems to be Linux-specific

Yes, but we have equivalent FreeBSD-specific capabilities provided by the jail interface in the kernel and we are defining the FreeBSD-specific config section here.

> I'm not sure how it handles the firewall rules which are typically added for NAT and port forwarding.

It doesn't.  This is also true for the jail interface in FreeBSD; that's left up to things like pf.

dfr:
> I think I'm currently 👎 on this PR as-is. There are two known OCI jail implementations (well, they're not _currently_ OCI given the spec doesn't have FreeBSD support yet, but they are planned to be): runj and ocijail. The config defined here does not satisfy the requirements of runj as it exists today.
> 
> I think it'd be a reasonable requirement from the spec maintainers to say that there should at least be a POC PR that demonstrates the proposed changes working in both runj and ocijail before merging. I'm tracking that for runj in [samuelkarp/runj#60](https://github.com/samuelkarp/runj/issues/60) and welcome contributions to that (I also have some WIP code on my laptop, but my time here is extremely limited right now).

Sam, that is a very good point. I do have a branch in ocijail which tracks the progress of this PR but its currently private. I'll bring that up to date with the latest iteration of the PR and make it public on github.

kolyshkin (commented):
LGTM overall, left a couple of nits (and opened https://github.com/opencontainers/runtime-spec/pull/1298 to fix a pre-existing issue).

specs-go/config.go
```diff
@@ -991,3 +993,73 @@ const (
...
+
+// These values are used to control access to features in the container, either
+// disabling the feature, sharing state with the parent or creating new private
+// state in the container.
+type FreeBSDSharing string
+
+const (
+	ShareDisable FreeBSDSharing = "disable"
+	ShareNew     FreeBSDSharing = "new"
+	ShareInherit FreeBSDSharing = "inherit"
```
samuelkarp: These consts are not namespaced to FreeBSD.  Not sure if a runtime-spec maintainer would want to have namespacing here?

samuelkarp (approved):
I've implemented this for runj in https://github.com/samuelkarp/runj/pull/63, so from my perspective this now LGTM.

specs-go/config.go
dfr: Good point - I will change the constant makes along the lines suggested by @AkihiroSuda.

AkihiroSuda (approved):

kolyshkin (approved):
LGTM

AkihiroSuda:
> [cyphar](https://github.com/cyphar) modified the milestones

@cyphar Aren't we ready to merge this in the spec v1.3.0 ? (which will correspond to runc v1.4.0)

cyphar:
Ah okay, I was going off the discussion we had a few weeks ago, I didn't notice that we'd resolved the outstanding issues. I'm okay to merge this then.

cyphar (dismissed):

specs-go/config.go
```diff
@@ -991,3 +993,75 @@ const (
...
+	Devices []FreeBSDDevice `json:"devices,omitempty"`
+	// Jail definition for this container
+	Jail *FreeBSDJail `json:"jail,omitempty"`
+}
+
+type FreeBSDDevice struct {
+	// Path to the device, relative to /dev.
+	Path string `json:"path"`
+	// FileMode permission bits for the device.
+	Mode *os.FileMode `json:"mode,omitempty"`
```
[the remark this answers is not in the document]
kolyshkin: I've noticed that too but since the max value is octal 777 (it was actually 1000 before the fix in https://github.com/opencontainers/runtime-spec/pull/1298/commits/9efd9f23e02338b75b6eddc16df1d56c9c7e46f1), I guess it's a non-issue.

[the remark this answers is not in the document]
cyphar: Ah, fair enough.

cyphar:
I was hoping to get more information about whether the runj approach was acceptable to @dfr but as I said, I don't have a strong opinion on it. I'm a bit more confused why there was a perceived rush to merge this for 1.3 (we want 1.3 for runc 1.4 -- though I also don't think this should be a hard requirement -- but the FreeBSD stuff is not really relevant for that either).

At the end of the day, I'm fine with the merged version.

dfr:
Sorry, I am a little late to this part of the discussion and failed to reply before this was merged.

I have been thinking a lot on the suggested structure which splits out the ip4 and ip6 parameters into sub-objects. I don't think this adds much in readability - in particular, the way it moves the namespace parameters (e.g. `vnet=new`, `ip4=inherit`) could be confusing given that the other namespace parameters still live in the jail object.

The underlying jail parameter set has no sub-structure - its just a list of key/value pairs where some of the keys names happen to have "." characters in them, e.g. `ip4.addr`. This is why I had a single jail object in the runtime-spec proposal which is (mostly) a simple 1-1 mapping, renaming some keys to fit with the style of other parts of runtime-spec.

Personally, I am happy with the merged version as it represents the consensus approach for our working group and has benefitted from this review process. If there are any concerns regarding the open discussion points, perhaps this can be addressed in followup pull requests if there is time before 1.3.

Maja-GHub (approved):

sherrybrownagent-bit (approved):

PR #1287, opened 2025-06-03
config-linux: add intelRdt.enableMonitoring

marquiz:
Add a parameter for enabling per-container resctrl monitoring.

This supersedes the previous `enableCMT` and `enableMBM` settings whose functionality was very vaguely specified. Separate parameter for every monitoring metric does not seem to make much sense, in particular because in the resctrl filesystem it is not possible to selectively enable a subset of the monitoring features. You always get all the metrics that the system provides. Also, with separate settings (and corresponding check if the specific metric is available) the user cannot specify "enable whatever is available" - setting everything to "true" might fail because one of the metrics is not available on the platform. In addition, having separate parameters is very future-unproof, making support for new monitoring metrics unnecessarily cumbersome to add. New metrics are certain to be added in new hardware generations, e.g. perf/energy monitoring in the near future (https://lkml.org/lkml/2025/5/21/1631), and requiring an update to the runtime-spec for each one of them feels like an overkill without much benefits. It is easier to have one switch for "enable container-specific metrics" and let the user read whatever metrics the platform provides.

Moreover, it is not even possible to turn off monitoring (from the resctrl filesystem). For example, you always get the metrics for all CTRL_MON groups (closIDs). However, that is not always very useful as there likely are a lot of applications packed in the same group. The new intelRdt.enableMontoring parameter will enable creation of a MON group specific to a single container allowing monitoring of resctrl metrics on per-container granularity.

The `intelRdt.enableCMT` and `intelRdt.enableMBM` are removed from the spec. They are vaguely specified, problematic and have no known implementations.

**NOTE:** I checked all the runtimes listed in [`implementeations.md`](https://github.com/opencontainers/runtime-spec/blob/main/implementations.md) and none of them have implemented the `enableCMT` and `enableMBM` so I wonder if there would be an option to just drop those(?)

### DIFF
--- a/config-linux.md
+++ b/config-linux.md
@@ -748,6 +748,7 @@ The following parameters can be specified for the container:
 * **`memBwSchema`** *(string, OPTIONAL)* - specifies the schema of memory bandwidth per L3 cache id.
     The value MUST start with `MB:` and MUST NOT contain newlines.
 * **`schemata`** *(array of strings, OPTIONAL)* - specifies the schemata to be written to the `schemata` file in resctrlfs. Each element represents one line in the `schemata` file. The value MUST NOT contain newlines.
+* **`enableMonitoring`** *(boolean, OPTIONAL)* - enables resctrl monitoring for the container.
 
 The following rules on parameters MUST be applied:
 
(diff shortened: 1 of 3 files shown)


kad (approved):
LGTM.
And if there is no existing implementations that are using old/deprecated flags, I'd propose to drop them completely, as it is proven to be impractical over all those years.

giuseppe (approved):
LGTM

marquiz:
> +1 from crun, they are not implemented/used.

Check. I added a separate patch that drops these fields (this could be of course debated in a separate PR, too)

config-linux.md
```diff
@@ -769,13 +770,14 @@ The following rules on parameters MUST be applied:
 
 * If `closID` is set, and none of `l3CacheSchema`, `memBwSchema` or `schemata` are set, runtime MUST check if corresponding pre-configured directory `closID` is present in mounted `resctrl`. If such pre-configured directory `closID` exists, runtime MUST assign container to this `closID` and [generate an error](runtime.md#errors) if directory does not exist.
 
-* **`enableCMT`** *(boolean, OPTIONAL)* - specifies if Intel RDT CMT should be enabled:
-    * CMT (Cache Monitoring Technology) supports monitoring of the last-level cache (LLC) occupancy
-      for the container.
-
-* **`enableMBM`** *(boolean, OPTIONAL)* - specifies if Intel RDT MBM should be enabled:
-    * MBM (Memory Bandwidth Monitoring) supports monitoring of total and local memory bandwidth
```
AkihiroSuda, line 777: The history of this technically-breaking changes should be documented as a note?

marquiz: :+1:   thanks @AkihiroSuda, I think that makes sense (I was thinking something along the lines I described in the commit message(s)). Any suggestion where to put that note?

marquiz: Got it. I now added a NOTE to config-linux.md (in the IntelRdt section).

@AkihiroSuda PTAL

utam0k (approved):
LGTM

PR #1293, opened 2025-08-16
Add vTPM specification

everzakov:
Add the vTPM (virtual Trusted Platform Module) specification to the documentation, config.go and schema description. Runtime uses this specification to create vTPMs and pass them to the container. This virtual module can be used to create quotes, signatures and perform Direct Anonymous Attestation.

Also, users can specify that vTPM should be manufactured with precreated certs / activated PCR banks and populated Endorsement Key Pair.

The following is an example of a vTPM description that is found under the path /linux/resources/vtpms:
```json
    "vtpms": [
        {
            "statePath": "/var/lib/runc/myvtpm1",
            "statePathIsManaged": false,
            "vtpmVersion": "2",
            "createCerts": false,
            "runAs": "tss",
            "pcrBanks": "sha1,sha512",
            "encryptionPassword": "mysecret",
            "vtpmName": "tpm0",
            "vtpmMajor": 100,
            "vtpmMinor": 1
        }
    ]
```

This PR is based on https://github.com/opencontainers/runtime-spec/pull/920


tianon:
If I understand correctly, the idea is that a runtime is expected to start an instance of `swtpm` behind the scenes, and wire up the result inside the container. Is that accurate? 

This is perhaps mirroring some of the concerns expressed in https://github.com/opencontainers/runtime-spec/pull/920, but what's the benefit of doing that over running `swtpm` explicitly and mapping the device or socket from it?

To maybe help explain why this makes me nervous, what do we do if the container dies? The runtime is typically long gone at that point, so what makes sure `swtpm` shuts down?  What if `swtpm` has a problem and shuts down before the container does?  At the level of the runtime, there's no "orchestrator" monitoring processes, there's just a container process and a bunch of kernel resources tied to that process (most of which clean themselves up pretty reasonably when the container exits or dies).

Another aspect is how non-container runtimes (VMs, etc) are expected to implement this.  If they can't support this, they should probably simply error, right?  The same if `swtpm` is not installed?

So in short, why is the runtime layer the appropriate place for this and not, say, the orchestrators like containerd, Docker, kubernetes, etc?

everzakov:
> Another aspect is how non-container runtimes (VMs, etc) are expected to implement this.

we assign runc to create vtpm because we want to allign the same architecture design as VM platform.
In VM scenario it is vBIOS do the job. While in container, from the component call sequence, it is runc that works as a similar role and position.

vtpm vm vs containerd pic2

everzakov:
> 1. With that example runtime-spec config you shared, what would runc need to do from a high-level point of view? Is it a bind mount of the hostPath in the containerPath? In that case, why do we need the major/minor?

Thank you for your comment and reminding me that we can also use mount config to pass device :)

We choose DRA because we want to pass swtpm config to container by resource claims.
In that case, the user only need to set params like vtpm version, vtpmName and etc in the resource claim config.

At the start vTPM plugin will publish resource slice with possible devices with their major / minor values.
The scheduler will allocate some devices from this resource slice for the specified pod.
Their major / minor will be passed to start swtpm_cuse.

In my opinion, after CUSE has created the host device, runc only need to pass / create this device in the test container
(and if possible with specified uid / gid in the test container).
Also, runc need to add the created device major/minor which are keys for virutal devices managed by the kernel in allow list cgroup
(otherwise any operation from user to this device in the created container will be rejected due to permission error)
that's why additional fields like vtpmMajor and vtpmMinor is passed.
It can be done by using NRI plugin https://github.com/containerd/nri/blob/main/pkg/runtime-tools/generate/generate.go#L287 .
However, the necessary major / minor is fixed only when pod is scheduled and the device is allocated (we can not pass them as an annotation).
If we use CDI to pass the device as a mount, it won't be possible because
we need to add the additional device node https://github.com/cncf-tags/container-device-interface/blob/main/pkg/cdi/container-edits.go#L112 .
Or we can add VTPM as a container edit :)

Also, I tried to use mount the device and i have a "bug":
1. If i understand correctly, when tpm2-tools package is installed on the host,
then 60-tpm-udev.rules https://packages.ubuntu.com/jammy/all/tpm-udev/filelist file is created.
The rule https://salsa.debian.org/debian/tpm-udev/-/blob/master/debian/tpm-udev.udev?ref_type=heads is setting tss owner on tpm[0-9]* device name.
2. if we mount the necessary device to the container rootfs, then the device will have the uid of host tss user (not the container).
3. We will have Permission denied error, if we try to use this device when the host and container tss user have the different UIDs.

If we use mknod command, the device will have a root uid.

rata:
@everzakov thanks! 

1. I'm not sure what is the point of a hostPath in the json if we will create it with mknod just using the major/minor. What am I missing?
1. It feels like an abuse to use DRA with something that has "infinite" capacity, DRA is currently designed to assign consumable devices to containers and track that. We don't have that problem here. It seems like using DRA for something it is not designed to do.
1. uid_mappings and those won't work with character devices. Those are designed to be used with filesystems. The device will need to have a uid/gid and be set to that. I guess that is missing in the json you proposed. In the userns case, you want to use the hostUID/hostGID that the root user is mapped to (or whichever user inside the container you want to be the owner of it).

I'm probably missing something obvious, but why can't we just use the existing devices section instead of adding a vtpm section? It seems to have everything we need already:
 * It has a path to use inside the container
 * major/minor
 * uid/gid
 * supports character devices

If the CUSE daemon is running on the host and the config is set with the right major/minor, then IIUC runc will do the mknod, the CUSE driver will do its thing to provision this, and it will just work.

Am I missing something?

If that works, we can later think a little bit about the vtpm state: how can we give the same state to this container when it runs on another node later? Can we do that? Or what would be the limitations? Is there something we need to take into account to "isolate" the vtpm of different containers?

But let's leave those questions for later

PR #1294, opened 2025-08-27
docs: fix and elaborate the nodes field in Linux memory policy

askervin:
Nodes is required only in some memory policy modes, while some other modes require that there must be no nodes.

This issue was pointed out by @fabbione in PR #1282 conversation.


giuseppe (approved):
LGTM

marquiz (approved):
LGTM

tianon (approved):

PR #1296, opened 2025-09-15
Introduce CPUAffinity process property instead of execCPUAffinity

kad:
This change introduces more generic `CPUAffinity` property of `Process` to specify desired CPU affinities while performing operations on create, start and exec operations.

As it was originally discussed in PR #1253, the existing implementation covers only `exec` usecase, where setting affinity for OCI hooks and initial container process will benefit wider set of workloads.

### DIFF
--- a/config.md
+++ b/config.md
@@ -340,18 +340,35 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
-* **`execCPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+* **`cpuAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute processes in the runtime and the container.
+    All the properties in this setting expects as argument a string that contain a comma-separated list, with dashes to represent ranges.
+    For example, `0-3,7` represents CPUs 0,1,2,3, and 7. Special value `all` can be used to reset CPU affinity to the value that includes all possible CPUs in the system.
+    If omitted or empty for particular [lifecycle](runtime.md#lifecycle) stage, the runtime SHOULD NOT change process' CPU affinity, and the actual affinity is determined by the Linux kernel.
+    The following properties are available:
+    * **`createRuntime`** (string, OPTIONAL) is a list of CPUs the runtime parent process should run on during the runtime creation stage, before entering the container's namespaces or cgroups.
+    CPU affinity should be applied on early stages of container creation, so that any potential CPU consuming operations inside runtime (e.g. [`createRuntime` hooks](#createRuntime-hooks)) will be run on the specified list of CPUs.
+    * **`createContainer`** (string, OPTIONAL) is a list of CPUs the process should run on during the container creation stage, after entering the container's namespaces but before the user process or any of [`createContainer` hooks](#createContainer-hooks) are executed.
+    * **`startContainer`** (string, OPTIONAL) is a list of CPUs the process should be run on during the container start stage, inside the container's namespace during `start` operation. The affinity should be applied before executing [`startContainer` hooks](#startContainer-hooks).
+    * **`execRuntime`** (string, OPTIONAL) is a list of CPUs the runtime parent process to be run during `exec` operation, before the transition to container's cgroup.
+    * **`execContainer`** (string, OPTIONAL) is a list of CPUs the process will be run on during `exec` operationon after the transition to container's cgroup.
+* **`execCPUAffinity`** (object, OPTIONAL, **DEPRECATED**) specifies CPU affinity used to execute the process.
+
+    Note: `execCPUAffinity` is deprecated in favor of `cpuAffinity`. If `cpuAffinity` is specified and supported by the runtime, the content of `execCPUAffinity` should be ignored.
+
     This setting is not applicable to the container's init process.
     The following properties are available:
     * **`initial`** (string, OPTIONAL) is a list of CPUs a runtime parent
       process to be run on initially, before the transition to container's
       cgroup. This is a a comma-separated list, with dashes to represent
       ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7.
+      Deprecated in favor of `execRuntime` in `cpuAffinity`.
     * **`final`** (string, OPTIONAL) is a list of CPUs the process will be run
       on after the transition to container's cgroup. The format is the same as
       for `initial`. If omitted or empty, runtime SHOULD NOT change process'
       CPU affinity after the process is moved to container's cgroup, and the
       final affinity is determined by the Linux kernel.
+      Deprecated in favor of `execContainer` in `cpuAffinity`.
+    
 
 ### <a name="configZOSProcess" />z/OS Process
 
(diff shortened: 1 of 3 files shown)


giuseppe:
@kolyshkin PTAL

config.md
```diff
@@ -340,18 +340,38 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
-* **`execCPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+* **`CPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute processes in the runtime and the container.
+    All the properties in this setting expects as argument a string that contain a comma-separated list, with dashes to represent
+    ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7. If omitted or empty for particular [lifecycle](runtime.md#lifecycle) stage,
+    the runtime SHOULD NOT change process' CPU affinity, and the affinity is determined by the Linux kernel.
+    The following properties are available:
+    * **`runtimeCreate`** (string, OPTIONAL) is a list of CPUs the runtime parent process should run on during the runtime creation stage,
```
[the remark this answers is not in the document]
cyphar: For better or worse, this might be non-trivial to do if someone is using a tool that wraps an OCI runtime and doesn't provide a knob for it (basically the same argument as runtime hooks).

I thin matching the lifecycle stages (which also match the hook names) as done here is probably the simplest approach.

cyphar:
Given issues like https://github.com/golang/sys/pull/259, I think it might be nice to have a way to indicate "all CPUs" as a way to reset affinity. In runc we implicitly do this now, but users might want to reset affinity for other stages explicitly.

config.md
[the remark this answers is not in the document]
kad: theoretically it is possible, e.g. via systemd to restrict container runtime to run on subset of CPUs, but this gives granularity on splitting CPU affinity for containerd/cri-o itself vs. per-container hook execution. In other words, do not affect higher level runtime' CPU quota with operations that might be local to subset of containers.

regarding names: currently it is in form "whereWhat", but OCI hooks are named "whatWhere". if we feel that OCI hook naming is better, I can rename proposed fields in this PR to have more consistency with the hooks.

specs-go/config.go
```diff
@@ -94,8 +94,12 @@ type Process struct {
 	SelinuxLabel string `json:"selinuxLabel,omitempty" platform:"linux"`
 	// IOPriority contains the I/O priority settings for the cgroup.
 	IOPriority *LinuxIOPriority `json:"ioPriority,omitempty" platform:"linux"`
+	// ExecCPUAffinity is Deprecated.
```
kolyshkin, line 97: If you want various linters to warn about using deprecated functionality (and other linters, that insist that description starts with the `Id`, to keep working fine), the deprecation notice should be added like this:

```go
// ExecCPUAffinity specifies CPU affinity for exec processes.
//
// Deprecated: use [CPUAffinity] instead.
```

(I.e. after the description, on a separate paragraph, etc.)

cyphar:
@kad

> theoretically, the higher-level runtime that generates OCI spec might be feeling it with right number based on detected system information. e.g. use result of sched_getaffinity(2) of parent process

My experience is that this rarely happens -- usually higher-level runtimes either hide new knobs like this (requiring you to specify patch `config.json` through experimental or unsupported hacks) or they transparently forward the values to the lower-level runtime without adding new functionality. Even if they do implement it, there is no guarantee that the behaviour or syntax will be standardised between runtimes, which leads to more problems than it solves.

Given that we have had seen practical issues with container runtimes being spawned with suboptimal CPU affinity values, I would suggest that having an "all" or "max" special value would be a good idea.

config.md
```diff
@@ -340,18 +340,38 @@ For Linux-based systems, the `process` object supports the following process-spe
 
     * **`class`** (string, REQUIRED) specifies the I/O scheduling class. Possible values are `IOPRIO_CLASS_RT`, `IOPRIO_CLASS_BE`, and `IOPRIO_CLASS_IDLE`.
     * **`priority`** (int, REQUIRED) specifies the priority level within the class. The value should be an integer ranging from 0 (highest) to 7 (lowest).
-* **`execCPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute the process.
+* **`CPUAffinity`** (object, OPTIONAL) specifies CPU affinity used to execute processes in the runtime and the container.
+    All the properties in this setting expects as argument a string that contain a comma-separated list, with dashes to represent
+    ranges. For example, `0-3,7` represents CPUs 0,1,2,3, and 7. If omitted or empty for particular [lifecycle](runtime.md#lifecycle) stage,
+    the runtime SHOULD NOT change process' CPU affinity, and the affinity is determined by the Linux kernel.
```
cyphar, line 346: I think it would be very useful to have an `all` special value which indicates that the CPU affinity should be reset to include as many CPUs as possible.

cyphar (changes requested):
Also, when it comes to formatting -- in the OCI specs we do not hard-wrap lines at N columns. Each complete sentence should be one line (this makes diffs easier to read).

cyphar:
@kolyshkin I think the point is to have the affinity change at different stages rather than it be hook-specific -- hooks will probably inherit them but hooks can also set their own affinity if they want to. The naming is similar to hooks because both correspond to runtime lifecycle stages.

specs-go/config.go
kad: Updated, similarly to how deprecation comment written in `Hooks` section

kad:
@kolyshkin @giuseppe @cyphar update pushed with following changes, as suggested:
- formating of Markdown fixed
- introduced special value `all`
- names of fields renamed to match hooks naming schema
- codespell fixes (sigh)

config.md
```diff
@@ -340,18 +340,31 @@ For Linux-based systems, the `process` object supports the following process-spe
...
+    * **`createContainer`** (string, OPTIONAL) is a list of CPUs the process should run on during the container creation stage, after entering the container's namespaces but before the user process or any of [`createContainer` hooks](#createContainer-hooks) are executed.
+    * **`startContainer`** (string, OPTIONAL) is a list of CPUs the process should be run on during the container start stage, inside the container's namespace during `start` operation. The affinity should be applied before executing [`startContainer` hooks](#startContainer-hooks).
+    * **`execRuntime`** (string, OPTIONAL) is a list of CPUs the runtime parent process to be run during `exec` operation, before the transition to container's cgroup.
+    * **`execContainer`** (string, OPTIONAL) is a list of CPUs the process will be run on during `exec` operationon after the transition to container's cgroup.
+* **`execCPUAffinity`** (object, OPTIONAL, **DEPRECATED**) specifies CPU affinity used to execute the process.
```
AkihiroSuda: What happens when both the old and new fields are specified?

AkihiroSuda (approved):

utam0k (approved):

AkihiroSuda:
Shall we merge this or shall we release v1.3.1 before doing that?

utam0k:
+1

kad:
> Shall we merge this or shall we release v1.3.1 before doing that? @opencontainers/runtime-spec-maintainers

If there are no objections on such change in the spec, it would be great to get it into v1.3.1, thus we would be able to add corresponding fields into NRI protocol.

AkihiroSuda:
@cyphar PTAL

AkihiroSuda:

AkihiroSuda:
@opencontainers/runtime-spec-maintainers Thoughts?

PR #1299, opened 2025-10-16
runc: support multi-container delete and update runtime-spec

yzewei:
### Summary

This PR updates the `delete` operation in the runtime-spec to support deleting multiple containers at once.

### Changes

- Lifecycle section updated to reference one or more container IDs for delete
- Delete operation signature updated to `delete <container-id> [<container-id> ...]`
- Errors must be reported per container; one failing container does not block others

### Motivation

Maintainers requested an update to the specification to clarify multi-container delete semantics before implementing multi-delete support in runtimes.

### DIFF
--- a/runtime.md
+++ b/runtime.md
@@ -73,7 +73,7 @@ The lifecycle describes the timeline of events that happen from when a container
     If any `poststart` hook fails, the runtime MUST [generate an error](#errors), stop the container, and continue the lifecycle at step 12.
 10. The container process exits.
     This MAY happen due to erroring out, exiting, crashing or the runtime's [`kill`](runtime.md#kill) operation being invoked.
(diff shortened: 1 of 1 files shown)


runtime.md
```diff
@@ -136,11 +140,13 @@ Attempting to send a signal to a container that is neither [`created` nor `runni
 This operation MUST send the specified signal to the container process.
 
 ### <a name="runtimeDelete" />Delete
-`delete <container-id>`
+`delete <container-id> [<container-id> ...]`
```
AkihiroSuda: https://github.com/opencontainers/runtime-spec/blob/5610abdb9fac3b48b2c0ba6216d77320cbbbfb6f/runtime.md?plain=1#L95

I agree with @cyphar 

This seems to be the scope of the runc's implementation details.

PR #1300, opened 2025-10-21
ci: use supported Go versions

kolyshkin:
The upside is, we don't have to bump the versions here every 6 months.
    
The downside is, CI might break once the new Go is out.
    
Overall I think it is net positive.

PS I can't do something like @tianon did in https://github.com/opencontainers/image-spec/pull/1229 as this repo lacks go.mod file 🤷🏻

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        go: [1.21.x, 1.22.x]
+        go: [oldstable, stable]
 
     steps:
       - name: checkout source code


AkihiroSuda (approved):

tianon (approved):
One question (getting at the intent of the `go.mod` inclusion elsewhere), but otherwise LGTM

cyphar (approved):
I think I would like to have a `go.mod` but that's not particularly critical yet. I suspect one of the issues is that if we put the `go.mod` in the `specs-go` directory we would need to publish `specs-go/vX.Y.Z` tags?

Maja-GHub (approved):

PR #1310, opened 2026-02-17
Don't ignore poststop hook failures

fspv:
A proposed change for this issue: https://github.com/opencontainers/runtime-spec/issues/1309


giuseppe:

